PR #12, opened 2020-05-29
Add a maximum number of retries and minimum throttle delay between retrying requests

jharrington22:
This PR attempts to mitigate issues with throttling when checking permissions.

### DIFF
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -19,10 +19,12 @@ package aws
 import (
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


jharrington22:
@vkareh these are the throttling errors we see intermittently, they are hard to reproduce what this PR is doing is limiting to amount of times the client can retry if there is an error to avoid these problems:

```
E: Unable to validate SCP policies
E: error simulating policy: Throttling: Rate exceeded
status code: 400, request id: 406846a7-6b67-4adc-a3e3-6f48768a1b87
```

vkareh:
Got it, thanks! yes I think I've seen that error a while ago (haven't experienced it in a while though)

/lgtm

PR #15, opened 2020-06-04
CreateReporterOrExit function for commands

ashcrow:
- Add `CreateReporterOrExit` function to reporter
- Use it in commands rather than the same code block

Admittedly not well tested but seems like it's a safe change :smile:


jharrington22:
Nice!

PR #17, opened 2020-06-04
whoami: Prefix output with source API

vkareh:
To distinguish data from AWS versus OCM, we prefix the output to include
the source. Also renames "OpenShift API" to "OCM API" to avoid
confusion.

### DIFF
--- a/cmd/whoami/cmd.go
+++ b/cmd/whoami/cmd.go
@@ -140,17 +140,17 @@ func run(_ *cobra.Command, _ []string) {
 		account = response.Body()
 	}
 	fmt.Printf(""+
-		"AWS Account ID:           %s\n"+
-		"AWS Default Region:       %s\n"+
-		"AWS ARN:                  %s\n"+
-		"OpenShift API:            %s\n"+
-		"Account ID:               %s\n"+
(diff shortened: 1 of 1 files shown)


cmd/whoami/cmd.go
```diff
@@ -140,17 +140,17 @@ func run(_ *cobra.Command, _ []string) {
...
-		"Account Name:             %s %s\n"+
-		"Account Username:         %s\n"+
-		"Account Email:            %s\n"+
-		"Organization ID:          %s\n"+
-		"Organization Name:        %s\n"+
-		"Organization External ID: %s\n",
+		"AWS Account ID:               %s\n"+
+		"AWS Default Region:           %s\n"+
+		"AWS ARN:                      %s\n"+
+		"OCM API:                      %s\n"+
```
jharrington22, line 146: Is this what we want? `OCM Cluster API` maybe?

vkareh: Maybe? What would make sense for customers?

/cc @jeremyeder

vkareh: Let's leave as OCM API for now and we can change later if it causes confusion

PR #23, opened 2020-06-08
CreateLoggerOrExit function for commands

ashcrow:
Very similar to https://github.com/openshift/moactl/pull/15/

- Add CreateLoggerOrExit function to reporter
- Use it in commands rather than the same code block
- Update `whoami` to use the functions as well


vkareh:
works great! thanks!

/lgtm

PR #30, opened 2020-06-30
interactive: Add interactive flag to create idp

vkareh:
This adds support for the interactive flag on the `create idp` command.
Currently supported by GitHub and Google IdPs. The OpenID and LDAP
options will fall back to the legacy interactive mode.

Demo: https://asciinema.org/a/Koqdz81jtapuawl3rxRe7qOK8

### DIFF
--- a/cmd/create/idp/github.go
+++ b/cmd/create/idp/github.go
@@ -23,41 +23,71 @@ import (
 	"strings"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/spf13/cobra"
 
 	"github.com/openshift/moactl/pkg/interactive"
 )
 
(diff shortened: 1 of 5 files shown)


jharrington22:
@vkareh nice! A good amount of work to implement interactive for all these pieces but a nice feature.

cmd/create/idp/github.go
```diff
@@ -23,41 +23,71 @@ import (
...
 	organizations := args.githubOrganizations
 	teams := args.githubTeams
-	teamsOrOrgs := ""
 
 	if organizations != "" && teams != "" {
 		return idpBuilder, errors.New("GitHub IDP only allows either organizations or teams, but not both")
 	}
 
-	isInteractive := clientID == "" || clientSecret == "" || (organizations == "" && teams == "")
+	restrictType := ""
```
jharrington22, line 39: Small nit `var restrictType string` here instead?

vkareh: fixed

PR #33, opened 2020-07-02
addons: Add list and create commands for add-ons

vkareh:
When listing add-ons for a cluster, we show the list of enabled add-ons
that the current organization has quota for, as well as those already
installed in the cluster. From this list the user can use the ID to
manually install add-ons using the `create` command.

Since there can be confusion regarding the spelling of add-ons, I've
added aliases for "addon", "add-on", as well as their plurals. Also,
since the user is not "creating" and add-on, we can use the alias "add"
instead of "create", although both work similarly.

Demo: https://asciinema.org/a/rlF6g0IujtKZNwT1s1d3B6f0u

### DIFF
--- a/cmd/create/addon/cmd.go
+++ b/cmd/create/addon/cmd.go
@@ -1,0 +1,140 @@
+/*
+Copyright (c) 2020 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package addon
+
+import (
+	"os"
+
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/moactl/pkg/aws"
+	clusterprovider "github.com/openshift/moactl/pkg/cluster"
+	"github.com/openshift/moactl/pkg/logging"
+	"github.com/openshift/moactl/pkg/ocm"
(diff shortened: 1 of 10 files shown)


cmd/create/addon/cmd.go
```diff
@@ -0,0 +1,150 @@
...
+	_, err = clustersCollection.Cluster(cluster.ID()).
+		Addons().
+		Add().
+		Body(addOnInstallation).
+		Send()
+	if err != nil {
+		reporter.Errorf("Failed to add add-on installation '%s' for cluster '%s': %s", addOnID, clusterKey, err)
+		os.Exit(1)
+	}
+	reporter.Debugf("Installed add-on '%s' on cluster '%s'", addOnID, clusterKey)
```
meowfaceman, line 149: Could you pop this into something reusable in pkg? osde2e could definitely use this.

vkareh: Ah, good point - I'll add to the cluster provider

vkareh: How about now? I've added `clusterprovider.InstallAddOn()`, which I think you should be able to use straight up in osde2e without having to wrap around OCM

meowfaceman: Would it be possible to take the aws related code and push this inside that function as well? That way it more encapsulates the extra functionality that moactl is providing.

If not, we can work around, but I like the AWS related code sitting in the function because that's a lot of the primary value of moactl, I think.

vkareh: I think what you're describing is better accomplished in separate refactor commit. I'm currently following the same pattern as clusterprovider.GetCluster, UpdateCluster, DeleteCluster, etc. They all take the clusterKey and the ARN. The only exception is creating the cluster, which needs it for grabbing the key/secret pair and tagging the IAM user.

I'd be happy to have an offline chat to discuss a refactor that would make it easier to consume externally. What do you think?

meowfaceman: +1 to that, that makes sense. @jharrington22 has mentioned this in the past as well, but I totally agree. No need to discuss offline (especially since I'm going on PTO next week :-)) Maybe that's something we can take on. I'll file a JIRA for this against SDCICD to help refactor that bit.

vkareh: Sounds good, thanks! :) Enjoy your time off!

PR #41, opened 2020-07-07
adding a tldr section to quickstart

lamek:
Per hackday feedback, adding a tldr version to the top of the quickstart for extra eager or returning users.


jeremyeder:
LGTM

PR #42, opened 2020-07-09
logs improvements

vkareh:
* logs: Detach log watch when cluster is Ready
* create-cluster: Allow user to watch cluster installation logs

Depends on https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/1989

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -27,6 +27,7 @@ import (
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 
+	clusterdescribe "github.com/openshift/moactl/cmd/describe/cluster"
 	clusterlogs "github.com/openshift/moactl/cmd/logs/cluster"
 
 	"github.com/openshift/moactl/pkg/aws"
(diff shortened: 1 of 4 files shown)


vkareh:
/hold
As this depends on a change in OCM that hasn't been merged yet

vkareh:
/unhold

I just merged the necessary changes to CS. Will be available in int/stage shortly and will rollout to prod tomorrow

cmd/create/cluster/cmd.go
```diff
@@ -418,18 +431,27 @@ func run(cmd *cobra.Command, _ []string) {
 
 	clusterID := cluster.ID()
 	clusterName := cluster.Name()
-	reporter.Infof("Creating cluster with identifier '%s' and name '%s'", clusterID, clusterName)
-	reporter.Infof("To view list of clusters and their status, run 'moactl list clusters'")
-
+	reporter.Debugf("Creating cluster with identifier '%s' and name '%s'", clusterID, clusterName)
```
RiRa12621, line 434: This line does seem a bit misplaces even as debug since there's no action between this and "Cluster has been created"
If this moved to L425 it could stay on info level

The flow from a user perspective would be 
* creating cluster
* short wait
* cluster has been created 

or in case of failure
* creating cluster
* short wait
* failed to create cluster

vkareh: Good catch, thanks! I've removed that line, as we don't have the cluster ID prior to creating it

vkareh:
Agreed. Also, there are currently inconsistencies between this and the clusters service API...

vkareh:
Opened https://github.com/openshift/moactl/pull/78 to replace this

PR #45, opened 2020-07-14
.github: add super simple go workflow

RiRa12621:
This is running `go build -v .` as github action on every PR as a super simple way to ensure a PR doesn't break anything to a point where it cannot be built at least


RiRa12621:
GitHub actions are currently disabled globally in the open shift Organisation, so this wouldn't work.

PR #52, opened 2020-07-16
OSD-4362: add additional debug information verify permissions

RiRa12621:
This is very much a monkey patch to give users a pointer where to go if the permission verify fails.

### DIFF
--- a/cmd/verify/permissions/cmd.go
+++ b/cmd/verify/permissions/cmd.go
@@ -64,6 +64,8 @@ func run(cmd *cobra.Command, argv []string) {
 	if err != nil {
 		reporter.Errorf("Unable to validate SCP policies")
 		reporter.Errorf("%v", err)
+		reporter.Infof("Verify your user's AWS account permissions")
+		reporter.Infof("You can set it up correctly by running aws iam attach-user-policy --user-name <username> --policy-arn \"arn:aws:iam::aws:policy/AdministratorAccess\" ")
 		os.Exit(1)
 	}
 	if !ok {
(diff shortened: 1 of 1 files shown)


cmd/verify/permissions/cmd.go
```diff
@@ -64,6 +64,8 @@ func run(cmd *cobra.Command, argv []string) {
 	if err != nil {
 		reporter.Errorf("Unable to validate SCP policies")
 		reporter.Errorf("%v", err)
+		reporter.Infof("Verify your user's AWS account permissions")
```
jharrington22, line 67: It would be better if this error informed the user that their specific IAM user doesn't have the correct permissions and to add the "AdministratorAccess" policy to it. We could even refer to the ARN `arn:aws:iam::aws:policy/AdministratorAccess`. I don't think we should provide the CLI steps as a user my not be able to add that policy to their user if they don't have the access already.

jharrington22: The last sentence doesn't make much sense but what I'm trying to say is if they have been given an account and it doesn't have the AdministratorAccess policy attached to it, its unlikely they will be able to add it if they are not the account owner.

RiRa12621: Right that's why I only put "verify permissions" 
we don't know if the current user can actually those permissions for themselves, that's why i kept it generic

jharrington22:
@RiRa12621 could you look at the lint failure here so we can merge? Thanks.

RiRa12621:
it's not liking the line being too long
i'll fix that @jharrington22

jharrington22 (approved):
/lgtm

RiRa12621:
/hold
Needs fixing to please the lint check

RiRa12621:
/close
outdated

PR #57, opened 2020-07-21
updates from second moa hackday

lamek:


jharrington22:
Nice @lamek /lgtm

jharrington22:
Nice @lamek /lgtm

PR #58, opened 2020-07-21
Don't validate AWS Organization List Policies

jharrington22:
Don't validate AWS policy specifically for validating SCP changes and not required for cluster install.


jeremyeder:
This the one most likely to reduce throttling errors?  If so, merge it asap before tonight's hackfest and cut another release?  /cc @vkareh

jharrington22:
No this isn't causing the throttling errors however we did see 2 people hit an error with this one permission. The next PR removing the entire validation of the installer permissions will likely resolve the throttling errors.

PR #62, opened 2020-07-24
create-cluster: Set compute node defaults based on AZ

vkareh:
To avoid the API from rejecting a cluster creation request based on the
number of nodes, we set a reasonable default: 4 nodes for single AZ
clusters and 9 for multi AZ clusters.

### DIFF
--- a/docs/moactl_create_cluster.md
+++ b/docs/moactl_create_cluster.md
@@ -28,8 +28,7 @@ moactl create cluster [flags]
       --version string                Version of OpenShift that will be used to install the cluster, for example "4.3.10"
       --multi-az                      Deploy to multiple data centers.
       --compute-machine-type string   Instance type for the compute nodes. Determines the amount of memory and vCPU allocated to each compute node.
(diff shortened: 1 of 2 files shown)


docs/moactl_create_cluster.md
```diff
@@ -28,8 +28,7 @@ moactl create cluster [flags]
       --version string                Version of OpenShift that will be used to install the cluster, for example "4.3.10"
       --multi-az                      Deploy to multiple data centers.
       --compute-machine-type string   Instance type for the compute nodes. Determines the amount of memory and vCPU allocated to each compute node.
-      --compute-nodes int             Number of worker nodes to provision per zone. Single zone clusters need at least 4 nodes, while multizone clusters need at least 9 nodes (3 per zone) for resiliency.
-                                      
+      --compute-nodes int             Number of worker nodes to provision per zone. Single zone clusters need at least 4 nodes, multizone clusters need at least 9 nodes. (default 4)
```
cben, line 31: Maybe should say (default 4 or 9, respectively) or something like that.

vkareh: That's a tricky one, as the `(default 4)` is added automatically by the cobra library... the alternative would be to _not_ set a default (so nothing shows up in the help text) but then force a default value in the code...

PR #63, opened 2020-07-24
create-idp: Allow user to specify IdP name

vkareh:
If the user does not supply a name for the identity provider, one will
be generated based on the identity provider type.

### DIFF
--- a/cmd/create/idp/cmd.go
+++ b/cmd/create/idp/cmd.go
@@ -36,6 +36,7 @@ var args struct {
 	clusterKey string
 
 	idpType string
+	idpName string
 
 	clientID      string
 	clientSecret  string
@@ -98,14 +99,20 @@ func init() {
 		"type",
 		"t",
 		"",
-		fmt.Sprintf("Type of identity provider. Options are %s\n", validIdps),
+		fmt.Sprintf("Type of identity provider. Options are %s.", validIdps),
+	)
+	flags.StringVar(
+		&args.idpName,
+		"name",
+		"",
+		"Name for the identity provider.\n",
 	)
 
 	flags.StringVar(
 		&args.mappingMethod,
 		"mapping-method",
 		"claim",
-		"Specifies how new identities are mapped to users when they log in",
+		"Specifies how new identities are mapped to users when they log in.",
 	)
 	flags.StringVar(
 		&args.clientID,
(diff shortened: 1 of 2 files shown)


cmd/create/idp/cmd.go
```diff
@@ -98,14 +99,20 @@ func init() {
...
 		"t",
 		"",
-		fmt.Sprintf("Type of identity provider. Options are %s\n", validIdps),
+		fmt.Sprintf("Type of identity provider. Options are %s.", validIdps),
+	)
+	flags.StringVar(
+		&args.idpName,
+		"name",
+		"",
+		"Name for the identity provider.\n",
```
igoihman, line 108: nit: newline is not needed

vkareh: The newline is there to align the help output better. Otherwise all the options get blended and hard to read
```
$ moactl create idp --help
Add an Identity providers to determine how users log into the cluster.

Usage:
  moactl create idp [flags]

Examples:
  # Add a GitHub identity provider to a cluster named "mycluster"
  moactl create idp --type=github --cluster=mycluster

  # Add an identity provider following interactive prompts
  moactl create idp --cluster=mycluster --interactive

Flags:
  -c, --cluster string               Name or ID of the cluster to add the IdP to (required).
  -t, --type string                  Type of identity provider. Options are [github google ldap openid].
      --name string                  Name for the identity provider.
                                     
      --mapping-method string        Specifies how new identities are mapped to users when they log in. (default "claim")
      --client-id string             Client ID from the registered application.
      --client-secret string         Client Secret from the registered application.
                                     
      --hostname string              GitHub: Optional domain to use with a hosted instance of GitHub Enterprise.
      --organizations string         GitHub: Only users that are members of at least one of the listed organizations will be allowed to log in.
      --teams string                 GitHub: Only users that are members of at least one of the listed teams will be allowed to log in. The format is <org>/<team>.
                                     
      --hosted-domain string         Google: Restrict users to a Google Apps domain.
                                     
      --url string                   LDAP: An RFC 2255 URL which specifies the LDAP search parameters to use.
      --bind-dn string               LDAP: DN to bind with during the search phase.
      --bind-password string         LDAP: Password to bind with during the search phase.
      --id-attributes string         LDAP: The list of attributes whose values should be used as the user ID. (default "dn")
      --username-attributes string   LDAP: The list of attributes whose values should be used as the preferred username. (default "uid")
      --name-attributes string       LDAP: The list of attributes whose values should be used as the display name. (default "cn")
      --email-attributes string      LDAP: The list of attributes whose values should be used as the email address.
                                     
      --issuer-url string            OpenID: The URL that the OpenID Provider asserts as the Issuer Identifier. It must use the https scheme with no URL query parameters or fragment.
      --email-claims string          OpenID: List of claims to use as the email address.
      --name-claims string           OpenID: List of claims to use as the display name.
      --username-claims string       OpenID: List of claims to use as the preferred username when provisioning a user.
                                     
  -h, --help                         help for idp

Global Flags:
      --debug         Enable debug mode.
  -i, --interactive   Enable interactive mode.
  -v, --v Level       log level for V logs
```

PR #65, opened 2020-07-27
Return error if using root account

yeya24:
Signed-off-by: Ben Ye <yb532204897@gmail.com>

Fixes https://issues.redhat.com/browse/OSD-4403

This pr checks the IAM user of credentials, if it is a root account, return error.

### DIFF
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -17,6 +17,7 @@ limitations under the License.
 package aws
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"time"
@@ -95,11 +96,10 @@ func (b *ClientBuilder) Region(value string) *ClientBuilder {
 }
 
 // Build uses the information stored in the builder to build a new AWS client.
-func (b *ClientBuilder) Build() (result Client, err error) {
+func (b *ClientBuilder) Build() (Client, error) {
 	// Check parameters:
 	if b.logger == nil {
-		err = fmt.Errorf("Logger is mandatory")
-		return
+		return nil, fmt.Errorf("Logger is mandatory")
 	}
 
 	// Create the AWS logger:
@@ -107,7 +107,7 @@ func (b *ClientBuilder) Build() (result Client, err error) {
 		Logger(b.logger).
 		Build()
 	if err != nil {
-		return
+		return nil, err
 	}
 
 	// Create the AWS session:
@@ -127,6 +127,9 @@ func (b *ClientBuilder) Build() (result Client, err error) {
 			},
 		},
 	})
+	if err != nil {
+		return nil, err
+	}
 
 	if b.logger.IsLevelEnabled(logrus.DebugLevel) {
 		var dumper http.RoundTripper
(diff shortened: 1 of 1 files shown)


pkg/aws/client.go
```diff
@@ -158,7 +159,7 @@ func (b *ClientBuilder) Build() (result Client, err error) {
 	}
 
 	// Create and populate the object:
-	result = &awsClient{
+	c := &awsClient{
```
RiRa12621, line 162: any reason to change the variable name?

pkg/aws/client.go
```diff
@@ -168,7 +169,17 @@ func (b *ClientBuilder) Build() (result Client, err error) {
...
 
-	return
+	_, root, err := getClientDetails(c)
+	if err != nil {
+		err = fmt.Errorf("failed to get client details: %v", err)
+		return nil, err
+	}
+
+	if root {
+		return nil, errors.New("root account shouldn't be used, please use IAM account")
```
RiRa12621, line 179: Can we make that sound a little more friendly? end of the day that's a customer hitting something unexpected at the very beginning of their customer journey

pkg/aws/client.go
yeya24: The return value result https://github.com/openshift/moactl/pull/65/files#diff-45cbc761a9f9e83f2c1f69c292f06703R99 has type `Client`, it is an interface. While `getClientDetails` only takes `*awsClient` type. So I cannot pass `result` to that function and have to rename the var here.

```
func getClientDetails(awsClient *awsClient)
```

pkg/aws/client.go
yeya24: @RiRa12621  Updated, PTAL, thanks!

pkg/aws/client.go
vkareh: In that case you should probably make the return anonymous
```
func (b *ClientBuilder) Build() (Client, error)
```
Then you can keep
```
result := &awsClient{...}
```
Otherwise you have a named return that doesn't get used

yeya24: Yes, I think returning anonymous values is better. I will update this pr.

vkareh: Thanks, I think this works

yeya24:
Hey, I already tested this locally using my account.

That's the screenshot.

Screenshot from 2020-07-27 11-11-39

pkg/aws/client.go
```diff
@@ -168,7 +166,16 @@ func (b *ClientBuilder) Build() (result Client, err error) {
...
 	}
 
-	return
+	_, root, err := getClientDetails(c)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get client details: %v", err)
+	}
+
+	if root {
+		return nil, errors.New("using root account is not recommended, please use IAM account instead")
```
fahlmant, line 175: Just a small nit, ` an IAM user` is a bit better as far as AWS Terminology goes

yeya24: Great! Thanks, I will update the pr.

acathrow:
Are we using the term "recommended" ? This implies that the user has a choice when they do not.

yeya24:
> Are we using the term "recommended" ? This implies that the user has a choice when they do not.

Hello, @acathrow. That makes sense, thank you! I am not a native speaker :cry:  and I am not sure what words should I use, it would be great if you could give some suggestions.

pkg/aws/client.go
```diff
@@ -168,7 +166,16 @@ func (b *ClientBuilder) Build() (result Client, err error) {
...
 	}
 
-	return
+	_, root, err := getClientDetails(c)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get client details: %v", err)
+	}
+
+	if root {
+		return nil, errors.New("using root account is not recommended, please use an IAM user instead")
```
RiRa12621, line 175:
```suggestion
		return nil, errors.New("using a root account is not support, please use an IAM user instead")
```

RiRa12621:
no worries about that @yeya24 
i've made a suggestion.
by saying that it's not supported, we imply that it wont work, but do still sound friendly and supportive

yeya24:
> no worries about that @yeya24
> i've made a suggestion.
> by saying that it's not supported, we imply that it wont work, but do still sound friendly and supportive

Thanks! :100:

PR #67, opened 2020-07-30
logs: Add uninstall logs

vkareh:
When a cluster is uninstalling, we swap the type of logs being queried
to return the uninstall/deprovision logs from Hive.

Also, since the only "logs" we get are for the cluster (rather than IdP,
ingress, etc), the user should be able to just call 'moactl logs'
directly on the cluster and that should be sufficient for the command to
understand what logs the user is requesting


pkg/ocm/logs.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+		return
+	}
+
+	return response.Body(), nil
+}
+
+func PollInstallLogs(client *cmv1.ClustersClient, clusterID string,
+	cb func(*cmv1.LogGetResponse) bool) (logs *cmv1.Log, err error) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Hour)
```
jharrington22, line 63: What are the chances of an install taking longer than an hour? I guess you can just restart the logs.

vkareh: I figured an hour was a good compromise since the log detach is not currently working, so it will exit after an hour regardless. But I can bump this to 4 hours or something once I get the log detach correctly implemented

jharrington22: :+1: Sounds good to me, I think its likely an install will take more than an hour occasionally. Probably ok for the uninstall log.

vkareh:
I've rebased against master and fix conflicts. Let me know if this looks good to merge

vkareh:
Based on team feedback, I've split the logs into two distinct resources: `moactl logs install` and `moactl logs uninstall`

PR #70, opened 2020-08-04
Check user exists

NautiluX:
* If stack already exists, and is not CREATE_COMPLETED, throw an error
* If stack already exists, but is DELETE_COMPLETED, allow creation
* If stack is CREATE_COMPLETED do not attempt to create it
* Add tests for EnsureOsdCcsAdminUser

see OSD-4363


pkg/aws/client.go
```diff
@@ -253,6 +274,27 @@ func (c *awsClient) EnsureOsdCcsAdminUser(stackName string) (bool, error) {
...
+		return false, err
+	}
+
+	for _, summary := range stackList.StackSummaries {
+		if *summary.StackName == stackName {
+			if *summary.StackStatus == cloudformation.StackStatusCreateComplete {
+				return true, nil
+			}
+			if *summary.StackStatus != cloudformation.StackStatusDeleteComplete {
+				return false, fmt.Errorf("Error creating user: Cloudformation stack %s existing with status %s. Expected status %s.\n Ensure user osdCcsAdmin does not exist, then retry with\n moactl init --delete-stack; moactl init",
```
vkareh, line 289: This line is too long:
```
pkg/aws/client.go:289: line is 221 characters (lll)
				return false, fmt.Errorf("Error creating user: Cloudformation stack %s existing with status %s. Expected status %s.\n Ensure user osdCcsAdmin does not exist, then retry with\n moactl init --delete-stack; moactl init",
```

Makefile
```diff
@@ -55,3 +55,10 @@ generate:
 docs: moactl
 	rm -rf docs && mkdir docs
 	./moactl docs -d ./docs -f markdown
+
+mocks:
```
vkareh, line 59: I think it might make sense for `mocks` to be a dependency of `test`:
```diff
diff --git a/Makefile b/Makefile
index 9402f5e..f538192 100644
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ moactl: generate
        go build .
 
 .PHONY: test
-test:
+test: mocks
        ginkgo -r cmd pkg
 
 .PHONY: fmt
```

NautiluX: you need to run this only if you change a underlying interface, so I think it's better to not run it on every make to have it more solid.

vkareh: makes sense

PR #71, opened 2020-08-04
interactive: Fix function call from broken dependency

### LINKED ISSUE #69: `go install` moactl fails
$ go get -u github.com/openshift/moactl
$ ls ${GOROOT}/bin/moactl //does not exist
$ go install github.com/openshift/moactl

> github.com/openshift/moactl/pkg/interactive
> /root/go/src/github.com/openshift/moactl/pkg/interactive/interactive.go:224:11: assignment mismatch: 2 variables but 
> core.RunTemplate returns 3 values

https://github.com/openshift/moactl/blob/master/pkg/interactive/interactive.go#L224
the line is question.

I would guess that go get doesn't check the versions correctly and takes the latest version of https://github.com/AlecAivazis/survey/blob/master/core/template.go#L31
which returns coloredString, nonColorString, err

vkareh:
A recent minor release of the AlecAivazis/survey module broke
compatibility, causing a `go get -u` to error out.

This updates to the latest version of the module and fixes the broken
function call.

* Relevant: https://github.com/AlecAivazis/survey/commit/025716185c739dd70633ca8c3fa47c2c1aed0dfc#r40567362
* Fixes #69


jharrington22:
Ouch.

/lgtm

vkareh:
Rebased and force-pushed

PR #72, opened 2020-08-04
addons: Remove global list of add-ons

vkareh:
We should not show the entire list of add-ons in our database. Instead,
we should just gate the list of add-ons by cluster and take into account
the organization's quota.


vkareh:
rebased

PR #73, opened 2020-08-04
addons: Confirm add-on installation

vkareh:
Since add-ons cannot be uninstalled, we warn the user and ask for
confirmation before installing the add-on.

### DIFF
--- a/cmd/create/addon/cmd.go
+++ b/cmd/create/addon/cmd.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/openshift/moactl/pkg/aws"
 	clusterprovider "github.com/openshift/moactl/pkg/cluster"
+	"github.com/openshift/moactl/pkg/confirm"
 	"github.com/openshift/moactl/pkg/logging"
 	"github.com/openshift/moactl/pkg/ocm"
 	rprtr "github.com/openshift/moactl/pkg/reporter"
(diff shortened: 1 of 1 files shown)


cmd/create/addon/cmd.go
```diff
@@ -130,12 +131,15 @@ func run(_ *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
-	reporter.Debugf("Installing add-on '%s' on cluster '%s'", addOnID, clusterKey)
-	err = clusterprovider.InstallAddOn(clustersCollection, clusterKey, awsCreator.ARN, addOnID)
-	if err != nil {
-		reporter.Errorf("Failed to add add-on installation '%s' for cluster '%s': %s", addOnID, clusterKey, err)
-		os.Exit(1)
+	reporter.Warnf("Once installed, add-ons cannot be uninstalled")
+	if confirm.Confirm("install add-on '%s' on cluster '%s'", addOnID, clusterKey) {
```
jharrington22, line 135: Should we add a global `--force` or `--confirm` flag that ignores these for users who might script this?

vkareh: Good idea, working on it now

vkareh: done

PR #74, opened 2020-08-04
create-cluster: Deprecate --name in favor of --cluster-name

vkareh:
This makes the flag shorthand to be '-c', which makes it consistent
across all other cluster commands.


jharrington22:
/lgtm

@lamek could you make sure docs are up to date?

PR #80, opened 2020-08-10
addons: Hide addons until it's feature-complete

vkareh:
Since addons do not support deletion yet, we hide the feature to prevent
confusion and unnecessary support tickets.


lamek:
LGTM.

PR #81, opened 2020-08-12
Fix/verify rate limits

boranx:
Fixes: OSD-4671

I've realized a `Retryer: client.DefaultRetryer`  was defined but `NumMaxRetries` was missing. By default, the value is `0` which means there will be no retry unless we increase the number. 
Results are looking good, I've tested it in while loop on 5 different tabs, didn't see any `400` but responses may take longer as a result of retry mechanism.

### DIFF
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -140,6 +140,7 @@ func (b *ClientBuilder) Build() (Client, error) {
 			MaxRetries: aws.Int(25),
 			// Set MinThrottleDelay to 1 second
 			Retryer: client.DefaultRetryer{
+				NumMaxRetries:    5,
 				MinThrottleDelay: 1 * time.Second,
 			},
 			Logger: logger,
--- a/go.mod
+++ b/go.mod
@@ -5,12 +5,9 @@ go 1.13
 require (
 	github.com/AlecAivazis/survey/v2 v2.1.0
 	github.com/aws/aws-sdk-go v1.29.17
-	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
 	github.com/golang/mock v1.4.3
-	github.com/json-iterator/go v1.1.10 // indirect
-	github.com/konsorten/go-windows-terminal-sequences v1.0.3 // indirect
 	github.com/mattn/go-colorable v0.1.7 // indirect
 	github.com/mgutz/ansi v0.0.0-20200706080929-d51e80ef957d // indirect
 	github.com/mitchellh/go-homedir v1.1.0


georgettica:
nice change! should there be tests for this code segment?

boranx:
for `checkPermissionsUsingQueryClient` and related functions, there are no tests but I can add! That would be a nice value-adding

boranx:
/hold
Adding tests for permissions

pkg/aws/permissions.go
```diff
@@ -68,25 +68,22 @@ func checkPermissionsUsingQueryClient(queryClient, targetClient *awsClient, poli
 		}
 	}
 
+	resp, err := client.iamClient.SimulatePrincipalPolicy(input)
```
jharrington22, line 71: we use `SimulatePrincipalPolicyPages()` here to ensure we capture all results. If the returned results are too long AWS will paginate the response, without it we could risk not validating all permissions.

pkg/aws/permissions.go
```diff
@@ -36,11 +36,11 @@ type SimulateParams struct {
 
 // checkPermissionsUsingQueryClient will use queryClient to query whether the credentials in targetClient can perform
 // the actions listed in the statementEntries. queryClient will need iam:GetUser and iam:SimulatePrincipalPolicy
-func checkPermissionsUsingQueryClient(queryClient, targetClient *awsClient, policyDocument PolicyDocument,
+func checkPermissionsUsingQueryClient(client *awsClient, policyDocument PolicyDocument,
```
jharrington22, line 39: We have both `queryClient` and `targetClient` here because we should actually be validating the ARN of the user created with cloudformation. I suggest leaving this and using the `targetClient` ARN when completing [OSD-4708](https://issues.redhat.com/browse/OSD-4708)

pkg/aws/client.go
```diff
@@ -140,6 +140,7 @@ func (b *ClientBuilder) Build() (Client, error) {
 			MaxRetries: aws.Int(25),
 			// Set MinThrottleDelay to 1 second
 			Retryer: client.DefaultRetryer{
+				NumMaxRetries:    5,
```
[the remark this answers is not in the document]
boranx: I've tested it with 2 different accounts. One is my personal and the other what @fahlmant gave me. The retry is not a permanent solution I agree, but still, the number of 400s significantly reduced(I didn't see in the logs any but still can occur..). Having said that, there's no response code that `resp` contains (`resp, err := client.iamClient.SimulatePrincipalPolicy(input)`) It only fills the err message with status code when anything happens except 200. Because of that, it's hard to adjust the retry numbers dynamically if continuously 400 happens

pkg/aws/permissions.go
boranx: OK, I'll revert here

pkg/aws/permissions.go
boranx: mm that's ambiguous because still hasn't been implemented yet so we're creating structures below which are poor in terms of readability. Though, I'll take that suggestion and revert here as well

```go
hasPermissions, err := validatePolicyDocuments(c, c, policyDocuments, sParams)
```

pkg/aws/client.go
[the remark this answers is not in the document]
boranx: what could be the other alternatives? We're making a single request for policy checks so I don't think we can reduce the number of requests at all.

vkareh:
@jharrington22 - you had a change request earlier, does this lgty now?

pkg/aws/client.go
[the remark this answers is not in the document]
jharrington22: Ok no problem, lets see how thing goes and look at it closer if we continue to see rate limiting errors.

PR #84, opened 2020-08-19
Ensure region on cluster create, update SDK

vkareh:
When creating a cluster, we want to make sure we instantiate the AWS client using the same region that the cluster is set to.

This also updates the OCM SDK to the latest version, which adds extra logging output. After discussion with @cben, we've determined that most SDK logs belong to the debug level output.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -97,6 +97,7 @@ func CreateCluster(client *cmv1.ClustersClient, config Spec) (*cmv1.Cluster, err
 
 	// Create the AWS client:
 	awsClient, err := aws.NewClient().
+		Region(config.Region).
 		Logger(logger).
 		Build()
 	if err != nil {
--- a/pkg/logging/ocm_logger.go
+++ b/pkg/logging/ocm_logger.go
@@ -93,11 +93,11 @@ func (l *OCMLogger) Debug(ctx context.Context, format string, args ...interface{
 }
 
 func (l *OCMLogger) Info(ctx context.Context, format string, args ...interface{}) {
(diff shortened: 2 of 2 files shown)


go.sum
```diff
@@ -275,6 +278,8 @@ github.com/onsi/gomega v1.7.0/go.mod h1:ex+gbHU/CVuBBDIJjb2X0qEXbFg53c61hWP/1Cpa
 github.com/op/go-logging v0.0.0-20160315200505-970db520ece7/go.mod h1:HzydrMdWErDVzsI23lYNej1Htcns9BCg93Dk0bBINWk=
 github.com/openshift-online/ocm-sdk-go v0.1.117 h1:MbRDTDfa1ip7pmgp8zSaE/7X933OuysvcUYDPBWh0bk=
 github.com/openshift-online/ocm-sdk-go v0.1.117/go.mod h1:BBF7bNkfHkkVwKolft9AsbYXJBsFiSOFQ6bLSedQby0=
+github.com/openshift-online/ocm-sdk-go v0.1.122 h1:IooPjb+CWhPICuhSTMLxJsKyLmvVV5GRlHug/QRU0bU=
+github.com/openshift-online/ocm-sdk-go v0.1.122/go.mod h1:3i3a/LEYtC7DMor3N94PTPn1+0qq+Fk+iLjt1Z0WVCs=
```
cben, line 282: You may want to run `go mod tidy` to drop unused old lines

cben:
LGTM

go.sum
vkareh: good idea. done

PR #91, opened 2020-08-27
logs: Add progress indicator when waiting for logs

vkareh:
To avoid users being concerned about their cluster not starting install,
we add a spinner while we poll the installation logs endpoint. Once the
logs are available, we remove the spinner and show the logs instead.

https://asciinema.org/a/WoaLz9coGMc5quqKjTBDRPBJY

### DIFF
--- a/cmd/logs/install/cmd.go
+++ b/cmd/logs/install/cmd.go
@@ -20,7 +20,9 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
+	"github.com/briandowns/spinner"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 	errors "github.com/zgalor/weberr"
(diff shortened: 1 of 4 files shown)


jeremyeder:
Some feedback on the video:

Drop the "Logs for cluster ... are not available yet"

Replace "it usually takes up to 5 minutes" with "$product installation beginning.  installation logs will show up within 5 minutes"

Drop the word "Waiting" and just put the dots

Drop the "cluster installation is taking longer than usual".  just continue emitting dots

vkareh:
@jeremyeder thanks for the feedback. Done making the changes now and recorded a new video: https://asciinema.org/a/0jSlv8z7uDIkyRWav0qeEOwEl

cmd/logs/install/cmd.go
```diff
@@ -212,6 +209,8 @@ func printLog(logs *cmv1.Log) {
 	lines := findNextLines(logs)
 	if lines != "" {
 		fmt.Printf("%s\n", lines)
+	} else {
+		fmt.Printf(".")
```
jharrington22, line 213: I was thinking instead we could use something like .9 [here](https://godoc.org/github.com/briandowns/spinner) so we'd just have a spinner `... show up within 5 minutes  <spinner>`

Thought @vkareh @jeremyeder ?

vkareh: I like it - will play around with it and push a change

vkareh:
@jharrington22 @jeremyeder - I've added the spinner, and here's the new recording: https://asciinema.org/a/WoaLz9coGMc5quqKjTBDRPBJY

jharrington22:
Nice @vkareh

jeremyeder:
lgtm

PR #92, opened 2020-08-27
Use default region for CloudFormation stack

vkareh:
Since a CloudFormation stack is region-specific, we avoid using the
user-set flag and hard-code a default AWS region. This ensures that the
stack is queried always from the same region, regardless where the
cluster is being created.

PR #94, opened 2020-08-28
refactor(create): add credential check for osdCcsAdmin when cluster starts to be created

boranx:
Fixes: OSD-4707

The credential check has been added to cluster create command.
Example output when the creds for `osdCcsAdmin` are being changed manually

```shell
 └─ ▶./moactl create cluster -c boran-test
I: Validating AWS credentials for CFUser...
E: Error validating AWS credentials: Invalid CloudFormation stack credentials: osdCcsAdmin is not valid
you can recreate the CloudFormation stack with:
moactl init --delete-stack && moactl init

```

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -252,6 +252,25 @@ func run(cmd *cobra.Command, _ []string) {
 		reporter.Errorf("Error getting region: %v", err)
 		os.Exit(1)
 	}
+
+	// Create the AWS client:
+	client, err := aws.NewClient().
+		Logger(logger).
+		Region(aws.DefaultRegion).
+		Build()
+	if err != nil {
+		reporter.Errorf("Error creating AWS client: %v", err)
+		os.Exit(1)
+	}
+
+	// Validate AWS credentials for current user
+	reporter.Infof("Validating AWS credentials for CFUser...")
+	if err = client.ValidateCFUserCredentials(); err != nil {
+		reporter.Errorf("Error validating AWS credentials: %v", err)
+		os.Exit(1)
+	}
+	reporter.Infof("AWS credentials are valid!")
+
 	regionList, regionAZ, err := getRegionList(ocmClient, multiAZ)
 	if err != nil {
 		reporter.Errorf(fmt.Sprintf("%s", err))
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -58,6 +58,7 @@ const (
 type Client interface {
 	GetRegion() string
 	ValidateCredentials() (bool, error)
+	ValidateCFUserCredentials() error
 	EnsureOsdCcsAdminUser(stackName string) (bool, error)
 	DeleteOsdCcsAdminUser(stackName string) error
 	GetAccessKeyFromStack(stackName string) (*AccessKey, error)
@@ -238,6 +239,45 @@ func (c *awsClient) ValidateCredentials() (bool, error) {
 		return false, err
 	}
 	return true, nil
+}
+
+// ValidateCFUserCredentials checks if CF-IAM credentials are valid.
+// it gets stack's key and actual key and compares them
+// to get the stack credentials:
+// aws cloudformation describe-stack-resource \
+// --logical-resource-id osdCcsAdminAccessKeys --stack-name osdCcsAdminIAMUser
+func (c *awsClient) ValidateCFUserCredentials() error {
+	name := AdminUserName
+	accessKeyInput := &iam.ListAccessKeysInput{
+		UserName: &name,
+	}
+	accessKeyList, err := c.iamClient.ListAccessKeys(accessKeyInput)
+	if err != nil {
+		return err
+	}
+
+	OsdCcsAdminStackNamePtr := OsdCcsAdminStackName
+	LogicalResourceIDPtr := "osdCcsAdminAccessKeys"
+	stackResourceInput := &cloudformation.DescribeStackResourceInput{
+		StackName:         &OsdCcsAdminStackNamePtr,
+		LogicalResourceId: &LogicalResourceIDPtr,
+	}
+	resources, err := c.cfClient.DescribeStackResource(stackResourceInput)
+	if err != nil {
+		return err
+	}
+	cfAccessKey := resources.StackResourceDetail.PhysicalResourceId
+
+	for _, key := range accessKeyList.AccessKeyMetadata {
+		if *key.AccessKeyId == *cfAccessKey && *key.Status == "Active" {
+			return nil
+		}
+	}
+
+	return fmt.Errorf(
+		"Invalid CloudFormation stack credentials: %s is not valid \n"+
+			"you can recreate the CloudFormation stack with: \n"+
+			"moactl init --delete-stack && moactl init \n", name)
 }
 
 // Ensure osdCcsAdmin IAM user is created
--- a/pkg/aws/client_test.go
+++ b/pkg/aws/client_test.go
@@ -3,6 +3,7 @@ package aws_test
 import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
+	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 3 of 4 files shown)


pkg/aws/client.go
```diff
@@ -235,6 +236,46 @@ func (c *awsClient) ValidateCredentials() (bool, error) {
...
+
+	meta := accessKeyList.AccessKeyMetadata
+	keyList := make([]string, len(meta))
+	for _, key := range meta {
+		keyList = append(keyList, *key.AccessKeyId)
+	}
+
+	OsdCcsAdminStackNamePtr := OsdCcsAdminStackName
+	LogicalResourceIDPtr := "osdCcsAdminAccessKeys"
+	stackResourceInput := &cloudformation.DescribeStackResourceInput{
```
boranx, line 260: Is there a better way to get the key?

pkg/aws/client.go
```diff
@@ -235,6 +236,46 @@ func (c *awsClient) ValidateCredentials() (bool, error) {
...
+		StackName:         &OsdCcsAdminStackNamePtr,
+		LogicalResourceId: &LogicalResourceIDPtr,
+	}
+	resources, err := c.cfClient.DescribeStackResource(stackResourceInput)
```
boranx, line 264: is there a better way to get stack's user access key?

cmd/create/cluster/cmd.go
```diff
@@ -252,6 +252,25 @@ func run(cmd *cobra.Command, _ []string) {
 		reporter.Errorf("Error getting region: %v", err)
 		os.Exit(1)
 	}
+
+	// Create the AWS client:
+	client, err := aws.NewClient().
+		Logger(logger).
+		Region(region).
```
vkareh, line 259: I think this region should be removed from this instantiation. The cloudformation template is applied to a specific region (most likely the default one in the `~/.aws/config` or `$AWS_REGION`) and using the one set for the cluster will just cause it to not be found. See https://github.com/openshift/moactl/pull/92 for more details.

pkg/aws/client.go
```diff
@@ -235,6 +236,46 @@ func (c *awsClient) ValidateCredentials() (bool, error) {
...
+	}
+	cfAccessKey := resources.StackResourceDetail.PhysicalResourceId
+
+	for _, item := range keyList {
+		if item == *cfAccessKey {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("credentials are not valid for %s \n you can recreate the CloudFormation stack with `moactl init --delete-stack`", name)
```
vkareh, line 276: The command to recreate would be something like `moactl init --delete-stack && moactl init`. Or maybe you can say something like "you can delete the CloudFormation stack with `moactl init --delete-stack` and create it again", or whatever, but I think the more important detail is the fact that if the user creates the CF stack in one region, then tries to read it from another, they will get an error. Not sure if there's a way to have a region-agnostic CF stack, or check all regions until we find the stack, or what - I'm still trying to figure out what to do there...

pkg/aws/client_test.go
```diff
@@ -38,6 +41,57 @@ var _ = Describe("Client", func() {
...
+
+				mockIamAPI.EXPECT().ListAccessKeys(gomock.Any()).Return(&iam.ListAccessKeysOutput{
+					AccessKeyMetadata: []*iam.AccessKeyMetadata{
+						{
+							AccessKeyId: &IAMAccessKey,
+						},
+					},
+				}, nil)
+			})
+			It("should return nil", func() {
```
georgettica, line 66: I would add that returning `nil` is expected behaviour:
```suggestion
			It("should finish successfully and return nil", func() {
```

pkg/aws/client_test.go
```diff
@@ -38,6 +41,57 @@ var _ = Describe("Client", func() {
...
+				err := client.ValidateCFUserCredentials()
+				Expect(err).To(BeNil())
+			})
+		})
+
+		Context("when the creds doesn't match", func() {
```
georgettica, line 72: that is me seeing something grammatically incorrect and being triggered 🤓 
```suggestion
		Context("when the credentials don't match", func() {
```

pkg/aws/client_test.go
```diff
@@ -38,6 +41,57 @@ var _ = Describe("Client", func() {
...
+					AccessKeyMetadata: []*iam.AccessKeyMetadata{
+						{
+							AccessKeyId: &oddIAMAccessKey,
+						},
+					},
+				}, nil)
+			})
+			It("should return err", func() {
+				err := client.ValidateCFUserCredentials()
+				Expect(err.Error()).Should(ContainSubstring("you can recreate the CloudFormation stack with `moactl init --delete-stack`"))
```
georgettica, line 90: IMO the error should not be such a big section of the error itself, maybe reorder it so the front of the error will a short description of it you can search for and the end will include the dynamic parts:

`
Invalid CloudFormation stack credentials: AdminUser %s is not valid, to recreate run "moactl init --delete-stack && moactl init"
`

this will make the error querying:
```suggestion
				Expect(err.Error()).Should(ContainSubstring("Invalid CloudFormation stack credentials"))
```

which is what the user will be searching anyway online if it was me

georgettica (commented):
got some more fixes in, none are deal breaker but these are my two cents

cmd/create/cluster/cmd.go
boranx: hi, @vkareh thanks for the review! appreciated.
Now the region section has been removed.

pkg/aws/client.go
boranx: I've changed the message body as requested. I think the permission related concerns are mostly related to OSD-4708 (I might be wrong but it currently exists in our backlog).

pkg/aws/client.go
```diff
@@ -235,6 +236,46 @@ func (c *awsClient) ValidateCredentials() (bool, error) {
...
+		LogicalResourceId: &LogicalResourceIDPtr,
+	}
+	resources, err := c.cfClient.DescribeStackResource(stackResourceInput)
+	if err != nil {
+		return err
+	}
+	cfAccessKey := resources.StackResourceDetail.PhysicalResourceId
+
+	for _, key := range accessKeyList.AccessKeyMetadata {
+		if *key.AccessKeyId == *cfAccessKey {
```
boranx, line 267: I guess we also need to check the accesskey to see if it is or isn't enabled.

boranx: Because this check will pass even if the credentials for `osdCcsAdmin` marked as inactive via console

pkg/aws/client.go
jharrington22: @boranx unfortunately no. When you create an AWS IAM access key the secret access key is not saved and you cannot query the API for it

cmd/create/cluster/cmd.go
```diff
@@ -252,6 +252,24 @@ func run(cmd *cobra.Command, _ []string) {
 		reporter.Errorf("Error getting region: %v", err)
 		os.Exit(1)
 	}
+
+	// Create the AWS client:
+	client, err := aws.NewClient().
+		Logger(logger).
```
vkareh, line 258: I have an open pull request https://github.com/openshift/moactl/pull/92 that will add a new `aws.DefaultRegion` (`us-east-1`) which will be used for all requests to the cloudformation API, that way we no longer run into the scenario where a user creates the CF stack in one region, then the client tries to query for the credentials from a non-existent stack in a different region.

Once that's merged, this will need to have an extra `Region(aws.DefaultRegion).` to ensure that it's compatible with the changes in https://github.com/openshift/moactl/pull/92

boranx: ack @vkareh we can add that once #92 has been merged

pkg/aws/client.go
```diff
@@ -235,6 +236,46 @@ func (c *awsClient) ValidateCredentials() (bool, error) {
...
+		}
+	}
+
+	return fmt.Errorf(`
+	Invalid CloudFormation stack credentials: %s is not valid 
+	you can recreate the CloudFormation stack with 
+	"moactl init --delete-stack && moactl init"
+	`, name)
+}
+
```
NautiluX, line 278: I think this will produce spaces in the error message output. See [here](https://github.com/openshift/moactl/blob/3428aad22789b4155fe34224619a9ed29017e0a5/pkg/aws/client.go#L290) how we solved that problem earlier.

NautiluX (changes requested):

pkg/aws/client.go
```diff
@@ -235,6 +236,46 @@ func (c *awsClient) ValidateCredentials() (bool, error) {
 	return true, nil
 }
 
+// Checks osdCcsAdmin IAM credentials are valid.
```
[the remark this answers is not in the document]
boranx: hey, @NautiluX, appreciate for the review... Yeah you're right, I've fixed that section

pkg/aws/client.go
boranx: thanks, fixed now

NautiluX (approved):

PR #98, opened 2020-09-01
Add 'Channel Group' attribute to 'moactl describe cluster'

nimrodshn:
cc: @vkareh @jeremyeder @igoihman  ptal

### DIFF
--- a/cmd/describe/cluster/cmd.go
+++ b/cmd/describe/cluster/cmd.go
@@ -136,16 +136,17 @@ func run(_ *cobra.Command, argv []string) {
 
 	// Print short cluster description:
 	fmt.Printf(""+
-		"Name:        %s\n"+
-		"ID:          %s\n"+
-		"External ID: %s\n"+
-		"AWS Account: %s\n"+
-		"API URL:     %s\n"+
-		"Console URL: %s\n"+
-		"Nodes:       Master: %d, Infra: %d, Compute: %d\n"+
-		"Region:      %s\n"+
-		"State:       %s\n"+
-		"Created:     %s\n",
+		"Name:          %s\n"+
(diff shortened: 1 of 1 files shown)


vkareh:
@nimrodshn - if we're showing the version's channel group, do we want to also show the version itself?

nimrodshn:
@vkareh  Umm `ocm-cli` doesnt show the `version` (see https://github.com/openshift-online/ocm-cli/pull/130) so... I guess we will settle for just the `Channel Group`? 

This is just to align with the `ocm-cli`.

cmd/describe/cluster/cmd.go
```diff
@@ -136,16 +136,17 @@ func run(_ *cobra.Command, argv []string) {
...
+		"Name:          %s\n"+
+		"ID:            %s\n"+
+		"External ID:   %s\n"+
+		"AWS Account:   %s\n"+
+		"API URL:       %s\n"+
+		"Console URL:   %s\n"+
+		"Nodes:         Master: %d, Infra: %d, Compute: %d\n"+
+		"Region:        %s\n"+
+		"State:         %s\n"+
+		"Channel Group: %s\n"+
```
vkareh, line 148: As per offline conversation, do we want to change this to `Channel:`?

nimrodshn: cc @jeremyeder 

In our API (and everywhere else) its call `channel_group` b/c its not **exactly** what OCP calls channel (which AFAIK is a link between "versions" in the upgrade graph) but rather an orthogonal concept loosely similar that was invented to partition versions.

vkareh:
This works for me, unless @jeremyeder wants to change the output text.

/lgtm

PR #101, opened 2020-09-09
create-cluster: Allow selective override of the marketplace AMI

vkareh:
Since the MOA AMIs are not yet live in the AWS Marketplace, only
allowlisted accounts with a valid subscription will be able to use them.
Using the new --use-marketplace-ami flag will tell OCM that it should
use the MOA image, rather than the default OSD one.


vkareh:
/hold

This is pending merge and deployment of https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/2146

jharrington22 (approved):
/lgtm

PR #102, opened 2020-09-10
versions: Allow querying for channel-groups

vkareh:
We keep the flag hidden for now, and we hard-code the default to
'stable' when creating clusters. Eventually we will want to expose this
to end-users and tell the OCM what channel group to use.


nimrodshn:
LGTM :+1:

PR #107, opened 2020-09-16
verify-quota: Avoid nil pointer dereferencing

vkareh:
When fetching quota for comparison, if the AWS response contains a nil
servicequota item, it crashes moactl.


jharrington22 (approved):
Ouch.

/lgtm

PR #109, opened 2020-09-17
OSD-4972

sulmer-redhat:


cmd/create/cluster/cmd.go
```diff
@@ -266,10 +265,10 @@ func run(cmd *cobra.Command, _ []string) {
 	// Validate AWS credentials for current user
 	reporter.Infof("Validating AWS credentials for CFUser...")
 	if err = client.ValidateCFUserCredentials(); err != nil {
-		reporter.Errorf("Error validating AWS credentials: %v", err)
+		reporter.Errorf("Error validating OSDCcsAdmin user AWS credentials: %v", err)
```
vkareh, line 268: The name is `osdCcsAdmin`, but really you should use the `aws.AdminUserName` const here:
```go
reporter.Errorf("Error validating AWS credentials for user '%s': %v", aws.AdminUserName, err)
```

cmd/describe/cmd.go
```diff
@@ -21,12 +21,14 @@ import (
...
 )
 
 var Cmd = &cobra.Command{
-	Use:   "describe RESOURCE [flags]",
-	Short: "Show details of a specific resource",
-	Long:  "Show details of a specific resource",
+	Use:              "describe RESOURCE [flags]",
+	Short:            "Show details of a specific resource",
+	Long:             "Show details of a specific resource",
+	PersistentPreRun: v.Validations,
```
vkareh, line 31: I'm not sure the `describe` command should do this prerun validation, as it just reads from OCM based on the configured AWS credentials (it doesn't actually interact with the AWS resources)

cmd/validations/validations.go
```diff
@@ -0,0 +1,33 @@
...
+		os.Exit(1)
+	}
+
+	reporter.Infof("Validating cloudformation stack is exist")
+	stackExist, err := client.CheckStackReadyOrNotExisting(aws.OsdCcsAdminStackName)
+	if !stackExist || err != nil {
+		reporter.Errorf("Cloudformation stack is not exist. Run `moactl init` first")
+		os.Exit(1)
+	}
+	reporter.Infof("cloudformation stack is valid!")
```
vkareh, line 32: I think these messages should be `Debugf`, otherwise every command will just output extra confusing output

cmd/edit/cmd.go
```diff
@@ -21,14 +21,16 @@ import (
...
 var Cmd = &cobra.Command{
-	Use:     "edit RESOURCE [flags]",
-	Aliases: []string{"update"},
-	Short:   "Edit a specific resource",
-	Long:    "Edit a specific resource",
+	Use:              "edit RESOURCE [flags]",
+	Aliases:          []string{"update"},
+	Short:            "Edit a specific resource",
+	Long:             "Edit a specific resource",
+	PersistentPreRun: v.Validations,
```
vkareh, line 33: This is probably not necessary in here, as it will show up for _all_ resources being editted. Also, this command needs an existing running cluster anyway, and if the cluster exists in OCM, we can be certain that the corresponding CF Stack also exists

cmd/dlt/cmd.go
```diff
@@ -23,15 +23,16 @@ import (
...
 var Cmd = &cobra.Command{
-	Use:     "delete RESOURCE [flags]",
-	Aliases: []string{"remove"},
-	Short:   "Delete a specific resource",
-	Long:    "Delete a specific resource",
+	Use:              "delete RESOURCE [flags]",
+	Aliases:          []string{"remove"},
+	Short:            "Delete a specific resource",
+	Long:             "Delete a specific resource",
+	PersistentPreRun: v.Validations,
```
vkareh, line 35: Same for `delete` as in `edit`, if the cluster exists, then the CF stack was already created and still exists

cmd/create/cmd.go
```diff
@@ -24,14 +24,16 @@ import (
...
 var Cmd = &cobra.Command{
-	Use:     "create RESOURCE [flags]",
-	Aliases: []string{"add"},
-	Short:   "Create a resource from stdin",
-	Long:    "Create a resource from stdin",
+	Use:              "create RESOURCE [flags]",
+	Aliases:          []string{"add"},
+	Short:            "Create a resource from stdin",
+	Long:             "Create a resource from stdin",
+	PersistentPreRun: v.Validations,
```
vkareh, line 36: This should probably live only under `create cluster`, putting it here makes it run for all resources (e.g. `moactl create user`) and these operations do not need to validate the stack, as we know it already exists (due to the existence of the cluster itself)

cmd/validations/validations.go
sulmer-redhat: Addressed and resolved

cmd/describe/cmd.go
sulmer-redhat: Addressed and resolved

cmd/edit/cmd.go
sulmer-redhat: Addressed and resolved

cmd/create/cmd.go
sulmer-redhat: Addressed and resolved

cmd/dlt/cmd.go
sulmer-redhat: Addressed and resolved

cmd/validations/validations.go
```diff
@@ -0,0 +1,33 @@
...
+	client, err := aws.NewClient().
+		Logger(logger).
+		Region(aws.DefaultRegion).
+		Build()
+	if err != nil {
+		reporter.Errorf("Error creating AWS client: %v", err)
+		os.Exit(1)
+	}
+
+	reporter.Infof("Validating cloudformation stack is exist")
```
vkareh, line 26: This should also be `Debugf`. Also, the message should probably read "Validating that cloudformation stack exists".

cmd/validations/validations.go
```diff
@@ -0,0 +1,33 @@
...
+	stackExist, err := client.CheckStackReadyOrNotExisting(aws.OsdCcsAdminStackName)
+	if !stackExist || err != nil {
+		reporter.Errorf("Cloudformation stack is not exist. Run `moactl init` first")
```
vkareh, line 29: message should be: "Cloudformation stack does not exist. Run `moactl init` first"

PR #112, opened 2020-09-17
versions: Expose channel-group

vkareh:
When listing versions or creating a cluster, we can now specify the
channel group to use, with a default of 'stable'. Interactive mode now
reads the channel group flag and displays the correct list when creating
the cluster.

https://asciinema.org/a/b4vW7EtlDp0vRBbljSbnnGwBa

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -57,6 +57,7 @@ var args struct {
 	clusterName        string
 	region             string
 	version            string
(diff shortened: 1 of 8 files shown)


vkareh:
@jeremyeder - we gate the channel groups directly in CS: https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/cmd/clusters-service/service/versions.go#L41-52

PR #116, opened 2020-09-24
verify-quota: Check for only 100 vCPU

vkareh:
To allow users to create a single cluster, we can decrease the quota
check for only 100 vCPU.


jeremyeder:
lgtm

PR #118, opened 2020-09-25
 Rotate osdCcsAdmin credentails on creation of each cluster

jharrington22:
Rotate IAM access keys for the IAM user osdCcsAdmin user each time we use the client to create a cluster. There is no need to permanently store these credentials since they are only used on create, the cluster uses a completely different set of IAM credentials provisioned by this user.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -294,23 +294,6 @@ func run(cmd *cobra.Command, _ []string) {
 		reporter.Errorf("Error getting region: %v", err)
 		os.Exit(1)
 	}
-	// Create the AWS client:
-	client, err := aws.NewClient().
-		Logger(logger).
-		Region(aws.DefaultRegion).
-		Build()
-	if err != nil {
-		reporter.Errorf("Error creating AWS client: %v", err)
-		os.Exit(1)
-	}
-
-	// Validate AWS credentials for current user
-	reporter.Debugf("Validating AWS credentials...")
-	if err = client.ValidateCFUserCredentials(); err != nil {
-		reporter.Errorf("Error validating AWS credentials for user '%s': %v", aws.AdminUserName, err)
-		os.Exit(1)
-	}
-	reporter.Debugf("AWS credentials are valid!")
 
 	regionList, regionAZ, err := regions.GetRegionList(ocmClient, multiAZ)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


vkareh:
I'm wondering what happens during `moactl edit cluster`, the credentials we used to create are stored in a secret in Hive, but will no longer be valid when calling the API...

pkg/ocm/regions/regions.go
```diff
@@ -20,38 +20,27 @@ import (
...
-		Logger(logger).
-		Region(aws.DefaultRegion).
-		Build()
-	if err != nil {
-		return nil, fmt.Errorf("Failed to create AWS client: %v", err)
-	}
+	// Retrieve AWS credentials from the local AWS user
+	// pass these to OCM to validate what regions are available
+	// in this AWS account
+	localAWSCredentials := credentials.NewEnvCredentials()
```
vkareh, line 33: Here we should grab the credentials from the client, which can also honor the `--profile` flag.

vkareh:
Apart from my comment about `GetRegions`, this looks fine to me

pkg/ocm/regions/regions.go
jharrington22: > Apart from my comment about `GetRegions`, this looks fine to me

Thanks will take a look!

pkg/ocm/regions/regions.go
```diff
@@ -20,18 +20,34 @@ import (
...
+
+	// Build AWS client and retrieve credentials
+	// This ensures we use the profile flag if passed to rosa
+	// Create the AWS client:
+	reporter := rprtr.CreateReporterOrExit()
+	logger := logging.CreateLoggerOrExit(reporter)
+
+	awsClient, err := aws.NewClient().
+		Logger(logger).
+		Region("us-east-1").
```
vkareh, line 43: /nitpick You can use `aws.DefaultRegion` here

cmd/create/cluster/cmd.go
```diff
@@ -517,6 +501,16 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
+	if args.dryRun {
+		reporter.Infof(
+			"Creating cluster '%s' should succeed. Run without the '--dry-run' flag to create the cluster.",
+			clusterName)
+		os.Exit(0)
+	}
```
vkareh, line 509: This block needs to be moved back down, it should only display if `CreateCluster` succeeded.

jharrington22:
/hold cancel

@vkareh comments addressed, thanks!

vkareh:
/lgtm

Thanks @jharrington22! :)

PR #127, opened 2020-09-30
Makefile: only download go-bindata when not available

vkareh:
Also, remove it from the repository module configuration, since we only
need the local binary


vkareh:
@jharrington22 - please take a look

jharrington22 (approved):
/lgtm

PR #128, opened 2020-09-30
Support dry-run parameter for 'init' and 'create cluster'

vkareh:
* create-cluster: Add --dry-run flag   
    With the dry-run flag we allow the customer to simulate creating
    a cluster. This runs through the entire process of validating, preflight
    and quota checks and returns with either success or error. The cluster
    is not persisted nor written to the OCM database.
* init: Simulate cluster creation
    To ensure that a customer has all necessary permissions and quota in
    their OCM account, we attempt creating a very basic cluster with the
    dry-run flag and warn the customer in case of failuer.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -47,6 +47,9 @@ var args struct {
 	// Watch logs during cluster installation
 	watch bool
 
+	// Simulate creating a cluster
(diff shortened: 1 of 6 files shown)


vkareh:
/hold

This needs the `dryRun` parameter in OCM deployed to prod

PR #135, opened 2020-10-08
init: Ensure osdCcsAdmin exists before attempting cluster dry-run

vkareh:
If the osdCcsAdmin user does not exist, the dry-run will always fail. We
should always ensure that the cloudformation stack is set before
attempting a cluster install.


igoihman (approved):
/lgtm

PR #136, opened 2020-10-08
cluster: Add flag to disable SCP checks

vkareh:
This flag indicates if cloud permission checks are disabled when
attempting installation of the cluster by setting mint credential mode
in the installer config.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -65,6 +65,9 @@ type Spec struct {
 
 	// Simulate creating a cluster but don't actually create it
 	DryRun *bool
+
+	// Disable SCP checks in the installer by setting credentials mode as mint
+	DisableSCPChecks *bool
 }
 
 func IsValidClusterKey(clusterKey string) bool {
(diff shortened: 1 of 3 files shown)


pkg/cluster/cluster.go
```diff
@@ -65,6 +65,9 @@ type Spec struct {
 
 	// Simulate creating a cluster but don't actually create it
 	DryRun *bool
+
+	// Disable SCP checks in the installer by setting credentials mode as mint
+	DisableSCPChecks *bool
```
zvikorn, line 70: Question - Will it be nicer if instead having a new field in the cluster, we set the given argument in CCS.DisableSCPChecks? Can we do that at all?

vkareh: We could, but it's probably unnecessary, this is a temporary cluster object only used to capture the command-line flags. What do you think?

zvikorn: Oh..I see. /lgtm then.

zvikorn: lgtm

PR #137, opened 2020-10-09
Fix idp name generation

NautiluX:
Fixes OSD-5454


vkareh:
This looks good, but now I'm wondering if we should auto-name the IDP as `{IDPType}-{Org,Team}-{index}` (e.g. `GitHub-AppSRE-1`) or something like that, as it's far more meaningful in the console login page than seeing `github-1`, `github-2`, etc. WDYT?

jharrington22 (approved):
/lgtm

jharrington22:
@NautiluX held for @vkareh comments.

georgettica:
if it counts, I love the idea of having an auto-name for the idp with the pervious format

NautiluX:
I can adjust the name generation @vkareh , but I'd prefer to do that in a separate PR, as it will be independent of the bug fix. Added some tests to it, please have another look @vkareh @jharrington22

cmd/create/idp/cmd.go
```diff
@@ -29,9 +29,14 @@ import (
 	"github.com/openshift/moactl/pkg/interactive"
 	"github.com/openshift/moactl/pkg/logging"
 	"github.com/openshift/moactl/pkg/ocm"
+	"github.com/openshift/moactl/pkg/reporter"
 	rprtr "github.com/openshift/moactl/pkg/reporter"
 )
 
+type IdentityProvider interface {
+	Name() string
+}
```
vkareh, line 38: Is this interface used for the tests/mocks? Otherwise it's probably unnecessary (that's just me nitpicking, definitely not a blocker)

georgettica: as I know @NautiluX, this is to make that string more meaningful.. the usage in the code does not require the interface

NautiluX: to be able to mock the `IdentityProvider` from the OCM lib used, it's necessary to decouple it, yes. I don't like that we have to convert the array because you can't automatically use a `[]SomeIf` as `[]*SomeType` (SomeType implements SomeIf) unfortunately. The interface itself is not a bad thing, but I don't like the conversion logic. Happy to change it if somebody has a better idea :)

georgettica: Maybe create a type alias?
type IdentityProvider string

See example https://github.com/openshift/moactl/blob/45d6607f2ef01db3b1ce9ce6c0631832b644c1ca/OWNERS#L6

vkareh: No, this is fine. Let's leave it as is for now and get this bugfix merged.

PR #144, opened 2020-10-12
refactor(init): add permissions check for osdccsadmin after creation

boranx:
Fixes: [OSD-4708](https://issues.redhat.com/browse/OSD-4708)


pkg/aws/client_test.go
```diff
@@ -41,6 +41,71 @@ var _ = Describe("Client", func() {
...
 	})
 
+	Context("EnsureOsdCcsAdminUserPermissions", func() {
+		var (
+			policyArn   = "sut"
+			testPolicy  = "sut"
+			adminPolicy = "AdministratorAccess"
+			adminArn    = "AdministratorAccess"
+		)
+		Context("when listAttachedUserPolicies() returns multiple policy", func() {
```
georgettica, line 51:
```suggestion
		Context("when listAttachedUserPolicies() returns multiple policies", func() {
```

pkg/aws/client_test.go
```diff
@@ -41,6 +41,71 @@ var _ = Describe("Client", func() {
...
+							PolicyName: &adminPolicy,
+							PolicyArn:  &adminArn,
+						},
+					},
+				}, nil)
+			})
+			It("should fail as only Administrator policy needs to be attached", func() {
+				err := client.EnsureOsdCcsAdminUserPermissions()
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).Should(ContainSubstring("only AdministratorAccess needs to be attached"))
```
georgettica, line 69: maybe change to https://onsi.github.io/gomega/#haveprefixprefix-string-args-interfacee

georgettica (commented):
small nits, code looks good altogether but not sure about logic

pkg/aws/client.go
```diff
@@ -289,8 +290,38 @@ func (c *awsClient) ValidateCFUserCredentials() error {
 
 	return fmt.Errorf(
 		"Invalid CloudFormation stack credentials: %s is not valid \n"+
-			"you can recreate the CloudFormation stack with: \n"+
-			"moactl init --delete-stack && moactl init \n", name)
+			adviseRecreate(), name)
+}
+
+func (c *awsClient) EnsureOsdCcsAdminUserPermissions() error {
+	AdminUserNamePtr := AdminUserName
```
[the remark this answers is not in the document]
boranx: `AdminUserName` is actually a const and can't be used as a pointer so that another object needs to be created :)) you can have a look at the rest of the file and will come across several definements like that.. Is there a better way of getting rid of it?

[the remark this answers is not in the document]
boranx: I've changed it to `adminUserName` now, thanks!

pkg/aws/client_test.go
boranx: fixed, thanks!

pkg/aws/client_test.go
boranx: I've replaced it with `HavePrefix` now

NautiluX (commented):
/lgtm

NautiluX:
/assign @jharrington22 
@jharrington22 @vkareh can we get some eyes on this please?

jharrington22:
@boranx This PR validates that the `osdCcsAdmin` user has the `AdministratorAccess` policy attached as a managed policy, it doesn't validate that the user isn't part of a group that has the `AdministratorAccess` policy attached or that the policy isn't inline.

Using the AWS policy simulator you can `simulate` all the required OCD CCS permissions (see https://www.openshift.com/dedicated/ccs) against a AWS session. The `rosa verify permissions` CMD does that, I'd expect that we use that to validate that the `osdCcsAdmin` user has the permissions we require.

The policy simulator will also validate that there are no AWS SCP policies blocking any permissions we might need. AWS SCP policies allow administrators of an AWS Organization to block IAM users from certian permissions, even if they have "AdministratorAccess" with the action `*:*`

boranx:
hey @jharrington22 
Thanks for the clarification. I'll close that PR and will open a fresh one which will have the permissions be checked in `verify permissions` way

PR #145, opened 2020-10-12
ocm: Keep error opIds and codes behind debug flag

vkareh:
Now errors only output the relevant message to the user

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -79,7 +79,7 @@ func HasClusters(client *cmv1.ClustersClient, creatorARN string) (bool, error) {
 		Size(1).
 		Send()
 	if err != nil {
-		return false, fmt.Errorf("Failed to list clusters: %v", err)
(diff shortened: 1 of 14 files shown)


pkg/cluster/cluster.go
```diff
@@ -79,7 +79,7 @@ func HasClusters(client *cmv1.ClustersClient, creatorARN string) (bool, error) {
 		Size(1).
 		Send()
 	if err != nil {
-		return false, fmt.Errorf("Failed to list clusters: %v", err)
+		return false, fmt.Errorf(response.Error().Reason())
```
igoihman, line 82: missing `eporter.Debugf` call?

vkareh: This is part of the clusterprovider, and doesn't have access to the reporter. Reporter errors happen at the `cmd` level (so whichever command calls this code should already have a call to `reporter.Debugf` inside their err-catch

PR #151, opened 2020-10-19
Osd 5227

sulmer-redhat:


pkg/aws/client.go
```diff
@@ -344,18 +351,33 @@ func (c *awsClient) CheckStackReadyOrNotExisting(stackName string) (stackReady b
...
 
+func (c *awsClient) CheckAdminUserNotExisting(userName string) (err error) {
+	userList, err := c.iamClient.ListUsers(&iam.ListUsersInput{})
+	if err != nil {
+		return err
+	}
+	for _, user := range userList.Users {
+		if *user.UserName == userName {
+			return fmt.Errorf("Error creating user: IAM user '%s' already exists."+
+				"Ensure user osdCcsAdmin IAM user does not exist, then retry with\n"+
```
vkareh, line 373: Here you should change `osdCcsAdmin` to `'%s'` and inject `userName`.

pkg/aws/client.go
```diff
@@ -344,18 +351,33 @@ func (c *awsClient) CheckStackReadyOrNotExisting(stackName string) (stackReady b
 				return true, nil
 			}
 			if *summary.StackStatus != cloudformation.StackStatusDeleteComplete {
-				return false, fmt.Errorf("Error creating user: Cloudformation stack %s exists with status %s. "+
+				return false, fmt.Errorf("Error creating CloudFormation Stack: Cloudformation stack %s exists with status %s. "+
 					"Expected status is %s.\n"+
-					"Ensure user osdCcsAdmin does not exist, then retry with\n"+
+					"Ensure user osdCcsAdmin CloudFormation Stack does not exist, then retry with\n"+
```
vkareh, line 356: Here you should change `osdCcsAdmin` to `'%s'` and inject `stackName`, as the stack's name is different. Also drop the term `user`.

pkg/aws/client.go
```diff
@@ -344,18 +351,33 @@ func (c *awsClient) CheckStackReadyOrNotExisting(stackName string) (stackReady b
...
+func (c *awsClient) CheckAdminUserNotExisting(userName string) (err error) {
+	userList, err := c.iamClient.ListUsers(&iam.ListUsersInput{})
+	if err != nil {
+		return err
+	}
+	for _, user := range userList.Users {
+		if *user.UserName == userName {
+			return fmt.Errorf("Error creating user: IAM user '%s' already exists."+
+				"Ensure user osdCcsAdmin IAM user does not exist, then retry with\n"+
+				"moactl init --delete-stack; moactl init",
```
vkareh, line 374: I don't think this function should advise the user to delete the stack. If we determine that the user exists, and that the stack is ready (we get this from the previous function call) then the user existing should return success

templates/cloudformation/iam_user_osdCcsAdmin.json
```diff
@@ -6,7 +6,8 @@
         "ManagedPolicyArns": [
           "arn:aws:iam::aws:policy/AdministratorAccess"
         ],
-        "UserName": "osdCcsAdmin"
+        "UserName": "osdCcsAdmin",
+        "Tags" : [{"Key": "MOA", "Value": "true"}]
```
vkareh, line 10: Since you're tagging the user, might as well check for the tag to verify that the `osdCcsAdmin` user was created by `moactl init`

vkareh (changes requested):
I think this breaks the `ensureFoo` pattern. Running `moactl init` should be idempotent, and so we should not tell the user to delete the stack and retry if everything is fine.

templates/cloudformation/iam_user_osdCcsAdmin.json
sulmer-redhat: This is the plan, however right now the tagging is not working.
I filed an issue on AWS GO SDK repository: 
https://github.com/aws/aws-sdk-go/issues/3594

pkg/aws/client.go
sulmer-redhat: If the CF stack is ready then this function shouldn't be called. 
This function will only be called prior to creating a new stack, hence the need to verify there is no IAM user with this name.

templates/cloudformation/iam_user_osdCcsAdmin.json
jharrington22: @sulmer-redhat so this tag property isn't working?

sulmer-redhat: It is not working as of right now. should I remove it?

pkg/aws/client.go
sulmer-redhat: I removed the part that asks for `moactl init --delete-stack` from the error.

NautiluX:
@vkareh @jharrington22 can you have another look at this?

vkareh (approved):
Okay, I finally understand this PR. If the osdCcsAdmin user exists _after_ verifying that the CF stack _doesn't_ exist, that's when it throws an error! This makes sense to me now :)

/lgtm

PR #154, opened 2020-10-19
errors: Fall back to full error message

vkareh:
If the user-friendly error Reason is not available, show the OCM
api-friendly error string to avoid empty errors.


jharrington22 (approved):
/lgtm

PR #158, opened 2020-10-20
idp: Always use interactive mode on unset required flags

vkareh:
For any required attributes, if the user has not set it as a flag in the
CLI, we fallback to displaying the interactive mode for it instead of
failing the request. This also means that for any attribute that the
user has explicitly set through a flag, the tool will not show it even
if the user requested interactive mode.

### DIFF
--- a/cmd/create/idp/cmd.go
+++ b/cmd/create/idp/cmd.go
@@ -336,7 +336,7 @@ func run(cmd *cobra.Command, _ []string) {
 
 	// Grab all the IDP information interactively if necessary
 	idpType := args.idpType
-	if interactive.Enabled() || idpType == "" {
+	if idpType == "" {
 		if idpType == "" {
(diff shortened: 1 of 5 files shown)


vkareh:
/hold

@wgordon17 - this should address https://issues.redhat.com/browse/SDA-2940, but I'm not entirely sure if this is the desired behavior. Maybe we can have a chat to discuss the interactive flow.

wgordon17:
Ugh, sorry, my GoLang is a bit rusty 😉 

Based on your description above, it sounds like it matches my expectations...if you can build this for testing, I'll be happy to validate it locally :)

vkareh:
@wgordon17 - no rush, just let me know whenever you've had a chance to test this: [moactl.tar.gz](https://github.com/openshift/moactl/files/5424778/moactl.tar.gz)

PR #159, opened 2020-10-20
admin: Rename IDP to Cluster-Admin

vkareh:
This makes it more user-friendly when the user opens the console.


vkareh:
/hold

This depends on https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/2289

PR #177, opened 2020-11-02
Remove paid AMI flag and finalize ROSA transition

vkareh:
Since we're now moving paid AMIs to be a feature flag in Unleash, we
remove the attribute from here and let OCM make the decision based on
the user's organization.

We also finalize the transition from MOA to ROSA, since all dependencies
are now deployed.


vkareh:
/hold

Needs https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/2341 to be promoted to prod before merging.

jharrington22 (approved):
Looks good with the CS MR is ready.

vkareh:
@jharrington22 - any chance that we can add the `rosa` label to a production clusterimageset? It's blocking this change as well as the counterpart in CS

PR #180, opened 2020-11-05
machinepool: Allow managing 'default' machinepool

vkareh:
Since we have the machinery for managing machinepools, we can scale the
cluster itself using this resource and so remove the scaling options on
'edit cluster'.

### DIFF
--- a/cmd/dlt/machinepool/cmd.go
+++ b/cmd/dlt/machinepool/cmd.go
@@ -91,6 +91,11 @@ func run(_ *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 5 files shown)


jeremyeder:
Don't we have code in OCM enforcing the minimum nodes in the default worker pool?  That will continue to be enforced even if we allow the pool to be edited, right?

vkareh:
@jeremyeder correct:
```
$ rosa edit machinepool -c vkareh-rosa default --replicas=1
E: Failed to update machine pool 'default' on cluster 'vkareh-rosa': Invalid number of compute nodes: 1. Minimal requirement is 2 compute nodes.
```

PR #184, opened 2020-11-18
Add support for existing VPC

nimrodshn:
JIRA: https://issues.redhat.com/browse/SDA-3271

~~## Why is this WIP?~~

No longer WIP please take a look.

cc: @vkareh @igoihman


nimrodshn:
@vkareh Current CLI parameter for `AvailabilityZones` and `SubnetIDs` is using `StringSlice` which accepts params as comma separated string (e.g. "subnet-1, subnet2") splitting them in the background. This is the same for the interactive case where you're asked:

pkg/cluster/cluster.go
```diff
@@ -409,6 +409,23 @@ func createClusterSpec(config Spec, awsClient aws.Client) (*cmv1.Cluster, error)
 		}
 	}
 
+	if config.AvailabilityZones != nil {
+		clusterBuilder = clusterBuilder.Nodes(
+			cmv1.NewClusterNodes().AvailabilityZones(config.AvailabilityZones...),
```
vkareh, line 414: I think this might override the existing Nodes object already built a few blocks above.

vkareh (commented):
There are a few linting errors. Make sure you run `make lint` to catch those!

Since `rosa` is heavily integrated with the AWS SDK, and we have a session open already during cluster creation, it should be fairly easy to query for a list of availabilty zones and present those as a [MultiSelect](https://github.com/AlecAivazis/survey#multiselect) prompt. Having selected AZs, then we can query AWS for the subnet IDs and have a new  prompt for the user to select those as well

In case the user skips the AZ selection, then we can just skip subnet IDs entirely. What do you think?

pkg/cluster/cluster.go
nimrodshn: > I think this might override the existing Nodes object already built a few blocks above.

> Since rosa is heavily integrated with the AWS SDK, and we have a session open already during cluster creation, it should be fairly easy to query for a list of availabilty zones and present those as a MultiSelect prompt. Having selected AZs, then we can query AWS for the subnet IDs and have a new prompt for the user to select those as well

I guess we can do that - the main drawback is that **when** the user created her subnets aws has created them in _specific_ az's (we have preflight to verify that the subnets are in the az's the user supplied) so in fact its not really a choice for the user; they are simply filling in the az's from the aws console. I think that perhaps it could be easier UX if we do show them a list rather ask them to fill that in but am not sure - WDYT?

In that case we can also query for the subnets as well..

vkareh: Can we just show a list of subnets for the `region` and let the tool auto-fill the AZs without user interaction?

nimrodshn: Ummm no ..

vkareh: You can do something like:
```
subnetList = aws.ec2client.DescribeSubnets() // fetch subnets from AWS
subnets = interactive.GetMultiOptions(subnetList) // let user select subnets
foreach subnet in subnets {
  foreach subnetObj in subnetList {
    if subnet == subnetObj.id {
      az = append(az, subnetObj.AvailabilityZone) // create an internal list of matching AZs
    }
  }
}
createCluster({subnets: subnets, availabilityZones: az}) // send to cluster creation
```

vkareh: > Ummm no ..

Why not? Do we need the user to select the AZs explicitly?

vkareh: Alternatively we can do something like:
```
azList = aws.ec2client.DescribeAvailabilityZones(region)
azs = interactive.GetMultiOptions(azList)

subnetList = aws.ec2client.DescribeSubnets(region, azs)
subnets = interactive.GetMultiOptions(subnetList)

createCluster(subnets, azs)
```
that way the user selects AZs first and we show the list of subnets from the AZs they've already selected

nimrodshn: @vkareh I looked that thsubnet list and I think one can even do better
```
subnetList = aws.ec2client.DescribeSubnets(region, azs)
subnets = interactive.GetMultiOptions(subnetList)
for subnet := range subnets.Items {
    availabilityZones = append(availabilityZones, aws.String(subnet.AvailabilityZone))
}

createCluster(subnets, availabilityZones)
```

nimrodshn: Oh its essentially youre first options :facepalm:

vkareh: haha we're on the same page! good! :)

nimrodshn: @vkareh Using a `map` I was able to optimize the code a little bit, please take a look.

pkg/cluster/cluster.go
```diff
@@ -409,6 +409,23 @@ func createClusterSpec(config Spec, awsClient aws.Client) (*cmv1.Cluster, error)
 		}
 	}
 
+	if config.AvailabilityZones != nil {
+		clusterBuilder = clusterBuilder.Nodes(
```
vkareh, line 413: We already set `clusterBuilder.Nodes` above (line 375), I think you should add this code inside the `clusterNodesBuilder` code:
```
if config.ComputeMachineType != "" || config.ComputeNodes != 0 || if config.AvailabilityZones != nil {
  clusterNodesBuilder := cmv1.NewClusterNodes()
  // etc...
}

File: pkg/interactive/interactive.go [hunk 3]
```diff
@@ -106,6 +108,31 @@ func parseInt(str string) (num int, err error) {
...
 }
 
+// Asks for multiple options selection
+func GetMultipleOptions(input Input) (res []string, err error) {
+	dflt, ok := input.Default.(string)
+	if !ok {
+		dflt = ""
+	}
+	question := input.Question
+	if !input.Required {
```
vkareh, line 118: This should be `if !input.Required && dflt == "" {`, otherwise it shouls the prompt with both "(optional)" and the pre-populated value, which has caused confusion in the past.

pkg/interactive/interactive.go
```diff
@@ -278,3 +305,13 @@ func PrintHelp(help Help) error {
 	fmt.Fprint(terminal.NewAnsiStdout(os.Stdout), out)
 	return nil
 }
+
+// Trims all white spaces from a given string
+func TrimWhitSpaces(str string) string {
```
vkareh, line 310: I don't think you're using this any more, we should probably remove it

go.sum
```diff
@@ -281,6 +281,8 @@ github.com/onsi/gomega v1.7.0/go.mod h1:ex+gbHU/CVuBBDIJjb2X0qEXbFg53c61hWP/1Cpa
 github.com/op/go-logging v0.0.0-20160315200505-970db520ece7/go.mod h1:HzydrMdWErDVzsI23lYNej1Htcns9BCg93Dk0bBINWk=
 github.com/openshift-online/ocm-sdk-go v0.1.142 h1:wt+8DnLFUlqETZ4eskpDH/c4jxr9x3p0NPGXCUMpdG4=
 github.com/openshift-online/ocm-sdk-go v0.1.142/go.mod h1:3i3a/LEYtC7DMor3N94PTPn1+0qq+Fk+iLjt1Z0WVCs=
```
vkareh, line 283: run `go mod tidy` to remove old versions of `ocm-sdk-go`

vkareh (commented):
I get this when running it:
```
~/g/s/g/o/moactl (⎇ pull/184) $ rosa create cluster -i
I: Interactive mode enabled.
? Cluster name: foo
? Multiple availability zones (optional): No
? AWS region: us-east-1
? OpenShift version (optional): 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xf2f23a]

goroutine 1 [running]:
github.com/openshift/moactl/pkg/aws.(*awsClient).GetSubnetIDs(0xc000125b00, 0x15b8700, 0xc000125b00, 0x0, 0x0, 0x0)
        /home/vkareh/go/src/github.com/openshift/moactl/pkg/aws/client.go:270 +0x3a
github.com/openshift/moactl/cmd/create/cluster.run(0x1e10020, 0xc0003314f0, 0x0, 0x1)
        /home/vkareh/go/src/github.com/openshift/moactl/cmd/create/cluster/cmd.go:385 +0x919
github.com/spf13/cobra.(*Command).execute(0x1e10020, 0xc0003314e0, 0x1, 0x1, 0x1e10020, 0xc0003314e0)
	/home/vkareh/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:846 +0x2c2
github.com/spf13/cobra.(*Command).ExecuteC(0x1e178e0, 0x1402ef2, 0x35, 0x12)
	/home/vkareh/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
	/home/vkareh/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887
main.main()
	/home/vkareh/go/src/github.com/openshift/moactl/cmd/rosa/main.go:87 +0x92
```

Also linting fails currently

vkareh:
I know. You're missing this:
```diff
diff --git a/pkg/aws/client.go b/pkg/aws/client.go
index dea6335..aa2ce23 100644
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -238,6 +238,7 @@ func (b *ClientBuilder) Build() (Client, error) {
        c := &awsClient{
                logger:              b.logger,
                iamClient:           iam.New(sess),
+               ec2Client:           ec2.New(sess),
                orgClient:           organizations.New(sess),
                stsClient:           sts.New(sess),
                cfClient:            cloudformation.New(sess),

```

vkareh:
Looking good so far!
```
~/g/s/g/o/moactl (⎇ pull/184 *) $ rosa create cluster -i --dry-run 
I: Interactive mode enabled.
? Cluster name: foo
? Multiple availability zones (optional): No
? AWS region: us-east-1
? OpenShift version (optional): 
? Subnet IDs (optional): subnet-050b5a070768d2737, subnet-074572b18f5698f1f, subnet-0f7041e8bbfcb605c, subnet-0ba3f0ead1f486510, subnet-04a0a9a3def2d7701, subnet-0c7b531f9375931b1
? Compute nodes instance type (optional): 
? Compute nodes: 2
? Machine CIDR (optional): 
? Service CIDR (optional): 
? Pod CIDR (optional): 
? Host prefix (optional): 
? Private cluster (optional): No
I: Creating cluster 'foo'
I: To view a list of clusters and their status, run 'rosa list clusters'
E: Creating cluster 'foo' should fail: Only a single availability zone can be provided to a single-availability-zone cluster, instead received 6
```
However, that last error makes me think that we should make some UX changes here:
* If the user selects single AZ, maybe we should display an AZ _single select_, then we can show subnets _just_ for that AZ
* If the user selects multi AZ, we could either:
  - display a multi-select for AZs, then subnets for those selected, or
  - display a multi-select for subnets, with the AZ in parenthesis:
```
? Subnet IDs (optional):  [Use arrows to move, space to select, <right> to all, <left> to none, type to filter, ? for more help]
> [ ]  subnet-050b5a070768d2737 (us-east-1f)
  [ ]  subnet-074572b18f5698f1f (us-east-1a)
  [ ]  subnet-0f7041e8bbfcb605c (us-east-1a)
  [ ]  subnet-0ba3f0ead1f486510 (us-east-1a)
  [ ]  subnet-04a0a9a3def2d7701 (us-east-1a)
  [ ]  subnet-0c7b531f9375931b1 (us-east-1f)
```
What do you think?

vkareh:
Also, you need to add
```
mockgen -package mocks -destination=pkg/aws/mocks/ec2api.go github.com/aws/aws-sdk-go/service/ec2/ec2iface EC2API
```
in the Makefile and run `make mocks` to regenerate the mock files

nimrodshn:
@vkareh I dont think we should invest too much in a multi / single az complex selection logic - I think that if the user selected multi-az they should pick the correct number of subnets - this is essentially derived **from** that choice.

I chose you're third suggestion - showing the az's in the subnet selection - I think it make the most sense.

cmd/create/cluster/cmd.go
```diff
@@ -355,6 +373,55 @@ func run(cmd *cobra.Command, _ []string) {
...
+		options[i] = setSubnetOption(subnetID, availabilityZone)
+		mapSubnetToAZ[subnetID] = availabilityZone
+		mapAZCreated[availabilityZone] = false
+	}
+	if interactive.Enabled() {
+		subnetIDs, err = interactive.GetMultipleOptions(interactive.Input{
+			Question: "Subnet IDs",
+			Help:     cmd.Flags().Lookup("subnet-ids").Usage,
+			Required: false,
+			Options:  options,
```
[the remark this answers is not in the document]
vkareh: Also, we should have a check if the user selected mult-az. If they did, and `len(az)` is less than 3, then there is no way for them to select subnets without the request failing. We need something like
```go
if interactive && (!multiAZ || len(az) >= 3) {
  // ...
}
```
Does that make sense?

vkareh (commented):
This is looking great!! I found a case where the application crashes, please take a look at my review comments.

Also make sure you run `make docs` to update the markdown files with documentation

cmd/create/cluster/cmd.go
```diff
@@ -355,6 +373,55 @@ func run(cmd *cobra.Command, _ []string) {
...
+	options := make([]string, len(subnets))
+	for i, subnet := range subnets {
+		subnetID := awssdk.StringValue(subnet.SubnetId)
+		availabilityZone := awssdk.StringValue(subnet.AvailabilityZone)
+
+		// Create the options to prompt the user.
+		options[i] = setSubnetOption(subnetID, availabilityZone)
+		mapSubnetToAZ[subnetID] = availabilityZone
+		mapAZCreated[availabilityZone] = false
+	}
```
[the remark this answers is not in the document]
nimrodshn: Only the options need to be inside the block. I will move it

cmd/create/cluster/cmd.go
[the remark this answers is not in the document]
nimrodshn: @vkareh Yes.  Fixed.

nimrodshn:
@vkareh please re-tal

cmd/create/cluster/cmd.go
```diff
@@ -355,6 +373,55 @@ func run(cmd *cobra.Command, _ []string) {
...
+	if interactive.Enabled() && (!multiAZ || len(mapAZCreated) >= 3) {
```
vkareh, line 400: I think we still need to add `len(options) > 0` to this condition check

nimrodshn: Done.

vkareh:
This is great stuff, thanks a bunch @nimrodshn!

/lgtm

vkareh:
great, now we have a conflict :(

Can you please rebase?

nimrodshn:
@vkareh done.

vkareh:
/lgtm

Thanks @nimrodshn :+1:

PR #186, opened 2020-11-20
machinepool: Fix interactive mode

vkareh:
When creating a new machinepool, we now add a default instance type
(m5.xlarge) and drop the user into interactive mode if there is missing
required data.

Also provide faster validation on the number of compute nodes when
editing the default machine pool.


nimrodshn:
LGTM :+1:

PR #187, opened 2020-11-24
Return clear error message when no version is found

nimrodshn:
When a user provides some random channel-group (or one that she isn't authorized to see) she receives the following error:
```
[nshneor@localhost moactl]$ ./rosa create cluster --channel-group=foo
I: Enabling interactive mode
? Cluster name: foo
? Multiple availability zones (optional): No
? AWS region: us-west-1
E: Expected a valid OpenShift version: please provide options to select from
```

This patch fixes this behavior by returning a more clear error message for this scenario:
```
...
E: Failed to find versions for the provided channel-group: 'foo'.
```

JIRA: https://issues.redhat.com/browse/SDA-2968

cc: @vkareh @igoihman

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -655,6 +655,11 @@ func getVersionList(client *cmv1.Client, channelGroup string) (versionList []str
 		versionList = append(versionList, strings.Replace(v.ID(), "openshift-v", "", 1))
 	}
 
+	if len(versionList) == 0 {
+		err = fmt.Errorf("Could not find versions for the provided channel-group: '%s'", channelGroup)
+		return
+	}
+
 	return
 }
 


cmd/create/cluster/cmd.go
```diff
@@ -655,6 +655,10 @@ func getVersionList(client *cmv1.Client, channelGroup string) (versionList []str
 		versionList = append(versionList, strings.Replace(v.ID(), "openshift-v", "", 1))
 	}
 
+	if len(versionList) == 0 {
+		err = fmt.Errorf("Failed to find versions for the provided channel-group: '%s'.", channelGroup)
```
vkareh, line 659: /nitpick
I would add a `return` inside this block just to be consistent. If someone else decides to add more code after this block, they might cause an unintended bug

nimrodshn: Done. thx!

cmd/create/cluster/cmd.go
```diff
@@ -655,6 +655,11 @@ func getVersionList(client *cmv1.Client, channelGroup string) (versionList []str
...
```
vkareh, line 659: This has a linting error:
```
cmd/create/cluster/cmd.go:659:20: error strings should not be capitalized or end with punctuation or a newline (golint)
		err = fmt.Errorf("Failed to find versions for the provided channel-group: '%s'.", channelGroup)
		                 ^
make: *** [Makefile:45: lint] Error 1
```

nimrodshn: Yeah I see - but the error just above it is capitalized :sad

cmd/create/cluster/cmd.go
```diff
@@ -655,6 +655,11 @@ func getVersionList(client *cmv1.Client, channelGroup string) (versionList []str
 		versionList = append(versionList, strings.Replace(v.ID(), "openshift-v", "", 1))
 	}
 
+	if len(versionList) == 0 {
+		err = fmt.Errorf("could not find versions for the provided channel-group: '%s'", channelGroup)
```
vkareh, line 659: You can keep it capitalized, just remove the `.` at the end of the sentence:
```
err = fmt.Errorf("Failed to find versions for the provided channel-group: '%s'", channelGroup)
```

PR #189, opened 2020-11-24
create: Ask user before showing subnets

vkareh:
To install the cluster in an existing VPC should be an explicit option
by the user. Showing the list of subnets without asking can cause
confusion as to the required cluster configuration. Here we add a simple
confirmation asking the user whether they want to use an existing VPC.
The list of subnets is only shown if the answer Yes.

https://asciinema.org/a/Wv0xh61TD0TE7bjSb0458Izqp


nimrodshn:
@vkareh LGTM :100: 
Thanks a lot!

PR #191, opened 2020-11-29
add taints to machinepool commands

igoihman:


igoihman:
@vkareh PTAL

igoihman:
@vkareh PTAL

vkareh:
This LGTM. Can you run `make docs` as well?

PR #192, opened 2020-11-30
[rosa create cluster] Verify provided subnets for Existing VPC exist in AWS

nimrodshn:
CC: @vkareh  @igoihman

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -418,7 +418,23 @@ func run(cmd *cobra.Command, _ []string) {
 		mapSubnetToAZ := make(map[string]string)
 		mapAZCreated := make(map[string]bool)
 		options := make([]string, len(subnets))
-		optionDefaults := make([]string, len(subnetIDs))
+		defaultOptions := make([]string, len(subnetIDs))
+
+		// Verify subnets provided exist.
+		if subnetsProvided {
+			for _, subnetArg := range subnetIDs {
+				verifiedSubnet := false
+				for _, subnet := range subnets {
+					if awssdk.StringValue(subnet.SubnetId) == subnetArg {
+						verifiedSubnet = true
+					}
+				}
+				if !verifiedSubnet {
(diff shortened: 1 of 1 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -428,9 +444,7 @@ func run(cmd *cobra.Command, _ []string) {
 			options[i] = setSubnetOption(subnetID, availabilityZone)
 			if subnetsProvided {
 				for _, subnetArg := range subnetIDs {
-					if subnetArg == subnetID {
-						optionDefaults = append(optionDefaults, setSubnetOption(subnetID, availabilityZone))
-					}
+					defaultOptions = append(defaultOptions, setSubnetOption(subnetArg, availabilityZone))
```
vkareh, line 447: Doesn't this make it so that _all_ subnets are selected, regardless of which were sent in the CLI?

nimrodshn: @vkareh why? only those passed via the CLI (i.e. `subnetArg`) are selected..

vkareh: duh, makes sense :stuck_out_tongue_closed_eyes:

cmd/create/cluster/cmd.go
```diff
@@ -418,7 +418,23 @@ func run(cmd *cobra.Command, _ []string) {
...
+
+		// Verify subnets provided exist.
+		if subnetsProvided {
+			for _, subnetArg := range subnetIDs {
+				verifiedSubnet := false
+				for _, subnet := range subnets {
+					if awssdk.StringValue(subnet.SubnetId) == subnetArg {
+						verifiedSubnet = true
+					}
+				}
```
vkareh, line 431: Can we do this check inside the next loop? We already have the same loop doing
```
if subnetsProvided {
// ...
```
It's the exact same loop, let's reuse it!

nimrodshn: I wanted to error early..

PR #198, opened 2020-12-10
create-cluster: Set default version

vkareh:
If no version is set in the CLI, the interactive mode does not set
a proper default. This means that a user pressing [Enter] through the
prompts will end up _not_ selecting a version, even though it appears as
they do.


jharrington22 (approved):
/lgtm

PR #199, opened 2020-12-10
multi-az: Validate that compute nodes are multiple of 3

vkareh:
Instead of waiting for the API request to return a failure, we can
easily validate that the user requests compute nodes in multiples of
3 for multi-az clusters.


jharrington22 (approved):
/lgtm

PR #200, opened 2020-12-10
login: Hide 'env' parameter

vkareh:
Since end-users should not have access to our integration and staging
environments, there's no need to expose the '--env' parameter.


jharrington22 (approved):
/lgtm

PR #201, opened 2020-12-11
cluster: Show warnings when user makes cluster private

vkareh:
Since users will not be able to access private clusters unless they
configure their network settings in AWS, we warn them whenever creating
a private cluster (or editing a cluster to make it private).

We also display whether the cluster is private in the 'describe cluster'
output.

Demo: https://asciinema.org/a/EtUqEPWae07r2Fj8stIzJ9Wa0


pvasant:
lgtm

PR #202, opened 2020-12-14
replicas: Fix local validation for worker nodes and machinepool replicas

vkareh:
This change disallows entering negative numbers on any machine pool
replicas, and will validate the minimum required number of nodes based
on the cluster type and machine pool type.


jharrington22:
@vkareh looks good, just one comment.

vkareh:
@jharrington22 
> just one comment

What comment?

PR #204, opened 2020-12-18
login: Add link to retrieve tokens

thequailman:
As a user, it's unclear where I get a token when trying to use `rosa` programatically.

### DIFF
--- a/cmd/login/cmd.go
+++ b/cmd/login/cmd.go
@@ -55,7 +55,8 @@ var Cmd = &cobra.Command{
 		"\t3. Environment variable (OCM_TOKEN)\n"+
 		"\t4. Configuration file\n"+
 		"\t5. Command-line prompt\n", uiTokenPage),
-	Example: `  # Login to the OpenShift API with an existing token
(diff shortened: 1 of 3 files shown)


cmd/login/cmd.go
```diff
@@ -55,7 +55,8 @@ var Cmd = &cobra.Command{
 		"\t3. Environment variable (OCM_TOKEN)\n"+
 		"\t4. Configuration file\n"+
 		"\t5. Command-line prompt\n", uiTokenPage),
-	Example: `  # Login to the OpenShift API with an existing token
+	Example: "  # Login to the OpenShift API with an existing token generated from" +
```
vkareh, line 58: You should leave a space after `generated from`, but before the closing `"`, otherwise the docs output ends up like:
```
Examples:
  # Login to the OpenShift API with an existing token generated fromhttps://cloud.redhat.com/openshift/token/rosa
  rosa login --token=$OFFLINE_ACCESS_TOKEN
```
You can also run `make docs` and commit the result along with this.

vkareh:
ooh very nice! Thanks!

/lgtm

PR #205, opened 2020-12-28
Add autoscaling support

zgalor:
Add autoscaling to 
cluster create, cluster describe 
machinepool create list and edit

Related: https://issues.redhat.com/browse/SDA-3365

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -68,6 +68,9 @@ var args struct {
 	// Scaling options
 	computeMachineType string
 	computeNodes       int
+	autoscalingEnabled bool
+	minReplicas        int
+	maxReplicas        int
 
 	// Networking options
 	hostPrefix  int
@@ -167,6 +170,27 @@ func init() {
 		2,
 		"Number of worker nodes to provision per zone. Single zone clusters need at least 2 nodes, "+
 			"multizone clusters need at least 3 nodes.",
+	)
+
+	flags.BoolVar(
+		&args.autoscalingEnabled,
+		"enable-autoscaling",
+		false,
+		"Enable autoscaling of compute nodes.",
+	)
+
+	flags.IntVar(
+		&args.minReplicas,
+		"min-replicas",
+		0,
+		"Minimum number of compute nodes.",
+	)
+
+	flags.IntVar(
+		&args.maxReplicas,
+		"max-replicas",
+		0,
+		"Maximum number of compute nodes.",
 	)
 
 	flags.IPNetVar(
@@ -493,13 +517,53 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	// Autoscaling
+	autoscaling := args.autoscalingEnabled
+	if !autoscaling && interactive.Enabled() {
+		autoscaling, err = interactive.GetBool(interactive.Input{
+			Question: "Enable autoscaling",
+			Help:     cmd.Flags().Lookup("enable-autoscaling").Usage,
+			Default:  autoscaling,
+			Required: false,
+		})
+		if err != nil {
+			reporter.Errorf("Expected a valid value for enable-autoscaling: %s", err)
+			os.Exit(1)
+		}
+	}
+
+	minReplicas := args.minReplicas
+	maxReplicas := args.maxReplicas
+	if autoscaling && interactive.Enabled() {
+		minReplicas, err = interactive.GetInt(interactive.Input{
+			Question: "Min replicas",
+			Help:     cmd.Flags().Lookup("min-replicas").Usage,
+			Default:  minReplicas,
+			Required: true,
+		})
+		if err != nil {
+			reporter.Errorf("Expected a valid number of min replicas: %s", err)
+			os.Exit(1)
+		}
+		maxReplicas, err = interactive.GetInt(interactive.Input{
+			Question: "Max replicas",
+			Help:     cmd.Flags().Lookup("max-replicas").Usage,
+			Default:  maxReplicas,
+			Required: true,
+		})
+		if err != nil {
+			reporter.Errorf("Expected a valid number of max replicas: %s", err)
+			os.Exit(1)
+		}
+	}
+
 	// Compute nodes:
 	computeNodes := args.computeNodes
 	// Compute node requirements for multi-AZ clusters are higher
 	if multiAZ && !cmd.Flags().Changed("compute-nodes") {
 		computeNodes = 3
 	}
-	if interactive.Enabled() {
+	if !autoscaling && interactive.Enabled() {
 		computeNodes, err = interactive.GetInt(interactive.Input{
 			Question: "Compute nodes",
 			Help:     cmd.Flags().Lookup("compute-nodes").Usage,
@@ -621,6 +685,9 @@ func run(cmd *cobra.Command, _ []string) {
 		Expiration:         expiration,
 		ComputeMachineType: computeMachineType,
 		ComputeNodes:       computeNodes,
+		Autoscaling:        autoscaling,
+		MinReplicas:        minReplicas,
+		MaxReplicas:        maxReplicas,
 		MachineCIDR:        machineCIDR,
 		ServiceCIDR:        serviceCIDR,
 		PodCIDR:            podCIDR,
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -182,14 +182,14 @@ func init() {
 	flags.IntVar(
 		&args.minReplicas,
 		"min-replicas",
-		0,
+		2,
 		"Minimum number of compute nodes.",
 	)
 
 	flags.IntVar(
 		&args.maxReplicas,
 		"max-replicas",
-		0,
+		2,
 		"Maximum number of compute nodes.",
 	)
 
@@ -539,13 +539,20 @@ func run(cmd *cobra.Command, _ []string) {
 	minReplicas := args.minReplicas
 	maxReplicas := args.maxReplicas
 	if autoscaling {
-
 		// if the user set compute-nodes and enabled autoscaling
 		if cmd.Flags().Changed("compute-nodes") {
 			reporter.Errorf("Compute-nodes can't be set when autoscaling is enabled")
 			os.Exit(1)
 		}
 
+		if multiAZ {
+			if !isMinReplicasSet {
+				minReplicas = 3
+			}
+			if !isMaxReplicasSet {
+				maxReplicas = minReplicas
+			}
+		}
 		if interactive.Enabled() || !isMinReplicasSet {
 			minReplicas, err = interactive.GetInt(interactive.Input{
 				Question: "Min replicas",
@@ -571,7 +578,11 @@ func run(cmd *cobra.Command, _ []string) {
 			}
 		}
 
-		if minReplicas < 2 {
+		if multiAZ && minReplicas < 3 {
+			reporter.Errorf("Multi AZ cluster requires at least 3 compute nodes")
+			os.Exit(1)
+		}
+		if !multiAZ && minReplicas < 2 {
 			reporter.Errorf("Cluster requires at least 2 compute nodes")
 			os.Exit(1)
 		}
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -518,8 +518,9 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	// Autoscaling
+	isAutoscalingSet := cmd.Flags().Changed("enable-autoscaling")
 	autoscaling := args.autoscalingEnabled
-	if !autoscaling && interactive.Enabled() {
+	if !autoscaling && !isAutoscalingSet && interactive.Enabled() {
 		autoscaling, err = interactive.GetBool(interactive.Input{
 			Question: "Enable autoscaling",
 			Help:     cmd.Flags().Lookup("enable-autoscaling").Usage,
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -539,6 +539,13 @@ func run(cmd *cobra.Command, _ []string) {
 	minReplicas := args.minReplicas
 	maxReplicas := args.maxReplicas
 	if autoscaling {
+
+		// if the user set compute-nodes and enabled autoscaling
+		if cmd.Flags().Changed("compute-nodes") {
+			reporter.Errorf("Compute-nodes can't be set when autoscaling is enabled")
+			os.Exit(1)
+		}
+
 		if interactive.Enabled() || !isMinReplicasSet {
 			minReplicas, err = interactive.GetInt(interactive.Input{
 				Question: "Min replicas",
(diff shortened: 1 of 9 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -493,13 +517,53 @@ func run(cmd *cobra.Command, _ []string) {
...
+		})
+		if err != nil {
+			reporter.Errorf("Expected a valid value for enable-autoscaling: %s", err)
+			os.Exit(1)
+		}
+	}
+
+	minReplicas := args.minReplicas
+	maxReplicas := args.maxReplicas
+	if autoscaling && interactive.Enabled() {
```
igoihman, line 537: what if min/max replicas value was provided in the command line?

cmd/create/cluster/cmd.go
```diff
@@ -610,6 +674,9 @@ func run(cmd *cobra.Command, _ []string) {
 		Expiration:         expiration,
 		ComputeMachineType: computeMachineType,
 		ComputeNodes:       computeNodes,
+		Autoscaling:        autoscaling,
+		MinReplicas:        minReplicas,
```
igoihman, line 678: min/max replicas value will be set to 0 it wasn't provided. will the backend support it?

zgalor: 0 is illegal in the backend and considered "NOT SET"

cmd/create/cluster/cmd.go
```diff
@@ -493,13 +517,53 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	// Autoscaling
+	autoscaling := args.autoscalingEnabled
+	if !autoscaling && interactive.Enabled() {
```
vkareh, line 522: You should check for `!cmd.Flags().Changed("enable-autoscaling")` - if the user explicitly set `--enable-autoscaling=false` we probably shouldn't force the interactive question on them

cmd/create/cluster/cmd.go
vkareh: Since these are required, I would make it so that they are shown even if the interactive mode is not set. Something like
```go
if autoscaling && !cmd.Flags().Changed("min-replicas") {
...
}
```
Unless there's a default, in which case we can skip the interactive (unless the user sets `--interactive`) and just set the default.

cmd/create/cluster/cmd.go
```diff
@@ -493,13 +517,53 @@ func run(cmd *cobra.Command, _ []string) {
...
+		minReplicas, err = interactive.GetInt(interactive.Input{
+			Question: "Min replicas",
+			Help:     cmd.Flags().Lookup("min-replicas").Usage,
+			Default:  minReplicas,
+			Required: true,
+		})
+		if err != nil {
+			reporter.Errorf("Expected a valid number of min replicas: %s", err)
+			os.Exit(1)
+		}
```
vkareh, line 547: Since `0` is not a valid value, you should check
```go
if minReplicas <= 0 {
  reporter.Errorf(...)
  os.Exit(1)
}
```
Better to exit early rather than let the user go through the entire flow and still get an API failure

cmd/create/cluster/cmd.go
vkareh: The value should be validated before getting here. Also keep in mind that if the user does something like `rosa create cluaster --min-replicas=0`, they didn't set `--enable-autoscaling`, so these values here may be misleading (see above comment regarding forcing interactive mode on the user)

cmd/list/machinepool/cmd.go
```diff
@@ -153,6 +155,22 @@ func run(_ *cobra.Command, _ []string) {
 	writer.Flush()
 }
 
+func printAutoscaling(autoscaling *cmv1.MachinePoolAutoscaling) string {
+	if autoscaling != nil {
```
vkareh, line 159: Can this be an explicit `&false` or will it always be `nil` for a non-autoscaling machinepool?

cmd/create/machinepool/cmd.go
```diff
@@ -198,9 +222,49 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	// Autoscaling
+	autoscaling := args.autoscalingEnabled
+	if !autoscaling && interactive.Enabled() {
```
vkareh, line 227: Most of my thoughts on this file are similar to the `create cluster` one

cmd/create/cluster/cmd.go
zgalor: fixed

cmd/create/cluster/cmd.go
zgalor: Added a validation

cmd/create/cluster/cmd.go
zgalor: Fixed

cmd/list/machinepool/cmd.go
zgalor: it will always be nil

cmd/create/machinepool/cmd.go
zgalor: I added validations on both

cmd/create/cluster/cmd.go
zgalor: fixed

cmd/edit/machinepool/cmd.go
```diff
@@ -199,6 +220,7 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
+<<<<<<< HEAD
```
vkareh, line 223: Conflict marker - need to remove

cmd/edit/machinepool/cmd.go
```diff
@@ -210,13 +232,37 @@ func run(cmd *cobra.Command, argv []string) {
 	}
 	if cluster.MultiAZ() && replicas%3 != 0 {
 		reporter.Errorf("Multi AZ clusters require that the number of machine pool replicas be a multiple of 3")
+=======
```
vkareh, line 235: Conflict market - might want to check what's changed...

cmd/edit/machinepool/cmd.go
```diff
@@ -210,13 +232,37 @@ func run(cmd *cobra.Command, argv []string) {
...
+	autoscaling, replicas, minReplicas, maxReplicas := getReplicas(cmd, reporter, machinePoolID,
+		machinePool.Autoscaling())
+
+	if cluster.MultiAZ() &&
+		(!autoscaling && replicas%3 != 0 ||
+			(autoscaling && (minReplicas%3 != 0 || maxReplicas%3 != 0))) {
+		reporter.Errorf("Multi AZ clusters require that the number of MachinePool replicas be a multiple of 3")
+>>>>>>> ade85b9 (Handle edit and some other fixes)
```
vkareh, line 243: Another one here

cmd/edit/machinepool/cmd.go
zgalor: fixed

cmd/edit/machinepool/cmd.go
zgalor: fixed

cmd/create/cluster/cmd.go
```diff
@@ -493,37 +517,115 @@ func run(cmd *cobra.Command, _ []string) {
...
-		if computeNodes < 3 {
-			reporter.Errorf("The number of compute nodes needs to be at least 3")
+
+	isMinReplicasSet := cmd.Flags().Changed("min-replicas")
+	isMaxReplicasSet := cmd.Flags().Changed("max-replicas")
+
+	minReplicas := args.minReplicas
+	maxReplicas := args.maxReplicas
+	if autoscaling {
+
```
vkareh, line 542: The linter doesn't like this newline here...

cmd/edit/machinepool/cmd.go
```diff
@@ -238,16 +276,92 @@ func run(cmd *cobra.Command, argv []string) {
...
-	// Number of replicas:
-	replicas := args.replicas
-	if interactive.Enabled() || !cmd.Flags().Changed("replicas") {
-		return interactive.GetInt(interactive.Input{
+func getReplicas(cmd *cobra.Command,
+	reporter *rprtr.Object,
+	machinePoolID string,
+	existingAutoscaling *cmv1.MachinePoolAutoscaling) (autoscaling bool,
+	replicas, minReplicas, maxReplicas int) {
+
```
vkareh, line 284: The linter doesn't like this newline here...

docs/rosa_create_machinepool.md
```diff
@@ -19,6 +19,9 @@ rosa create machinepool [flags]
   # Add a machine pool mp-1 with 3 replicas of m5.xlarge to a cluster
   rosa create machinepool --cluster=mycluster --name=mp-1 --replicas=3 --instance-type=m5.xlarge
 
+  # Add a machine pool mp-1 with autoscaling enabled and 3 to 6 replicas of m5.xlarge to a cluster
+  rosa create machinepool --cluster=mycluster --name=mp-1 --enable-autoscaling --min-replicas=3 --max-replicas=6 --instance-type=m5.xlarge
```
vkareh, line 23: All these docs are auto-generated (`make docs`). You should put this inside the `Example` under each corresponding `&cobra.Command`

docs/rosa_create_machinepool.md
```diff
@@ -31,7 +34,10 @@ rosa create machinepool [flags]
       --instance-type string   Instance type that should be used. (default "m5.xlarge")
       --labels string          Labels for machine pool. Format should be a comma-separated list of 'key=value'. This list will overwrite any modifications made to Node labels on an ongoing basis.
       --name string            Name for the machine pool (required).
-      --replicas int           Count of machines for this machine pool (required).
+      --replicas int           Count of machines for this machine pool (required when autoscaling is disabled).
+      --enable-autoscaling     Enable autoscaling for the machine pool.
+      --min-replicas int       Minimun number  machines for this machine pool (required when autoscaling is enabled).
+      --max-replicas int       Maximun number  machines for this machine pool (required when autoscaling is enabled).
```
vkareh, line 40: Run `make docs` to automatically generate these

cmd/create/cluster/cmd.go
```diff
@@ -169,6 +172,27 @@ func init() {
...
+		&args.autoscalingEnabled,
+		"enable-autoscaling",
+		false,
+		"Enable autoscaling of compute nodes.",
+	)
+
+	flags.IntVar(
+		&args.minReplicas,
+		"min-replicas",
+		0,
```
vkareh, line 185: Maybe we should default to `2` here, otherwise since there's no other indication the user might enter 0 or 1. We can treat this in the same way as the compute nodes, where the hard-coded default is 2, but we update it to 3 in the code if they select multi-az. This reduces the possibility of input errors.

cmd/create/cluster/cmd.go
```diff
@@ -493,37 +517,115 @@ func run(cmd *cobra.Command, _ []string) {
...
+	// Compute nodes:
+	computeNodes := args.computeNodes
+	// Compute node requirements for multi-AZ clusters are higher
+	if multiAZ && !autoscaling && !cmd.Flags().Changed("compute-nodes") {
+		computeNodes = 3
+	}
+	if !autoscaling {
+		// if the user set min/max replicas and hasn't enabled autoscaling
+		if isMinReplicasSet || isMaxReplicasSet {
+			reporter.Errorf("Autoscaling must be enabled in order to set min and max replicas")
```
vkareh, line 599: In playing with this, if I do something like `rosa create cluster -c foo --min-replicas=2 --max-replicas=4` the user will probably expect the `--enable-autoscaling` to be set automatically, so maybe forcing interactive mode on the user to confirm that's what they want would be a good idea. What do you think?

cmd/create/cluster/cmd.go
```diff
@@ -493,37 +517,115 @@ func run(cmd *cobra.Command, _ []string) {
...
+				Default:  maxReplicas,
+				Required: true,
+			})
+			if err != nil {
+				reporter.Errorf("Expected a valid number of max replicas: %s", err)
+				os.Exit(1)
+			}
+		}
+
+		if minReplicas < 2 {
```
vkareh, line 574: This number should be dependent on the multi-az flag too

cmd/create/cluster/cmd.go
zgalor: I can, though edit behavior would be different.

If the machinepool already has autoscaling enabled and the user only wants to set min replicas - they should be able set 
`rosa edit machinepool foo -c bar --min-replicas=3` without setting `--enable-autoscaling`. However if autoscaling is disabled I don't want to turn to interactive because I want to user to expilicitly enable it.

cmd/create/cluster/cmd.go
zgalor: fixed

cmd/edit/machinepool/cmd.go
zgalor: fixed

docs/rosa_create_machinepool.md
zgalor: fixed

cmd/create/cluster/cmd.go
zgalor: Done

cmd/create/cluster/cmd.go
zgalor: Done

cmd/create/cluster/cmd.go
zgalor: On second thought, I don't want to go there.
If we start complicating validations - potential bugs are coming.
We will have to check also that the user didn't set `--min-replicas` with `--replicas`

If the user set `--enable-autoscaling` and not the min,max replicas they will get prompted for them, but I think we should keep the `enable-autoscaling flag explicit` (except for edit, when it is unchanged)

cmd/create/cluster/cmd.go
zgalor: fixed

zgalor:
@vkareh I fixed all issues but the one I commented on.
PTAL

cmd/create/cluster/cmd.go
vkareh: this sounds reasonable

cmd/edit/machinepool/cmd.go
```diff
@@ -142,32 +168,29 @@ func run(cmd *cobra.Command, argv []string) {
...
 	// Editing the default machine pool is a different process
 	if machinePoolID == "default" {
-		replicas, err = getReplicas(cmd)
-		if err != nil {
-			reporter.Errorf("Expected a valid number of replicas: %s", err)
+		autoscaling, replicas, minReplicas, maxReplicas := getReplicas(cmd, reporter, machinePoolID,
+			cluster.Nodes().AutoscaleCompute())
+
+		if !autoscaling && replicas < 2 ||
+			(autoscaling && cmd.Flags().Changed("min-replicas") && minReplicas < 2) {
```
vkareh, line 177: We should check here for multi-az and compare against 3 instead of 2

vkareh:
This is almost ready to merge, thanks a bunch for this work! I just posted one last nitpick, but I might still merge it later today and fix it in a different commit.

cmd/edit/machinepool/cmd.go
zgalor: Fixed

vkareh (approved):
Looks great!

PR #206, opened 2020-12-28
Disable `maligned` linter

cben:
It limits the order of struct members, e.g.:

    struct of size 320 bytes could be of size 312 bytes (maligned)

Such micro-optimizations are not interesting for this project IMHO.

@zgalor @vkareh please review.


zgalor:
@igoihman do you have merge rights here?

PR #207, opened 2021-01-04
clean up addons

vkareh:
This PR exposes addons to end-users. It also changes the command verb from `create` to `install` (e.g. `rosa install addon foo`). Finally it adds a `rosa list addons` command that shows a list of all addons available to the current org, regardless of installation status on any cluster.


vkareh:
@igoihman - please take a look

This doesn't _yet_ support addon parameters, but I've created [SDA-3527](https://issues.redhat.com/browse/SDA-3527) to address that.

jharrington22 (commented):
/lgtm

PR #210, opened 2021-01-12
Require min/max replicas iff existing machinepool autoscaling is disabled

zgalor:
Require min/max replicas on interactive mode iff existing machinepool autoscaling is disabled

Related: https://issues.redhat.com/browse/SDA-3543


zgalor:
@nimrodshn @vkareh PTAL

PR #232, opened 2021-02-09
arguments: Move region and profile flags

vkareh:
Instead of blanket-enabling the `--region` and `--profile` flags, we add
them to the commands that benefit from them. This avoid confusion by
removing them from the help entries of non-relevant commands.


jharrington22:
@vkareh I'd like to test this one before we merge, I was looking at the region flag as it doesn't work with the list command and somewhere else that I don't remember :slightly_smiling_face: - Looks like you've fixed the list command though :+1:

PR #233, opened 2021-02-10
interactive: Output command to rerun cluster creation

vkareh:
In the case where a user went through the interactive mode and the
cluster failed installation, they would need to go through the entire
interactive prompt again until their cluster succeeds. To simplify this
scenario, the cluster creation process now prints out a populated
command to re-create a cluster with the same parameters.

This is useful even if the cluster does not fail installation to allow
for reproducible clusters.

Demo: https://asciinema.org/a/qu9pOxdo49lF9iJO4JraIPgUn


wgordon17 (approved):
+1

PR #268, opened 2021-02-25
addons: Allow editing of addon parameters

vkareh:
Since some addon parameters support being edited, we can run through the
parameter list to allow the user to update them, defaulting to the
values currently set up in the existing addon installation.

### DIFF
--- a/cmd/edit/addon/cmd.go
+++ b/cmd/edit/addon/cmd.go
@@ -195,10 +195,10 @@ func run(_ *cobra.Command, argv []string) {
 				cidrVal, err = interactive.GetIPNet(input)
 				val = cidrVal.String()
 			case "number":
-				var numVal int
-				input.Default, _ = strconv.ParseInt(dflt, 10, 0)
-				numVal, err = interactive.GetInt(input)
-				val = strconv.Itoa(numVal)
+				var numVal float64
+				input.Default, _ = strconv.ParseFloat(dflt, 64)
+				numVal, err = interactive.GetFloat(input)
+				val = fmt.Sprintf("%f", numVal)
 			case "string":
 				input.Default = dflt
 				val, err = interactive.GetString(input)
--- a/cmd/edit/addon/cmd.go
+++ b/cmd/edit/addon/cmd.go
@@ -1,0 +1,235 @@
+/*
+Copyright (c) 2021 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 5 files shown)


cmd/edit/addon/cmd.go
```diff
@@ -0,0 +1,226 @@
...
+				if err != nil || !isValid {
+					reporter.Errorf("Expected %v to match /%s/", val, param.Validation())
+					os.Exit(1)
+				}
+			}
+
+			params = append(params, clusterprovider.AddOnParam{Key: param.ID(), Val: val})
+
+			return true
+		})
```
mikenairn, line 216: I might just missing it in here, but i cant see how this handles the case where a new param is added to the addon after creating the addon installation and has a default value, should input.Default be param.Default in that case?

cmd/edit/addon/cmd.go
```diff
@@ -0,0 +1,226 @@
...
+
+			if val != "" {
+				val = strings.Trim(val, " ")
+			}
+			if val != "" && param.Validation() != "" {
+				isValid, err := regexp.MatchString(param.Validation(), val)
+				if err != nil || !isValid {
+					reporter.Errorf("Expected %v to match /%s/", val, param.Validation())
+					os.Exit(1)
+				}
```
mikenairn, line 210: I assume you are doing this here to avoid an unnecessary API call? Should be fine as long as we never change how we process param.Validation

cmd/edit/addon/cmd.go
```diff
@@ -0,0 +1,226 @@
...
+					_, defaultIDR, _ := net.ParseCIDR(addOnInstallationParam.Value())
+					input.Default = *defaultIDR
+				}
+				cidrVal, err = interactive.GetIPNet(input)
+				val = cidrVal.String()
+			case "number":
+				var numVal int
+				input.Default, _ = strconv.ParseInt(addOnInstallationParam.Value(), 10, 0)
+				numVal, err = interactive.GetInt(input)
+				val = strconv.Itoa(numVal)
```
mikenairn, line 192: Backend would accept a float value as input, not sure if that matters.

cmd/edit/addon/cmd.go
vkareh: Good catch, I've updated this to let new params through, using the `param.DefaultValue()`

cmd/edit/addon/cmd.go
vkareh: Yes, we want to fail early if possible, to avoid unnecessarily leading the user through all the options if we already know it will fail

cmd/edit/addon/cmd.go
vkareh: Thanks, I've added a new commit to convert numeric inputs to float

PR #272, opened 2021-03-02
addons: Allow installation parameters in CLI

vkareh:
To enable scriptin/automation for addon installation, we now support
passing arbitrary flags to the 'install|edit addon' commands. When these
flags match the addon parameter ID, they get set as the chosen value and
so the interactive mode can be skipped.


vkareh:
A lot of the command boilerplate code in this PR can be simplified a lot if this gets merged/released: https://github.com/spf13/cobra/pull/1359 but I didn't want ROSA depend on my own fork in the meantime.

PR #274, opened 2021-03-08
versions: Align version list with cluster creation

vkareh:
When listing available versions, the format should match the expected
input when creating a cluster.


jharrington22:
Nice, I wanted to change this other other day.

/lgtm

PR #278, opened 2021-03-12
addons: Disallow editing addons without parameters

vkareh:
If an addon has no parameters, there is nothing to edit and so we should
exit early.


ciaranRoche:
LGTM

PR #281, opened 2021-03-12
machinepools: Allow editing labels and taints

vkareh:


cmd/edit/machinepool/cmd.go
```diff
@@ -103,6 +106,22 @@ func init() {
 		0,
 		"Maximum number of machines for the machine pool.",
 	)
+
+	flags.StringVar(
+		&args.labels,
+		"labels",
+		"",
+		"Labels for machine pool. Format should be a comma-separated list of 'key=value'. "+
+			"This list will overwrite any modifications made to Node labels on an ongoing basis.",
```
zgalor, line 115: np: `Node` -> `node`

cmd/edit/machinepool/cmd.go
```diff
@@ -103,6 +106,22 @@ func init() {
...
+	)
+
+	flags.StringVar(
+		&args.taints,
+		"taints",
+		"",
+		"Taints for machine pool. Format should be a comma-separated list of 'key=value:ScheduleType'. "+
+			"This list will overwrite any modifications made to Node taints on an ongoing basis.",
```
zgalor, line 123: np: `Node`->`node`

cmd/edit/machinepool/cmd.go
```diff
@@ -245,8 +273,76 @@ func run(cmd *cobra.Command, argv []string) {
...
+			Help:     cmd.Flags().Lookup("labels").Usage,
+			Default:  labels,
+		})
+		if err != nil {
+			reporter.Errorf("Expected a valid comma-separated list of attributes: %s", err)
+			os.Exit(1)
+		}
+	}
+	labels = strings.Trim(labels, " ")
+	if labels != "" {
```
zgalor, line 298: How can the user remove the existing labels? not by sending `""` ?
You might want to allow `""` as input and change this chek to `Changed("labels")` bool

cmd/edit/machinepool/cmd.go
```diff
@@ -245,8 +273,76 @@ func run(cmd *cobra.Command, argv []string) {
...
+			}
+			tokens := strings.Split(label, "=")
+			labelMap[strings.TrimSpace(tokens[0])] = strings.TrimSpace(tokens[1])
+		}
+	}
+
+	taints := args.taints
+	taintBuilders := []*cmv1.TaintBuilder{}
+	if interactive.Enabled() {
+		if taints == "" {
```
zgalor, line 312: maybe use `Changed("taints")` to allow resetting taints?

cmd/edit/machinepool/cmd.go
```diff
@@ -245,8 +273,76 @@ func run(cmd *cobra.Command, argv []string) {
...
+			tokens := strings.FieldsFunc(taint, Split)
+			taintBuilders = append(taintBuilders, cmv1.NewTaint().Key(tokens[0]).Value(tokens[1]).Effect(tokens[2]))
+		}
+	}
+
 	mpBuilder := cmv1.NewMachinePool().
-		ID(machinePool.ID())
+		ID(machinePool.ID()).
+		Labels(labelMap).
+		Taints(taintBuilders...)
```
zgalor, line 345: So labels and taints are always sent? 
How does the API handle the difference between no change and reset to empty?

nimrodshn: Both are sent to the backend.

cmd/edit/machinepool/cmd.go
```diff
@@ -171,8 +190,17 @@ func run(cmd *cobra.Command, argv []string) {
...
 	if machinePoolID == "default" {
+		if cmd.Flags().Changed("labels") {
+			reporter.Errorf("Labels cannot be updated on the default machine pool")
+			os.Exit(1)
+		}
+		if cmd.Flags().Changed("taints") {
+			reporter.Errorf("Taints are not supported on the default machine pool")
+			os.Exit(1)
+		}
+
```
nimrodshn, line 201: Nice. I was looking for this ^^

cmd/edit/machinepool/cmd.go
vkareh: I changed how this is done (which addresses all the comments above). If the user explicitly passes the flag (or if the interactive mode is used, since that always displays labels and taints), then we send it to the API regardless of whether there's a value or not. That way, if the user sends `--labels ""` (or enters an empty string  in interactive mode) it clears the labels.

PR #285, opened 2021-03-15
addons: Send empty string when CIDR is nil

vkareh:
When a nil IPNet is parsed, it returns a string "<nil>". This breaks
validation in the API when determining whether the addon parameter has
any data, so we set it as "" in that case.


ciaranRoche:
lgtm

PR #286, opened 2021-03-15
machinepool: Skip autoscaling prompt when setting replicas

vkareh:
If the user sets an explicit replica when creating a machine pool, we
should not ask about auto-scaling, as it should always be set to false.


jharrington22 (approved):
/lgtm

PR #287, opened 2021-03-15
machinepool: Error out on invalid min-replica

vkareh:
Instead of continuing with max-replicas, if we can already determine
that the min-replicas are wrong, we should error out before moving on.


jharrington22 (approved):
/lgtm

PR #288, opened 2021-03-15
cluster-admin: Format the success message

vkareh:
This makes the copy-pasteable command to work on Windows and makes it
clear that it will take some time to sync the new user credentials.


jharrington22 (approved):
/lgtm

PR #290, opened 2021-03-15
edit-cluster: Skip interactive mode if any flag is set

vkareh:


jharrington22 (approved):
/lgtm

Would be nice if we could get that list from somewhere and it not be static.

PR #293, opened 2021-03-15
interactive: Remove flag from global create

vkareh:
Not all create commands support interactive mode, so instead of adding
it to all sub-commands, we add it to those that do.


jharrington22 (approved):
/lgtm

PR #307, opened 2021-03-23
addons: Error when editing non-editable parameters

vkareh:
Since editing does not require a full interactive flow if at least one
parameter is supplied, we just verify that any CLI flags that match an
addon parameter are editable.


jharrington22 (approved):
/lgtm

PR #309, opened 2021-03-24
logs: Display logs when cluster is in error state

vkareh:


cmd/logs/install/cmd.go
```diff
@@ -168,7 +168,7 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(0)
 	}
 
-	if cluster.State() == cmv1.ClusterStateError || cluster.State() == cmv1.ClusterStateUninstalling {
```
jharrington22, line 171: I'm not understanding how this is displaying the logs? It will let a customer know there are logs available and also the logic has changed. Don't we only want to display this message if the cluster state is in `error`?  

Or are you assuming an error if in this case the cluster is installing and its in an uninstalling state? (opposite for uninstall)

jharrington22: Seems to me you want `if cluster.State() == cmv1.ClusterStateError` for both cases no?

vkareh: No, this was a mistake on my part when I added the code. This entire condition block didn't exist two weeks ago, then I was asked to filter out installation logs when a cluster is uninstalling (and vice-versa), but I also unnecessarily added the error state check.

If the cluster errors out during installation, we want to display the logs so that the user can report the failure.

jharrington22: Makes sense. Thanks for clarifying.

PR #311, opened 2021-03-25
logs: Filter out misleading output

vkareh:
Since logs are forwarded directly from the installer, we remove log
lines containing misleading KUBECONFIG instrutions as well as lines
already redacted by Hive.


jharrington22:
/lgtm

Verified that I see only a debug line and info line in the raw Hive output with `KUBECONFIG` in it and for `REDACTED` I see a few lines in the raw output containing just the string "REDACTED LINE OF OUTPUT".

The concatenated multi-line string would be caught with either `KUBECONFIG` or `REDACTED`.

PR #313, opened 2021-03-25
aws: Add helpful error message when using STS credentials

jharrington22:
Currently we don't support using STS users to create ROSA clusters. This PR will better inform the user of the issue. 

Old error message:

 `E: Error creating AWS client: ValidationError: Must specify userName when calling with non-User credentials`

New error message:

```
E: Error creating AWS client: The AWS Access Key ID AAAFFFDSDDDSSSSSS is associated with with STS credentials ARN arn:aws:sts::11223344556612:assumed-role/Role/SREAccess
STS users are not supported, please set AWS credentials to use an IAM user
```

Feel free to make suggestions on the error message.

### DIFF
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -195,6 +195,9 @@ func (b *ClientBuilder) Build() (Client, error) {
 	}
 
 	// Check that the AWS credentials are available:
+	// TODO: No need to do this twice, we're essentially doing the
+	// same thing in getClientDetails()
+	// We should implement getClientDetails() here or a new validation func
 	_, err = sess.Config.Credentials.Get()
 	if err != nil {
 		b.logger.Debugf("Failed to find credentials: %v", err)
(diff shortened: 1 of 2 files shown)


pkg/aws/client.go
```diff
@@ -300,10 +303,32 @@ func (c *awsClient) GetCreator() (*Creator, error) {
...
+		return false, err
+	}
+
+	// At this time we don't support creating clusters with STS credentials
+	// Inform the user to try again using a IAM User, provide access key to help
+	// the user debug where the credentials are coming from
+	if parsedArn.Service == "sts" {
+		return false, fmt.Errorf("The AWS Access Key ID %s is associated"+
+			" with with STS credentials ARN %s\n"+
+			"STS users are not supported, please set AWS credentials to use an IAM user", iamCredentials.AccessKeyID, parsedArn)
```
vkareh, line 330: `STS users are not supported` -> `STS users are not currently supported`

vkareh: I updated the copy based on @acathrow's comment

jharrington22: :+1:

PR #315, opened 2021-03-25
logs: Redact KUBECONFIG line

vkareh:
This causes confusion to users and is unnecessary for logs.

Fixes https://github.com/openshift/openshift-docs/issues/30903

### DIFF
--- a/cmd/logs/install/cmd.go
+++ b/cmd/logs/install/cmd.go
@@ -19,6 +19,7 @@ package install
 import (
 	"fmt"
 	"os"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


vkareh:
@jharrington22 - the issue was that the previous matching didn't account for multi-line log lines, so it didn't actually capture the `KUBECONFIG` string. This fixes it (tested locally with a full log run)

jharrington22:
> @jharrington22 - the issue was that the previous matching didn't account for multi-line log lines, so it didn't actually capture the `KUBECONFIG` string. This fixes it (tested locally with a full log run)

Don't you want to catch `REDACTED` lines too?

PR #321, opened 2021-04-06
init: Track ad-hoc authenticated events

vkareh:
During the init phase (as well as the verify phases) we log specific
events to help us track where users are having issues. These events are
tracked in an authenticated OCM endpoint, but are not related to any
specific cluster. Only the error key is sent to the API.


vkareh:
@jharrington22 - when reviewing, note that there are 2 commits: one to bump the SDK (a million files changed) and the actual event logger (only a handful of changes).

jharrington22 (approved):
/lgtm

PR #334, opened 2021-05-06
PrivateLink: Hide references to PrivateLink

vkareh:
To avoid confusion due to non-GA'd features, we hide all mentions of
PrivateLink when creating a cluster. Only if the user explicitly sets
the --private-link flag, we then show all available options and help
text.


pvasant:
LGTM. thanks

PR #337, opened 2021-05-12
Add support for STS clusters

vkareh:
This adds support for passing an AWS Role ARN to be used as
authentication for creating the cluster. Also, it accepts a list of
operator IAM roles so that the OpenShift installer can assume role in
the various components.

Finally, it allows custom tags to be added to all AWS resources created
by the installer.


pkg/logging/ocm_logger.go
```diff
@@ -22,7 +22,7 @@ import (
 	"context"
 	"fmt"
 
-	sdk "github.com/openshift-online/ocm-sdk-go"
+	// sdk "github.com/openshift-online/ocm-sdk-go"
```
jharrington22, line 25: Delete if not needed?

pkg/logging/ocm_logger.go
```diff
@@ -40,7 +40,7 @@ type OCMLogger struct {
 }
 
 // Make sure that we implement the OCM logger interface.
-var _ sdk.Logger = &OCMLogger{}
+// var _ sdk.Logger = &OCMLogger{}
```
jharrington22, line 43: Delete here too?

pkg/ocm/connection.go
```diff
@@ -74,17 +74,17 @@ func (b *ConnectionBuilder) Build() (result *sdk.Connection, err error) {
...
+	// 	Build()
+	// if err != nil {
+	// 	return
+	// }
 
 	// Prepare the builder for the connection adding only the properties that have explicit
 	// values in the configuration, so that default values won't be overridden:
 	builder := sdk.NewConnectionBuilder()
-	builder.Logger(logger)
+	// builder.Logger(logger)
```
jharrington22, line 87: Are you intentionally not creating a logger here?

jharrington22 (approved):
/lgtm

PR #345, opened 2021-05-20
describe: Display description during Pending state

vkareh:
By default OCM is 'Preparing account' during 'Pending', but if we
override the state description, we should show that instead.

Example:
```
$ rosa describe cluster -c vkareh-sts | grep State
State:                      pending (Waiting for OIDC configuration)
```


pvasant:
lgtm

PR #346, opened 2021-05-20
Release v1.0.7

vkareh:
- Allow setting 0 replicas to autoscaling machine pool (Not default)
- Updated the details link
- Added custom IAM Roles
- Remove default region
- describe: Display description during Pending state


pvasant:
lgtm

PR #348, opened 2021-05-24
Disable IAM user checks for STS

vkareh:
When a customer intends on using ROSA for clusters with STS
authentication, we do not need the 'osdCcsAdmin' user. We skip those
checks entirely, as well as the cluster simulation.


pvasant:
LGTM

PR #351, opened 2021-05-25
Support STS users

jharrington22:
This PR allows users with STS credentials to use the ROSA CLI client

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -359,6 +359,21 @@ func run(cmd *cobra.Command, _ []string) {
 	}()
 	ocmClient := ocmConnection.ClustersMgmt().V1()
 
+	tempAWSClient := aws.GetAWSClientForUserRegion(reporter, logger)
+
+	creator, err := tempAWSClient.GetCreator()
+	if err != nil {
+		reporter.Errorf("Unable to get IAM credentials: %v", err)
+		os.Exit(1)
+	}
+
+	// Creating non STS clusters is not supported
+	if creator.IsSTS && args.roleARN == "" {
+		reporter.Errorf("IAM STS credentials can only create STS clusters --role-arn cannot be blank")
+		reporter.Errorf("Your AWS credentials are returning an STS ARN: %s", creator.ARN)
+		os.Exit(1)
+	}
+
 	//Currently roleARN is hidden in interactive mode
 	//Might need to change later
 	if args.roleARN == "" {
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -304,7 +304,7 @@ func (c *awsClient) GetCreator() (*Creator, error) {
 
 // Checks if given credentials are valid.
 func (c *awsClient) ValidateCredentials() (bool, bool, error) {
-	// Validate the AWS credentails by calling STS GetCallerIdentity
+	// Validate the AWS credentials by calling STS GetCallerIdentity
 	// This will fail if the AWS access key and secret key are invalid. This
 	// will also work for STS credentials with access key, secret key and session
 	// token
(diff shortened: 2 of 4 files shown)


pkg/aws/client.go
```diff
@@ -304,33 +304,15 @@ func (c *awsClient) GetCreator() (*Creator, error) {
...
 
-	// Get IAM credentials only to support a better error message
-	// to help a user identity which AWS credentials the AWS client
-	// is using
-	iamCredentials, err := c.GetIAMCredentials()
-	if err != nil {
-		return false, false, err
-	}
-
-	parsedArn, err := arn.Parse(*callerIdentityOutput.Arn)
```
vkareh, line 320: I wonder if we should retain this call and `return true, parsedArn.Service == "sts", nil` at the end of the function, since during `rosa init` we have some code to capture in grafana how many users are using STS credentials. Probably not as important now that STS support is a thing

jharrington22: I managed to find a solution in the latest commit. Let me know what you think! Can do this too.

cmd/create/cluster/cmd.go
```diff
@@ -359,6 +359,20 @@ func run(cmd *cobra.Command, _ []string) {
 	}()
 	ocmClient := ocmConnection.ClustersMgmt().V1()
 
+	tempAWSClient := aws.GetAWSClientForUserRegion(reporter, logger)
+
+	creator, err := tempAWSClient.GetCreator()
+	if err != nil {
+		reporter.Errorf("Unable to get IAM credentials: %v", err)
```
vkareh, line 366: Do we want to `os.Exit(1)` here?

cmd/create/cluster/cmd.go
```diff
@@ -359,6 +359,20 @@ func run(cmd *cobra.Command, _ []string) {
...
+	}
+
+	// Creating non STS clusters is not supported
+	if creator.IsSTS && args.roleARN == "" {
+		reporter.Errorf("IAM STS credentials can only create STS clusters --role-arn cannot be blank")
+		reporter.Errorf("You are AWS credentials returning an STS ARN: %s please create an STS cluster", creator.ARN)
```
vkareh, line 372: Change to `Your AWS credentials are returning an STS ARN: '%s'.` (Remove the `please...` part)

jharrington22: Fixed.

cmd/create/cluster/cmd.go
jharrington22: Done

cmd/create/cluster/cmd.go
```diff
@@ -359,6 +359,21 @@ func run(cmd *cobra.Command, _ []string) {
...
+	creator, err := tempAWSClient.GetCreator()
+	if err != nil {
+		reporter.Errorf("Unable to get IAM credentials: %v", err)
+		os.Exit(1)
+	}
+
+	// Creating non STS clusters is not supported
+	if creator.IsSTS && args.roleARN == "" {
+		reporter.Errorf("IAM STS credentials can only create STS clusters --role-arn cannot be blank")
+		reporter.Errorf("You are AWS credentials returning an STS ARN: %s", creator.ARN)
```
vkareh, line 373: `You are AWS credentials` -> `Your AWS credentials are`

jharrington22: If I told you my linting failures were all spelling would you believe me?

jharrington22: Fixed.

vkareh: I have a code review checklist that includes:
- [ ] review spelling mistakes on @jharrington22's code

PR #354, opened 2021-05-27
sts: Normalize instace role parameters

vkareh:
To avoid possible confusion, we now expect a full ARN to be set for the
instance IAM roles. This removes inconsistencies in some parameters
using Role ARNs and some using Role Names.


vkareh:
We could add it, but the API already does this with the new attribute. I'd say it's out of scope for this specific PR, but I'd like to add proper ARN validation using the AWS SDK itself very soon (and specifically for the interactive mode)

pvasant:
Thanks @vkareh  .current changes lgtm

PR #355, opened 2021-05-27
sts: Expose all flags

vkareh:
To enable customers at large to create STS clusters, we expose all
STS-related flags, along with interactive help.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -24,16 +24,15 @@ import (
 	"strings"
 	"time"
 
+	awssdk "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
-	clusterdescribe "github.com/openshift/rosa/cmd/describe/cluster"
-	installLogs "github.com/openshift/rosa/cmd/logs/install"
-
 	"github.com/spf13/cobra"
 
-	awssdk "github.com/aws/aws-sdk-go/aws"
-	"github.com/openshift/rosa/pkg/aws"
-
+	clusterdescribe "github.com/openshift/rosa/cmd/describe/cluster"
+	installLogs "github.com/openshift/rosa/cmd/logs/install"
 	"github.com/openshift/rosa/pkg/arguments"
+	"github.com/openshift/rosa/pkg/aws"
 	clusterprovider "github.com/openshift/rosa/pkg/cluster"
 	"github.com/openshift/rosa/pkg/confirm"
 	"github.com/openshift/rosa/pkg/interactive"
@@ -136,37 +135,33 @@ func init() {
 		"",
 		"The Amazon Resource Name of the role that OpenShift Cluster Manager will assume to create the cluster.",
 	)
-	flags.MarkHidden("role-arn")
 	flags.StringVar(
 		&args.externalID,
 		"external-id",
 		"",
-		"A unique identifier that might be required when you assume a role in another account",
+		"An optional unique identifier that might be required when you assume a role in another account.",
 	)
-	flags.MarkHidden("external-id")
 	flags.StringArrayVar(
 		&args.operatorIAMRoles,
 		"operator-iam-roles",
 		nil,
-		"",
+		"List of OpenShift name and namespace, and role ARNs used to perform credential "+
+			"requests by operators needed in the OpenShift installer.",
 	)
-	flags.MarkHidden("operator-iam-roles")
 
 	flags.StringVar(
 		&args.masterRoleARN,
 		"master-iam-role",
 		"",
-		"The IAM role ARN that will be attached to master instances",
+		"The IAM role ARN that will be attached to master instances.",
 	)
-	flags.MarkHidden("master-iam-role")
 
 	flags.StringVar(
 		&args.workerRoleARN,
 		"worker-iam-role",
 		"",
-		"The IAM role ARN that will be attached to worker instances",
+		"The IAM role ARN that will be attached to worker instances.",
 	)
-	flags.MarkHidden("worker-iam-role")
 
 	flags.StringSliceVar(
 		&args.tags,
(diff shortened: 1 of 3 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -367,19 +361,6 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
-	// Creating non STS clusters is not supported
-	if creator.IsSTS && args.roleARN == "" {
-		reporter.Errorf("IAM STS credentials can only create STS clusters --role-arn cannot be blank")
-		reporter.Errorf("Your AWS credentials are returning an STS ARN: %s", creator.ARN)
-		os.Exit(1)
-	}
```
jharrington22, line 375: This was added to prevent users from creating regular ROSA clusters with STS credentials I don't think we want to remove that.

vkareh: Agreed - since we cannot guarantee that the user will send the `--role-arn` in the initial payload, if the creator ARN is STS, then we make the Role ARN required

cmd/create/cluster/cmd.go
```diff
@@ -418,40 +399,89 @@ func run(cmd *cobra.Command, _ []string) {
...
+		if creator.IsSTS {
+			help = fmt.Sprintf("%s Since your AWS credentials are returning an STS ARN you can only "+
+				"create STS clusters. Otherwise, switch to IAM credentials.", help)
+		}
 		roleARN, err = interactive.GetString(interactive.Input{
 			Question: "Role ARN",
-			Help:     cmd.Flags().Lookup("role-arn").Usage,
+			Help:     help,
 			Default:  roleARN,
+			Required: creator.IsSTS,
```
vkareh, line 412: Here the Role ARN is marked as required if the customer uses STS credentials

cmd/create/cluster/cmd.go
```diff
@@ -418,40 +399,98 @@ func run(cmd *cobra.Command, _ []string) {
...
+			Message: "Since your AWS credentials are returning an STS ARN you can only " +
+				"create STS clusters. Otherwise, switch to IAM credentials.",
+		})
+		if err != nil {
+			reporter.Errorf("%s", err)
+			os.Exit(1)
+		}
+		interactive.Enable()
+		reporter.Infof("Enabling interactive mode")
+	}
```
vkareh, line 414: @jharrington22 - just added this to ensure that STS credentials without a Role ARN have to enter it interactively.

cmd/create/cluster/cmd.go
jharrington22: :+1:

cmd/create/cluster/cmd.go
```diff
@@ -418,40 +399,98 @@ func run(cmd *cobra.Command, _ []string) {
...
 			Default:  roleARN,
+			Required: creator.IsSTS,
 		})
 		if err != nil {
 			reporter.Errorf("Expected a valid ARN: %s", err)
 			os.Exit(1)
 		}
 	}
+	if roleARN != "" {
+		_, err = arn.Parse(roleARN)
```
jharrington22, line 429: For a subsequent PR but I noticed this doesn't 100% parse/validate as you or I would expect, ie this ARN passes `arn:aws:sts::"123456789123":assumed-role/OrganizationAccountAccessRole/SREAccess` :disappointed: fine for now.

cmd/create/cluster/cmd.go
```diff
@@ -136,37 +135,33 @@ func init() {
 		"",
 		"The Amazon Resource Name of the role that OpenShift Cluster Manager will assume to create the cluster.",
 	)
-	flags.MarkHidden("role-arn")
 	flags.StringVar(
 		&args.externalID,
 		"external-id",
 		"",
-		"A unique identifier that might be required when you assume a role in another account",
+		"A unique identifier that might be required when you assume a role in another account.",
```
jharrington22, line 142: This might be better to facilitate the help in interactive mode since we don't explicitly say its "optional" `Optional identifier information when assuming a IAM role`

cmd/create/cluster/cmd.go
```diff
@@ -418,40 +399,98 @@ func run(cmd *cobra.Command, _ []string) {
...
+		os.Exit(1)
+	}
 
-		operatorIAMRoles := args.operatorIAMRoles
+	operatorIAMRoles := args.operatorIAMRoles
+	operatorIAMRoleList := []clusterprovider.OperatorIAMRole{}
+	if roleARN != "" {
 		for _, role := range operatorIAMRoles {
+			if !strings.Contains(role, ",") {
+				reporter.Errorf("Expected operator IAM roles to be a comma-separated " +
```
jharrington22, line 484: Use the flag usage here?

cmd/create/cluster/cmd.go
```diff
@@ -136,37 +135,33 @@ func init() {
...
-		"A unique identifier that might be required when you assume a role in another account",
+		"A unique identifier that might be required when you assume a role in another account.",
 	)
-	flags.MarkHidden("external-id")
 	flags.StringArrayVar(
 		&args.operatorIAMRoles,
 		"operator-iam-roles",
 		nil,
-		"",
+		"List of name, namespace, and role ARNs used to perform credential "+
```
jharrington22, line 148: I feel like we should mention its the OpenShift name and namespace to not confuse it with AWS?

cmd/create/cluster/cmd.go
jharrington22: The role ARN should be required though if the user is creating a STS cluster with non STS credentials.

We're preventing STS credentials from creating clusters that aren't STS. Non STS credentials should be able to create STS or non STS clusters.

vkareh: Correct, we are making the Role ARN required if the user uses STS credentials. However, whether the user is creating an STS cluster is determined by whether they enter a Role ARN or not, so it cannot be a required field.

cmd/create/cluster/cmd.go
vkareh: Fixed

cmd/create/cluster/cmd.go
vkareh: Maybe, but this explains the syntax, whereas the flag usage explains the purpose. The flag usage already shows up in interactive mode when pressing `?`

cmd/create/cluster/cmd.go
vkareh: Fixed

jharrington22:
Thanks @vkareh

PR #356, opened 2021-06-02
sts: Add support role ARN attribute

vkareh:
To enable SREs to access an STS cluster's AWS account in order to
provide support, we expose a support role ARN attribute that gets set in
the accountclaim.


jharrington22 (approved):
/lgtm

PR #360, opened 2021-06-07
versions: Ensure versions with STS support

vkareh:
When creating STS clusters we should only allow 4.7.11 and higher, as
lower OpenShift versions will fail installation when using STS
attributes.

### DIFF
--- a/pkg/ocm/versions/versions.go
+++ b/pkg/ocm/versions/versions.go
@@ -19,12 +19,17 @@ package versions
 import (
 	"errors"
 	"fmt"
+	"sort"
+
+	ver "github.com/hashicorp/go-version"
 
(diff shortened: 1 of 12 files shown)


pkg/ocm/versions/versions.go
```diff
@@ -51,9 +56,40 @@ func GetVersions(client *cmv1.Client, channelGroup string) (versions []*cmv1.Ver
 		}
 		page++
 	}
+
+	// Sort list in descending order, ensuring the 'default' version at the top
+	sort.Slice(versions, func(i, j int) bool {
```
pvasant, line 61: what is the purpose of this? don't we already have the order by in the query?

vkareh: The query uses database sort, which doesn't understand semver, so it ends up sorting like
```
4.7.2
4.7.11
4.7.1
```

PR #362, opened 2021-06-07
sts: Ensure operator IAM roles

vkareh:
When creating an STS cluster, operator IAM roles are mandatory to ensure
proper cluster installation. We now validate that there are operator IAM
roles included in the payload, otherwise we error out with a message
describing the process for extracting the credentials requests.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -525,22 +525,25 @@ func run(cmd *cobra.Command, _ []string) {
 				RoleARN:   roleData[2],
 			})
 		}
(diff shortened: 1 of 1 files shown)


pvasant:
LGTM

vkareh:
@pvasant thanks - you are actually part of the `OWNERS` file in this repo, so saying `/lgtm` should be sufficient to approve/merge automatically :) (https://github.com/openshift/rosa/pull/358)

vkareh:
> @vkareh: you cannot LGTM your own PR.

:facepalm:

PR #364, opened 2021-06-09
hack: Add directory with development scripts

vkareh:


hack/commit-release.sh
```diff
@@ -0,0 +1,70 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# This generates the commit necessary for release of the next z-stream tag.
+
+current=$(rosa version)
```
jharrington22, line 20:
```suggestion
current=$(git describe --tags `git rev-list --tags --max-count=1`)
```

Should we instead base this off the repo in case someone doesn't have an updated binary?

jharrington22: We might want a `git fetch --tags` before that.

hack/tag-release.sh
```diff
@@ -0,0 +1,42 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# This tags the latest release
+
+current=$(git tag --list --sort version:refname | tail -n1)
+next="v$(rosa version)"
```
jharrington22, line 21: Are we expecting the binary to be rebuilt shouldn't get this from `pkg/info/info.go`

hack/tag-release.sh
```diff
@@ -0,0 +1,42 @@
...
+read -r -p "Create release tag '$next'? [y/N] " response
+if [[ "$response" =~ ^([yY][eE][sS]|[yY])$ ]]
+then
+  git tag --annotate --message "Release $next" --message "$log" $next
+else
+  echo -e "\tgit tag --annotate $next"
+fi
+
+# Push git release tag to upstream GitHub repository
+upstream=$(git remote --verbose | grep "github\.com.openshift\/rosa\.git" | tail -n1 | awk '{print $1}')
```
jharrington22, line 35: while `github\.com.openshift\/rosa\.git` is valid wont it always be `:` if it ends in `.git` ?

vkareh: No, you can use `https://github.com/openshift/rosa.git` as well as a remote to your local fork clone - using `.` takes care of that.

hack/commit-release.sh
vkareh: Good idea. Fixed now

hack/tag-release.sh
vkareh: Good point - never considered that use case. Fixed.

hack/tag-release.sh
jharrington22: Ah there you go. Np!

PR #366, opened 2021-06-14
Refactor OCM client code

vkareh:
This is a series of commits that refactors most of the code that wraps the OCM SDK. Instead of having the various commands passing the SDK clients as function call arguments, we create a single client during the command, and the client will contain wrappers around all necessary API calls. All these wrappers now live inside `pkg/ocm`, with each resource type in its own file. This makes the ROSA code a lot simpler to traverse and make changes to.


pvasant:
@vkareh  checkout and tried few functionalities and looks good ..  I think it is good to merge.

PR #369, opened 2021-06-17
output: Add flag for JSON and YAML output

vkareh:
To use this, you can call `rosa describe cluster -c foo -o json` or `rosa list regions --output yaml`, etc. It also has autocomplete (if you have it enabled) so that you can do `--output [tab][tab]` and auto-complete the `json`/`yaml` parts.

### DIFF
--- a/pkg/output/output.go
+++ b/pkg/output/output.go
@@ -1,0 +1,119 @@
+/*
+Copyright (c) 2021 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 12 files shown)


pkg/output/output.go
```diff
@@ -0,0 +1,119 @@
...
+			cmv1.MarshalMachineTypeList(machineTypes, &b)
+		}
+	case "[]*v1.Version":
+		if versions, ok := resource.([]*cmv1.Version); ok {
+			cmv1.MarshalVersionList(versions, &b)
+		}
+	}
+	str, err := parseResource(b)
+	if err != nil {
+		fmt.Println(err)
```
jharrington22, line 71: Seems odd to print the err if we can't parse the json/yaml. The command would exit `0` in this case which would be miss leading? Maybe we just want to exit `1` after this?

vkareh: Fixed. It doesn't feel right to call `os.Exit` outside of the command itself, so I decided to just return the error and check for it in every call to `output.Print`.

PR #371, opened 2021-06-22
aws: Filter clusters by AWS account ID

vkareh:
To determine what clusters a user should be able to see, we no longer
limit the filtering to the creator ARN. Instead, we determine what AWS
account was the cluster installed in based on the creator ARN and the
role ARN (for STS clusters). This allows OCM organizations to see and
manage clusters created by different users within the same OCM
organization and same AWS account.

### DIFF
--- a/pkg/ocm/clusters.go
+++ b/pkg/ocm/clusters.go
@@ -94,8 +94,18 @@ type OperatorIAMRole struct {
 	RoleARN   string
 }
 
-func (c *Client) HasClusters(creatorARN string) (bool, error) {
-	query := fmt.Sprintf("properties.%s = '%s'", properties.CreatorARN, creatorARN)
+// Generate a query that filters clusters running on the current AWS session account
+func getClusterFilter(creator *aws.Creator) string {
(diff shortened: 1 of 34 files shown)


pkg/ocm/clusters.go
```diff
@@ -94,8 +94,18 @@ type OperatorIAMRole struct {
...
-	query := fmt.Sprintf("properties.%s = '%s'", properties.CreatorARN, creatorARN)
+// Generate a query that filters clusters running on the current AWS session account
+func getClusterFilter(creator *aws.Creator) string {
+	return fmt.Sprintf(
+		"product.id = 'rosa' AND (properties.%s LIKE '%%:%s:%%' OR aws.sts.role_arn LIKE '%%:%s:%%')",
+		properties.CreatorARN,
+		creator.AccountID,
+		creator.AccountID,
+	)
+}
```
vkareh, line 105: @jharrington22 @acathrow - this is the actual change in filtering here. It returns clusters within the same OCM org that are installed in the same AWS account ID.

PR #377, opened 2021-07-09
create account-roles to support STS clusters

vkareh:
Bunch of commits to allow customers to create roles for installing STS clusters. Best to read commit by commit:
- aws: Refactor AWS client code
- init: Replace --delete-stack flag with --delete
- init: Confirm delete operation
- create: Add new account-roles resource
- vendor: Update AWS SDK
- account-roles: Add tags to AWS resources
- init: Add 'account' to init command

Demos:
- manual mode: https://asciinema.org/a/eCjeilh3987RSzfXKnukEbTt7
- automatic mode: https://asciinema.org/a/Elrow80sc1jSWsOd5mHR9ZWnu


cmd/initialize/cmd.go
```diff
@@ -284,9 +238,82 @@ func run(cmd *cobra.Command, argv []string) {
 		}
 	}
 
+	// Verify version of `oc`
 	oc.Cmd.Run(cmd, argv)
 }
 
+func ocmLogin(cmd *cobra.Command, argv []string, reporter *rprtr.Object) error {
```
jharrington22, line 245: This is repeated from account sub command. We can't abstract some of this to the ocm package?

pkg/ocm/versions.go
```diff
@@ -88,6 +89,16 @@ func HasSTSSupport(rawID string, channelGroup string) bool {
...
 }
 
+func HasSTSSupportMinor(minor string) bool {
+	a, erra := ver.NewVersion(minor)
+	b, errb := ver.NewVersion(LowestSTSMinor)
+	if erra != nil || errb != nil {
+		return minor >= LowestSTSMinor
+	}
+
+	return a.GreaterThanOrEqual(b)
```
jharrington22, line 99: semver has a nice function for this https://pkg.go.dev/github.com/masterminds/semver#Version.GreaterThan

templates/policies/4.7/sts_installer_permission_policy.json
```diff
@@ -0,0 +1,177 @@
...
+                "elasticloadbalancing:DescribeTargetHealth",
+                "elasticloadbalancing:ModifyLoadBalancerAttributes",
+                "elasticloadbalancing:ModifyTargetGroup",
+                "elasticloadbalancing:ModifyTargetGroupAttributes",
+                "elasticloadbalancing:RegisterInstancesWithLoadBalancer",
+                "elasticloadbalancing:RegisterTargets",
+                "elasticloadbalancing:SetLoadBalancerPoliciesOfListener",
+                "iam:AddRoleToInstanceProfile",
+                "iam:CreateInstanceProfile",
+                "iam:DeleteInstanceProfile",
```
jharrington22, line 112: We should be able to remove these which I'll test. We'll need to repeat it through the other versions.

templates/policies/4.7/sts_installer_permission_policy.json
```diff
@@ -0,0 +1,177 @@
...
+                "iam:GetInstanceProfile",
```
jharrington22, line 113: Possibly this too.

templates/policies/4.7/sts_installer_permission_policy.json
```diff
@@ -0,0 +1,177 @@
...
+                "iam:GetUser",
+                "iam:ListAttachedRolePolicies",
+                "iam:ListInstanceProfiles",
+                "iam:ListInstanceProfilesForRole",
+                "iam:ListRolePolicies",
+                "iam:ListRoles",
+                "iam:ListUserPolicies",
+                "iam:ListUsers",
+                "iam:PassRole",
+                "iam:RemoveRoleFromInstanceProfile",
```
jharrington22, line 125: And this permission too.

cmd/create/accountroles/cmd.go
```diff
@@ -0,0 +1,472 @@
...
+		if err != nil {
+			reporter.Errorf("There was an error generating the policy files: %s", err)
+			os.Exit(1)
+		}
+
+		reporter.Infof("All policy files saved to the current directory")
+		reporter.Infof("Run the following commands to create the account roles and policies:\n")
+
+		commands := buildCommands(prefix, version)
+		fmt.Println(commands)
```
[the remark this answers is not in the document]
vkareh: you mean by using `reporter`? No I don't think I do - it's not an `Info` message (so I don't want it prefixed with `I: `), just the output with a list of commands to run

cmd/create/accountroles/cmd.go
```diff
@@ -0,0 +1,472 @@
...
+		}
+	}
+
+	switch mode {
+	case "auto":
+		creator, err := awsClient.GetCreator()
+		if err != nil {
+			reporter.Errorf("Unable to get IAM credentials: %s", err)
+			os.Exit(1)
+		}
```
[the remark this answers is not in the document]
vkareh: Ah good point.

To be honest, if the user selects `--mode manual`, there's no need to even create the AWS client. So we have three options:
1. leave as is (kinda silly though, creating the AWS client even if we don't need it)
2. Move both calls to the top: create client, then get creator before anything else happens
3. Move both calls inside the `case "auto"`, there's no need to create AWS client unless it's `auto`

I'm leaning towards 2 as any other call to `rosa` commands fail without credentials anyway, but what do you think?

cmd/initialize/account/cmd.go
```diff
@@ -0,0 +1,168 @@
...
+	if err != nil {
+		reporter.Errorf("Error creating AWS client: %v", err)
+		os.Exit(1)
+	}
+
+	// Validate AWS credentials for current user
+	reporter.Infof("Validating AWS credentials...")
+	ok, err := client.ValidateCredentials()
+	if err != nil {
+		ocmClient.LogEvent("ROSAInitCredentialsFailed")
```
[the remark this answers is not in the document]
vkareh: We should put them somewhere accessible to ROSA. The source of truth lives in the OCM/CS code: https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/cmd/clusters-service/service/event_service.go#L15-27

Since this is not new code, but reusing the existing `rosa init` code, I would punt on it to another PR.

cmd/initialize/cmd.go
vkareh: Good point, I'll fix that (I'm thinking inside `cmd/login/`, as I don't want to feed `cobra.Command` to the `ocm` package.

pkg/ocm/versions.go
vkareh: I'm using https://github.com/hashicorp/go-version instead. I have no strong opinion on either, but the OCM API already uses the hashicorp lib anyway.

cmd/create/accountroles/cmd.go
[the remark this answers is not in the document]
vkareh: I moved it to the top, right after we create `awsClient`. Let me know if you think it's a good idea to chage it.

cmd/initialize/cmd.go
vkareh: Done. Moved to `cmd/login/cmd.go`

templates/policies/4.7/sts_installer_permission_policy.json
vkareh: We should defer this to another PR with its own Jira, since this also affects documentation (these policies are straight from the docs).

templates/policies/4.7/sts_installer_permission_policy.json
vkareh: We should defer this to another PR with its own Jira, since this also affects documentation (these policies are straight from the docs).

templates/policies/4.7/sts_installer_permission_policy.json
vkareh: We should defer this to another PR with its own Jira, since this also affects documentation (these policies are straight from the docs).

vkareh:
@jharrington22 - I think I've addressed all comments. Let me know what you think.

pvasant:
@vkareh  LGTM

jharrington22 (commented):
@vkareh your comments are :+1: however a couple of things.

`rosa init --help` or `rosa init account` don't show the sub command or help as well as it looks like we've regressed on the STS user issue:

```
rosa init account
I: Logged in as 'xxxxx' on 'https://api.openshift.com'                                                                                                                 E: Error creating AWS client: The AWS Access Key ID XXXXXXXXXXXX is associated with with STS credentials ARN arn:aws:sts::xxxxxxxxxx:assumed-role/OrganizationAccountAcces
sRole/SREAccess                                                                                                                                                                     
STS users are not currently supported, please set AWS credentials to use an IAM user   
```

jharrington22:
> @vkareh your comments are  however a couple of things.
> 
> `rosa init --help` or `rosa init account` don't show the sub command or help as well as it looks like we've regressed on the STS user issue:
> 
> ```
> rosa init account
> I: Logged in as 'xxxxx' on 'https://api.openshift.com'                                                                                                                 E: Error creating AWS client: The AWS Access Key ID XXXXXXXXXXXX is associated with with STS credentials ARN arn:aws:sts::xxxxxxxxxx:assumed-role/OrganizationAccountAcces
> sRole/SREAccess                                                                                                                                                                     
> STS users are not currently supported, please set AWS credentials to use an IAM user   
> ```

Ignore this comment. This is actually working correctly it was a testing issue.

cmd/create/accountroles/cmd.go
```diff
@@ -0,0 +1,472 @@
...
+		path := fmt.Sprintf("templates/policies/%s", filename)
+
+		policy, err := aws.ReadPolicyDocument(path, map[string]string{
+			"aws_account_id": aws.JumpAccounts[env],
+		})
+		if err != nil {
+			return err
+		}
+
+		reporter.Infof("Creating role '%s'", name)
```
jharrington22, line 403: @vkareh since `rosa create cluster` takes flags with role ARN's should we output them here so a user doesn't have to go look them up?

vkareh: that's a good idea, yes

pkg/aws/policies.go
```diff
@@ -0,0 +1,215 @@
...
+	})
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case iam.ErrCodeEntityAlreadyExistsException:
+				return nil
+			}
+		}
+		return err
+	}
```
jharrington22, line 172: Is it a conscious decision to not return an error if the policy already exists? How would this work when a customer upgrades from `4.7` to `4.8` for instance? Applying the `4.8` policies over the `4.7` policies for instance results in no change even though the policy for `4.8` adds a permission `elasticloadbalancing:DeregisterTargets` which is silently dropped in this case.

vkareh: Good question. I'll add a way to not just ensure the role exists, but make sure it's the right version. Maybe I can add it in a follow-up PR?

jharrington22: My concern is we push this release and we're close to the 4.8 upgrade. It should be easy enough to validate the `rosa_openshift_version` tag and return an error if it doesn't match? If its a bigger change then ok let's do it in a separate PR.

jharrington22: @vkareh lets merge this one and call out the risk before we cut a release.

PR #381, opened 2021-07-13
There is no "user" anymore

jjasghar:
To create an administrator user it's `admin` not `user`.
It also seems that `user` isn't available anymore either.

Signed-off-by: JJ Asghar <awesome@ibm.com>


vkareh (commented):
Good catch.

`rosa create admin` is used to create an HTPasswd IDP. The original `create user` command was moved to `grant user`, so I think we should use that.

PR #383, opened 2021-07-16
output: Ensure that JSON output for empty arrays looks correct

vkareh:
Also send warning output to stderr to make it all consistent.


jharrington22 (approved):
/lgtm

PR #386, opened 2021-07-20
Reduce EBS quota checks

jharrington22:
Recent changes in AWS quota have caused the EBS quota checks to fail. This PR allows the quota checks to pass, this requires testing as the low quota likely will prevent a cluster install.

I'm going to open this PR and put it on hold until we work with AWS to figure out the correct path forward.

PR #387, opened 2021-07-20
Add init flag to disable quota checks

jharrington22:
Given issues with certain AWS quotas see #386 this PR adds an additional `init` flag to disable quota checks. 

Putting this on hold for discussion @vkareh - Maybe we want to make this so you disable certain servicequota codes `L-D18FCD1D` the reason I suggest this is because we don't want uses disabling quota if they are not sure as it will case a lot of toil for SRE. Asking them to be specific might be better.

### DIFF
--- a/cmd/initialize/cmd.go
+++ b/cmd/initialize/cmd.go
@@ -37,10 +37,11 @@ import (
 )
 
 var args struct {
-	deleteStack      bool
-	disableSCPChecks bool
-	sts              bool
-	region           string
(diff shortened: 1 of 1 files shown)


vkareh:
I think we should not do this and rely on the changes in #386. If that's not sufficient, skipping quota check will probably not help us since cluster installation will very likely fail

jharrington22:
Agree @vkareh #386 has merged. Closing this.

PR #388, opened 2021-07-21
Add commands to create operator roles and OIDC provider

vkareh:


cmd/create/oidcprovider/cmd.go
```diff
@@ -0,0 +1,261 @@
...
+var args struct {
+	clusterKey string
+	mode       string
+}
+
+var Cmd = &cobra.Command{
+	Use:     "oidc-provider",
+	Aliases: []string{"oidcprovider"},
+	Short:   "Create OIDC provider for an STS cluster.",
+	Long:    "Create OIDC provider for operators to authenticate against an STS cluster.",
```
jharrington22, line 51:
```suggestion
	Long:    "Create OIDC provider for operators to authenticate against in an STS cluster.",
```

cmd/create/oidcprovider/cmd.go
```diff
@@ -0,0 +1,261 @@
...
+
+	// Try to find the cluster:
+	reporter.Debugf("Loading cluster '%s'", clusterKey)
+	cluster, err := ocmClient.GetCluster(clusterKey, creator)
+	if err != nil {
+		reporter.Errorf("Failed to get cluster '%s': %v", clusterKey, err)
+		os.Exit(1)
+	}
+
+	if cluster.State() != cmv1.ClusterStatePending {
```
jharrington22, line 146: Is the only reason for a cluster to be pending OIDC configuration? Can we be more specific and look at the `oidc_ready` status? Should we also validate that this cluster is an STS cluster?

cmd/create/operatorroles/cmd.go
```diff
@@ -0,0 +1,379 @@
...
+		})
+		if err != nil {
+			reporter.Errorf("Expected a valid role prefix: %s", err)
+			os.Exit(1)
+		}
+	}
+	if prefix == "" {
+		prefix = cluster.Name()
+	}
+	if len(prefix) > 32 {
```
jharrington22, line 179: I'm not sure if I asked this on the last PR, is there any reason behind `32` being the length of the prefix? `64` is the ultimate limit and I assume you just picked an arbitrary limit? (I don't disagree with it just wanted to ask)

jharrington22 (commented):
Looks good so far just a couple of questions. Let me test it locally.

cmd/create/oidcprovider/cmd.go
vkareh: Done

cmd/create/operatorroles/cmd.go
vkareh: it's arbitrary - but if we allow too long of a prefix, we run the risk of duplicate role names. For example
`thisisaverylongrolenameprefixjusttoshowwhyitsabadidea-openshif` <-- all operator roles, when truncated to 64, would end up with the same name.

cmd/create/oidcprovider/cmd.go
vkareh: It doesn't really matter. Currently 'Pending' means waiting for Account Claims or OIDC, however the OIDC can be done at any stage _prior_ to 'Installing' (which implies OIDC is already set up). If the OIDC is set up before it gets to the 'Waiting for OIDC configuration' state, then it just moves on to 'Installing' right away.

Think of it as a `if cluster.State() < cmv1.ClusterStateInstalling`, except for the cluster states are not numeric

vkareh: > Should we also validate that this cluster is an STS cluster?
We already do that in the next condition (`cluster.AWS().STS().RoleARN() == ""`)

cmd/create/operatorroles/cmd.go
jharrington22: Make sense, I was curious. Thanks.

cmd/create/oidcprovider/cmd.go
```diff
@@ -0,0 +1,261 @@
...
+
+	conn, err := tls.Dial("tcp", fmt.Sprintf("%s:443", connect.Host), nil)
+	if err != nil {
+		return "", err
+	}
+	defer conn.Close()
+
+	// Grab the last certificate in the chain
+	certChain := conn.ConnectionState().PeerCertificates
+	cert := certChain[len(certChain)-1]
```
jharrington22, line 248: @vkareh can you think of a way to ensure that this is the CA certificate and not the intermediate certificate? I can't right now. Would be happy to mess around tomorrow and try to figure it out. Point being CA certs are less likely to expire or get rotated so thumbprints should be more stable.

vkareh:
The AWS docs want you to use the last cert (out of the two returned) and they validate it in the request, AFAIK. It should be the root CA.

cmd/create/oidcprovider/cmd.go
vkareh: Fixed now. I'm looking for the CA, and if it's not there then it falls back to the last one in the chain.

cmd/create/operatorroles/cmd.go
```diff
@@ -0,0 +1,379 @@
...
+	if cluster.State() != cmv1.ClusterStatePending {
+		reporter.Errorf("Cluster '%s' is %s and does not need additional configuration.",
+			clusterKey, cluster.State())
+		os.Exit(1)
+	}
+
+	if cluster.AWS().STS().RoleARN() == "" {
+		reporter.Errorf("Cluster '%s' is not an STS cluster.", clusterKey)
+		os.Exit(1)
+	}
```
jharrington22, line 161: The cluster could be `installing` if someone configures the OIDC provider but not the roles. I think we need to validate that the IAM roles exist in AWS here for the operators.

jharrington22: Will be resolved in subsequent PR.

jharrington22:
Nice @vkareh :partying_face:

PR #391, opened 2021-07-21
account-roles: Ensure that roles and policies can be upgraded

vkareh:
If a customer creates roles and policies for OpenShift 4.7 clusters and
then decides to upgrade to 4.8, we need to ensure that they can re-run
the initialization commands that will update the resources to be
compatible. Since later versions are a superset of older versions, we
enforce upgrading and ignore downgrade requests.

### DIFF
--- a/pkg/aws/policies.go
+++ b/pkg/aws/policies.go
@@ -22,10 +22,13 @@ import (
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/iam"
+	semver "github.com/hashicorp/go-version"
 
 	"github.com/openshift/rosa/assets"
+	"github.com/openshift/rosa/pkg/aws/tags"
 )
 
 type Operator struct {
@@ -129,7 +132,7 @@ type PolicyStatementPrincipal struct {
 	AWS []string `json:"AWS"`
 }
 
-func (c *awsClient) EnsureRole(name string, policy string, tagList map[string]string) (string, error) {
+func (c *awsClient) EnsureRole(name string, policy string, version string, tagList map[string]string) (string, error) {
 	output, err := c.iamClient.GetRole(&iam.GetRoleInput{
 		RoleName: aws.String(name),
 	})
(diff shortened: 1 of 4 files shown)


jharrington22:
@vkareh this looks good however the IAM role tag `rosa_openshift_version` doesn't appear to update? The policy does.

```
aws iam list-policy-tags --policy-arn arn:aws:iam::884684697417:policy/james3-openshift-machine-api-aws-cloud-credentials | jq '.Tags[] | select(.Key=="rosa_openshift_version")'      
{
  "Key": "rosa_openshift_version",
  "Value": "4.8"
}
(venv_3) (⎈ |default/hivep01ue1/jaharrin.openshift:default)➜  rosa git:(vkareh-SDA-4507/create-account-roles-versions) ✗ aws iam get-role --role-name james3-openshift-machine-api-aws-cloud-credentials  | jq '.Role.Tags[] | select(.Key=="rosa_openshift_version")'
{
  "Key": "rosa_openshift_version",
  "Value": "4.7"
}
```

Validated that the policy itself was updated though.

vkareh:
> @vkareh this looks good however the IAM role tag `rosa_openshift_version` doesn't appear to update? The policy does.
> 
> ```
> $ aws iam list-policy-tags --policy-arn arn:aws:iam::884684697417:policy/james3-openshift-machine-api-aws-cloud-credentials | jq '.Tags[] | select(.Key=="rosa_openshift_version")'      
> {
>   "Key": "rosa_openshift_version",
>   "Value": "4.8"
> }
> $ aws iam get-role --role-name james3-openshift-machine-api-aws-cloud-credentials  | jq '.Role.Tags[] | select(.Key=="rosa_openshift_version")'
> {
>   "Key": "rosa_openshift_version",
>   "Value": "4.7"
> }
> ```
> 
> Validated that the policy itself was updated though.

Interesting find. The `create account-roles` command only creates the policies. To create those roles (which are cluster-specific) you use `create operator-roles`. What this means is that the _roles_ are created based on the version of the current cluster, whereas those policies are set to the latest version that the customer wants in their account. Since you're updating policies that are used account-wide, regardless of clusters, the roles themselves (which belong to the cluster) don't get updated.

Does that make sense?

pkg/aws/policies.go
```diff
@@ -162,8 +190,37 @@ func (c *awsClient) createRole(name string, policy string, tagList map[string]st
...
+			}
+			currentVersion, err := semver.NewVersion(aws.StringValue(tag.Value))
+			if err != nil {
+				return false, err
+			}
+			return currentVersion.GreaterThanOrEqual(wantedVersion), nil
+		}
+	}
+
+	return false, err
```
ciaranRoche, line 223: This block is repeated in `isPolicyCompatible` might be worth breaking out into a seperate function.

ciaranRoche (commented):
Just one comment, other than that lgtm :+1:

pkg/aws/policies.go
vkareh: Ah, good point. I'll do that :+1:

jharrington22:
> > @vkareh this looks good however the IAM role tag `rosa_openshift_version` doesn't appear to update? The policy does.
> > ```
> > $ aws iam list-policy-tags --policy-arn arn:aws:iam::884684697417:policy/james3-openshift-machine-api-aws-cloud-credentials | jq '.Tags[] | select(.Key=="rosa_openshift_version")'      
> > {
> >   "Key": "rosa_openshift_version",
> >   "Value": "4.8"
> > }
> > $ aws iam get-role --role-name james3-openshift-machine-api-aws-cloud-credentials  | jq '.Role.Tags[] | select(.Key=="rosa_openshift_version")'
> > {
> >   "Key": "rosa_openshift_version",
> >   "Value": "4.7"
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Validated that the policy itself was updated though.
> 
> Interesting find. The `create account-roles` command only creates the policies. To create those roles (which are cluster-specific) you use `create operator-roles`. What this means is that the _roles_ are created based on the version of the current cluster, whereas the _policies_ are set to the latest version that the customer wants in their account. Since you're updating policies that are used account-wide, regardless of clusters, the roles themselves (which belong to the cluster) don't get updated.
> 
> Does that make sense?

It makes sense, could be confusing to the customer but this for the moment does what is required by updating the policy.

jharrington22 (approved):

pkg/aws/policies.go
vkareh: Done

PR #395, opened 2021-07-23
SDA-4457 : Fix support for conditional addon parameters

ciaranRoche:
Fixes : https://issues.redhat.com/browse/SDA-4457

This change : 
- Bumps ocm-sdk to v0.1.197
- Adds `resource` type to addon parameter interactive menu
- Replaces getAddonParameter call to use `addon-inquiries` endpoint

To verify 
- Create ROSA cluster
- Add Managed API Addon
```
rosa install addon managed-api-service --cluster=<<cluster-name>> --addon-managed-api-service=1 --addon-resource-required=true --cidr-range=10.1.0.0/26 -y 
```
- Verify no errors, and addon is created


cmd/install/addon/cmd.go
```diff
@@ -218,6 +218,11 @@ func run(cmd *cobra.Command, argv []string) {
 				case "string":
 					input.Default = param.DefaultValue()
 					val, err = interactive.GetString(input)
+				case "resource":
+					var resourceVal int
+					input.Default, _ = strconv.Atoi(param.DefaultValue())
+					resourceVal, err = interactive.GetInt(input)
+					val = fmt.Sprintf("%d", resourceVal)
```
vkareh, line 225: This is the same as the `number` case. You can always do something like
```
case "number", "resource":
// etc...
```

vkareh:
/hold since the addon inquiries is not yet deployed in production

Other than that and my comment above, this looks good to me :+1:

ciaranRoche:
@vkareh can we merge this for the release too?

vkareh:
This is deployed in prod now, right?

ciaranRoche:
Yeah, it was pushed yesterday

PR #396, opened 2021-07-23
roles: Update trust policy

vkareh:
When creating roles in AWS, if the role already exists, instead of
ignoring the update we ensure that the new trust policy contains the
combination of what was there and the new one. That way the new policy
does not override the old one.

### DIFF
--- a/assets/bindata.go
+++ b/assets/bindata.go
@@ -347,7 +347,9 @@ var _templatesPolicies47Operator_iam_role_policyJson = []byte(`{
       "Principal": {
         "Federated": "%{oidc_provider_arn}"
       },
-      "Action": "sts:AssumeRoleWithWebIdentity",
+      "Action": [
+        "sts:AssumeRoleWithWebIdentity"
+      ],
       "Condition": {
         "StringEquals": {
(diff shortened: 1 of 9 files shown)


vkareh:
@jharrington22 this is awful, but basically the issue is that if you run `rosa create account-roles` in staging and then run it again in production, since the role already exists, your trust policy will not get updated. This solves it by getting the existing policy doc and adding the new one to the list, then updating it.

Please take a look/test

jharrington22:
> @jharrington22 this is awful, but basically the issue is that if you run `rosa create account-roles` in staging and then run it again in production, since the role already exists, your trust policy will not get updated. This solves it by getting the existing policy doc and adding the new one to the list, then updating it.
> 
> Please take a look/test

Oh, text parsing the the principals :( let me test this.

vkareh:
> Oh, text parsing the the principals :( let me test this.

Yes, this sucks. The alternative is to enforce new roles per environment or something like that (not sure how to actually _enforce_ that though).

jharrington22:
Validated the principal ARN's get updated as expected.

/lgtm

PR #399, opened 2021-07-26
create-operatorroles: Make default flag empty

vkareh:
Instead of using the ManagedOpenShift prefix as the default for operator
roles, since they are cluster-scoped we make the default an empty
string that we then pre-fill with the cluster name.


vkareh:
This PR is incorrect. Closing.

PR #400, opened 2021-07-27
create-operatorroles: Auto-find policies for roles

vkareh:
Since the operator role policies created through 'rosa create
account-roles' are tagged with the operator metadata, we can drop the
requirement for a user-set prefix and instead automatically find the
corresponding policy.

### DIFF
--- a/pkg/aws/policies.go
+++ b/pkg/aws/policies.go
@@ -32,6 +32,8 @@ import (
 	"github.com/openshift/rosa/pkg/aws/tags"
 )
 
+var DefaultPrefix = "ManagedOpenShift"
+
 type Operator struct {
 	Name                string
 	Namespace           string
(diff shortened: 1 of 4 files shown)


pkg/aws/policies.go
```diff
@@ -210,6 +213,52 @@ func (c *awsClient) AttachRolePolicy(roleName string, policyARN string) error {
 	return nil
 }
 
+func (c *awsClient) FindPolicyARN(operator Operator, version string) (string, error) {
+	listPoliciesOutput, err := c.iamClient.ListPolicies(&iam.ListPoliciesInput{
```
jharrington22, line 217: I see that we're applying a scope here but larger organizations might have > 100 policies. We might want to use `[ListPoliciesPagesWithContext](https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#IAM.ListPoliciesPagesWithContext)` instead to paginate.

jharrington22, line 217: I see that we're applying a scope here but larger organizations might have > 100 policies. We might want to use [ListPoliciesPagesWithContext](https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#IAM.ListPoliciesPagesWithContext) instead to paginate.

pkg/aws/policies.go
```diff
@@ -210,6 +213,52 @@ func (c *awsClient) AttachRolePolicy(roleName string, policyARN string) error {
...
+				}
+			case tags.OpenShiftVersion:
+				isTagged = true
+				if tagValue != version {
+					skip = true
+					break
+				}
+			}
+		}
+		if isTagged && !skip {
```
ciaranRoche, line 255: I dont know if I am misreading the logic above, but what happens if there are multiple policies with the same tag `Key(s)`? 
Would the loop above always return `skip` as true, even if one policy is valid?

vkareh: No, here I want to make sure that this particular policy has tags, and that _all_ three tags (namespace, name, version) match. If any of those tags do not match, I skip it and the first `for` loop moves on to the next policy. As soon as I find a policy that has all 3 tags matching, I use it

ciaranRoche: Perfect, thanks for the explanation.

pkg/aws/policies.go
vkareh: Done

jharrington22 (commented):
```
rosa create operator-roles -c jamesh-foo-48 --mode auto -y
E: Expected a valid role prefix matching ^[\w+=,.@-]+$
```

Regex validation fails because the prefix is undefined. Should run `FindPolicyARN` immediately after the interactive mode check for a `prefix` if its blank? The the rest of the parsing would complete.

vkareh:
Fixed it, I was validating the prefix regex even if a prefix wasn't set. This new version only checks the regex if there's a prefix

PR #402, opened 2021-07-27
create-oidcprovider: Verify if OIDC Provider already exists

vkareh:
If a cluster already has the OIDC provider configured, we skip trying to
re-create it. Also, if there is a configuration issue we expose it to
the user.

### DIFF
--- a/cmd/create/oidcprovider/cmd.go
+++ b/cmd/create/oidcprovider/cmd.go
@@ -167,6 +167,18 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(0)
 	}
 
+	oidcEndpointURL := cluster.AWS().STS().OIDCEndpointURL()
+	oidcProviderExists, err := awsClient.HasOpenIDConnectProvider(oidcEndpointURL, creator.AccountID)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/create/oidcprovider/cmd.go
```diff
@@ -167,6 +167,18 @@ func run(cmd *cobra.Command, _ []string) {
...
 	}
 
+	oidcEndpointURL := cluster.AWS().STS().OIDCEndpointURL()
+	oidcProviderExists, err := awsClient.HasOpenIDConnectProvider(oidcEndpointURL, creator.AccountID)
+	if err != nil {
+		reporter.Errorf("Failed to verify if OIDC provider exists: %s", err)
+		os.Exit(1)
+	}
+	if oidcProviderExists {
+		reporter.Warnf("Cluster '%s' already has OIDC provider but has not yet started installation. "+
```
igoihman, line 177: shouldn't it be an error?

vkareh: Not necessarily - OCM has a worker that verifies OIDC configuration and sometimes it may take a minute or two, but there's no way for `rosa` to know whether that's the case or if there's a config error: it needs to wait for OCM to determine that

PR #404, opened 2021-07-30
create-cluster: Ensure all role ARNs are required

vkareh:
STS clusters require the instance roles for proper installation and
support role for SRE access.


jharrington22 (approved):
/lgtm

PR #405, opened 2021-07-30
clusters: Ensure blocking pending clusters are non-STS

vkareh:
When blocking a pending cluster from installing based on a waiting
account claim, ensure that the blocking cluster is not STS.


jharrington22 (approved):
/lgtm

PR #415, opened 2021-08-04
Interactive validation

vkareh:
This adds interactive validation so that the program does not exit on bad input, instead it displays the error and re-prompts the user.

Demo: https://asciinema.org/a/sFXGyyEcv1y6inoGiBEEXrjWI


zgalor:
Very nice!
How can we avoid duplicating the validation code between interactive mode and user flag inputs?

vkareh:
@zgalor
> How can we avoid duplicating the validation code between interactive mode and user flag inputs? 

I've been thinking about this, and I want to try writing an `input` wrapper that takes the flag info and the validator, then depending on whether the interactive mode is enabled or not, it would display the prompt or just validate the existing data. This would be a large change, probably better suited for a future improvement.

pvasant:
this is nice. looks good to me

PR #417, opened 2021-08-05
config: Use standard config path for ocm.json

vkareh:
If the legacy ~/.ocm.json file already exists, it continues using that.
Otherwise it prefers the standard XDG config directory.


vkareh:
relevant: https://github.com/openshift-online/ocm-cli/pull/302

igoihman (approved):

PR #420, opened 2021-08-06
Remove unnecessary SCP checks

vkareh:
Since we only use SCP checks to verify whether a cluster can be created,
we need to only query for the strictly necessary permissions.

Ref: https://cloud.redhat.com/dedicated/ccs


jharrington22 (approved):
/lgtm

PR #424, opened 2021-08-10
Bump OCM SDK version to v0.1.199

vkareh:
This new version of the OCM SDK uses golang-jwt/jwt instead of
dgrijalva/jwt-go. We match the behaviour by swapping the module on our
side as well.


jhernand (approved):
Looks good to me.

jhernand:
@vkareh the new JWT library uses some Go 1.15 methods, so you will need to update the project to use at least Go 1.15.

vkareh:
> @vkareh the new JWT library uses some Go 1.15 methods, so you will need to update the project to use at least Go 1.15.

Ah, thanks @jhernand - Also realised that openshift/release forces golang-1.14 on the pipeline

vkareh:
@jhernand - this is needed too: https://github.com/openshift/release/pull/21032

jhernand:
While at it it may be worth to update to Go 1.16 already.

vkareh:
The problem with 1.16 is that I'm not entirely sure if RHEL already supports 1.16.6, and I need to build it in a version supported by RHEL. Any ideas how to check?

jhernand (approved):
Looks good to me.

PR #430, opened 2021-08-17
region: Move flag up a level

vkareh:
Also simplify code by removing the explicit set in every client
instantiation, and read the flag directly from the AWS client builder.


jharrington22:
Nice. 

/lgtm

PR #431, opened 2021-08-17
edit: Add automatic upgrade policy

vkareh:


wgordon17:
Hey @vkareh, I have a general _feeling_ that this is a better fit for `rosa edit cluster`, since this is about setting a _policy_, as opposed to actually actioning an _upgrade_.

vkareh:
that's not a bad idea. I'll update this

jharrington22:
/hold

@vkareh adding this in light of the STS upgrade discussions.

vkareh:
@jharrington22 I've updated this to block setting automatic upgrade schedules to STS clusters.

vkareh:
@wgordon17 @jharrington22 - now that OCM only does automatic upgrades for z-streams this can probably be unblocked and merged. What do you think?

tzvatot:
@vkareh this has been stale for a while. Do you like to revive it?

tzvatot:
@vkareh MR has conflicts. Can you fix those?

cmd/edit/cluster/cmd.go
```diff
@@ -48,14 +51,41 @@ var args struct {
 	httpsProxy                string
 	noProxySlice              []string
 	additionalTrustBundleFile string
+
+	// Upgrade schedule options
+	upgradeScheduleDay  string
+	upgradeScheduleTime string
+}
+
+var dayOptions = []string{
```
tzvatot, line 60: Do we really need this an also `daysMap`? seems like a duplication. We can get the map keys (keys := reflect.ValueOf(daysMap).MapKeys())

PR #434, opened 2021-08-19
add check and prompt for required true addon parameters

ciaranRoche:
Part of - https://issues.redhat.com/browse/SDA-4588

This change adds a prompt if an addon parameter is a boolean and required to be true.

Follow up work to this is dependant on https://github.com/openshift-online/ocm-api-model/pull/406 where we can update the regex error returned to return `validation_error` if it exists.


ciaranRoche:
@vkareh PTAL

cmd/edit/addon/cmd.go
```diff
@@ -206,9 +206,13 @@ func run(cmd *cobra.Command, argv []string) {
 			if addOnInstallationParam != nil {
 				dflt = addOnInstallationParam.Value()
 			}
-
+			question := param.Name()
+			// add a prompt to question name to indicate if the boolean param is required to be true
+			if param.ValueType() == "boolean" && param.Validation() == "^true$" {
+				question = fmt.Sprintf("%s (required to be true)", param.Name())
+			}
```
vkareh, line 213: What's the use-case for having a parameter whose only possible value is `true`?

ciaranRoche: Its basically, agree to terms and conditions box. The RHOAM team are currently using it https://gitlab.cee.redhat.com/service/managed-tenants/-/blob/main/addons/rhoams/metadata/stage/addon.yaml#L105-112

It translates well in the UI, as it is flagged as required and you can not proceed until you select it, but in the CLI and API it seems to get lost in translation. As it is not picked up until validation is done. 

This change will add the prompt that it needs to be true, allowing the user to drill down `?` for the description for more info if they wish. I am also adding a `validation_error` to the addon metadata https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/3259, so the addon maintainer can specify a better error message if the validation fails. My thinking is having a clearer validation error message instead of saying `it must match <<regular expression>>` might negate the need for having the prompt.

vkareh: okay, this makes sense. Now, isn't having a Boolean field marked as Required the same as having the Validation `^true$`? Due to how the data is processed, you can't really mark a boolean as `false` if it's required (I guess unless we take the value as a pointer and consider it a tri-state).

In any case, I think this is fine.

ciaranRoche: Not really, because `Required` only checks if a value is passed, and it doesn't check the actual values value. In the UI, there is no way to not send a boolean value, as the checkbox defaults to false. This is why we are supporting this piece of regex validation to allow for a boolean that has to be true, and mimic a confirmation box in the UI. Its just unfortunate it does not really translate too good to the CLI or API.  

I guess ideally, we probably should have a new parameter type, something like `confirmation` that is a required boolean that has to be true.

vkareh: okay I see - right, we take parameters all as strings and then parse/validate in the API :facepalm: 

Alright, let's not rewrite all this API logic for the sake of a CLI prompt :)

cmd/edit/addon/cmd.go
```diff
@@ -206,9 +206,13 @@ func run(cmd *cobra.Command, argv []string) {
 			if addOnInstallationParam != nil {
 				dflt = addOnInstallationParam.Value()
 			}
-
+			question := param.Name()
+			// add a prompt to question name to indicate if the boolean param is required to be true
+			if param.ValueType() == "boolean" && param.Validation() == "^true$" {
+				question = fmt.Sprintf("%s (required to be true)", param.Name())
```
vkareh, line 212: Can you change the string to `"%s (required)"`, I think the rest is implicit.

cmd/edit/addon/cmd.go
```diff
@@ -206,9 +206,13 @@ func run(cmd *cobra.Command, argv []string) {
...
+			}
 			input := interactive.Input{
-				Question: param.Name(),
+				Question: question,
 				Help:     fmt.Sprintf("%s: %s", param.ID(), param.Description()),
 				Required: param.Required(),
```
vkareh, line 217: Can you ensure that `Required` is always `true` for these scenarios? Otherwise it will show `"Foo Param (required) (optional)"`

ciaranRoche: thats not confusing at all :laughing:

cmd/edit/addon/cmd.go
```diff
@@ -206,9 +206,13 @@ func run(cmd *cobra.Command, argv []string) {
 			if addOnInstallationParam != nil {
 				dflt = addOnInstallationParam.Value()
 			}
-
+			question := param.Name()
+			// add a prompt to question name to indicate if the boolean param is required to be true
+			if param.ValueType() == "boolean" && param.Validation() == "^true$" {
```
vkareh, line 211: Since we have a regexp validation, can you add an interactive validator? Basically this:
```
if param.ValueType() == "boolean" && param.Validation() == "^true$" {
  truthValidator := interactive.RegExp(param.Validation())
}
input := interactive.Input{
  Question: question,
  Help:     fmt.Sprintf("%s: %s", param.ID(), param.Description()),
  Required: param.Required(),
    truthValidator,
  },
}
```
Or something like that... this way the prompt will only let the user through if they select `true`

pkg/interactive/validation.go
```diff
@@ -101,3 +100,22 @@ func RegExp(restr string) Validator {
...
+	return func(val interface{}) error {
+		if boolVal, ok := val.(bool); ok {
+			var val string
+			if boolVal {
+				val = "true"
+			} else {
+				val = "false"
+			}
+			if !re.MatchString(val) {
+				return fmt.Errorf("%s does not match regular expression %s", val, re.String())
```
vkareh, line 115: Since you have a function specifically for this, why not have an error that represents the boolean nature: `"%s can only be true or false"` or something along those lines.

pkg/interactive/validation.go
```diff
@@ -20,12 +20,11 @@ package interactive
 
 import (
 	"fmt"
+	"github.com/AlecAivazis/survey/v2"
 	"net"
 	"net/url"
 	"os"
 	"regexp"
-
-	"github.com/AlecAivazis/survey/v2"
```
vkareh, line 28: noooo :laughing:

cmd/edit/addon/cmd.go
```diff
@@ -213,6 +213,14 @@ func run(cmd *cobra.Command, argv []string) {
 				Required: param.Required(),
 			}
 
+			// add a prompt to question name to indicate if the boolean param is required and check validation
+			if param.ValueType() == "boolean" && param.Validation() == "^true$" {
+				input.Question = fmt.Sprintf("%s (required)", param.Name())
```
vkareh, line 218: You're missing `input.Required = true`

pkg/interactive/validation.go
ciaranRoche: Yeah I played around with that, I am unsure as it is a boolean it can only be true or false anyway. The regular expression will only be forcing it to be explicitly one of them. In this case it might be better to show the expression. 

We will soon have the custom validation error messages, maybe we can refactor this to return that error instead :thinking:

PR #446, opened 2021-08-31
use default version on create account-roles

ciaranRoche:
Closes : https://issues.redhat.com/browse/SDA-4690


ciaranRoche:
@vkareh if `--mode` is not passed, it will go into interactive mode, do we want the version to be present here, or do we want to add a check `if interactive.Enabled() || version == ""` so we just default to the version. Or is it okay to go interactive if `mode` flag is not set

vkareh:
@ciaranRoche Nah, if the lack of `--mode` drops you into interactive mode, then it's okay to see the version prompt.

PR #454, opened 2021-09-03
Autocomplete cluster names on --cluster flag

vkareh:
Also since the --cluster logic is the same across most commands, we
centralize the flag definition.


vkareh:
To test this, make sure there are clusters in your account, then run `rosa describe cluster -c <TAB><TAB>` - it also works to autocomplete partial names of clusters.

vkareh:
@ciaranRoche - I've added support for bash (default), zsh, fish, and powershell. Use `rosa completion -h` to view your configuration options.

PR #455, opened 2021-09-03
SDA-4651: list operator roles for cluster

ciaranRoche:
Closes : https://issues.redhat.com/browse/SDA-4651

Usage : 
- `rosa list operator-roles`
- `rosa list operator-roles -c mycluster`


pvasant:
@ciaranRoche  @vkareh Maybe we should skip a couple of columns that are long to display the result.

pvasant:
@ciaranRoche  also i was thinking if  we have the cluster id and we anyways go get the cluster details we can directly  use the operator name and fetch the roles that would help us with delay

ciaranRoche:
> @ciaranRoche also i was thinking if we have the cluster id and we anyways go get the cluster details we can directly use the operator name and fetch the roles that would help us with delay . Thoughts?

@pvasant Yeah I like this idea, once https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/3301 is merged, we can get back the list of operator roles, and fetch those specific roles instead. 
For listing all roles, it will still be slow.

> @ciaranRoche @vkareh Maybe we should skip a couple of columns that are long to display the result.
> 
> Are we supporting jso/yaml output. looks like it is needed

I added support for json/yaml output. I also reduced the columns to operator name/ arn/ cluster id. WDYT?

vkareh:
I'm wondering if the new endpoint in https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/3301 should also have `GET /clusters/archived` in addition to `GET /clusters/archived/<id>`. This way we can get a filtered list and do something like `GET /clusters/archived&product_id=rosa&aws_account_id=123456789012&etc` and thus get the full list of operator roles from active and archived clusters. Then we can highlight in the output those that are unused and need cleanup.

It may also help with speed as we can fetch a list of known role names, rather than get the full AWS list (which can be thousands).

Thoughts?

ciaranRoche:
> I'm wondering if the new endpoint in https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/3301 should also have `GET /clusters/archived` in addition to `GET /clusters/archived/<id>`. This way we can get a filtered list and do something like `GET /clusters/archived&product_id=rosa&aws_account_id=123456789012&etc` and thus get the full list of operator roles from active and archived clusters. Then we can highlight in the output those that are unused and need cleanup.
> 
> It may also help with speed as we can fetch a list of known role names, rather than get the full AWS list (which can be thousands).

Yeah I like this idea, it would definitely speed up listing roles. Alternatively, have we thought about using `paths`, if we have all rosa roles under a specific path we could really speed up and potentially reduce calls to aws. 
Like to list all rosa roles we could do something like:
```
aws iam list-role --path-prefix /rosa/
```
Just thinking as you said there could be thousands of defined roles in an account, organising all roles we create under a specific path may help, as filtering on tags is very slow at the moment. Down side I guess would be roles created before we added the path, we would lose backward compatibility

vkareh:
@ciaranRoche
> Alternatively, have we thought about using paths, if we have all rosa roles under a specific path we could really speed up and potentially reduce calls to aws.

Yes, that would be awesome, but unfortunately we cannot enforce paths for roles since customer can always opt-out of those and leave us going back to searching by tags. The reality is that until the AWS IAM API supports searching roles by tags we're stuck having to improvise.

I think we should use the OCM API to determine current and past roles, as if it's not in our database, then we have no business cleaning it up.

vkareh:
/hold
Until we merge the archived clusters API endpoint

ciaranRoche:
@vkareh @pvasant PTAL. I removed fetching the policies and it has become much faster. I have added the spinner too. WDYT?

cmd/list/operatorroles/cmd.go
```diff
@@ -0,0 +1,193 @@
...
+	activeClusters, err := ocmClient.GetClusters(awsCreator, 1000)
+	if err != nil {
+		reporter.Errorf("Failed to list clusters : '%v'", err)
+		os.Exit(1)
+	}
+
+	// get all archieved clusters
+	archivedClusters, err := ocmClient.ListAccountArchivedClusters(awsCreator.AccountID)
+	if err != nil {
+		reporter.Errorf("Failed to list archieved clusters : '%v'", err)
```
vkareh, line 131: archieved -> archived

cmd/list/operatorroles/cmd.go
```diff
@@ -0,0 +1,193 @@
...
+		os.Exit(1)
+	}
+
+	// print message to indicate there may be some wait time
+	totalClusters := len(activeClusters) + len(archivedClusters)
+	reporter.Infof("Calculating roles for '%d' clusters", totalClusters)
```
vkareh, line 137: This message is wrong when used with `-c foo`. Maybe something like
```go
if clusterKey != "" {
  reporter.Infof("Fetching roles for cluster '%s'", clusterKey)
} else {
  reporter.Infof("Fetching roles for %d clusters", totalClusters)
}
```
(also, don't put the %d in quotes, it looks strange)

pvasant:
@ciaranRoche  do we  need a spinner for listing without operator roles? also, can we remove some of the columns? may role ARN is too big and occupies more space

cmd/list/operatorroles/cmd.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+
+		// Try to find the cluster:
+		reporter.Debugf("Loading cluster '%s'", clusterKey)
+		cluster, err = ocmClient.GetClusterOrArchived(clusterKey, awsCreator)
+		if err != nil {
+			fmt.Println(err.Error())
+			reporter.Errorf("Failed to get cluster '%s': %v", clusterKey, err)
+			os.Exit(1)
+		}
```
vkareh, line 118: Should check here if cluster is _not_ STS, we should error out with an appropriate message

cmd/list/operatorroles/cmd.go
```diff
@@ -0,0 +1,216 @@
...
+		reporter.Errorf("Failed to list archived clusters : '%v'", err)
+		os.Exit(1)
+	}
+
+	// print message to indicate there may be some wait time
+	totalClusters := len(activeClusters) + len(archivedClusters)
+	if clusterKey != "" {
+		reporter.Infof("Fetching roles for cluster '%s'", clusterKey)
+	} else {
+		reporter.Infof("Fetching roles for %d clusters", totalClusters)
```
vkareh, line 140: It dawned on my that this number may not make sense to people... I just ran it and it said `Fetching roles for 86 clusters`, but then the list had only 5 clusters in it. Maybe skip the number or change the message to something like `Fetching roles for %d active clusters and %d archived clusters`, what do you think?

ciaranRoche:
@vkareh @pvasant PTAL

pkg/ocm/clusters.go
```diff
@@ -604,3 +639,44 @@ func (c *Client) ResumeCluster(clusterID string) error {
...
+		return clusters, err
+	}
+
+	response.Items().Each(func(cluster *cmv1.Cluster) bool {
+		clusters = append(clusters, cluster)
+		return true
+	})
+
+	return clusters, nil
+}
```
vkareh, line 682: This function does not fetch all clusters, only the first 100 in the database or so. Can we change this? I tested it and it failed to find a cluster that I created and deleted last night. (It worked if I specificied the `-c $id`)

pkg/ocm/clusters.go
```diff
@@ -604,3 +639,44 @@ func (c *Client) ResumeCluster(clusterID string) error {
...
+
+func (c *Client) ListAccountArchivedClusters(accountID string) ([]*cmv1.Cluster, error) {
+	var clusters []*cmv1.Cluster
+
+	query := fmt.Sprintf("product.id like '%%rosa%%' and aws.sts.role_arn like '%%%s%%'", accountID)
```
vkareh, line 668: I think this should be `product.id = 'rosa'`, no need to use `like` since it's an exact string.

pkg/ocm/clusters.go
```diff
@@ -604,3 +639,44 @@ func (c *Client) ResumeCluster(clusterID string) error {
 
 	return nil
 }
+
+//Get all the archived ROSA STS clusters
+func (c *Client) GetArchivedCluster(clusterKey string) (clusters []*cmv1.Cluster, err error) {
+	query := fmt.Sprintf("id = '%s' OR name = '%s' OR external_id = '%s'",
+		clusterKey, clusterKey, clusterKey,
+	)
```
vkareh, line 647: This should also filter by `rosa` and clusters in the current account, to be consistent:
```
	query := fmt.Sprintf("%s AND (id = '%s' OR name = '%s' OR external_id = '%s')",
		getClusterFilter(creator),
		clusterKey, clusterKey, clusterKey,
	)
```

ciaranRoche:
closing for now, due to being de-prioritised

PR #461, opened 2021-09-15
cluster: Support custom properties

vkareh:
This hidden flag allows for the internal use of cluster properties.

Usage:
```
$ rosa create cluster --properties key:val --properties boolean --properties support:'{"json":"yes"}'
{
  ...
  "properties": {
    "key": "val",
    "boolean": "true"
    "support": "{\"json\":\"yes\"}"
  }
}
```


jeefy:
/lgtm
You da best

PR #463, opened 2021-09-16
aws: Force manual mode when getting AccessDenied

vkareh:
Since the resulting AWS commands are useful even without the correct
permissions in the current account, we force the mode to manual if the
current user is not allowed to verify the existence of
operator-roles/oidc-provider.

### DIFF
--- a/cmd/create/oidcprovider/cmd.go
+++ b/cmd/create/oidcprovider/cmd.go
@@ -151,8 +151,12 @@ func run(cmd *cobra.Command, _ []string) {
 	// Check to see if IAM operator roles have already created
 	missingRoles, err := validateOperatorRoles(awsClient, cluster)
 	if err != nil {
-		reporter.Errorf("Failed to validate operator roles exist: %s", err)
-		os.Exit(1)
+		if strings.Contains(err.Error(), "AccessDenied") {
+			reporter.Debugf("Failed to verify if operator roles exist: %s", err)
+		} else {
+			reporter.Errorf("Failed to verify if operator roles exist: %s", err)
+			os.Exit(1)
+		}
 	}
 
 	if len(missingRoles) > 0 {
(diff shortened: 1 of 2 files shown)


ciaranRoche:
Just one small comment, it is a little confusing in the flow as an error comes up and the user can continue with the prompt. 

```
rosa create operator-roles --cluster croche-test
E: Failed to validate if operator roles exist: AccessDenied: User: arn:aws:iam::765374464689:user/test-user-no-permissions is not authorized to perform: iam:GetRole on resource: role croche-test-i7w6-openshift-cloud-credential-operator-cloud-crede
        status code: 403, request id: 0de750f4-f753-4942-89fd-95b1693d71ba
? Operator policy prefix (optional): 
? Permissions boundary ARN (optional): 
W: Forcing manual mode.
I: Run the following commands to create the operator roles:
```
Would it be worth a prompt between the error and the optional params :thinking: 

Other than that, works well so wouldn't class it as a blocker to this LGTM :+1:

cmd/create/oidcprovider/cmd.go
```diff
@@ -148,11 +148,19 @@ func run(cmd *cobra.Command, _ []string) {
...
+	mode := args.mode
+	autoMode := cmd.Flags().Changed("mode") && mode == "auto"
+	forceManual := false
+
 	// Check to see if IAM operator roles have already created
 	missingRoles, err := validateOperatorRoles(awsClient, cluster)
 	if err != nil {
-		reporter.Errorf("Failed to validate operator roles exist: %s", err)
-		os.Exit(1)
+		reporter.Errorf("Failed to validate if operator roles exist: %s", err)
```
pvasant, line 158: Can we change this message to just print for other errors and for access denied we can print saying that "access is denied" trying running manually .. something like that?

vkareh:
@pvasant @ciaranRoche - I've changed the approach here. Instead of having to force `--mode manual` or any of that, I just silently ignore `AccessDenied` for non-essential API calls. This way the user can just move on with creating the resources, either manually or automatically (they _may_ still have perms to create, but not to list)

pvasant:
LGTM. but this might conflict with other merge

PR #470, opened 2021-09-28
SDA-4829 update getThumbprints to use http package instead of tls

ciaranRoche:
Closes : https://issues.redhat.com/browse/SDA-4829

Copied thumbprints from http to tls to show no differences in the peer certificates returned : https://gist.github.com/ciaranRoche/3c0054a28c4d7a608a7d8ebc4e06ffa0


ciaranRoche:
@vkareh PTAL

PR #475, opened 2021-10-05
SDA-4744 : Add account role validation on cluster create

ciaranRoche:
closes : https://issues.redhat.com/browse/SDA-4744

Validate : 
- create account roles for 4.7 version
- create sts cluster 4.8 with 4.7 account role
- ensure error is returned


cmd/create/cluster/cmd.go
```diff
@@ -722,6 +723,58 @@ func run(cmd *cobra.Command, _ []string) {
...
+						ARN, policyVersion.String(), openshiftVersion.String())
+					os.Exit(1)
+				}
+			}
+		}
+		// if no version tag is found, error out as we can not guarantee policy's are compatible with openshift version
+		if !foundVersion {
+			reporter.Errorf("Could not validate openshift policy version from arn '%s'", ARN)
+			os.Exit(1)
+		}
```
vkareh, line 775: I think we have a function `isRoleCompatible` that does this already. Can you reuse that function (and modify it if necessary), I'd rather have all this code inside `pkg/aws/policies`

cmd/create/cluster/cmd.go
```diff
@@ -722,6 +722,36 @@ func run(cmd *cobra.Command, _ []string) {
...
+
+		validVersion, err := awsClient.HasCompatibleVersionTags(role.Tags, getVersionMinor(version))
+		if err != nil {
+			reporter.Errorf("Could not validate Role '%s' : %v", ARN, err)
+			os.Exit(1)
+		}
+		if !validVersion {
+			reporter.Errorf("Can not proceed with cluster install, policy '%s' version not "+
```
ciaranRoche, line 748: @vkareh what kind of error should we give back here, should we prompting to create new account roles?

vkareh: Yes, let's try a message like
```
Account role '%s' is not compatible with version %s. Run 'rosa create account-roles' to create compatible roles and try again.
```

PR #476, opened 2021-10-06
Hide reporter Info and Warn log when redirecting to a file

zgalor:
Related: SDA-4832


zgalor:
@vkareh @pvasant @ciaranRoche PTAL

vkareh:
I think we should not do it this generally. Some users may pipe the entire output for whatever reason, and the use-case in this ticket is specifically when there's an operation that prints out commands to run (`rosa create account-roles`, etc.) if the user selects `--mode manual` and they pipe it, the can technically do something `rosa create account-roles --mode manual | sh` to auto-run the commands.

zgalor:
> I think we should not do it this generally. Some users may pipe the entire output for whatever reason, and the use-case in this ticket is specifically when there's an operation that prints out commands to run (`rosa create account-roles`, etc.) if the user selects `--mode manual` and they pipe it, the can technically do something `rosa create account-roles --mode manual | sh` to auto-run the commands.

@vkareh what other command should be handled, other than `create account-roles` ?

zgalor:
@vkareh rePTAL

PR #485, opened 2021-10-14
sts: Group account roles by prefix

vkareh:
When auto-selecting account roles to use on a cluster, ensure that all
roles match the same prefix group. This ensures consistency and avoids
the auto-find to fall into a mix-and-match approach.

Also extract the prefix from said roles and feed it to the 'create
operator-roles' command so that the correct policies are attached to
those as well.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -584,7 +584,6 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	hasRoles := false
-	rolePrefix := aws.DefaultPrefix
 
 	if isSTS && roleARN == "" {
 		minor := getVersionMinor(version)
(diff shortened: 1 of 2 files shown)


vkareh:
@ciaranRoche I cherry-picked some of your changes from #486 into this PR. Please take a look

cmd/create/cluster/cmd.go
```diff
@@ -587,49 +587,93 @@ func run(cmd *cobra.Command, _ []string) {
...
+				case aws.InstallerAccountRole:
+					roleARN = selectedARN
+				case aws.SupportAccountRole:
+					supportRoleARN = selectedARN
+				case aws.ControlPlaneAccountRole:
+					controlPlaneRoleARN = selectedARN
+				case aws.WorkerAccountRole:
+					workerRoleARN = selectedARN
+				}
+				hasRoles = true
```
ciaranRoche, line 675: We run into an issue here I think, if one of the roles does not have the prefix, hasRoles can still end up true if the last role in the list has the prefix. Which we will then skip the manual prompt to enter the role. 
Example : https://gist.github.com/ciaranRoche/53c19f395688f021048b4aa25ef68564#file-when_on_role_does_not_exist

Maybe we can drop the `hasRoles` flag and explicitly check if the `role == ""` when determining to prompt the user input for the role. Will also need to add a validator to those prompts to ensure the correct prefix.

PR #487, opened 2021-10-15
SDA-4916 add validation to sts cluster create mode flag

ciaranRoche:
Currently mode is not validated until the operator roles func is called. This leads to a confusing error, and breaks the mode flow. This change validates mode value before the cluster is created 

Closes : https://issues.redhat.com/browse/SDA-4916


ciaranRoche:
@vkareh @pvasant PTAL

PR #492, opened 2021-10-19
add --mode to create command output

ciaranRoche:
This change : 

- Adds --mode to create command output
- Reduces message from error to info on delete operator roles if no roles are found
- Reduces message from error to info on delete oidc provider if no provider is found

[Verification](https://gist.github.com/ciaranRoche/a21d33a4e4d303fa268a508cb3f8df46)

### DIFF
--- a/cmd/dlt/oidcprovider/cmd.go
+++ b/cmd/dlt/oidcprovider/cmd.go
@@ -141,8 +141,8 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 	if providerARN == "" {
-		reporter.Errorf("Cluster '%s' doesn't have OIDC provider associated with it.", clusterKey)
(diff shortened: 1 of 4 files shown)


cmd/dlt/oidcprovider/cmd.go
```diff
@@ -141,8 +141,8 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 	if providerARN == "" {
-		reporter.Errorf("Cluster '%s' doesn't have OIDC provider associated with it.", clusterKey)
-		os.Exit(1)
+		reporter.Infof("Cluster '%s' doesn't have OIDC provider associated with it.", clusterKey)
+		return
```
pvasant, line 145: @ciaranRoche  what is the purpose of return instead of exit?

PR #512, opened 2021-11-02
Clarify `verify permissions` cmd is only for non-STS clusters

zgalor:
Related: [SDA-4729](https://issues.redhat.com/browse/SDA-4729)


zgalor:
@ciaranRoche @pvasant @vkareh PTAL

ciaranRoche:
/lgtm 

image

PR #514, opened 2021-11-03
Check for pre-existing operator roles and error if they exist

zgalor:
Related: [SDA-4624](https://issues.redhat.com/browse/SDA-4624)

```
$ ./rosa create cluster --sts --operator-roles-prefix zg-oct-14-u8z3 -c zg-nov-3
W: More than one Installer role found
? Installer role ARN: arn:aws:iam::765374464689:role/ManagedOpenShift-Installer-Role
I: Using arn:aws:iam::765374464689:role/ManagedOpenShift-ControlPlane-Role for the ControlPlane role
I: Using arn:aws:iam::765374464689:role/ManagedOpenShift-Worker-Role for the Worker role
I: Using arn:aws:iam::765374464689:role/ManagedOpenShift-Support-Role for the Support role
E: Error validating role: A role named 'zg-oct-14-u8z3-openshift-machine-api-aws-cloud-credentials' already exists. Please delete the existing role, or provide a different prefix
```

### DIFF
--- a/pkg/aws/policies.go
+++ b/pkg/aws/policies.go
@@ -238,6 +238,25 @@ func (c *awsClient) EnsureRole(name string, policy string, permissionsBoundary s
 	return roleArn, nil
 }
 
+func (c *awsClient) ValidateRoleNameAvailable(name string) (err error) {
+	_, err = c.iamClient.GetRole(&iam.GetRoleInput{
+		RoleName: aws.String(name),
+	})
+	if err == nil {
+		// If we found an existing role with this name we want to error
+		return fmt.Errorf("A role named '%s' already exists. "+
+			"Please delete the existing role, or provide a different prefix", name)
+	}
+
+	if aerr, ok := err.(awserr.Error); ok {
+		switch aerr.Code() {
+		case iam.ErrCodeNoSuchEntityException:
+			// This is what we want
+			return nil
+		}
+	}
+	return fmt.Errorf("Error validating role name '%s': %v", name, err)
+}
 func (c *awsClient) updateAssumeRolePolicyPrincipals(policy string, role *iam.Role) (string, bool, error) {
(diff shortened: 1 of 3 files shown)


zgalor:
@vkareh @pvasant @ciaranRoche PTAL

pkg/aws/policies.go
```diff
@@ -238,6 +238,24 @@ func (c *awsClient) EnsureRole(name string, policy string, permissionsBoundary s
 	return roleArn, nil
 }
 
+func (c *awsClient) ValidateRoleNameAvailable(name string) (err error) {
+	_, err = c.iamClient.GetRole(&iam.GetRoleInput{
+		RoleName: aws.String(name),
+	})
+	if err == nil {
+		// If we found an existing role with this name we want to error
+		return fmt.Errorf("A role named '%s' already exists", name)
```
ciaranRoche, line 247: should we give some action for the user to take here? Feels like the error is kinda confusing, would it be better to return an info message with some options to either create cluster with different role prefix, or delete existing role prefix if its not in use :thinking:

zgalor: Something like 'Please delete the existing role, or use a different prefix'?

ciaranRoche: yeah i think its a common pattern in the cli, to provide some [next steps](https://github.com/openshift/rosa/blob/master/cmd/create/cluster/cmd.go#L1701-L1721) especially if it can be solved with the CLI

zgalor: fixed

pvasant:
@zgalor  i think we need to  do this 

"if when creating operator roles for a cluster rosa detects that they exist, we can compare the OIDC provider that they're attached to and error out"

zgalor:
> @zgalor i think we need to do this
> 
> "if when creating operator roles for a cluster rosa detects that they exist, we can compare the OIDC provider that they're attached to and error out"

IIUC per my offline chat with @vkareh we can skip it and just error.
@vkareh ?

PR #524, opened 2021-11-10
ROSA: Support editing cluster-wide proxy

zvikorn:

https://issues.redhat.com/browse/SDA-4733

### DIFF
--- a/cmd/edit/cluster/cmd.go
+++ b/cmd/edit/cluster/cmd.go
@@ -19,19 +19,21 @@ package cluster
 import (
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"time"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
-	"github.com/spf13/cobra"
-
 	"github.com/openshift/rosa/pkg/aws"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/interactive/confirm"
 	"github.com/openshift/rosa/pkg/logging"
 	"github.com/openshift/rosa/pkg/ocm"
 	rprtr "github.com/openshift/rosa/pkg/reporter"
+	"github.com/spf13/cobra"
 )
+
+const doubleQuotesToRemove = "\"\""
 
 var args struct {
 	// Basic options
@@ -41,6 +43,9 @@ var args struct {
 	// Networking options
 	private                   bool
 	disableWorkloadMonitoring bool
+	httpProxy                 string
+	httpsProxy                string
+	additionalTrustBundleFile string
 }
 
 var Cmd = &cobra.Command{
@@ -92,6 +97,26 @@ func init() {
 		"Enables you to monitor your own projects in isolation from Red Hat Site Reliability Engineer (SRE) "+
 			"platform metrics.",
 	)
+	flags.StringVar(
+		&args.httpProxy,
+		"http-proxy",
+		"",
+		"A proxy URL to use for creating HTTP connections outside the cluster. The URL scheme must be http.",
+	)
+
+	flags.StringVar(
+		&args.httpsProxy,
+		"https-proxy",
+		"",
+		"A proxy URL to use for creating HTTPS connections outside the cluster.",
+	)
+
+	flags.StringVar(
+		&args.additionalTrustBundleFile,
+		"additional-trust-bundle-file",
+		"",
+		"A file contains a PEM-encoded X.509 certificate bundle that will be "+
+			"added to the nodes' trusted certificate store.")
 }
 
 func run(cmd *cobra.Command, _ []string) {
@@ -167,6 +192,47 @@ func run(cmd *cobra.Command, _ []string) {
 			"Any optional fields can be ignored and will not be updated.")
 	}
 
+	/*There are three possible options of input from the user when a prompt shows up:
+	1) The user presses the 'enter' button ---> interactive 'getString' method returns either an existing value if exists
+	   (the one that shows up as part of the question, i.e. - ? HTTP proxy: http://site.com),
+	   or double quotes ("") if no existing value. In that case, we send to OCM nil as we do not want any change.
+	2) In case the user wants to remove an existing value, an empty string ("") should be entered by the user -->
+	   interactive 'getString' method returns "\"\"". In that case, we send OCM double quotes to remove the existing value.
+	3) The user enters any other value ---> a simple and straightforward case. */
+
+	enableProxy := false
+	useExistingVPC := false
+	var httpProxy *string
+	var httpProxyValue string
+	if cmd.Flags().Changed("http-proxy") {
+		httpProxyValue = args.httpProxy
+		httpProxy = &httpProxyValue
+	}
+	var httpsProxy *string
+	var httpsProxyValue string
+	if cmd.Flags().Changed("https-proxy") {
+		httpsProxyValue = args.httpsProxy
+		httpsProxy = &httpsProxyValue
+	}
+	var additionalTrustBundleFile *string
+	var additionalTrustBundleFileValue string
+	if cmd.Flags().Changed("additional-trust-bundle-file") {
+		additionalTrustBundleFileValue = args.additionalTrustBundleFile
+		additionalTrustBundleFile = &additionalTrustBundleFileValue
+	}
+
+	if httpProxy != nil || httpsProxy != nil || additionalTrustBundleFile != nil {
+		enableProxy = true
+		useExistingVPC = true
+	}
+
+	if len(cluster.AWS().SubnetIDs()) == 0 &&
+		((httpProxy != nil && *httpProxy != "") || (httpsProxy != nil && *httpsProxy != "") ||
+			(additionalTrustBundleFile != nil && *additionalTrustBundleFile != "")) {
+		reporter.Errorf("Cluster-wide proxy is not supported on clusters using the default VPC")
+		os.Exit(1)
+	}
+
 	var private *bool
 	var privateValue bool
 	if cmd.Flags().Changed("private") {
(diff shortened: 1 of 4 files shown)


cmd/edit/cluster/cmd.go
```diff
@@ -189,10 +216,185 @@ func run(cmd *cobra.Command, _ []string) {
...
+	}
+
+	if httpProxy != nil || httpsProxy != nil || additionalTrustBundleFile != nil {
+		enableProxy = true
+		useExistingVPC = true
+	}
+
+	if len(cluster.AWS().SubnetIDs()) > 0 {
+		useExistingVPC = true
+	}
```
vkareh, line 247: This is likely wrong. Since cluster-wide proxy only applies to BYO-VPC clusters, if the user sets any of the proxy flags, but there are no subnets in the cluster, then the command should error out.

pkg/interactive/validation.go
```diff
@@ -65,6 +68,9 @@ func IsCert(filepath interface{}) error {
 		if s == "" {
 			return nil
 		}
+		if s == "\"\"" {
+			return nil
+		}
```
vkareh, line 73: Why do you need this check? If the user enters "" as the actual string then that's wrong, we shouldn't let it through.

zvikorn: @vkareh  so what should the user enter in order to delete the proxy value?

cmd/edit/cluster/cmd.go
zvikorn: @vkareh 
```
if len(cluster.AWS().SubnetIDs()) > 0 {
	useExistingVPC = true
}

if useExistingVPC && !enableProxy && interactive.Enabled() {
	enableProxyValue, err := interactive.GetBool(interactive.Input{
		Question: "Update cluster-wide proxy",  **<---- I want to present this question only if I know that this is BYOVPC cluster which is indicated by the first 'if' check above**.
```

vkareh: Correct, but my point is that we should go further than this and have something like
```go
if len(cluster.AWS().SubnetIDs()) > 0 && (httpProxy != "" || httpsProxy != "" || additionalTrustBundle != "") {
  reporter.Errorf("Cluster-wide proxy is not supported on clusters using the default VPC")
  os.Exit(1)
}
```

pkg/interactive/validation.go
vkareh: It should be either `nil` or an empty string - if it's `nil`, no action is taken; if it's an empty string, the attribute is updated to an empty string

zvikorn: in case I remove this check, I get:
**X Sorry, your reply was invalid: file '""' does not have a valid file extension**
? Additional trust bundle file path (optional): [? for help] 

There are three possible options of input from the user:
1) the user presses the '**enter**' button ---> rosa cli code returns either an existing value if exists, or double quotes ("") if no existing value. This is what happens regardless of the proxy feature.
2) the user enters double quotes ("') to remove an existing value --> rosa cli code returns "\"\"".  Again, regardless the proxy feature.
3) the user enters a valid value  ---> simple and straightforward case.

So the check above is for a case the user wants tp remove the current certificate by entering "" (like in OCM palyload), and we do not want to parse it.

zvikorn:
@vkareh I replied to your comments. Also, added more comments inside the code to make the logic clear. PTAL

pkg/ocm/clusters.go
```diff
@@ -33,6 +33,8 @@ import (
 	errors "github.com/zgalor/weberr"
 )
 
+const DoubleQuotesToRemove = "\"\""
```
zvikorn, line 36: @vkareh not sure about this term. Maybe you have something more clear....

vkareh: no, just remove this and have the user actually delete the contents of the field:

if the cluster has data in httpProxy, the prompt will be populated:
- if they leave the populated data in place, we don't modify it
- if they change the populated data, we update the cluster
- if they delete the populated data, we delete it from the cluster

if the cluster does not have data, the prompt will show up empty:
- if they leave the prompt empty, we don't modify it
- if they enter data, we send it to the cluster

What other options do you see as necessary here?

zvikorn: @vkareh I still do not understand how the user deletes the existing value. The prompt is like:
```
? HTTP proxy: [? for help] (http://aa.com)
```
You cannot delete it. Pressing backspace does not delete the value as it's part of the question.

zvikorn:
@vkareh it looks like that now:
```
? Update cluster-wide proxy (optional): Yes
? To remove any existing cluster-wide proxy value or an existing additional-trust-bundle value, enter a set of double quotes ("")
? HTTP proxy (optional): 
? HTTPS proxy (optional): [? for help]
```
PTAL

pkg/interactive/validation.go
```diff
@@ -20,6 +20,7 @@ package interactive
 
 import (
 	"fmt"
+	"github.com/openshift/rosa/pkg/ocm"
```
vkareh, line 23: This needs to be moved to the internal imports section below

pkg/interactive/validation.go
```diff
@@ -65,6 +69,9 @@ func IsCert(filepath interface{}) error {
 		if s == "" {
 			return nil
 		}
+		if s == ocm.DoubleQuotesToRemove {
```
vkareh, line 72: This creates an unnecessary dependency on the `ocm` package. I think this `const` should live inside this package instead so that the import above can be removed.

pkg/ocm/clusters.go
```diff
@@ -410,6 +412,42 @@ func (c *Client) UpdateCluster(clusterKey string, creator *aws.Creator, config S
 		)
 	}
 
+	if config.HTTPProxy != nil || config.HTTPSProxy != nil || config.AdditionalTrustBundleFile != nil {
+		clusterProxyBuilder := cmv1.NewProxy()
+		if config.HTTPProxy != nil {
+			if *config.HTTPProxy == "" {
+				// user pressed the 'enter' button. No change is required.
+				config.HTTPProxy = nil
+			} else if *config.HTTPProxy == DoubleQuotesToRemove {
```
vkareh, line 421: You should not have to compare to `DoubleQuotesToRemove` here - that's why you have pointers already: if the user entered double quotes, then you set the value to an empty string, if the user didn't do anything, then you send `nil`. The point is that the double-quotes is an artifact of the interactive mode and the `ocm` package should not be aware of it.

pkg/ocm/clusters.go
```diff
@@ -410,6 +412,42 @@ func (c *Client) UpdateCluster(clusterKey string, creator *aws.Creator, config S
...
+				clusterProxyBuilder = clusterProxyBuilder.HTTPProxy("")
+			} else {
+				clusterProxyBuilder = clusterProxyBuilder.HTTPProxy(*config.HTTPProxy)
+			}
+		}
+		if config.HTTPSProxy != nil {
+			if *config.HTTPSProxy == "" {
+				// user pressed the 'enter' button. No change is required.
+				config.HTTPSProxy = nil
+			} else if *config.HTTPSProxy == DoubleQuotesToRemove {
```
vkareh, line 432: Same here, as well as below in the trust bundle

pkg/ocm/clusters.go
zvikorn: I must do that, otherwise, if I remove the check to double quotes, cs will get """" (double double quotes)

pkg/interactive/validation.go
zvikorn: defined it locally.

pkg/ocm/clusters.go
vkareh: No, I mean that in the `cmd/edit/cluster/cmd.go` file, you should be checking and replacing this:
```go
if httpProxy != nil {
  if *httpProxy == ocm.DoubleQuotesToRemove {
    *httpProxy = ""
  }
  clusterConfig.HTTPProxy = httpProxy
}
```
This means that double quotes do not need to be checked outside of the `cmd` file (maybe in the interactive validation function, but certainly not in the `ocm` package)

zvikorn: @vkareh ok, I think I can do something. Still trying....

zvikorn:
@vkareh updated the code PTAL

cmd/edit/cluster/cmd.go
```diff
@@ -41,6 +43,10 @@ var args struct {
 	// Networking options
 	private                   bool
 	disableWorkloadMonitoring bool
+	enableProxy               bool
```
vkareh, line 46: `enableProxy` is not used, we should remove it.

cmd/edit/cluster/cmd.go
```diff
@@ -220,12 +287,201 @@ func run(cmd *cobra.Command, _ []string) {
...
+				"'https-proxy', additional-trust-bundle",
+			Default: enableProxy,
+		})
+		if err != nil {
+			reporter.Errorf("Expected a valid proxy-enabled value: %s", err)
+			os.Exit(1)
+		}
+		enableProxy = enableProxyValue
+	}
+	if enableProxy {
```
vkareh, line 306: This should be
```suggestion
	if enableProxy && interactive.Enabled() {
```

cmd/edit/cluster/cmd.go
```diff
@@ -220,12 +287,201 @@ func run(cmd *cobra.Command, _ []string) {
...
+			Default:  def,
+		})
+		if err != nil {
+			reporter.Errorf("Expected a valid http proxy: %s", err)
+			os.Exit(1)
+		}
+
+		if len(httpProxyValue) == 0 { //user skipped the prompt by pressing 'enter'
+			httpProxy = nil
+		} else if len(httpProxyValue) == 2 { //user entered double quotes ("") to remove the existing value
```
vkareh, line 341: I think this should be
```suggestion
		} else if httpProxyValue == doubleQuotesToRemove { //user entered double quotes ("") to remove the existing value
```
no?

cmd/edit/cluster/cmd.go
```diff
@@ -220,12 +287,201 @@ func run(cmd *cobra.Command, _ []string) {
...
+			Help:     cmd.Flags().Lookup("https-proxy").Usage,
+			Default:  def,
+		})
+		if err != nil {
+			reporter.Errorf("Expected a valid https proxy: %s", err)
+			os.Exit(1)
+		}
+		if len(httpsProxyValue) == 0 { //user skipped the prompt by pressing 'enter'
+			httpsProxy = nil
+		} else if len(httpsProxyValue) == 2 { //user entered double quotes ("") to remove the existing value
```
vkareh, line 380: same here

cmd/edit/cluster/cmd.go
```diff
@@ -220,12 +287,201 @@ func run(cmd *cobra.Command, _ []string) {
...
+			Default:  def,
+		})
+		if err != nil {
+			reporter.Errorf("Expected a valid additional trust bundle file name: %s", err)
+			os.Exit(1)
+		}
+
+		if len(additionalTrustBundleFileValue) == 0 { //user skipped the prompt by pressing 'enter'
+			additionalTrustBundleFile = nil
+		} else if len(additionalTrustBundleFileValue) == 2 { //user entered double quotes ("") to remove the existing value
```
vkareh, line 436: and here

vkareh:
This LGTM

@pvasant can you do a last review and merge if it all looks good to you?

PR #533, opened 2021-11-23
create-cluster: Allow FIPS mode support

vkareh:
Since we can support FIPS cluster installations, we add the flag for
FIPS mode. However, we keep it hidden from the end-user since it's still
in experimental mode. Enabling FIPS forces etcd encryption to be
enabled.

https://issues.redhat.com/browse/SDA-5060


cmd/create/cluster/cmd.go
```diff
@@ -1988,7 +2028,9 @@ func buildCommand(spec ocm.Spec, operatorRolesPrefix string) string {
 	if len(spec.SubnetIds) > 0 {
 		command += fmt.Sprintf(" --subnet-ids %s", strings.Join(spec.SubnetIds, ","))
 	}
-	if spec.EtcdEncryption {
+	if spec.FIPS {
+		command += " --fips"
+	} else if spec.EtcdEncryption {
```
pvasant, line 2033: I don't understand this because if fips is enabled etcdEncryption should be enabled is what i infer from the above code

cmd/create/cluster/cmd.go
```diff
@@ -1493,6 +1501,37 @@ func run(cmd *cobra.Command, _ []string) {
...
+			Default:  fips,
+		})
+		if err != nil {
+			reporter.Errorf("Expected a valid fips value: %v", err)
+			os.Exit(1)
+		}
+	}
+
+	etcdEncryption := args.etcdEncryption
+	if interactive.Enabled() && !fips {
```
pvasant, line 1518: why only !fips ? if that is the case  will it help to add the validation 
```
if fips && cmd.Flags().Changed("etcd-encryption") && !etcdEncryption {
		reporter.Errorf("etcd encryption cannot be disabled on clusters with FIPS mode")
		os.Exit(1)
	}
```

before as a warning and give them interactive mode ?  or is it too much

pvasant:
Other than two clarification requested code looks good to me

cmd/create/cluster/cmd.go
vkareh: Yes, but I want it to be implicit - if you set `--fips` you don't need to set etcd encryption, as it will happen automatically, so no need to have both flags.

cmd/create/cluster/cmd.go
vkareh: No, the interactive mode for `etcd-encryption` is hidden if they set `--fips`, however, if the user sets something like `--fips --etcd-encryption=false` that's when this error should show up

PR #537, opened 2021-11-24
link: Allow linking multiple role ARNs

vkareh:
To avoid overridding linked role ARNs, we fetch the label specifically
using its known name, and then verify whether the ARN is in place.
Otherwise we append it to the existing list.


vkareh:
@pvasant 
> @vkareh this fails for the first time when we don't have the sts_user_role . we can may be handled resp.Status() == 404 then assume label doesn't exists

Good catch, thank. It's fixed now.

pvasant:
thank you for fixing. verified and works well.

PR #545, opened 2021-12-02
Add --admin option to create ocm-role command

zgalor:
Related: [SDA-4992](https://issues.redhat.com/browse/SDA-4992)

### DIFF
--- a/cmd/create/ocmrole/cmd.go
+++ b/cmd/create/ocmrole/cmd.go
@@ -351,10 +351,29 @@ func createRoles(reporter *rprtr.Object, awsClient aws.Client, prefix string, po
 		return "", err
 	}
 	if exists {
+		isExistingRoleAdmin, err := awsClient.IsAdminRole(roleName)
+		if err != nil {
+			return "", err
+		}
 		reporter.Warnf("Role '%s' already exists", roleName)
-		return roleARN, nil
+
+		if !isAdmin {
+			if isExistingRoleAdmin {
+				return "", fmt.Errorf("The existing role is an admin role."+
+					" To remove admin capabilities please delete the admin policy and the '%s' tag",
+					tags.AdminRole)
+			}
+			return roleARN, nil
+		}
+
+		if isExistingRoleAdmin {
+			return roleARN, nil
+		}
+
+		if !confirm.Prompt(true, "Add admin policies to '%s' role?", roleName) {
+			return roleARN, nil
+		}
 	}
-	reporter.Debugf("Creating role '%s'", roleName)
 
 	iamTags := map[string]string{
 		tags.RolePrefix:  prefix,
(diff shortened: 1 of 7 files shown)


zgalor:
@vkareh @pvasant PTAL

cmd/create/ocmrole/cmd.go
```diff
@@ -308,46 +355,48 @@ func createRoles(reporter *rprtr.Object, awsClient aws.Client, prefix string, po
...
-			tags.RolePrefix:  prefix,
-			tags.RoleType:    aws.OCMRolePolicyFile,
-			tags.Environment: env,
-		})
-	if err != nil {
-		return "", err
-	}
+	if isAdmin {
+		// tag role with admin tag
+		err = awsClient.AddRoleTag(roleName, tags.AdminRole, "true")
```
vkareh, line 381: Why not add the tag directly to the `iamTags` map and avoid an extra API call?

zgalor: I find the code cleaner like that when all admin changes are in one place, like keeping a separate policy, that we also tag with the admin tag (and we don't tag the regular permission policy with it).

pvasant:
@zgalor  it works fine. @vkareh  Should we have an option for the user to add   admin role to the existing role?

zgalor:
> @zgalor it works fine. @vkareh Should we have an option for the user to add admin role to the existing role? I personally think we need it. We can even use the same command and detect if the user exists and not an admin we can may be show an option. WDYT?

@pvasant I don't understand what you mean by using the same command.
This seems more suitable for `rosa edit ocm-role`

zgalor:
@pvasant cmd is now idempotent. Existing role can be upgraded to admin

cmd/create/ocmrole/cmd.go
```diff
@@ -305,49 +352,69 @@ func createRoles(reporter *rprtr.Object, awsClient aws.Client, prefix string, po
...
-	path = fmt.Sprintf("templates/policies/%s", filename)
+		isExistingRoleAdmin, err := awsClient.IsAdminRole(roleName)
+		if err != nil {
+			return "", err
+		}
 
-	policy, err = aws.ReadPolicyDocument(path)
-	if err != nil {
-		return "", err
+		if isExistingRoleAdmin {
```
pvasant, line 364: @zgalor  I verified and works well. One thing is if the user already has admin role and they choose "No" in the response we are not warning the user that we are not making any changes to the policy. Either we can error out stating they cannot choose "No" to admin policy if it already exists or maybe a warning here that says "Admin already exists and not changes made"

pvasant:
@zgalor Except that one comment above it looks good to me. If we can get approved from @vkareh  we can merge.

cmd/create/ocmrole/cmd.go
vkareh: This is a good idea - if the command is idempotent, it should either set the user requirements or error out:
- `rosa create ocm-role --admin` always ensures that the ocm role is admin
- `rosa create ocm-role` always ensures that the ocm role is non-admin <--- since this is destructive if the role is admin, I would just error out and tell the user to delete the policy instead

zgalor: If it already has admin role, users will not be prompted. 
they will just get the warning `W: Role 'prefix-OCM-Role-id' already exists`, like we currently print for creating a role that already exists

zgalor: @vkareh ok. I'll go with the second option of erroring. I don't think we should revoke admin policies

zgalor:
```
$ ./rosa create ocm-role --prefix zg2 --mode auto
I: Creating ocm role
I: Creating role using 'arn:aws:iam::765374464689:user/zgalor'
? Create the 'zg2-OCM-Role-12541229' role? Yes
W: Role 'zg2-OCM-Role-12541229' already exists
E: There was an error creating the ocm role: The existing role is an admin role. To remove admin capabilities please delete the admin policy and the 'rosa_admin_role' tag
```

pvasant:
@zgalor looks good to me

PR #547, opened 2021-12-03
aws: Remove hard dependency on default region

vkareh:
This forces a fall back on the region configured in the current AWS
session and allows the user to override the region with the use of the
'profile' and 'region' flags.

https://issues.redhat.com/browse/OSD-9245


zgalor:
LGTM

PR #552, opened 2021-12-22
idp: Enable interactive mode when missing required flags

vkareh:
To ensure a consistent experience in interactive mode, instead of
prompting the user for only the required fields, we verify if the flags
are missing and enable interactive mode globally. This ensures that when
dropped into interactive mode no options are skipped.

https://issues.redhat.com/browse/SDA-5229


pvasant:
Works well. Merging.

pvasant:
\lgtm

PR #556, opened 2022-01-05
Generate static assets for STS support permissions

vkareh:
We forgot to re-generate static assets before merging
https://github.com/openshift/rosa/pull/553. This commit just does that
to ensure the latest build contains the new permissions.


jharrington22 (approved):
Thanks @vkareh 

/lgtm

PR #567, opened 2022-01-19
Add gate support in rosa cli cluster upgrade

zgalor:
Related: [SDA-5205](https://issues.redhat.com/browse/SDA-5205)

### DIFF
--- a/cmd/upgrade/cluster/cmd.go
+++ b/cmd/upgrade/cluster/cmd.go
@@ -23,6 +23,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/openshift/rosa/pkg/interactive/confirm"
+
 	"github.com/briandowns/spinner"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
@@ -426,6 +428,31 @@ func run(cmd *cobra.Command, _ []string) {
 	if err != nil {
 		reporter.Errorf("Failed to schedule upgrade for cluster '%s': %v", clusterKey, err)
 		os.Exit(1)
+	}
+
+	// check if the cluster upgrade requires gate agreements
+	gates, err := ocmClient.GetMissingGateAgreements(cluster.ID(), upgradePolicy)
+	if err != nil {
+		reporter.Errorf("Failed to check for missing gate agreements upgrade for cluster '%s': %v", clusterKey, err)
+		os.Exit(1)
+	}
+
+	for _, gate := range gates {
+		reporter.Infof("Gate: %v", gate)
+		if !gate.STSOnly() {
+			// for non sts gates we require user agreement
+			if !confirm.Prompt(true,
+				"I acknowledge that my workloads are no longer using removed APIs. (Background:'%s)",
+				gate.Description()) {
+				os.Exit(0)
+			}
+		}
+		err = ocmClient.AckVersionGate(cluster.ID(), gate.ID())
+		if err != nil {
+			reporter.Errorf("Failed to acknowledge version gate '%s' for cluster '%s': %v",
+				gate.ID(), clusterKey, err)
+			os.Exit(1)
+		}
 	}
 
 	err = ocmClient.ScheduleUpgrade(cluster.ID(), upgradePolicy)
--- a/pkg/ocm/upgrades.go
+++ b/pkg/ocm/upgrades.go
@@ -17,6 +17,8 @@ limitations under the License.
 package ocm
 
 import (
+	"fmt"
+
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 )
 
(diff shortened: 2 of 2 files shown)


pkg/ocm/upgrades.go
```diff
@@ -96,3 +98,79 @@ func (c *Client) CancelUpgrade(clusterID string) (bool, error) {
...
+		Add().
+		Parameter("dryRun", true).
+		Body(upgradePolicy).
+		Send()
+	if err == nil {
+		return []*cmv1.VersionGate{}, nil
+	}
+
+	// parse gates list
+	details, ok := response.Error().GetDetails()
```
zgalor, line 117: @pvasant @vkareh @jhernand @tzvatot I really don't like the following solution.

For our purpose, it would be much better to add a query endpoint to CS with a proprietary API in the SDK like `GetMissingGateAgreements()`, and that would return the list of gates.

However, since we decided to rely on `Error().Details()` I can't use the returned map to read the object, since the sdk doesn't have json annotaion for the fields, so simplest way was to just read the fields we need from the map.
That was the quickest solution.

If we decide to add the API to CS and the SDK, we can change this

cmd/upgrade/cluster/cmd.go
```diff
@@ -428,6 +430,31 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	// check if the cluster upgrade requires gate agreements
+	gates, err := ocmClient.GetMissingGateAgreements(cluster.ID(), upgradePolicy)
+	if err != nil {
+		reporter.Errorf("Failed to check for missing gate agreements upgrade for cluster '%s': %v", clusterKey, err)
```
tzvatot, line 436: "Failed to check for missing gate agreements *required for* upgrade for cluster '%s': %v"

cmd/upgrade/cluster/cmd.go
```diff
@@ -428,6 +430,31 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	// check if the cluster upgrade requires gate agreements
+	gates, err := ocmClient.GetMissingGateAgreements(cluster.ID(), upgradePolicy)
```
zgalor, line 434: @vkareh @pvasant I wanted to add this call earlier for STS clusters before when we check if a role upgrade is required but wasn't sure we need to prompt the user on manual mode.

Also, that would really mess up the code. In what cases of sts do we prompt the user?

pkg/ocm/upgrades.go
jhernand: You can marshal the details to JSON and then unmarshal the type you need:

```go
data, err := json.Marshal(details)
if err != nil {
        ...
}
gates := cmv1.UnmarshalVersionGateList(details)
if err != nil {
        ...
}
```

cmd/upgrade/cluster/cmd.go
pvasant: @zgalor  why cant we check after the version is selected

zgalor: Because we use the upgrade policy API we need all the fields

cmd/upgrade/cluster/cmd.go
```diff
@@ -428,6 +430,31 @@ func run(cmd *cobra.Command, _ []string) {
...
+		reporter.Errorf("Failed to check for missing gate agreements upgrade for cluster '%s': %v", clusterKey, err)
+		os.Exit(1)
+	}
+
+	for _, gate := range gates {
+		reporter.Infof("Gate: %v", gate)
+		if !gate.STSOnly() {
+			// for non sts gates we require user agreement
+			if !confirm.Prompt(true,
+				"I acknowledge that my workloads are no longer using removed APIs. (Background:'%s)",
```
gregsheremeta, line 445: this message is specific to OCP 4.9. When 4.10 comes around and we add (via configuration) a gate for something completely different, this message makes no sense.

Why not just hardcode the entire description if we're hardcoding half of it here?

Better to get this string from the `gates[]` imo

cmd/upgrade/cluster/cmd.go
```diff
@@ -428,6 +430,31 @@ func run(cmd *cobra.Command, _ []string) {
...
+				gate.Description()) {
+				os.Exit(0)
```
gregsheremeta, line 447: The UI here should also include the `documentation_link` from the json payload. The link has important help info for the customer

cmd/upgrade/cluster/cmd.go
pvasant: @zgalor  ok  I was thinking something like this  wdyt? 
```

	upgradePolicyBuilder := cmv1.NewUpgradePolicy().
		ScheduleType("manual").
		Version(version)

	upgradePolicy, err := upgradePolicyBuilder.Build() 
// check if the cluster upgrade requires gate agreements
	gates, err := ocmClient.GetMissingGateAgreements(cluster.ID(), upgradePolicy)
	if err != nil {
		reporter.Errorf("Failed to check for missing gate agreements upgrade for cluster '%s': %v", clusterKey, err)
		os.Exit(1)
	}
upgradePolicyBuilder.NextRun(nextRun)
```

cmd/upgrade/cluster/cmd.go
vkareh: I agree with @gregsheremeta - the API already returns the correct messaging on purpose so that the CLI does not need to know the context for the ACK (apart from the STS one)

pvasant:
Will be continued in https://github.com/openshift/rosa/pull/570

PR #588, opened 2022-01-30
Refactor MachineTypes handling

tbrisker:
This got a bit larger than expected, broken down into small commits for easier review. The main changes:
- Introduce a new MachineTypeList type
- Break down the logic into smaller functions

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -1300,7 +1300,7 @@ func run(cmd *cobra.Command, _ []string) {
 		computeMachineType, err = interactive.GetOption(interactive.Input{
 			Question: "Compute nodes instance type",
 			Help:     cmd.Flags().Lookup("compute-machine-type").Usage,
-			Options:  ocm.GetAvailableMachineTypeList(computeMachineTypeList, multiAZ),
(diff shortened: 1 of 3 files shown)


tbrisker:
@pvasant PTAL

pvasant:
@tbrisker One thing I dont understand is why is it needed only in rosa cli?  If we need to display the machines based on the region then I would think we need to implement this logic in CS as inquiries endpoint that way we can access it in both rosa cli, OCM cli and UI. @vkareh  wdyt?

tbrisker:
@pvasant that's a fair concern, and I agree long-term we should also have this or similar logic on the back-end. 
Right now the MachineTypeList endpoint does not have any logic inside it regarding regions, only returns what is in the DB. I tried to minimize the changes needed to get a resolution of the issue first, as otherwise we would likely need to modify the backend, api, sdk and both clis to get the full fix (i.e. also listing only the available types, not just erroring out with a nicer message).

pvasant:
@tbrisker  Can you please create a new jira and attach it with this one. I m ok with code changes and we can merge this one. Thanks.

tbrisker:
thanks @pvasant, created SDA-5560 for this

PR #589, opened 2022-01-31
Fix bug - remove duplicated error message when deleting a cluster

oriAdler:
Related: [SDA-5208](https://issues.redhat.com/browse/SDA-5208)


oriAdler:
@zvikorn @pvasant PTAL
I think this is the origin of the duplicate error message.
Though, I wasn't sure how to reproduce the error, do we have an account **which is with sts_ocm_role on the org and without the sts_user_role on the account**, or there is a way to make these modifications on our account?

zvikorn:
@oriAdler rosa create ocm-role and rosa create user-role are two different commands. righ? you can validate that the organization/labels contains the ocm-role and that the account/labels does not include the user-role, if this is what you need

oriAdler:
@zvikorn Thanks, I have verified the changes.

PR #592, opened 2022-02-01
Fix bug - add warning when creating ocm-role with duplicate name

oriAdler:
In case there is already an admin role name, and the user tries
to create (manually) a non-admin role with the same name,
We will prompt a warning message to the terminal.

Related: [SDA-5140](https://issues.redhat.com/browse/SDA-5140)

### DIFF
--- a/cmd/create/ocmrole/cmd.go
+++ b/cmd/create/ocmrole/cmd.go
@@ -259,6 +259,10 @@ func run(cmd *cobra.Command, argv []string) {
 		}
 	case aws.ModeManual:
 		ocmClient.LogEvent("ROSACreateOCMRoleModeManual", map[string]string{})
+		_, _, err = checkRoleExists(reporter, awsClient, roleNameRequested, isAdmin, aws.ModeManual)
+		if err != nil {
+			reporter.Warnf("Creating ocm role '%s' should fail: %s", roleNameRequested, err)
+		}
 		err = generateOcmRolePolicyFiles(reporter, env, orgID, isAdmin)
 		if err != nil {
 			reporter.Errorf("There was an error generating the policy files: %s", err)
@@ -358,33 +362,12 @@ func createRoles(reporter *rprtr.Object, awsClient aws.Client, prefix string, ro
 		return "", err
 	}
 
-	exists, roleARN, err := awsClient.CheckRoleExists(roleName)
+	roleARN, exists, err := checkRoleExists(reporter, awsClient, roleName, isAdmin, aws.ModeAuto)
 	if err != nil {
 		return "", err
 	}
 	if exists {
-		isExistingRoleAdmin, err := awsClient.IsAdminRole(roleName)
-		if err != nil {
-			return "", err
-		}
-		reporter.Warnf("Role '%s' already exists", roleName)
-
-		if !isAdmin {
-			if isExistingRoleAdmin {
-				return "", fmt.Errorf("The existing role is an admin role."+
-					" To remove admin capabilities please delete the admin policy and the '%s' tag",
-					tags.AdminRole)
-			}
-			return roleARN, nil
-		}
-
-		if isExistingRoleAdmin {
-			return roleARN, nil
-		}
-
-		if !confirm.Prompt(true, "Add admin policies to '%s' role?", roleName) {
-			return roleARN, nil
-		}
+		return roleARN, nil
 	}
 
 	iamTags := map[string]string{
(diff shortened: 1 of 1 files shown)


cmd/create/ocmrole/cmd.go
```diff
@@ -268,6 +268,15 @@ func run(cmd *cobra.Command, argv []string) {
...
 		}
 		if reporter.IsTerminal() {
+			err = checkAdminRoleExists(reporter, awsClient, roleNameRequested, isAdmin)
+			if err != nil {
+				reporter.Errorf(
+					"There was an error checking if the required role name is already an admin role name: %s", err)
+				ocmClient.LogEvent("ROSACreateOCMRoleModeManual", map[string]string{
+					ocm.Response: ocm.Failure,
+				})
+				os.Exit(1)
```
oriAdler, line 278: @zgalor I wasn't sure if we want to exit here without printing the commands to the user, WDYT?

zgalor: exit

cmd/create/ocmrole/cmd.go
```diff
@@ -499,3 +508,24 @@ func createPermissionPolicy(reporter *rprtr.Object, awsClient aws.Client, policy
...
+	if err != nil {
+		return err
+	}
+	if exists {
+		reporter.Warnf("Role '%s' already exists", roleName)
+		isExistingRoleAdmin, err := awsClient.IsAdminRole(roleName)
+		if err != nil {
+			return err
+		}
+		if !isAdmin && isExistingRoleAdmin {
```
zgalor, line 523: This is a problem. You will return this error to the user but won't exit

cmd/create/ocmrole/cmd.go
```diff
@@ -268,6 +268,15 @@ func run(cmd *cobra.Command, argv []string) {
 			os.Exit(1)
 		}
 		if reporter.IsTerminal() {
+			err = checkAdminRoleExists(reporter, awsClient, roleNameRequested, isAdmin)
```
zgalor, line 271: This shouldn't be called inside isTeminal

cmd/create/ocmrole/cmd.go
```diff
@@ -499,3 +508,24 @@ func createPermissionPolicy(reporter *rprtr.Object, awsClient aws.Client, policy
 
 	return nil
 }
+
+func checkAdminRoleExists(reporter *rprtr.Object, awsClient aws.Client, roleName string, isAdmin bool) error {
```
zgalor, line 512: see if you can make use of this function also in auto mode. Right now this is a code duplication

cmd/create/ocmrole/cmd.go
oriAdler: I did it on purpose because it is for manual mode, and the user doesn't actually try to create a role yet.
I wanted only to warn the user and continue and display the commands to create roles.
WDYT?

zgalor: I think it should exit. That was the intention of the card. If the user gets the commands and run them, they would fail, and that is what we want to prevent

zgalor: Per @vkareh 's [comment](https://issues.redhat.com/browse/SDA-5140?focusedCommentId=19428457&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-19428457
) I mislead you and you should warn and continue. 

@vkareh can you confirm?

vkareh: Correct. This is for manual mode only, so printing a warning before the commands makes sense. The user can decide how to proceed after that.

oriAdler:
@pvasant PTAL

cmd/create/ocmrole/cmd.go
```diff
@@ -267,6 +267,10 @@ func run(cmd *cobra.Command, argv []string) {
 			})
 			os.Exit(1)
 		}
+		_, _, err = checkRoleExists(reporter, awsClient, roleNameRequested, isAdmin, aws.ModeManual)
+		if err != nil {
+			reporter.Warnf("There was an error while checking if the ocm role already exists: %s", err)
```
pvasant, line 272: reporter.Warnf(("Creating ocm role '%s' should should fail : %s", rolename, err) same as dry run option

cmd/create/ocmrole/cmd.go
```diff
@@ -267,6 +267,10 @@ func run(cmd *cobra.Command, argv []string) {
 			})
 			os.Exit(1)
 		}
+		_, _, err = checkRoleExists(reporter, awsClient, roleNameRequested, isAdmin, aws.ModeManual)
```
pvasant, line 270: Can we move this before we generate the policy file so we don't generate  in case of run time error

cmd/create/ocmrole/cmd.go
oriAdler: Changed, thanks.

PR #593, opened 2022-02-02
Create command 'rosa list ocm-roles'

oriAdler:
The new command is implemented similarly to the `rosa list account-roles` command.
The code checks if a `role name` contains the `OCM-Role` infix to identify an ocm-role.

image

Related: [SDA-5405](https://issues.redhat.com/browse/SDA-5405)

### DIFF
--- a/cmd/list/ocmroles/cmd.go
+++ b/cmd/list/ocmroles/cmd.go
@@ -1,0 +1,129 @@
+/*
+Copyright (c) 2022 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ocmroles
+
+import (
+	"fmt"
+	"os"
+	"text/tabwriter"
+
+	"github.com/openshift/rosa/pkg/aws"
+	"github.com/openshift/rosa/pkg/helper"
+	"github.com/openshift/rosa/pkg/logging"
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/output"
+	rprtr "github.com/openshift/rosa/pkg/reporter"
+	"github.com/spf13/cobra"
+)
+
+var Cmd = &cobra.Command{
+	Use:     "ocm-roles",
+	Aliases: []string{"ocmrole", "ocm-role", "ocmroles", "ocm-roles"},
+	Short:   "List ocm roles",
+	Long:    "List ocm roles for the current AWS account.",
+	Example: ` # List all ocm roles
+rosa list ocm-roles`,
+	Run:    run,
+	Hidden: true,
+}
+
+func init() {
+	output.AddFlag(Cmd)
+}
+
+func run(_ *cobra.Command, _ []string) {
+	reporter := rprtr.CreateReporterOrExit()
+	logger := logging.CreateLoggerOrExit(reporter)
+
+	// Create the AWS client:
+	awsClient, err := aws.NewClient().
+		Logger(logger).
+		Build()
+	if err != nil {
+		reporter.Errorf("Failed to create AWS client: %v", err)
+		os.Exit(1)
+	}
+
+	// Create the client for the OCM API:
+	ocmClient, err := ocm.NewClient().
+		Logger(logger).
+		Build()
+	if err != nil {
+		reporter.Errorf("Failed to create OCM connection: %v", err)
+		os.Exit(1)
+	}
+	defer func() {
+		err = ocmClient.Close()
+		if err != nil {
+			reporter.Errorf("Failed to close OCM connection: %v", err)
+		}
+	}()
+
+	ocmRoles, err := listOCMRoles(awsClient, ocmClient)
+	if err != nil {
+		reporter.Errorf("Failed to get ocm roles: %v", err)
+		os.Exit(1)
+	}
+
+	if len(ocmRoles) == 0 {
+		reporter.Infof("No ocm roles available")
+		os.Exit(0)
+	}
+	if output.HasFlag() {
+		err = output.Print(ocmRoles)
+		if err != nil {
+			reporter.Errorf("%s", err)
+			os.Exit(1)
+		}
+		os.Exit(0)
+	}
+
+	// Create the writer that will be used to print the tabulated results:
+	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
+	fmt.Fprint(writer, "ROLE NAME\tROLE ARN\tLINKED\tADMIN\n")
+	for _, ocmRole := range ocmRoles {
+		fmt.Fprintf(writer, "%s\t%s\t%s\t%s\n", ocmRole.RoleName, ocmRole.RoleARN, ocmRole.Linked, ocmRole.Admin)
+	}
+	writer.Flush()
+}
+
+func listOCMRoles(awsClient aws.Client, ocmClient *ocm.Client) ([]aws.Role, error) {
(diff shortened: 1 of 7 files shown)


oriAdler:
@zgalor @pvasant @zvikorn @leshemnoa @vkareh PTAL

pvasant:
@oriAdler  do you mind posting the result in the description

pkg/ocm/helpers.go
```diff
@@ -354,9 +356,19 @@ func (c *Client) LinkOrgToRole(orgID string, roleARN string) (bool, error) {
 	return true, nil
 }
 
+func (c *Client) IsRoleLinked(orgID string, roleARN string) (bool, error) {
+	resp, err := c.ocm.AccountsMgmt().V1().Organizations().Organization(orgID).
+		Labels().Labels(STSOCMRole).Get().Send()
+	if err != nil {
```
oriAdler, line 362: @zgalor @pvasant I wasn't sure if in the case of **404** or **403**: 
Should we ignore the error and return `linked==false` for this particular role? 
Or return the error and exit, so no ocm-roles are returned to the user? in case we can't figure out if a certain role is linked.

zgalor: 404 means no label - so linked=false.
You shouldn't be getting 403 when getting the user org. Did that happen or are you just asking hypothetically?

oriAdler:
@pvasant @arendej I've changed the command output and added a **Linked** column.

pkg/ocm/helpers.go
```diff
@@ -354,9 +356,19 @@ func (c *Client) LinkOrgToRole(orgID string, roleARN string) (bool, error) {
...
+		return false, err
+	}
+
+	return roleARN == resp.Body().Value(), nil
```
oriAdler, line 366: Can we assume that the `sts_ocm_role` label value contains only a single roleARN?

pkg/ocm/helpers.go
oriAdler: That right! thanks.
Didn't happen, just asking hypothetically.

cmd/list/ocmroles/cmd.go
```diff
@@ -0,0 +1,128 @@
+/*
+Copyright (c) 2021 Red Hat, Inc.
```
pvasant, line 2: 2022

cmd/list/ocmroles/cmd.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/output"
+	rprtr "github.com/openshift/rosa/pkg/reporter"
+	"github.com/spf13/cobra"
+)
+
+var Cmd = &cobra.Command{
+	Use:     "ocm-roles",
+	Aliases: []string{"ocmrole", "ocm-role", "ocmroles", "ocm-roles"},
+	Short:   "List ocm roles and policies",
```
pvasant, line 35: We are not listing policies

pkg/aws/policies.go
```diff
@@ -674,14 +676,75 @@ func parsePolicyDocument(path string) (PolicyDocument, error) {
...
+		return ocmRoles, err
+	}
+
+	for _, role := range roles {
+		if isOCMRole(role.RoleName) {
+			var ocmRole Role
+			ocmRole.RoleName = aws.StringValue(role.RoleName)
+			ocmRole.RoleARN = aws.StringValue(role.Arn)
+
+			policies, err := c.listPolicies(role)
```
pvasant, line 696: We are not listing policies why do we need this ?

pkg/ocm/helpers.go
pvasant: No you have to iterate through it and check for the aws account id and then list it

pkg/aws/policies.go
```diff
@@ -674,14 +676,75 @@ func parsePolicyDocument(path string) (PolicyDocument, error) {
 	return doc, nil
 }
 
+func isOCMRole(roleName *string) bool {
+	return strings.Contains(aws.StringValue(roleName), OCMRoleInfix)
```
pvasant, line 680: Can we use the same one from polciies.go or change it to some common place. we should be using the same constant while creating/listing/deleting so if it changes. SO i think it is safer to check the tag as well

pvasant:
> @pvasant @arendej I've changed the command output and added a **Linked** column.

Thanks @oriAdler  Can we please add isAdmin column as well.

pkg/aws/policies.go
oriAdler: I've seen that we list policies in `rosa list account-roles`, I didn't realize it's redundant in `ocm-roles`.
I'll remove this line.

pkg/aws/policies.go
oriAdler: Changed to use the same variable from **policies.go** (used in creating).
I've tried to check the tags as well, but it's really slowing down the command (~2 min), we have ~500 roles ending up in
500 requests to the **AWS client** to get the tags for each role.
image
As for checking the role name, [once created it can't be changed](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user.html), so if we are consistent on our side it should work.
WDYT?

pkg/ocm/helpers.go
oriAdler: I've noticed that the **roleARN** and the **sts-ocm-role** label both contains the AWS account id,
for example: "arn:aws:iam::**765374464689**:role/tavni26a-OCM-Role-12541229.
I've changed this line, is it sufficient? 
If we have multiple values in this label separated with commas, I think this check: 
`strings.Contains(resp.Body().Value(), roleARN)` will work. 
WDYT?

cmd/list/ocmroles/cmd.go
```diff
@@ -0,0 +1,127 @@
...
+		return nil, fmt.Errorf("failed to get organization account: %v", err)
+	}
+	linkedRoles, err := ocmClient.GetLinkedRoles(orgID)
+	if err != nil {
+		return nil, err
+	}
+
+	for i := range ocmRoles {
+		if ocmClient.IsRoleLinked(ocmRoles[i].RoleARN, linkedRoles) {
+			ocmRoles[i].Linked = "true"
```
zgalor, line 120: NP: in other list commands we use `Yes` and `No` for booleans. This is also what we present to the user when we ask them if they want to link the role

pkg/aws/policies.go
```diff
@@ -674,14 +676,91 @@ func parsePolicyDocument(path string) (PolicyDocument, error) {
...
+			ocmRole.RoleARN = aws.StringValue(role.Arn)
+
+			roleTags, err := c.iamClient.ListRoleTags(&iam.ListRoleTagsInput{
+				RoleName: role.RoleName,
+			})
+			if err != nil {
+				return nil, err
+			}
+			if roleHasTag(roleTags.Tags, tags.AdminRole, "true") {
+				ocmRole.Admin = "true"
```
zgalor, line 713: Same as previous comment regarding `Yes/No`

pkg/ocm/helpers.go
```diff
@@ -51,6 +51,8 @@ const (
 	Version   = "Version"
 	Username  = "Username"
 	URL       = "URL"
+
+	STSOCMRole = "sts_ocm_role"
```
zgalor, line 55: NP: Better user a meaningful name like `OCMRoleLabel`

pkg/ocm/helpers.go
```diff
@@ -354,9 +356,29 @@ func (c *Client) LinkOrgToRole(orgID string, roleARN string) (bool, error) {
...
+	resp, err := c.ocm.AccountsMgmt().V1().Organizations().Organization(orgID).
+		Labels().Labels(STSOCMRole).Get().Send()
+	if err != nil && resp.Status() != http.StatusNotFound {
+		return nil, err
+	}
+
+	return strings.Split(resp.Body().Value(), ","), nil
+}
+
+func (c *Client) IsRoleLinked(roleARN string, linkedRoles []string) bool {
```
zgalor, line 369: I don't think this should belong in ocm client. This doesn't go to OCM.

I also think this can be very generic func implementation `contains`, that might be used in other places. Maybe you can place it in a helper file.
See https://freshman.tech/snippets/go/check-if-slice-contains-element/

oriAdler: Good catch!

zgalor:
LGTM

PR #595, opened 2022-02-08
Modify function `handleErr` to include the error type in the new error

oriAdler:
In some places in our code, we display an error message to the user depending
on the error type.
Function `handleErr` omits the error type in the new error it generates.
This function is called in many places in our code, and this change might affect the
error message returned to the user.

These are some occurrences that I found that can be affected by this change:
`dlt/oidcprovider/cmd.go 101`
`dlt/oidcprovider/cmd.go 126`
`link/userrole/cmg.go 140`
`ocm/logs.go 40`
`logs/install/cmd.go 164`
`logs/uninstall/cmd.go 150`


oriAdler:
@zgalor @pvasant @vkareh @ciaranRoche PTAL

PR #596, opened 2022-02-08
Create command 'rosa list user-roles'

oriAdler:
The new command is implemented similarly to the `rosa list ocm-roles` command.
The code checks if a `role name` contains the `User` infix and in addition,
checks if the role has the tag `rosa_role_type: User`, because 'user' is a common word.

image

Related: [SDA-5412](https://issues.redhat.com/browse/SDA-5412)

### DIFF
--- a/cmd/list/cmd.go
+++ b/cmd/list/cmd.go
@@ -31,6 +31,7 @@ import (
 	"github.com/openshift/rosa/cmd/list/region"
 	"github.com/openshift/rosa/cmd/list/upgrade"
 	"github.com/openshift/rosa/cmd/list/user"
+	"github.com/openshift/rosa/cmd/list/userroles"
 	"github.com/openshift/rosa/cmd/list/version"
 	"github.com/openshift/rosa/pkg/arguments"
 )
@@ -55,6 +56,7 @@ func init() {
 	Cmd.AddCommand(instancetypes.Cmd)
 	Cmd.AddCommand(accountroles.Cmd)
 	Cmd.AddCommand(ocmroles.Cmd)
+	Cmd.AddCommand(userroles.Cmd)
 	flags := Cmd.PersistentFlags()
 	arguments.AddProfileFlag(flags)
 	arguments.AddRegionFlag(flags)
--- a/cmd/list/userroles/cmd.go
+++ b/cmd/list/userroles/cmd.go
@@ -1,12 +1,9 @@
 /*
 Copyright (c) 2022 Red Hat, Inc.
-
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
-
   http://www.apache.org/licenses/LICENSE-2.0
-
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -14,7 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package ocmroles
+package userroles
 
 import (
 	"fmt"
@@ -31,12 +28,12 @@ import (
 )
 
 var Cmd = &cobra.Command{
-	Use:     "ocm-roles",
-	Aliases: []string{"ocmrole", "ocm-role", "ocmroles", "ocm-roles"},
-	Short:   "List ocm roles",
-	Long:    "List ocm roles for the current AWS account.",
-	Example: ` # List all ocm roles
-rosa list ocm-roles`,
+	Use:     "user-roles",
+	Aliases: []string{"userrole", "user-role", "userroles", "user-roles"},
+	Short:   "List user roles",
+	Long:    "List user roles for current AWS account",
+	Example: `# List all user roles
+rosa list user-roles`,
 	Run:    run,
 	Hidden: true,
 }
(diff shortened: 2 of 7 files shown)


oriAdler:
@pvasant @vkareh @zgalor @zvikorn @leshemnoa PTAL

cmd/list/cmd.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package list
 
 import (
+	"github.com/openshift/rosa/cmd/list/userroles"
```
zgalor, line 20: import order

pkg/ocm/client.go
```diff
@@ -18,10 +18,12 @@ package ocm
 
 import (
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
 	sdk "github.com/openshift-online/ocm-sdk-go"
+	"github.com/openshift/rosa/pkg/reporter"
```
zgalor, line 26: np: import order

pkg/ocm/helpers.go
```diff
@@ -210,6 +212,28 @@ func (c *Client) GetCurrentAccount() (*amsv1.Account, error) {
 	return response.Body(), nil
 }
 
+// MarkLinkedRoles gets a slice of roles and populates the role's 'Linked' attribute with 'Yes'/'No'
+func (c *Client) MarkLinkedRoles(roles []aws.Role) error {
+	orgID, _, err := c.GetCurrentOrganization()
+	if err != nil {
+		return fmt.Errorf("failed to get organization account: %v", err)
+	}
+	linkedRoles, err := c.GetLinkedRoles(orgID)
```
zgalor, line 221: You need to pass the label as parameter to this func. 
This looks for `sts_ocm_role` label, but you need to look for `sts_user_role`

pkg/aws/policies.go
```diff
@@ -690,6 +690,47 @@ func isOCMRole(roleName *string) bool {
...
+		if err != nil {
+			return nil, err
+		}
+
+		if isUserRole {
+			var userRole Role
+			userRole.RoleName = aws.StringValue(role.RoleName)
+			userRole.RoleARN = aws.StringValue(role.Arn)
+
+			userRoles = append(userRoles, userRole)
```
zgalor, line 727: why not create the map here and return it as map

pkg/ocm/helpers.go
```diff
@@ -357,7 +358,17 @@ func (c *Client) LinkOrgToRole(orgID string, roleARN string) (bool, error) {
 	return true, nil
 }
 
-func (c *Client) GetLinkedRoles(orgID string) ([]string, error) {
+func (c *Client) GetAccountLinkedRoles(accountID string) ([]string, error) {
+	resp, err := c.ocm.AccountsMgmt().V1().Accounts().Account(accountID).
+		Labels().Labels(USERRoleLabel).Get().Send()
```
zgalor, line 363: If the label is hardcoded inside the func I would rename the func to `GetAccountLinkedUserRoles` so it's more verbose.
The other option is to keep the name as is, and pass the label as parameter, to allow it to be more generic, but I think it's obsolete, since we don't have another use case

pkg/ocm/helpers.go
```diff
@@ -357,7 +358,17 @@ func (c *Client) LinkOrgToRole(orgID string, roleARN string) (bool, error) {
...
+	if err != nil && resp.Status() != http.StatusNotFound {
+		return nil, handleErr(resp.Error(), err)
+	}
+
+	return strings.Split(resp.Body().Value(), ","), nil
+}
+
+func (c *Client) GetOrganizationLinkedRoles(orgID string) ([]string, error) {
```
zgalor, line 371: If the label is hardcoded inside the func I would rename the func to `GetOrganizationLinkedOCMRoles` so it's more verbose.

pkg/ocm/helpers.go
oriAdler: Good point. 
Changed to `GetAccountLinkedUserRoles`, thanks.

cmd/list/userroles/cmd.go
```diff
@@ -0,0 +1,96 @@
...
+	writer.Flush()
+}
+
+func listUserRoles(awsClient aws.Client, ocmClient *ocm.Client) ([]aws.Role, error) {
+	userRoles, err := awsClient.ListUserRoles()
+	if err != nil {
+		return nil, err
+	}
+
+	err = ocmClient.MarkLinkedRoles(userRoles)
```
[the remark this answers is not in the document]
oriAdler: @pvasant @zgalor I've changed the picture in the description to demonstrate the bug was fixed.

cmd/list/userroles/cmd.go
```diff
@@ -0,0 +1,112 @@
...
+func listUserRoles(awsClient aws.Client, ocmClient *ocm.Client) ([]aws.Role, error) {
+	userRoles, err := awsClient.ListUserRoles()
+	if err != nil {
+		return nil, err
+	}
+
+	// Check if roles are linked to account
+	account, err := ocmClient.GetCurrentAccount()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get organization account: %v", err)
```
pvasant, line 94: I think it should "Failed to get Redhat User Account" @zgalor   wdyt

pkg/aws/policies.go
```diff
@@ -690,6 +690,47 @@ func isOCMRole(roleName *string) bool {
 	return strings.Contains(aws.StringValue(roleName), fmt.Sprintf("%s-Role", OCMRole))
 }
 
+// isUserRole checks the role tags in addition to the role name, because the word 'user' is common
+func (c *awsClient) isUserRole(roleName *string) (bool, error) {
+	if strings.Contains(aws.StringValue(roleName), OCMUserRole) {
+		roleTags, err := c.iamClient.ListRoleTags(&iam.ListRoleTagsInput{
```
pvasant, line 696: Would you mind moving all the role related code into a different file than policies. 

Also do we need to list role Tags or are we getting role.tags now from aws?

oriAdler: Sure, I'll create a sequence PR to refactor to a new file, as some role-related functions are already merged with **rosa list ocm-roles** PR.
Unfortunately, when calling `c.iamClient.ListRolesPages` we don't get the tags,
only when retrieving a single role with `c.iamClient.GetRole` we do get also the tags in the response.

PR #603, opened 2022-02-14
Add 'rosa delete ocm-role' command

oriAdler:
The new command follows these steps:
1. Unlink role if needed.
2. Detach Attached role policies.
3. Delete the policies from step 2.
4. Delete permissions boundary (doesn't delete the policy).
5. Delete role.

image

Related: [SDA-5202](https://issues.redhat.com/browse/SDA-5202)

### DIFF
--- a/cmd/dlt/ocmrole/cmd.go
+++ b/cmd/dlt/ocmrole/cmd.go
@@ -1,0 +1,239 @@
+/*
+  Copyright (c) 2022 Red Hat, Inc.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+*/
+
+package ocmrole
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/spf13/cobra"
+
+	unlinkocmrole "github.com/openshift/rosa/cmd/unlink/ocmrole"
+	"github.com/openshift/rosa/pkg/aws"
+	"github.com/openshift/rosa/pkg/helper"
+	"github.com/openshift/rosa/pkg/interactive"
+	"github.com/openshift/rosa/pkg/interactive/confirm"
+	"github.com/openshift/rosa/pkg/logging"
+	"github.com/openshift/rosa/pkg/ocm"
+	rprtr "github.com/openshift/rosa/pkg/reporter"
+)
+
+var args struct {
+	roleARN string
+}
+
+var Cmd = &cobra.Command{
+	Use:     "ocm-role",
+	Aliases: []string{"ocmrole"},
+	Short:   "Delete ocm role",
+	Long:    "Delete ocm role from the current AWS account",
+	Example: ` # Delete ocm role
+rosa delete ocm-role --role-arn arn:aws:iam::123456789012:role/xxx-OCM-Role-1223456778`,
+	RunE: run,
+}
+
+func init() {
+	flags := Cmd.Flags()
+
+	flags.StringVar(
+		&args.roleARN,
+		"role-arn",
+		"",
+		"Role ARN to delete from the OCM organization account")
+
+	aws.AddModeFlag(Cmd)
+
+	confirm.AddFlag(flags)
+	interactive.AddFlag(flags)
+}
+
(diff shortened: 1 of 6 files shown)


oriAdler:
@zgalor @zvikorn @pvasant @leshemnoa @vkareh PTAL

cmd/dlt/ocmrole/cmd.go
```diff
@@ -0,0 +1,239 @@
...
+	Example: ` # Delete ocm role
+rosa delete ocm-role --role-arn arn:aws:iam::123456789012:role/xxx-OCM-Role-1223456778`,
+	RunE: run,
+}
+
+func init() {
+	flags := Cmd.Flags()
+
+	flags.StringVar(
+		&args.roleARN,
```
pvasant, line 56: what is the motivation behind getting roleARN instead of role name?

cmd/dlt/ocmrole/cmd.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+	}
+	if roleARN != "" {
+		_, err := arn.Parse(roleARN)
+		if err != nil {
+			reporter.Errorf("Expected a valid ocm role ARN to delete from the current organization: %s", err)
+			os.Exit(1)
+		}
+	}
+	if !confirm.Prompt(true, "Delete the '%s' role from organization '%s'?", roleARN, orgID) {
```
pvasant, line 132: We dont need to specify organization.  Delete  '%s' ocm role ?

cmd/dlt/ocmrole/cmd.go
```diff
@@ -0,0 +1,239 @@
...
+	roleName, err := aws.RoleARNToRoleName(roleARN)
+	if err != nil {
+		reporter.Errorf("%s", err)
+		os.Exit(1)
+	}
+	switch mode {
+	case aws.ModeAuto:
+		ocmClient.LogEvent("ROSADeleteOCMRoleModeAuto", nil)
+		if isLinked {
+			reporter.Warnf("Role ARN '%s' is linked to organization '%s'. Unlink the role before deletion", roleARN, orgID)
```
pvasant, line 166: Either remove Unlink the role before deletion or change it to Unlinking the role before deletion",

pkg/aws/sts.go
```diff
@@ -0,0 +1,118 @@
...
+	err = c.deletePermissionsBoundary(roleName)
+	if err != nil {
+		return err
+	}
+
+	return c.DeleteRole(roleName, aws.String(roleName))
+}
+
+func RoleARNToRoleName(roleARN string) (string, error) {
+	ARN, err := arn.Parse(roleARN)
```
pvasant, line 44: I would user parsedARN so it is snake case

cmd/dlt/ocmrole/cmd.go
oriAdler: In case the role is linked we'll want to call `rosa unlink ocm-role` and pass the `roleARN` as a parameter:
`err = unlinkocmrole.Cmd.RunE(unlinkocmrole.Cmd, []string{roleARN})`

Also to be consistent with other commands. 
In the future, if we'll want to make decisions based on other information embedded in the `roleARN`, we'll have it, otherwise, we'll need to change the argument in the command.

cmd/dlt/ocmrole/cmd.go
oriAdler: Changed.

cmd/dlt/ocmrole/cmd.go
oriAdler: Removed.
I've made a small change in `unlink/ocmrole/cmd.go` so now we ask for role ARN only once (updated the screenshot in the comment).
Thanks.

oriAdler: Though now the output of the `rosa unlink ocm-role` is a bit different:
image
In case the `roleARN` is provided, it won't prompt:
`? OCM Role ARN: arn:aws:iam::765374464689:role/ManagedOpenShift-OCM-Role-12541229`
WDYT?

Alternatively, I can pass a boolean argument from `rosa delete ocm-role` to determine if we should ask for `roleARN`.
Or there is a way to determine from with `rosa unlink ocm-role` if it was called by `rosa delete ocm-role`?

pkg/aws/sts.go
```diff
@@ -0,0 +1,118 @@
...
+			RoleName:  aws.String(roleName),
+		})
+		if err != nil {
+			return err
+		}
+
+		_, err = c.iamClient.DeletePolicy(&iam.DeletePolicyInput{PolicyArn: policy.PolicyArn})
+		if err != nil {
+			if awsErr, ok := err.(awserr.Error); ok {
+				if awsErr.Code() == iam.ErrCodeDeleteConflictException { // policy is attached to another entity
```
zgalor, line 109: Did you test this?

oriAdler: Yes, I have attached a policy for two different roles.
Detached the policy from the role that was deleted,
then got this error when trying to delete the policy.
[AWS SDK reference](https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#IAM.DeletePolicy)

PR #607, opened 2022-02-16
Add 'rosa delete user-role' command

oriAdler:
The new command is implemented similarly to `rosa delete ocm-role`.
1. Unlink user role if needed.
2. Detach attached role policies.
3. Delete permission boundary (doesn't delete the policy).
4. Delete role.

image

Related: [SDA-5437](https://issues.redhat.com/browse/SDA-5437)

### DIFF
--- a/cmd/dlt/cmd.go
+++ b/cmd/dlt/cmd.go
@@ -29,6 +29,7 @@ import (
 	"github.com/openshift/rosa/cmd/dlt/oidcprovider"
 	"github.com/openshift/rosa/cmd/dlt/operatorrole"
 	"github.com/openshift/rosa/cmd/dlt/upgrade"
+	"github.com/openshift/rosa/cmd/dlt/userrole"
 	"github.com/openshift/rosa/pkg/arguments"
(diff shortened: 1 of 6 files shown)


oriAdler:
@zgalor @tbrisker @pvasant @zvikorn @leshemnoa @vkareh PTAL

I've noticed that when creating a `user-role` we add a tag `rosa_environment: integration/staging/production` for our team purposes.
Should I add a check for the environment to this command? 
and enable deletion only for the same environment?

pvasant:
> @zgalor @tbrisker @pvasant @zvikorn @LeshemNoa @vkareh PTAL
> 
> I've noticed that when creating a `user-role` we add a tag `rosa_environment: integration/staging/production` for our team purposes. Should I add a check for the environment to this command? and enable deletion only for the same environment?

I think for ocm -role and user-role we can have the warning like how we have for account role i think. The warning should be only for "Stage" or "int" environment and not a customer-facing one.

oriAdler:
Because the check for the environment is for our internal use,
Priya and I decided it can be added in a sequence PR, both for **delete ocm-role** and **delete user-role**

PR #608, opened 2022-02-16
Add HTPasswd IDP to ROSA CLI

LeshemNoa:
In this MR I added the HTPasswd IDP type to the ROSA CLI, and it can be created with `rosa create idp --type="htpasswd"`.
The `rosa create admin` and `rosa delete admin` commands have been changed accordingly so that they can still be used with or without an existing HTPasswd IDP. 

I think I'll record a usage demo and post it here to best explain the contents of this MR - coming ASAP.

PTAL @vkareh @pvasant


cmd/create/idp/cmd.go
```diff
@@ -258,12 +266,28 @@ func init() {
...
 	interactive.AddFlag(flags)
+	reporter = rprtr.CreateReporterOrExit()
+	logger = logging.CreateLoggerOrExit(reporter)
 }
 
 func run(cmd *cobra.Command, _ []string) {
-	reporter := rprtr.CreateReporterOrExit()
-	logger := logging.CreateLoggerOrExit(reporter)
+	reporter = rprtr.CreateReporterOrExit()
+	logger = logging.CreateLoggerOrExit(reporter)
```
oriAdler, line 290: Do we need lines 289-290?
if we already initialized `reporter` and `logger` in lines 284-285

pkg/ocm/idps.go
```diff
@@ -45,6 +47,38 @@ func (c *Client) CreateIdentityProvider(clusterID string, idp *cmv1.IdentityProv
...
+	if err != nil {
+		return handleErr(response.Error(), err)
+	}
+	return nil
+}
+
+func (c *Client) DeleteHTPasswdUser(username, clusterID string, htpasswdIDP *cmv1.IdentityProvider) error {
+	var userID string
+	listResponse, err := c.ocm.ClustersMgmt().V1().Clusters().Cluster(clusterID).
+		IdentityProviders().IdentityProvider(htpasswdIDP.ID()).HtpasswdUsers().List().Send()
```
oriAdler, line 63:
```suggestion
		IdentityProviders().IdentityProvider(htpasswdIDP.ID()).HtpasswdUsers().List().Send()
		if err != nil {
		    return handleErr(response.Error(), err)
		}
```

cmd/create/admin/cmd.go
```diff
@@ -134,51 +133,60 @@ func run(cmd *cobra.Command, _ []string) {
...
-		MappingMethod(cmv1.IdentityProviderMappingMethod("claim")).
-		Htpasswd(htpasswdIDP).
-		Build()
-	if err != nil {
-		reporter.Errorf("Failed to create '%s' identity provider for cluster '%s'", idpName, clusterKey)
-		os.Exit(1)
-	}
+	// No HTPasswd IDP - create it with cluster-admin user.
+	if existingHTPasswdIDP == nil {
+		idpName := "htpasswd-1"
```
vkareh, line 152: This should probably be a constant

cmd/create/idp/cmd.go
```diff
@@ -74,9 +74,13 @@ var args struct {
...
 	openidUsername  string
 	openidScopes    string
+
+	// HTPasswd
+	htpasswdUsername string
+	htpasswdPassword string
 }
 
-var validIdps []string = []string{"github", "gitlab", "google", "ldap", "openid"}
+var validIdps []string = []string{"github", "gitlab", "google", "ldap", "openid", "htpasswd"}
```
vkareh, line 83: sort this alphabetically, it's less chaotic for the customer that way

cmd/create/idp/cmd.go
```diff
@@ -258,13 +266,26 @@ func init() {
...
+	flags.StringVar(
+		&args.htpasswdPassword,
+		"password",
+		"",
+		"HTPasswd: Password for provided username, to log into the cluster console with.\n",
+	)
+
 	interactive.AddFlag(flags)
+	reporter = rprtr.CreateReporterOrExit()
+	logger = logging.CreateLoggerOrExit(reporter)
```
vkareh, line 285: I would move these back inside the `run` function to be consistent, unless there's a reason to move it here?

cmd/create/idp/cmd.go
```diff
@@ -314,26 +335,20 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
-	// Grab all the IDP information interactively if necessary
-	idpType := args.idpType
-	if idpType == "" {
-		interactive.Enable()
-	}
-
```
vkareh, line 322: By moving this block, it prevents the `Interactive mode enabled` message from being shown

cmd/create/idp/cmd.go
```diff
@@ -314,26 +335,20 @@ func run(cmd *cobra.Command, _ []string) {
...
+	// Grab all the IDP information interactively if necessary
+	idpType := args.idpType
+	if idpType == "" {
+		interactive.Enable()
 		idpType, err = interactive.GetOption(interactive.Input{
 			Question: "Type of identity provider",
 			Options:  validIdps,
 			Required: true,
-			Default:  idpType,
+			Default:  validIdps[0],
```
vkareh, line 351: This is wrong. If a user runs `rosa create idp --type htpasswd --interactive` their prompts will revert back to `github`

cmd/create/idp/cmd.go
```diff
@@ -362,7 +377,7 @@ func run(cmd *cobra.Command, _ []string) {
 
 	// Auto-generate a name if none provided
 	if !cmd.Flags().Changed("name") {
-		idps := getIdps(reporter, ocmClient, cluster)
+		idps := getIdps(ocmClient, cluster)
```
vkareh, line 380: Okay, I see why you moved the scope of the reporter :+1:

cmd/create/idp/cmd.go
```diff
@@ -385,6 +400,10 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 	idpName = strings.Trim(idpName, " \t")
 
+	if idpType == "htpasswd" {
+		createHTPasswdIDP(cmd, cluster, clusterKey, idpName, ocmClient)
+		return
+	}
```
vkareh, line 406: Why not put this inside the `switch` statement below?

cmd/create/idp/htpasswd.go
```diff
@@ -0,0 +1,179 @@
...
+			if user.Username() == ClusterAdminUsername {
+				hasAdmin = true
+			}
+			return true
+		})
+	}
+	return hasAdmin
+}
+
+func FindExistingHTPasswdIDP(cluster *cmv1.Cluster, ocmClient *ocm.Client) (htpasswdIDP *cmv1.IdentityProvider) {
```
vkareh, line 165: nitpick: I wonder if this function should be part of `pkg/ocm/idps.go`

cmd/create/idp/htpasswd.go
```diff
@@ -0,0 +1,179 @@
...
+			Name(idpName).
+			Htpasswd(
+				cmv1.NewHTPasswdIdentityProvider().Users(
+					cmv1.NewHTPasswdUserList().Items(CreateHTPasswdUser(username, password)),
+				),
+			)
+		htpasswdIDP = doCreateIDP(idpName, *idpBuilder, cluster, clusterKey, ocmClient)
+	}
+
+	for shouldAddAnotherUser() {
```
vkareh, line 75: This should only be called if `interactive.Enabled()` otherwise it will prevent customers from using this in pipelines

LeshemNoa: I didn't know that. But that would only allow them to add one user without the `--interactive` flag. I guess that's not too too bad for now because later on we intend to allow batch user addition.

cmd/create/idp/cmd.go
LeshemNoa: good thing I did this little bit in a separate commit so I just dropped it.

cmd/create/idp/htpasswd.go
LeshemNoa: In the `ocm` package we wouldn't have access to the `idp` package's reporter. I could pass the reporter in as a parameter but this seems less correct to me.

LeshemNoa:
@vkareh, I addressed all of your comments, can you please take another look?

cmd/create/idp/htpasswd.go
vkareh: Makes sense, thanks

cmd/create/idp/cmd.go
```diff
@@ -393,6 +414,9 @@ func run(cmd *cobra.Command, _ []string) {
 		idpBuilder, err = buildGitlabIdp(cmd, cluster, idpName)
 	case "google":
 		idpBuilder, err = buildGoogleIdp(cmd, cluster, idpName)
+	case "htpasswd":
+		createHTPasswdIDP(cmd, cluster, clusterKey, idpName, ocmClient)
```
vkareh, line 418: This function calls `doCreateIDP`, but so does this file after the `switch` statement. Would it make sense to follow the same pattern as the rest of the IDPs, where you just call something like `buildHtpasswdIdp` that returns the `idpBuilder`?

LeshemNoa: I wanted to hijack the flow like that because unlike the other IDPs you can't just build the IDP, create it and finish. 

You may need to create the new IDP and then go through the interactive `shouldAddAnotherUser()` loop to post more users to the IDP's user collection. 

Or, the HTPasswd IDP might already exist, if the user already called `rosa create admin` for example. in that case on `rosa create idp --type="htpasswd"` we don't build the IDP at all, we just add the new users to the existing IDP. 

I hope this makes sense.

vkareh: Yes, this makes sense. But it seems like `doCreateIDP` is being called twice (once inside the `createHTPasswdIDP` and another time after the `switch`).

LeshemNoa: That's exactly why there's a return statement right after that. The second call to `doCreate` isn't reached.

vkareh: :facepalm: Of course!! thanks :)

I think you can change that `return` to be `os.Exit(0)`

LeshemNoa:
@pvasant, can you please re-stamp? I had some lint errors I needed to fix.

PR #612, opened 2022-02-24
Fix bug - forbid deletion of ocm-role in case user cannot unlink role

oriAdler:
In case a role is linked, and the user doesn't have permission to unlink the role,
print an error message and exit the program.
At the moment, if there is an error when unlinking the role, the role will be
deleted anyway.

Related: [SDA-5482](https://issues.redhat.com/browse/SDA-5482)


oriAdler:
@zgalor @pvasant @vkareh @leshemnoa @zvikorn PTAL

For some reason, when this line is called:
`			err = unlinkocmrole.Cmd.RunE(unlinkocmrole.Cmd, []string{roleARN})
`

If an error is returned from `unlinkocmrole`, the returned value is nil, and the error is ignored.
I'm trying to figure out what is causing this behavior.

Anyhow, this solution of exiting in case of an error should fix the bug in the meantime.

PR #613, opened 2022-02-24
List roles - display a spinner while fetching the roles

oriAdler:
The command of fetching the roles with the AWS client can take some time when
there are a lot of roles in an account.
A spinner was added to `list ocm-roles`, `list user-roles`, and `list account-roles`
to improve the user experience.

https://user-images.githubusercontent.com/57869309/155566837-cc94abab-804f-4984-b826-bce2e1c00392.mp4


oriAdler:
@ciaranRoche @zgalor @pvasant PTAL

PR #614, opened 2022-02-27
Fix bug - deletion of a role with the wrong account ID in role ARN

oriAdler:
To delete a role using the AWS client the role name is sufficient.

For `delete ocm-role` and `delete user-role`, we ask the user to enter the whole role ARN,
in order to unlink the role in case it's linked, and to be consistent with other ROSA commands.

This PR adds validation to check if the user account ID match the role ARN account ID,
if not, an error message will be displayed and the program will terminate.

Related: [SDA-5544](https://issues.redhat.com/browse/SDA-5544)

image

### DIFF
--- a/cmd/dlt/ocmrole/cmd.go
+++ b/cmd/dlt/ocmrole/cmd.go
@@ -130,6 +130,12 @@ func run(cmd *cobra.Command, argv []string) error {
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


oriAdler:
@zgalor @pvasant @zvikorn PTAL

PR #615, opened 2022-02-28
Fix bug - change the interactive message of `rosa delete user-role`

oriAdler:
Related: [SDA-5548](https://issues.redhat.com/browse/SDA-5548)


oriAdler:
@zgalor @zvikorn @pvasant PTAL

pvasant:
Do we need user role deletion mode?

oriAdler:
> Do we need user role deletion mode?

Changed to `User role deletion mode`
We have both **manual** and **auto** mode similarly to `delete ocm-role` and `delete account-role`

PR #616, opened 2022-02-28
HTPasswd bug fixes corresponding with some CS changes

LeshemNoa:
In this MR:

- Correctly fetch the HTPasswd user list to if idp indeed `HasClusterAdmin`
- Don't dispatch the request to delete the `cluster-admin` user from the `cluster-admins` group, since this is now done on the CS side.

### DIFF
--- a/cmd/create/idp/htpasswd.go
+++ b/cmd/create/idp/htpasswd.go
@@ -44,11 +44,16 @@ func createHTPasswdIDP(cmd *cobra.Command,
 	}
 
 	// Choose which way to create the IDP according to whether it already has an admin or not.
-	htpasswdIDP := FindExistingHTPasswdIDP(cluster, ocmClient)
+	htpasswdIDP, userList := FindExistingHTPasswdIDP(cluster, ocmClient)
 	if htpasswdIDP != nil {
-		containsAdminOnly := !HasClusterAdmin(htpasswdIDP) || htpasswdIDP.Htpasswd().Users().Len() == 0
+		// if existing idp has any users other than `cluster-admin`, then it was created as a proper idp
+		// and not as an admin container during `rosa create admin`. A cluster may only have one
+		// htpasswd idp so `create idp` should not continue.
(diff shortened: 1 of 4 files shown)


cmd/create/idp/htpasswd.go
```diff
@@ -44,11 +44,13 @@ func createHTPasswdIDP(cmd *cobra.Command,
 	}
 
 	// Choose which way to create the IDP according to whether it already has an admin or not.
-	htpasswdIDP := FindExistingHTPasswdIDP(cluster, ocmClient)
+	htpasswdIDP, userList := FindExistingHTPasswdIDP(cluster, ocmClient)
 	if htpasswdIDP != nil {
-		containsAdminOnly := !HasClusterAdmin(htpasswdIDP) || htpasswdIDP.Htpasswd().Users().Len() == 0
+		containsAdminOnly := HasClusterAdmin(userList) && userList.Len() == 1
 		if !containsAdminOnly {
```
pvasant, line 50: is it  `if containsAdminOnly {`

LeshemNoa: I added documentation for this line. 
The idea behind this is to allow users to call `rosa create admin` and afterwards call `rosa create idp` with type htpasswd without having the idp creation fail due to the preexisting idp they may have not known was created during `create admin`. If `!containsAdminOnly` we can infer that the idp was created via `ocm create idp` or `rosa create idp`, so logically creating a new htpasswd idp should fail. 
I hope it makes sense. We do intend to deprecate this command in the future since now htpasswd is a proper idp, this is only to maintain compatibility.

pvasant:
```
 
cmd/create/idp/htpasswd.go:187:4: printf: (*github.com/openshift/rosa/pkg/reporter.Object).Errorf call needs 1 arg but has 2 args (govet)
			reporter.Errorf("Failed to get user list of the HTPasswd IDP of '%s'", clusterKey, err)
			^
```

above one needs to be fixed. Otherwise changes look good to me.

PR #617, opened 2022-03-01
Fix bug - `delete ocm-role` should be hidden in rosa cli

oriAdler:
Related: [SDA-5558](https://issues.redhat.com/browse/SDA-5558)


oriAdler:
@zgalor @pvasant PTAL

PR #619, opened 2022-03-02
Fix bug - validate role type before deletion

oriAdler:
To delete a role using the AWS client, a role name is sufficient.

This PR adds the following validations:
- `delete ocm-role`: check that the provided role is an **ocm role**.
- `delete user-role`: check that the provided role is a **user role**.

Related: [SDA-5564](https://issues.redhat.com/browse/SDA-5564)

image


oriAdler:
@zgalor @pvasant @ciaranRoche PTAL

PR #620, opened 2022-03-02
Improve `rosa unlink user role` error message

oriAdler:
- Specify that a user can unlink a user role from its account,
also without `orgAdmin` permissions.
- Fix the argument in the error message, for case the `role ARN` is not
provided as a flag.

Related: [SDA-5546](https://issues.redhat.com/browse/SDA-5546)

image


oriAdler:
@pvasant @zgalor @LeshemNoa PTAL

PR #622, opened 2022-03-03
Sort roles to display linked ones first

oriAdler:
For the commands `list user-roles` and `list ocm-roles` linked roles
will be displayed first.

The order between the unlinked roles stays the same.
The same applied to the linked roles.

Related: [SDA-5574](https://issues.redhat.com/browse/SDA-5574)

image


oriAdler:

pvasant:
@oriAdler  i would use sort.Slice . wdyt ?

oriAdler:
> @oriAdler i would use sort.Slice . wdyt ?

Changed to ``sort.SliceStable`` to keep equal elements in their original order.
Thanks.

PR #626, opened 2022-03-06
Refactor `list ocm-role` to use a map of linked roles

oriAdler:
Implementation is similar to the command `list user-roles`.
This change improves the time complexity of marking the linked roles.


oriAdler:
@zgalor @pvasant PTAL

PR #636, opened 2022-03-08
Add quota service permissions to the `installer policy`

oriAdler:
This MR will enable CS to run preflight quota checks.

Related: [SDA-5429](https://issues.redhat.com/browse/SDA-5429)

### DIFF
--- a/templates/policies/sts_installer_permission_policy.json
+++ b/templates/policies/sts_installer_permission_policy.json
@@ -169,6 +169,8 @@
                 "s3:PutObject",
                 "s3:PutObjectAcl",
                 "s3:PutObjectTagging",
+                "servicequotas:GetServiceQuota",
+                "servicequotas:ListAWSDefaultServiceQuotas",
                 "sts:AssumeRole",
                 "sts:AssumeRoleWithWebIdentity",
                 "sts:GetCallerIdentity",
--- a/assets/bindata.go
+++ b/assets/bindata.go
@@ -739,6 +739,8 @@ var _templatesPoliciesSts_installer_permission_policyJson = []byte(`{
                 "s3:PutObject",
                 "s3:PutObjectAcl",
                 "s3:PutObjectTagging",
+                "servicequotas:GetServiceQuota",
+                "servicequotas:ListAWSDefaultServiceQuotas",
                 "sts:AssumeRole",
                 "sts:AssumeRoleWithWebIdentity",
                 "sts:GetCallerIdentity",


oriAdler:
@zgalor @pvasant PTAL

pvasant:
@jharrington22  @vkareh @arendej  PTAL

vkareh:
This looks good to me.

I would like @arendej's opinion before merging

templates/policies/sts_installer_permission_policy.json
```diff
@@ -169,6 +169,8 @@
                 "s3:PutObject",
                 "s3:PutObjectAcl",
                 "s3:PutObjectTagging",
+                "servicequotas:GetServiceQuota",
+                "servicequotas:ListAWSDefaultServiceQuotas",
```
jharrington22, line 173: @oriAdler do you only need these two? We were initially requesting these as part of the OSD service definition SCP policy

```
            "Action": [
                "servicequotas:ListServices",
                "servicequotas:GetRequestedServiceQuotaChange",
                "servicequotas:GetServiceQuota",
                "servicequotas:RequestServiceQuotaIncrease",
                "servicequotas:ListServiceQuotas"
            ],
```

pvasant: @jharrington22  this PR is primarily focused on the preflight requirement. So we kept it as just listing and read-only permissions. Why do we need   "servicequotas:ListServices" though 
https://issues.redhat.com/browse/SDE-1545

arendej:
LGTM

templates/policies/sts_installer_permission_policy.json
oriAdler: I'm currently working on adding the [`EIPs` preflight check](https://issues.redhat.com/browse/SDA-5288) and using the next API call:
`client.GetServiceQuota(ctx, QuotaCode, ServiceCode)`

So this permission `"servicequotas:GetServiceQuota"` is sufficient at the moment.

I assume that for `SDA-4852 Internet Gateway Exceeded`, `SDA-5015 - NAT Gateway Exceeded`, 
`SDA 5043 - Too many buckets`, `SDA-5287 -  ec2 1216C47A quota` we'll use the same API call
to get the resource limit on the account.

jharrington22: Ok this make sense, I thought you were listing and checking the same way the CLI does it but you have a defined list of service codes. We have the RequestServiceQuotaIncrease for the AWS account operator.

jharrington22 (approved):

PR #638, opened 2022-03-10
Restrict iam:PassRole to account wide instance IAM Roles

jharrington22:
Restrict the `iam:PassRole` permission to the specific roles created for the controlplane and worker nodes.

We cannot remove the specific `iam:PassRole` action because the installer is creating the controlplane IAM instance profiles and needs to assign the controlplane IAM role when the controlplane node is created. The cluster it self will create the worker nodes and needs to assign the IAM instance profile and the worker IAM roles when they are created.

https://issues.redhat.com/browse/OSD-10327

### DIFF
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -363,7 +363,17 @@ func createRoles(reporter *rprtr.Object, awsClient aws.Client,
 		filename = fmt.Sprintf("sts_%s_permission_policy.json", file)
 		path = fmt.Sprintf("templates/policies/%s", filename)
 
-		policy, err = aws.ReadPolicyDocument(path)
+		// For the installer permission policy scope the iam:PassRole permission
+		// to the worker and controlplane ARNs roles only
+		if strings.ToLower(role.Name) == aws.InstallerAccountRole {
+			policy, err = aws.ReadPolicyDocument(path, map[string]string{
+				"aws_account_id":   accountID,
+				"controlplane_arn": aws.GetRoleName(prefix, aws.AccountRoles[aws.ControlPlaneAccountRole].Name),
+				"worker_arn":       aws.GetRoleName(prefix, aws.AccountRoles[aws.WorkerAccountRole].Name),
+			})
+		} else {
+			policy, err = aws.ReadPolicyDocument(path)
+		}
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


cmd/create/accountroles/cmd.go
```diff
@@ -363,7 +363,18 @@ func createRoles(reporter *rprtr.Object, awsClient aws.Client,
...
+		// The cluster itslef will provision worker nodes see machine-api-operator
+		// permission.
+		if strings.ToLower(role.Name) == aws.InstallerAccountRole {
+			policy, err = aws.ReadPolicyDocument(path, map[string]string{
+				"aws_account_id":   accountID,
+				"controlplane_arn": aws.GetRoleName(prefix, aws.AccountRoles[aws.ControlPlaneAccountRole].Name),
+			})
+		} else {
+			policy, err = aws.ReadPolicyDocument(path)
+		}
```
vkareh, line 377: You probably don't need to wrap this in a conditional. The templating should replace only if the format string exists.

cmd/create/accountroles/cmd.go
```diff
@@ -393,7 +404,21 @@ func createRoles(reporter *rprtr.Object, awsClient aws.Client,
...
+			// Scope the IAM policies iam:PassRole permission to the worker
+			// IAM Role ARN
+			if "machine_api_aws_cloud_credentials" == credrequest {
+				policy, err = aws.ReadPolicyDocument(path, map[string]string{
+					"aws_account_id": accountID,
+					"worker_arn":     aws.GetRoleName(prefix, aws.AccountRoles[aws.WorkerAccountRole].Name),
+				})
+			} else {
+				policy, err = aws.ReadPolicyDocument(path)
+			}
```
vkareh, line 420: same here

cmd/create/accountroles/cmd.go
```diff
@@ -363,7 +363,16 @@ func createRoles(reporter *rprtr.Object, awsClient aws.Client,
 		filename = fmt.Sprintf("sts_%s_permission_policy.json", file)
 		path = fmt.Sprintf("templates/policies/%s", filename)
 
-		policy, err = aws.ReadPolicyDocument(path)
+		// For the installer permission policy scope the iam:PassRole permission
+		// to the controlplane ARN role only.
+		// The cluster itslef will provision worker nodes see machine-api-operator
+		// permission.
+		if strings.ToLower(role.Name) == aws.InstallerAccountRole {
```
vkareh, line 370: Remove the conditional, otherwise only the installer will have a policy :P

jharrington22: :facepalm:

jharrington22:
```
rosa delete account-roles --region us-east-1 --prefix ManagedOpenShift     
E: Instance Profiles are attached to the role. Please make sure it is deleted: iam-passrole-bvrmc-worker-profile,iam-passrole-xj7qr-worker-profile
```

Need to investigate this issue with role clean up.

vkareh:
> Need to investigate this issue with role clean up.

Good catch

PR #639, opened 2022-03-11
Adding message to update rosa CLI if incompatible roles are found

ravitri:
Based on feedback by end user, updating the message to suggest updating `rosa` CLI to latest version and try again if incompatible roles are found. Not sure if this would work and be valid for all versions and latest `rosa` CLI yet but I think its worth adding if its supported.


pvasant:
@ravitri account roles will also be not compatible if the user already has rosa 1.1.8 and created account roles for ocp 4.8. Planning to use the same rosa 1.1.8 to install a fresh ocp 4.9 cluster then they need to run the account roles again if not they will get the above error. Having said that i m not sure if we can ask the user to always download the latest rosa cli to create a cluster.  


@arendej  can you PTAL.

PR #641, opened 2022-03-13
Re-introducing HTPasswd IDP after revert

LeshemNoa:
Last week we had to revert the ROSA HTPasswd work for an urgent ROSA version release. 
I am now reintroducing the same changes.

PTAL @pvasant


pvasant:
@LeshemNoa  Does this include the fix MR as well ?

PR #644, opened 2022-03-16
Fix help for --compute-nodes

smalleni:
The flag --compute-nodes is used to pass total number
of compute nodes, not compute nodes per zone.

Signed-off-by: Sai Sindhur Malleni <smalleni@redhat.com>


ciaranRoche:
/lgtm 

Thanks @smalleni for the update.

PR #652, opened 2022-03-24
Introducing a command to create managed services

renan-campos:


jmolmo:
LGTM. 
Minor nit: Just i would like to have the clarification in the type parameter...

renan-campos:
New PR with all the commands: https://github.com/openshift/rosa/pull/654

PR #653, opened 2022-03-24
Fix bug - create ocm-role - prompt the role ARN

oriAdler:
In case the user deleted a linked ocm-role using the AWS console,
there is a scenario where the role is linked to the organization but doesn't exist.
We want the provide the user with the command to unlink it.

image

Related: [SDA-5689](https://issues.redhat.com/browse/SDA-5689)

### DIFF
--- a/cmd/create/ocmrole/cmd.go
+++ b/cmd/create/ocmrole/cmd.go
@@ -225,16 +225,26 @@ func run(cmd *cobra.Command, argv []string) {
 
 	roleNameRequested := aws.GetOCMRoleName(prefix, aws.OCMRole, externalID)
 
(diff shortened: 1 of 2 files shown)


oriAdler:
@pvasant @LeshemNoa @zvikorn PTAL

zvikorn:
Not related to this MR: why Don't we change the name of the function in https://github.com/oriAdler/rosa/blob/5a3ede3c992fa9d1aec2fa21b018a6762768e710/pkg/ocm/helpers.go#L492
to be something that will reflect the functionality of this method. The name 'c.CheckIfAWSAccountExists' tells me that we are going to check if the AWSacocunt exists. Is this what the functionality does? I doubt it. I think it looks for roles. Doesn't it? @pvasant

oriAdler:
@zvikorn I swapped the order of the messages as you suggested and added a screenshot of the flow.

PR #654, opened 2022-03-28
Introducing managed service commands

renan-campos:
This PR has the create, list, describe, and delete commands for managed services. 
The commands are seperated into different commits and we contributed by me, @jmolmo and @vkareh


cmd/create/service/cmd.go
```diff
@@ -0,0 +1,360 @@
...
+	var roleARN string
+	var supportRoleARN string
+	var controlPlaneRoleARN string
+	var workerRoleARN string
+
+	role := aws.AccountRoles[aws.InstallerAccountRole]
+
+	roleARNs, err := awsClient.FindRoleARNs(aws.InstallerAccountRole, minor)
+	if err != nil {
+		reporter.Errorf("Failed to find %s role: %s", role.Name, err)
```
jmolmo, line 186: In this kind of messages i see that usually we enclose the subject between ticks, '%s', instead only %s. It seems that makes more easy to search in logs.
Apart of that, if we follow this pattern ("relevant element/subjects" tick enclosed), I  think we can use directly `%q` instead `'%s'` (this would be applied to all the messages in the MR)

cmd/dlt/service/cmd.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+	}()
+
+	if !confirm.Confirm("delete service with id '%s'", args.ID) {
+		os.Exit(0)
+	}
+
+	// First get the service to report additional resources
+	// that must be manually deleted.
+	service, err := ocmClient.GetManagedService(ocm.DescribeManagedServiceArgs{ID: args.ID})
```
jmolmo, line 81: What if we do not have the service in the database? (and err management.)

cmd/dlt/service/cmd.go
```diff
@@ -0,0 +1,119 @@
...
+	_, err = ocmClient.DeleteManagedService(args)
+	if err != nil {
+		reporter.Errorf("%s", err)
+		os.Exit(1)
+	}
+	reporter.Infof("Service '%s' will start uninstalling now", args.ID)
+
+	if service.Cluster().AWS().STS().RoleARN() != "" {
+		reporter.Infof(
+			"Your service '%s' will be deleted byt the following objects may remain",
```
jmolmo, line 93: typo: r/byt/but

cmd/dlt/service/cmd.go
```diff
@@ -0,0 +1,119 @@
...
+					" - %s\n", str,
+					operatorIAMRole.RoleARN())
+			}
+			reporter.Infof("%s", str)
+		}
+		reporter.Infof("OIDC Provider : %s\n", service.Cluster().AWS().STS().OIDCEndpointURL())
+		reporter.Infof("Once the service is uninstalled use the following commands to remove the " +
+			"above aws resources.\n")
+		commands := buildCommands(service.Cluster())
+		fmt.Print(commands, "\n")
```
jmolmo, line 109: why do you use the `reporter` for part of the messages and f`mt.Print` for others ?

jmolmo (changes requested):

cmd/create/service/cmd.go
```diff
@@ -0,0 +1,360 @@
...
+		}
+	}
+
+	// Find all installer roles in the current account using AWS resource tags
+	var roleARN string
+	var supportRoleARN string
+	var controlPlaneRoleARN string
+	var workerRoleARN string
+
+	role := aws.AccountRoles[aws.InstallerAccountRole]
```
ciaranRoche, line 182: This looks to only support STS? Is this correct? Shouldn't we support a "normal" ROSA cluster and an STS ROSA cluster?

@renan-campos @vkareh

cmd/dlt/service/cmd.go
renan-campos: Good catch!

cmd/dlt/service/cmd.go
renan-campos: We are leveraging managed service to do all of these checks for us. Managed service will return an error and we will report it to the end user.

cmd/dlt/service/cmd.go
renan-campos: The report prepends the message with information before hand. `fmt.Print` is used when that prepended format isn't needed.

cmd/dlt/service/cmd.go
vkareh: I think it makes sense to add basic error reporting here:
```
if err != nil {
  reporter.Errorf("%s", err)
  os.Exit(1)
}
```

cmd/list/service/cmd.go
```diff
@@ -0,0 +1,87 @@
...
+		reporter.Errorf("Failed to retrieve list of managed services: %v", err)
+		os.Exit(1)
+	}
+
+	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
+	fmt.Fprintf(writer, "ID\tSERVICE\tSTATE\n")
+	servicesList.Each(func(srv *msv1.ManagedService) bool {
+		fmt.Fprintf(writer, "%s\t%s\t%s\n", srv.ID(), srv.Service(), srv.ServiceState())
+		return true
+	})
```
vkareh, line 85: This needs to include the cluster name, otherwise it's impossible to tell which appliances you are seeing

cmd/dlt/service/cmd.go
renan-campos: Oh! I didn't realize I didn't add any error handling to this call.

cmd/list/service/cmd.go
renan-campos: Since the cluster name isn't being passed I added the cluster ID for now.

cmd/create/service/cmd.go
```diff
@@ -0,0 +1,360 @@
...
+	rprtr "github.com/openshift/rosa/pkg/reporter"
+)
+
+var args ocm.CreateManagedServiceArgs
+
+var Cmd = &cobra.Command{
+	Use:     "service",
+	Aliases: []string{"appliance"},
+	Short:   "Creates a managed service.",
+	Long: `  Managed Services are OpenShift clusters that provide a specific function.
```
ciaranRoche, line 45: This command is not showing up in the help 
```
rosa create --help 
Create a resource from stdin

Usage:
  rosa create [command]

Aliases:
  create, add

Available Commands:
  account-roles  Create account-wide IAM roles before creating your cluster.
  admin          Creates an admin user to login to the cluster
  cluster        Create cluster
  idp            Add IDP for cluster
  ingress        Add Ingress to cluster
  machinepool    Add machine pool to cluster
  oidc-provider  Create OIDC provider for an STS cluster.
  operator-roles Create operator IAM roles for a cluster.

Flags:
  -h, --help             help for create
      --profile string   Use a specific AWS profile from your credential file.
      --region string    Use a specific AWS region, overriding the AWS_REGION environment variable.
  -y, --yes              Automatically answer yes to confirm operation.

Global Flags:
      --debug   Enable debug mode.

Use "rosa create [command] --help" for more information about a command.

```

cmd/create/service/cmd.go
```diff
@@ -0,0 +1,360 @@
...
+		"",
+		"Type of service.",
+	)
+
+	flags.StringVar(
+		&args.ClusterName,
+		"name",
+		"",
+		"Name of the service instance.",
+	)
```
ciaranRoche, line 82: Should we make these as required, so that we can validate in the CLI the request is as expected before sending it to the API. At the moment its a little weird the flow and not overly clear in the error IMO 

```
$ rosa create service 
W: More than one Installer role found, using "arn:aws:iam::765374464689:role/ManagedOpenShift-Installer-Role"
I: Using "arn:aws:iam::765374464689:role/ManagedOpenShift-Worker-Role" for the Worker role
I: Using "arn:aws:iam::765374464689:role/ManagedOpenShift-Support-Role" for the Support role
I: Using "arn:aws:iam::765374464689:role/ManagedOpenShift-ControlPlane-Role" for the ControlPlane role
E: Failed to create managed service: service is required

```

cmd/describe/service/cmd.go
```diff
@@ -0,0 +1,99 @@
...
+  %20s: %s
+  %20s: %s
+`,
+		"Id", service.ID(),
+		"Href", service.HREF(),
+		"Service type", service.Service(),
+		"Service State", service.ServiceState(),
+		"Cluster Name", service.Cluster().Name(),
+		"Created At", service.CreatedAt(),
+		"Updated At", service.UpdatedAt())
```
ciaranRoche, line 97: I think the formatting is slightly off on this, I think we should have something similar to describe cluster 
```
╭─croche@localhost ~/Projects/sda/scripts 
╰─$ rosa describe service --id 27R3zVEr2DBnMojF2qPIyspcLuE

                    Id: 27R3zVEr2DBnMojF2qPIyspcLuE
                  Href: /api/service_mgmt/v1/services/27R3zVEr2DBnMojF2qPIyspcLuE
          Service type: ocm-addon-test-operator
         Service State: waiting for cluster
          Cluster Name: croche
            Created At: 2022-04-06 18:22:13 +0000 UTC
            Updated At: 2022-04-06 18:30:03 +0000 UTC
╭─croche@localhost ~/Projects/sda/scripts 
╰─$ rosa describe cluster --cluster croche                
Name:                       croche
ID:                         1rdoojlna7fitub4vg054t3pnmuivtov
External ID:                
OpenShift Version:          
DNS:                        croche.q6pk.s1.devshift.org
API URL:                    
Console URL:                
Region:                     us-east-1
Nodes:
 - Control plane:           3
 - Infra:                   2
 - Compute:                 2
Network:
 - Service CIDR:            172.30.0.0/16
 - Machine CIDR:            10.0.0.0/16
 - Pod CIDR:                10.128.0.0/14
 - Host Prefix:             /23
STS Role ARN:               arn:aws:iam::765374464689:role/ManagedOpenShift-Installer-Role
Support Role ARN:           arn:aws:iam::765374464689:role/ManagedOpenShift-Support-Role
Instance IAM Roles:
 - Control plane:           arn:aws:iam::765374464689:role/ManagedOpenShift-ControlPlane-Role
 - Worker:                  arn:aws:iam::765374464689:role/ManagedOpenShift-Worker-Role
Operator IAM Roles:
 - arn:aws:iam::765374464689:role/croche-p8o4-openshift-cluster-csi-drivers-ebs-cloud-credentials
 - arn:aws:iam::765374464689:role/croche-p8o4-openshift-cloud-network-config-controller-cloud-cred
 - arn:aws:iam::765374464689:role/croche-p8o4-openshift-machine-api-aws-cloud-credentials
 - arn:aws:iam::765374464689:role/croche-p8o4-openshift-cloud-credential-operator-cloud-credential
 - arn:aws:iam::765374464689:role/croche-p8o4-openshift-image-registry-installer-cloud-credentials
 - arn:aws:iam::765374464689:role/croche-p8o4-openshift-ingress-operator-cloud-credentials
State:                      waiting (Waiting for OIDC configuration)
Created:                    Apr  6 2022 18:22:11 UTC
Details Page:               https://qaprodauth.cloud.redhat.com/openshift/details/s/27R3zDCAcAaU4HATZrV030Vlp5y
OIDC Endpoint URL:          https://rh-oidc-staging.s3.us-east-1.amazonaws.com/1rdoojlna7fitub4vg054t3pnmuivtov

```

cmd/describe/service/cmd.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/openshift/rosa/pkg/ocm"
+	rprtr "github.com/openshift/rosa/pkg/reporter"
+)
+
+var args ocm.DescribeManagedServiceArgs
+
+var Cmd = &cobra.Command{
+	Use:     "service",
+	Aliases: []string{"appliance"},
+	Short:   "Show details of a service",
```
ciaranRoche, line 35: Same as create, this command is not showing up in the help

cmd/list/service/cmd.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/output"
+	rprtr "github.com/openshift/rosa/pkg/reporter"
+	"github.com/spf13/cobra"
+)
+
+var Cmd = &cobra.Command{
+	Use:     "services",
+	Aliases: []string{"service", "appliance", "appliances"},
+	Short:   "List managed services",
```
ciaranRoche, line 35: same as the others, not showing up in the help

ciaranRoche (commented):
Looks like we are missing some options, not sure if its a blocker for this MR as it is pretty large as it is. But we need to be able to set config such as cluster.network, cluster.aws.subnets, cluster.multiaz etc

cmd/dlt/service/cmd.go
```diff
@@ -0,0 +1,124 @@
...
+		"id",
+		"",
+		"The ID of the service to be deleted.")
+}
+
+func run(cmd *cobra.Command, _ []string) {
+	reporter := rprtr.CreateReporterOrExit()
+	logger := logging.CreateLoggerOrExit(reporter)
+
+	if !confirm.Confirm("delete service with id '%s'", args.ID) {
```
ciaranRoche, line 61: I think we should make arge.ID required or add a nil check, to save the following error when no id is passed 
```
rosa delete service 
? Are you sure you want to delete service with id ''? Yes
E: expected response content type 'application/json' but received '' and content ''

```

cmd/create/service/cmd.go
vkareh: That's on purpose: it's a hidden command (see `Hidden: true` a few lines below)

cmd/create/service/cmd.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+	parameters := addOn.Parameters()
+
+	if parameters.Len() > 0 {
+		args.Parameters = map[string]string{}
+		// Determine if all required parameters have already been set as flags.
+		if arguments.HasUnknownFlags() {
+			parameters.Each(func(param *cmv1.AddOnParameter) bool {
+				flag := cmd.Flags().Lookup(param.ID())
+				if param.Required() && (flag == nil || flag.Value.String() == "") {
```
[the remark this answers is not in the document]
vkareh: I think the issue here is the `if arguments.HasUnknownFlags()` - since these are addon parameters, and we don't support interactive flow, I think we need to remove that conditional check and run the loop for _all_ parameters defined in the addon.

cmd/describe/service/cmd.go
vkareh: It's on purpose

cmd/list/service/cmd.go
vkareh: On purpose

cmd/create/service/cmd.go
renan-campos: Resolve?

cmd/describe/service/cmd.go
renan-campos: Updated: 
```
[rcampos@rh-laptop rosa]$ ./rosa describe service --id 27QUb9SfWsHqHtMNpgnf1xD2A9N
Id:                         27QUb9SfWsHqHtMNpgnf1xD2A9N
Href:                       /api/service_mgmt/v1/services/27QUb9SfWsHqHtMNpgnf1xD2A9N
Service type:               ocs-provider-qe
Service State:              error
Cluster Name:               rc-test
Created At:                 2022-04-06 13:31:11 +0000 UTC
Updated At:                 2022-04-06 18:25:03 +0000 UTC
```

cmd/dlt/service/cmd.go
renan-campos: https://github.com/openshift/rosa/pull/654/commits/93d12bd3a5c5a7baba448edd030c74e32be47baa#diff-8c41d30fb00687e5de664bfb4cb8fab7a8bb0fac106c9b2c677049d600ce0d1fR61

vkareh:
This LGTM

@ciaranRoche if your comments are addressed, can you stamp this?

PR #658, opened 2022-03-31
support for cmk multi-region keys

thimslugga:
regex taken from botocore efs service file

https://github.com/boto/botocore/blob/ec5f655064d1e597e6a658c01be6193b65e08e65/botocore/data/efs/2015-02-01/service-2.json#L1596

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -47,7 +47,7 @@ import (
 )
 
 //nolint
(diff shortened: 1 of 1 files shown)


pvasant:
@arendej  Can you please take a look at this? Are we ready to support multi-region keys?

cblecker:
/ok-to-test
/cc @jharrington22 

@pvasant Yes, we do want to.

Issue #662, opened 2022-04-01
json output support for `create oidc-provider ...` and `create operator-roles ...`


jbpratt:
These commands currently lack support for json output, with #660 + #661 applied, this results in cluster creation `--output=json` no longer being strictly json. Partially due to no guards on the log statements in the STS block as well as logging in the commands themselves.

https://github.com/openshift/rosa/blob/2eab206e73e074f1150c33a83d3a06ce54e18b6f/cmd/create/cluster/cmd.go#L1803-L1806
https://github.com/openshift/rosa/blob/2eab206e73e074f1150c33a83d3a06ce54e18b6f/cmd/create/oidcprovider/cmd.go#L172
https://github.com/openshift/rosa/blob/2eab206e73e074f1150c33a83d3a06ce54e18b6f/cmd/create/oidcprovider/cmd.go#L226
https://github.com/openshift/rosa/blob/2eab206e73e074f1150c33a83d3a06ce54e18b6f/cmd/create/operatorroles/cmd.go#L329

PR #665, opened 2022-04-04
fix json output for cluster creation (sts+mode=auto)

jbpratt:
When executing with `--output=json` and `--mode=auto`, plain info logs
are printed from the create commands of `oidc-provider` and
`operator-roles` causing the output to no longer parse as JSON. Bug was
surfaced in #661 and mentioned in #662.

Simply add the output flag/terminal check conditional to offending log
statements. This does not include support for the commands themselves,
the `output` flag variable is set when calling from the cluster creation
command and prevents plain text logs. 

Test command:

```
rosa create cluster -c <cluster-name> --mode=auto --output=json --sts --yes | jq .
```

Signed-off-by: Brady Pratt <bpratt@redhat.com>

### DIFF
--- a/cmd/create/oidcprovider/cmd.go
+++ b/cmd/create/oidcprovider/cmd.go
@@ -35,6 +35,7 @@ import (
 	"github.com/openshift/rosa/pkg/interactive/confirm"
 	"github.com/openshift/rosa/pkg/logging"
 	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/output"
 	rprtr "github.com/openshift/rosa/pkg/reporter"
 )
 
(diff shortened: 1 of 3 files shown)


cmd/create/oidcprovider/cmd.go
```diff
@@ -169,7 +170,9 @@ func run(cmd *cobra.Command, argv []string) {
 				"Verify that the cluster operator roles exist and are configured correctly.", clusterKey)
 			os.Exit(1)
 		}
-		reporter.Infof("Creating OIDC provider using '%s'", creator.ARN)
+		if !output.HasFlag() || reporter.IsTerminal() {
```
jbpratt, line 173: Worth noting I'm not actually adding output support to these commands, but since `output` is set globally then it is respected when running from the cluster creation command.

PR #666, opened 2022-04-04
Add max throttle delay to avoid exponential backoff

oriAdler:
AWS SDK uses exponential backoff to wait longer between retries for consecutive error responses.
Because exponential backoff grows quickly, we are adding a maximum delay interval.

Reference: [https://aws.amazon.com/blogs/mt/managing-monitoring-api-throttling-in-workloads/](https://aws.amazon.com/blogs/mt/managing-monitoring-api-throttling-in-workloads/)


oriAdler:
@pvasant @LeshemNoa @tzvatot PTAL

Not sure if this will solve the problem, but worth trying.
I've tried a couple of times to call `upgrade cluster` which triggers `listPolicyTags` and it was better than before this change.

pvasant:
FYI @jharrington22

PR #671, opened 2022-04-08
add cloudwatch:GetMetricData STS installer permission policy

zvikorn:

[SDA-5791](https://issues.redhat.com/browse/SDA-5791)

@pvasant


pvasant:
This is not needed given that we have policies in OCM . closing

PR #675, opened 2022-04-11
HTPasswd: Add username & password validations in CLI

LeshemNoa:
This is done in order to display meaningful error messages to ROSA CLI users early during the process of IDP creation. Doing the these validations on the CLI side while they users are constructing the IDP payload (to be sent and validated in the same manner in CS) provides better UX, for example:
```
I: Interactive mode enabled.
? Type of identity provider: htpasswd
? Identity provider name: htpasswd-1
I: At least one user is required to create the IDP.
X Sorry, your reply was invalid: invalid username 'user%%': username must not contain /, :, or %
```

PTAL @pvasant

### DIFF
--- a/cmd/create/idp/htpasswd.go
+++ b/cmd/create/idp/htpasswd.go
@@ -19,6 +19,8 @@ package idp
 import (
 	"fmt"
 	"os"
+	"regexp"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


tbrisker:
/lgtm
not a big fan of these password rules, but since they are what we're currently enforcing...

PR #681, opened 2022-04-14
Fix bug - delete account roles - make the `prefix` flag optional

oriAdler:
Modify `rosa delete account-roles` to permit running the command without the `prefix` flag.
If the flag isn't provided, change to interactive mode, and get the prefix from the user.

image

Related: [SDA-5795](https://issues.redhat.com/browse/SDA-5797)


oriAdler:
@ciaranRoche @zvikorn @pvasant PTAL

PR #682, opened 2022-04-14
Upgrade cluster to 4.10.* - add delay after roles creation

oriAdler:
CS first attempts to validate the new roles right after the creation fails.
Add 5 seconds delay with a spinner to reduce the number of retries on conflict in CS.
It's better to wait in the CLI, instead of blocking both CLI and CS.

image
image

Related: [SDA-5753](https://issues.redhat.com/browse/SDA-5753)

### DIFF
--- a/cmd/upgrade/operatorroles/cmd.go
+++ b/cmd/upgrade/operatorroles/cmd.go
@@ -20,8 +20,10 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/openshift/rosa/pkg/helper"
 	"github.com/spf13/cobra"
 
 	"github.com/openshift/rosa/pkg/aws"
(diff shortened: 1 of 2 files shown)


oriAdler:
@tzvatot @pvasant PTAL

cmd/upgrade/operatorroles/cmd.go
```diff
@@ -347,6 +349,11 @@ func createOperatorRole(mode string, reporter *rprtr.Object, awsClient aws.Clien
 		if err != nil {
 			return err
 		}
+		// if an operators upgrade is triggered from the 'upgrade cluster' command,
+		// wait before creating an upgrade policy to avoid the retry loop in CS
+		if isProgrammaticallyCalled {
```
pvasant, line 354: Why only programmaticallyCalled?

oriAdler: I thought that if a user calls `rosa upgrade operator-roles` from the console, and afterward `rosa upgrade cluster` there is going to be enough time for the roles to reconcile.

Should I remove this `if` statement so we add the delay for all cases?

tzvatot:
lgtm, best if @pvasant  will also TAL

cmd/upgrade/operatorroles/cmd.go
pvasant: I think so  we cover all scenarios

oriAdler: The `if` statement was removed.

PR #683, opened 2022-04-14
Adding byo vpc for managed services

renan-campos:


cmd/create/service/cmd.go
```diff
@@ -122,7 +137,10 @@ func run(cmd *cobra.Command, argv []string) {
 	awsClient := aws.GetAWSClientForUserRegion(reporter, logger)
 
 	// Get AWS region
-	args.AwsRegion, err = aws.GetRegion(arguments.GetRegion())
+	args.AwsRegion = region.Region()
+	if args.AwsRegion == "" {
+		args.AwsRegion, err = aws.GetRegion(arguments.GetRegion())
+	}
```
vkareh, line 143: This change is redundant, `arguments.GetRegion()` already call `region.Region()` internally.

pkg/ocm/managedservice.go
```diff
@@ -22,6 +22,10 @@ type CreateManagedServiceArgs struct {
 	AwsRegion              string
 
 	AwsOperatorIamRoleList []OperatorIAMRole
+
+	MultiAZ              bool
+	AwsAvailabilityZones []string
```
vkareh, line 27: I would rename this to `AvailabilityZones`, since everything in ROSA is implied AWS already.

vkareh (changes requested):

PR #684, opened 2022-04-18
Allow for additional, customer-provided "no_proxy" values for cluster-wide proxy

zvikorn:

[SDA-5612](https://issues.redhat.com//browse/SDA-5612)

@pvasant @vkareh  PTAL

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -119,6 +119,7 @@ var args struct {
 	enableProxy               bool
 	httpProxy                 string
 	httpsProxy                string
+	noProxySlice              []string
 	additionalTrustBundleFile string
 
 	tags []string
@@ -285,6 +286,14 @@ func init() {
 		"https-proxy",
 		"",
 		"A proxy URL to use for creating HTTPS connections outside the cluster.",
+	)
+
+	flags.StringSliceVar(
+		&args.noProxySlice,
+		"no-proxy",
+		nil,
+		"A comma-separated list of destination domain names, domains, IP addresses or "+
+			"other network CIDRs to exclude proxying. Use \"*\" to bypass proxy for all destinations",
 	)
 
 	flags.StringVar(
@@ -1166,8 +1175,9 @@ func run(cmd *cobra.Command, _ []string) {
 	enableProxy := false
 	httpProxy := args.httpProxy
 	httpsProxy := args.httpsProxy
+	noProxySlice := args.noProxySlice
 	additionalTrustBundleFile := args.additionalTrustBundleFile
-	if httpProxy != "" || httpsProxy != "" || additionalTrustBundleFile != "" {
+	if httpProxy != "" || httpsProxy != "" || len(noProxySlice) > 0 || additionalTrustBundleFile != "" {
 		useExistingVPC = true
 		enableProxy = true
 	}
@@ -1596,7 +1606,7 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	// Cluster-wide proxy configuration
-	if useExistingVPC && !enableProxy && interactive.Enabled() {
+	if (subnetsProvided || (useExistingVPC && !enableProxy)) && interactive.Enabled() {
 		enableProxy, err = interactive.GetBool(interactive.Input{
 			Question: "Use cluster-wide proxy",
 			Help: "To install cluster-wide proxy, you need to set one of the following attributes: 'http-proxy', " +
@@ -1646,6 +1656,43 @@ func run(cmd *cobra.Command, _ []string) {
 	err = interactive.IsURL(httpsProxy)
 	if err != nil {
 		reporter.Errorf("%s", err)
+		os.Exit(1)
+	}
+
+	if enableProxy && interactive.Enabled() {
+		noProxyInput, err := interactive.GetString(interactive.Input{
+			Question: "No proxy",
+			Help:     cmd.Flags().Lookup("no-proxy").Usage,
+			Default:  strings.Join(noProxySlice, ","),
+			Validators: []interactive.Validator{
+				aws.UserNoProxyValidator,
+				aws.UserNoProxyDuplicateValidator,
+			},
+		})
+		if err != nil {
+			reporter.Errorf("Expected a valid set of no proxy domains/CIDR's: %s", err)
+			os.Exit(1)
+		}
+		noProxySlice = strings.Split(noProxyInput, ",")
+	}
+
+	if len(noProxySlice) > 0 {
+		duplicate, found := aws.HasDuplicates(noProxySlice)
+		if found {
+			reporter.Errorf("Invalid no-proxy list, duplicate key '%s' found", duplicate)
+			os.Exit(1)
+		}
+		for _, domain := range noProxySlice {
+			err := aws.UserNoProxyValidator(domain)
+			if err != nil {
+				reporter.Errorf("%s", err)
+				os.Exit(1)
+			}
+		}
+	}
+
+	if httpProxy == "" && httpsProxy == "" && len(noProxySlice) > 0 {
+		reporter.Errorf("Expected at least one of the following: http-proxy, https-proxy")
 		os.Exit(1)
 	}
 
@@ -1732,6 +1779,10 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 	if httpsProxy != "" {
 		clusterConfig.HTTPSProxy = &httpsProxy
+	}
+	if len(noProxySlice) > 0 {
+		str := strings.Join(noProxySlice, ",")
+		clusterConfig.NoProxy = &str
 	}
 	if additionalTrustBundleFile != "" {
 		clusterConfig.AdditionalTrustBundleFile = &additionalTrustBundleFile
@@ -2099,6 +2150,9 @@ func buildCommand(spec ocm.Spec, operatorRolesPrefix string) string {
 		}
 		if spec.HTTPSProxy != nil && *spec.HTTPSProxy != "" {
 			command += fmt.Sprintf(" --https-proxy %s", *spec.HTTPSProxy)
+		}
+		if spec.NoProxy != nil && *spec.NoProxy != "" {
+			command += fmt.Sprintf(" --no-proxy \"%s\"", *spec.NoProxy)
 		}
 	}
 	if spec.AdditionalTrustBundleFile != nil && *spec.AdditionalTrustBundleFile != "" {
--- a/cmd/edit/cluster/cmd.go
+++ b/cmd/edit/cluster/cmd.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"strings"
 	"time"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
@@ -45,6 +46,7 @@ var args struct {
 	disableWorkloadMonitoring bool
 	httpProxy                 string
 	httpsProxy                string
+	noProxySlice              []string
 	additionalTrustBundleFile string
 }
 
@@ -111,6 +113,14 @@ func init() {
 		"A proxy URL to use for creating HTTPS connections outside the cluster.",
 	)
 
+	flags.StringSliceVar(
+		&args.noProxySlice,
+		"no-proxy",
+		nil,
+		"A comma-separated list of destination domain names, domains, IP addresses or "+
+			"other network CIDRs to exclude proxying. Use \"*\" to bypass proxy for all destinations",
+	)
+
 	flags.StringVar(
 		&args.additionalTrustBundleFile,
 		"additional-trust-bundle-file",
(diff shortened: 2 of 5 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -287,6 +288,13 @@ func init() {
 		"A proxy URL to use for creating HTTPS connections outside the cluster.",
 	)
 
+	flags.StringVar(
```
vkareh, line 291: Maybe this should be `flags.StringSliceVar`, that way you can do both comma-separated and multiple flags:
```
$ rosa ... --no-proxy foo --no-proxy bar
```
as well as
```
$rosa ... --no-proxy foo,bar
```
and they will automatically be stored in a `[]string`

cmd/create/cluster/cmd.go
```diff
@@ -119,6 +119,7 @@ var args struct {
 	enableProxy               bool
 	httpProxy                 string
 	httpsProxy                string
+	noProxy                   string
```
vkareh, line 122: This should be `[]string` - follow the same example as `args.tags`

cmd/create/cluster/cmd.go
```diff
@@ -287,6 +288,13 @@ func init() {
 		"A proxy URL to use for creating HTTPS connections outside the cluster.",
 	)
 
+	flags.StringVar(
+		&args.noProxy,
+		"no-proxy",
+		"",
+		"A comma-separated list of domains and CIDRs for which the proxy should not be used.",
```
vkareh, line 295: > A comma-separated list of destination domain names, domains, IP addresses or other network CIDRs to exclude proxying.

(from the [docs](https://docs.openshift.com/container-platform/4.10/networking/enable-cluster-wide-proxy.html#nw-proxy-configure-object_config-cluster-wide-proxy))

cmd/create/cluster/cmd.go
```diff
@@ -1649,6 +1658,26 @@ func run(cmd *cobra.Command, _ []string) {
...
+			Validators: []interactive.Validator{
+				interactive.IsURL,
+			},
+		})
+		if err != nil {
+			reporter.Errorf("Expected a valid no proxy: %s", err)
+			os.Exit(1)
+		}
+	}
+	err = interactive.IsURL(noProxy)
```
vkareh, line 1675: same here

cmd/edit/cluster/cmd.go
```diff
@@ -111,6 +112,13 @@ func init() {
 		"A proxy URL to use for creating HTTPS connections outside the cluster.",
 	)
 
+	flags.StringVar(
+		&args.noProxy,
+		"no-proxy",
+		"",
+		"A comma-separated list of domains and CIDRs for which the proxy should not be used.",
+	)
```
vkareh, line 120: Make sure this matches the one in the `create` command, if changed.

cmd/create/cluster/cmd.go
```diff
@@ -1649,6 +1658,26 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	if enableProxy && interactive.Enabled() {
+		noProxy, err = interactive.GetString(interactive.Input{
+			Question: "No proxy",
+			Help:     cmd.Flags().Lookup("no-proxy").Usage,
+			Default:  noProxy,
+			Validators: []interactive.Validator{
+				interactive.IsURL,
```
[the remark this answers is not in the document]
zvikorn: @vkareh Done. However, when I send --no-proxy *  I see ROSA sends out 'assets'. Any idea how '*' becomes 'assets'?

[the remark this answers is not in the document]
vkareh: Seems like the `*` is being expanded into the local directory (first directory in the `rosa` repo is `assets`).

cmd/create/cluster/cmd.go
```diff
@@ -287,6 +288,13 @@ func init() {
 		"A proxy URL to use for creating HTTPS connections outside the cluster.",
 	)
 
+	flags.StringSliceVar(
+		&args.noProxySlice,
+		"no-proxy",
+		nil,
+		"A comma-separated list of destination domain names, domains, IP addresses or other network CIDRs to exclude proxying.",
```
vkareh, line 295: This line is too long and fails linting
```
cmd/create/cluster/cmd.go:295: line is 122 characters (lll)
		"A comma-separated list of destination domain names, domains, IP addresses or other network CIDRs to exclude proxying.",
```

cmd/edit/cluster/cmd.go
```diff
@@ -111,6 +113,13 @@ func init() {
 		"A proxy URL to use for creating HTTPS connections outside the cluster.",
 	)
 
+	flags.StringSliceVar(
+		&args.noProxySlice,
+		"no-proxy",
+		nil,
+		"A comma-separated list of destination domain names, domains, IP addresses or other network CIDRs to exclude proxying.",
```
vkareh, line 120: This line is too long and fails linting:
```
cmd/edit/cluster/cmd.go:120: line is 122 characters (lll)
		"A comma-separated list of destination domain names, domains, IP addresses or other network CIDRs to exclude proxying.",
```

cmd/edit/cluster/cmd.go
```diff
@@ -215,20 +224,26 @@ func run(cmd *cobra.Command, _ []string) {
...
+	}
 	var additionalTrustBundleFile *string
 	var additionalTrustBundleFileValue string
 	if cmd.Flags().Changed("additional-trust-bundle-file") {
 		additionalTrustBundleFileValue = args.additionalTrustBundleFile
 		additionalTrustBundleFile = &additionalTrustBundleFileValue
 	}
 
-	if httpProxy != nil || httpsProxy != nil || additionalTrustBundleFile != nil {
+	if httpProxy != nil || httpsProxy != nil || (noProxySlice != nil && len(noProxySlice) > 0) ||
```
vkareh, line 238: Fails linting:
```
cmd/edit/cluster/cmd.go:238:47: S1009: should omit nil check; len() for nil slices is defined as zero (gosimple)
	if httpProxy != nil || httpsProxy != nil || (noProxySlice != nil && len(noProxySlice) > 0) ||
	                                             ^
```

cmd/edit/cluster/cmd.go
```diff
@@ -215,20 +224,26 @@ func run(cmd *cobra.Command, _ []string) {
...
+		additionalTrustBundleFile != nil {
 		enableProxy = true
 		useExistingVPC = true
 	}
 
 	if len(cluster.AWS().SubnetIDs()) == 0 &&
 		((httpProxy != nil && *httpProxy != "") || (httpsProxy != nil && *httpsProxy != "") ||
+			(noProxySlice != nil && len(noProxySlice) > 0) ||
```
vkareh, line 246: Fails linting:
```
cmd/edit/cluster/cmd.go:246:5: S1009: should omit nil check; len() for nil slices is defined as zero (gosimple)
			(noProxySlice != nil && len(noProxySlice) > 0) ||
			 ^
```

cmd/create/cluster/cmd.go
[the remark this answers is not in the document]
zvikorn: Any idea how to overcome this?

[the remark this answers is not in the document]
vkareh: No idea - maybe revert back to using `string` if this is an issue - I think using `"*"` (with quotes) should do the trick, but we can't assume users will know that.

Alternatively, check if the value is a local path, then set `*` internally - but that makes a lot of assumptions...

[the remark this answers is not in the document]
zvikorn: can't we have something like this No Proxy [?for help] (type 'all' to bypass all.....) as we have in -
Operator roles prefix: [? for help] (tavni-w5m2)
I know it's different, but I think about how to deliver this to the customer....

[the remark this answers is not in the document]
zvikorn: so reverting everting to a single string as it was before?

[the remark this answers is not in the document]
zvikorn: but wait... how a single string will help? Still need to tun regex on it...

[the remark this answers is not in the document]
zvikorn: @vkareh ^^
Also could you please start the build of the MR?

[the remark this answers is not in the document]
vkareh: String won't help. Bash automatically expands the `*` glob before the command is even called, so Cobra has no way of knowing that the user entered a naked asterisk. This is true of any CLI.

Possible solutions:
- Check if the value matches the first file in the current working directory, then replace by asterisk
- Check if the value matches the first file in the current working directory, then print an error/warning telling the user to surround asterisks in quotes
- Put in the help text that the value should be surrounded in quotes
- Do nothing and let the user figure it out based on the error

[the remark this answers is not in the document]
zvikorn: Added a warning before the 'No Proxy' prompt
```
? HTTPS proxy (optional): 
W: use "*" with double quotes to bypass proxy for all destinations
? No proxy (optional):
```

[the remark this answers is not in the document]
zvikorn: @vkareh can you approve the MR please?

cmd/create/cluster/cmd.go
```diff
@@ -1649,6 +1659,44 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	if enableProxy && interactive.Enabled() {
+		reporter.Warnf("use \"*\" with double quotes to bypass proxy for all destinations")
```
vkareh, line 1663: Since this is for interactive mode, I think we can use the `interactive.PrintHelp` here, as it flows better with the rest of the interactive prompts (it's not really a warning).

zvikorn: @vkareh do you see another way to show this?
```
? HTTPS proxy (optional): 
? use "*" with double quotes to bypass proxy for all destinations
? No proxy (optional): [? for help] 
```
I could not place the interactive.PrintHelp elsewhere as it returns an error that needs to be checked

vkareh: hmm... I just compiled this and the reality is that for interactive mode we don't need the double-quotes. In fact, entering double quotes gave me an error, but a naked asterisk worked:
```
? Use cluster-wide proxy (optional): Yes
? HTTP proxy (optional): http://localhost
? HTTPS proxy (optional): 
? No proxy: use "*" with double quotes to bypass proxy for all destinations
X Sorry, your reply was invalid: expected a valid user no-proxy value: '"*"' matching ^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$|^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/(3[0-2]|[1-2][0-9]|[0-9]))$|^.?(([a-zA-Z]|[a-zA-Z][a-zA-Z0-9-]*[a-zA-Z0-9])\.)*([A-Za-z]|[A-Za-z][A-Za-z0-9-]*[A-Za-z0-9])$|^\*$|^""$
? No proxy (optional): *
? Additional trust bundle file path (optional): 
```
I think just remove this warning entirely

cmd/create/cluster/cmd.go
```diff
@@ -2100,6 +2152,9 @@ func buildCommand(spec ocm.Spec, operatorRolesPrefix string) string {
 		if spec.HTTPSProxy != nil && *spec.HTTPSProxy != "" {
 			command += fmt.Sprintf(" --https-proxy %s", *spec.HTTPSProxy)
 		}
+		if spec.NoProxy != nil && *spec.NoProxy != "" {
+			command += fmt.Sprintf(" --no-proxy %s", *spec.NoProxy)
```
vkareh, line 2156: You should surround the value by quotes to be safe:
```
fmt.Sprintf(" --no-proxy \"%s\"", *spec.NoProxy)
```

cmd/create/cluster/cmd.go
zvikorn: So we are left with NON interactive mode where   --no-proxy *    still converts it to assets.  I'm in favor of adding a description to the 'rosa create cluster --help' no-proxy field and letting the user figure this out. WDYT? The only other way is to compare to the first directory name in the project, but it's not so easy, as the OS call returns also the hidden directories so the first directory is not 'assets' but a hidden one.

vkareh: > I'm in favor of adding a description to the 'rosa create cluster --help' no-proxy field and letting the user figure this out.

I think we have no choice, let's do that.

> The only other way is to compare to the first directory name in the project, but it's not so easy, as the OS call returns also the hidden directories so the first directory is not 'assets' but a hidden one.

That's opening a can of worms and I don't think we should pursue that.

zvikorn: so what's your solution then? Reading all directories and removing the hidden directories and files from the list, in order to get the first one?

vkareh: uh, no, we don't want to do that.

What you said in the previous comment: Let's add a description to the `rosa create cluster --help` no-proxy field and let the user figure it out.

zvikorn:
@vkareh LGTY?

PR #690, opened 2022-04-22
supporting different regions for create service command

renan-campos:


cmd/create/service/cmd.go
```diff
@@ -134,7 +134,14 @@ func run(cmd *cobra.Command, argv []string) {
 		}
 	}()
 
-	awsClient := aws.GetAWSClientForUserRegion(reporter, logger)
+	awsClient, err := aws.NewClient().
```
ciaranRoche, line 137: Is this something we want to do? Shouldn't we be using `GetAWSClientForUserRegion`? 

@pvasant what is the reasoning of defaulting to the `init` region if it does not match the `region` passed to the command, should we be following that pattern or should we be defaulting to the `region` passed. 

I am just wondering is this issue something that could be solved with documentation :thinking: like initializing rosa with the same region passed

PR #695, opened 2022-04-28
output validation error message when creating service with invalid parameter

renan-campos:
parameter

### DIFF
--- a/cmd/create/service/cmd.go
+++ b/cmd/create/service/cmd.go
@@ -187,7 +187,13 @@ func run(cmd *cobra.Command, argv []string) {
 				if val != "" && param.Validation() != "" {
 					isValid, err := regexp.MatchString(param.Validation(), val)
 					if err != nil || !isValid {
-						reporter.Errorf("Expected %v to match /%s/", val, param.Validation())
+						valErrMsg := param.ValidationErrMsg()
+						if valErrMsg != "" {
+							reporter.Errorf("Failed to process parameter --%s: %s", param.ID(), valErrMsg)
+						} else {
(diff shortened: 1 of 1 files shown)


cmd/create/service/cmd.go
```diff
@@ -187,7 +187,13 @@ func run(cmd *cobra.Command, argv []string) {
...
 					isValid, err := regexp.MatchString(param.Validation(), val)
 					if err != nil || !isValid {
-						reporter.Errorf("Expected %v to match /%s/", val, param.Validation())
+						valErrMsg := param.ValidationErrMsg()
+						if valErrMsg != "" {
+							reporter.Errorf("Failed to process parameter --%s: %s", param.ID(), valErrMsg)
+						} else {
+							reporter.Errorf("Failed to process parameter --%s: Expected %v to match /%s/",
+								param.ID(), val, param.Validation())
+						}
```
ciaranRoche, line 196: Changes look good to me, but I am wondering is it necessary to do this validation in the CLI? 

This validation and more is handled in both Managed Service and Clusters Service. So would it be better just to skip the validation in the CLI for the params and leave OCM handle this

renan-campos: That's true that we'll get the same validation error message back when we make the call to managed service. However, I think this approach has a better user experience since you get the error back immediately, whereas with making the call to managed service the error will come back after the sts roles have all been determined and outputted to the console.

ciaranRoche: I am just conscious of maintaining the same logic in multiple places. But I don't have a strong opinion on it. Lets merge it for now, but I would be against adding any more validations here in favour of using OCM for the validation

PR #696, opened 2022-05-03
allow addons be edited, regardless of addon state

ciaranRoche:
JIRA - [SDA-5935](https://issues.redhat.com//browse/SDA-5935)


ciaranRoche:
@vkareh PTAL

PR #698, opened 2022-05-03
adding new alias for managed service commands

renan-campos:


vkareh:
The alias doesn't show up when calling `--help`. Do we want to make `managed-service` the default one? @arendej

arendej:
> Do we want to make managed-service the default one? @arendej

Yes please

renan-campos:
> > Do we want to make managed-service the default one? @arendej
> 
> Yes please

Changed it! Just fyi though: these commands are still hidden from `help`

vkareh:
/lgtm

@arendej, I just want to voice my disapproval of the resource name. This should be `rosa create appliance`. Names like `"[Managed] Service"` are already overloaded... :stuck_out_tongue: 

Still, code change LGTM. Merging.

PR #700, opened 2022-05-04
adding private-link flag to managed service create

renan-campos:


vkareh:
@renan-campos that's a lot of lib updates for an SDK bump - are all of those necessary?

renan-campos:
> @renan-campos that's a lot of lib updates for an SDK bump - are all of those necessary?

Yes, they originate from bumping `github.com/openshift-online/ocm-sdk-go` to `v0.1.265` in `go.mod` and running `go mod vendor`. I needed to do these steps for rosa to compile successfully.

pkg/ocm/managedservice.go
```diff
@@ -45,6 +46,13 @@ func (c *Client) CreateManagedService(args CreateManagedServiceArgs) (*msv1.Mana
 			msv1.NewServiceParameter().ID(id).Value(val))
 	}
 
+	var listeningMethod msv1.ListeningMethod
+	if args.Privatelink {
+		listeningMethod = msv1.ListeningMethodInternal
+	} else {
+		listeningMethod = msv1.ListeningMethodExternal
```
tbrisker, line 53: do we want to handle this here? or should the cli not pass the `Listening` option at all to MS if it wasn't specified?

pkg/ocm/managedservice.go
```diff
@@ -66,7 +76,8 @@ func (c *Client) CreateManagedService(args CreateManagedServiceArgs) (*msv1.Mana
 								WorkerRoleARN(args.AwsWorkerRoleARN)).
 							OperatorIAMRoles(operatorIamRoles...)).
 						AccountID(args.AwsAccountID).
-						SubnetIDs(args.SubnetIDs...)).
+						SubnetIDs(args.SubnetIDs...).
+						PrivateLink(args.Privatelink)).
```
tbrisker, line 80: Why are we passing both `ListeningMethod` and `PrivateLink`? Shouldn't this logic be done in the MS? (i.e. if PrivateLink is set, set the ListeningMethod accordingly, or vice versa)

vkareh: We do this today to ensure that the interactive flow during `rosa create cluster` is consistent. The API should do its own validation as well to ensure that you can't have public privatelink clusters.

renan-campos: >  The API should do its own validation as well to ensure that you can't have public privatelink clusters.

I tested this out, and once the request gets to CS, cluster creation fails if the listening method isn't set to "Internal".

So are we ok with keeping this logic in Rosa, or should it be moved to MS?

vkareh: Yes, I'm okay with keeping this in ROSA for now

tbrisker:
@renan-campos please rebase, we've merged the sdk bump separately

renan-campos:
Thanks @tbrisker, just rebased!

pkg/ocm/managedservice.go
tbrisker: Just to confirm, this will effectively set `Listening(ListeningMethodExternal)` for all clusters that don't specify the `PrivateLink` flag explicitly. Is this the correct default behavior? If so, I'm ok with merging this PR

pkg/ocm/managedservice.go
tbrisker: I won't block on this, but I'm still not sure I understand why we need to pass two api parameters to MS here instead of one and letting MS set the other one accordingly. 
The MS api doesn't have to match the CS api. We can set the cluster listening mode inside MS according to the privatelink parameter instead of passing it from ROSA. 
Or do we want to expose the listening mode as another option for the customers? I much prefer to keep the MS api as minimal as possible.

pkg/ocm/managedservice.go
renan-campos: Marking this as resolved since we confirmed this in slack!

PR #701, opened 2022-05-06
cluster: Allow using local AWS credentials

vkareh:
This is an experimental internal feature where we allow local
credentials to be used for creating legacy IAM clusters.

https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/4148

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -61,6 +61,8 @@ var args struct {
 	fakeCluster bool
 	// Set custom properties in cluster spec
 	properties []string
+	// Use local AWS credentials instead of the 'osdCcsAdmin' user
+	useLocalCredentials bool
 
 	// Disable SCP checks in the installer
 	disableSCPChecks bool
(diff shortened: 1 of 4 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -1869,6 +1879,13 @@ func run(cmd *cobra.Command, _ []string) {
 	if args.fakeCluster {
 		props = append(props, properties.FakeCluster)
 	}
+	if args.useLocalCredentials {
+		if isSTS {
+			r.Reporter.Errorf("Local credentials are not supported for STS clusters")
```
yasun1, line 1884: Does it mean if the cluster is a rosa sts cluster, the osdCcsAdmin credential is still upserted?

vkareh: STS clusters do not use the `osdCcsAdmin` credentials. Only running `rosa init` creates those creds, but that command is not necessary for STS clusters.

PR #702, opened 2022-05-10
login: Allow tokens without 'typ' claim

vkareh:
Some authentication providers do not add a 'typ' claim in their token.
In this case we assume it's a Bearer token. We also verify that the
username is set in either the 'username' or 'preferred_username' claims.


pkg/aws/helpers.go
```diff
@@ -352,11 +354,30 @@ func GetOperatorPolicyARN(accountID string, prefix string, namespace string, nam
 }
 
 func GetPolicyARN(accountID string, name string) string {
-	return fmt.Sprintf("arn:aws:iam::%s:policy/%s", accountID, name)
+	partition := "aws"
+	region, err := GetRegion(arguments.GetRegion())
```
mjlshen, line 358: Given an AWS region you could also use `endpoints.PartitionForRegion` https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/#PartitionForRegion so that this wouldn't have to be hardcoded

something like:
```go
partition, ok := endpoints.PartitionForRegion(endpoints.DefaultPartitions(), region)
if !ok {
	 return nil, fmt.Errorf("no AWS partition found for region %s", region)
}
```

vkareh: Nice!! I was wondering if there was something like this, I was looking in the `arn` pkg and couldn't see anything.

vkareh:
@dustman9000 @mjlshen @tonytheleg @markturansky - please take a look and confirm that this latest version works for you (with the correct vanity URLs, on all environments). Would be nice to get this merged/released soon.

tonytheleg:
> @dustman9000 @mjlshen @tonytheleg @markturansky - please take a look and confirm that this latest version works for you (with the correct vanity URLs, on all environments). Would be nice to get this merged/released soon.

Current Integration is working with the vanity URLs but changes still need applying in Stage and Prod

tonytheleg:
@vkareh Vanity URLs are fixed and this is good to go!

PR #709, opened 2022-05-15
Create user-role - improve help message

oriAdler:
Add an example of creating a user role with a permission boundary.

Related: [SDA-6013](https://issues.redhat.com/browse/SDA-6013)

image


oriAdler:
@zgalor @tbrisker PTAL

PR #717, opened 2022-05-23
command to list parameters of add-on installation

renan-campos:
== Describe Add-on Installation sample output start ==

Id:                          bar
Href:                        /api/clusters_mgmt/v1/clusters/foo/addons/bar
Addon state:                 ready
Parameters:
        "id_1"                      : "value_1"
        "id_2"                      : "value_2"

== Describe Add-on Installation sample output end ==

### DIFF
--- a/cmd/describe/cmd.go
+++ b/cmd/describe/cmd.go
@@ -22,6 +22,7 @@ import (
 	"github.com/openshift/rosa/cmd/describe/addon"
 	"github.com/openshift/rosa/cmd/describe/admin"
 	"github.com/openshift/rosa/cmd/describe/cluster"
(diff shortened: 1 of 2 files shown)


ciaranRoche (commented):
Works well, can we split this into two MR's

Remove the reader interface and test to a seperate MR, would like to get the ROSA experts to have a look on that approach. Where as we can just merge the describe function on its own.

ciaranRoche:
@renan-campos If there is not a Jira for this, can you create one and link them, just so we can get this through QE

renan-campos:
> @renan-campos If there is not a Jira for this, can you create one and link them, just so we can get this through QE

will do

ciaranRoche:
!JIRA - SDA-6174

PR #719, opened 2022-05-26
Fix a bug when editing no-proxy field

zvikorn:

[SDA-6062](https://issues.redhat.com//browse/SDA-6062)

@zgalor @oriAdler @pvasant

### DIFF
--- a/cmd/edit/cluster/cmd.go
+++ b/cmd/edit/cluster/cmd.go
@@ -432,7 +432,7 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 		noProxySlice = strings.Split(noProxyInput, ",")
 	}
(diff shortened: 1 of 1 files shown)


cmd/edit/cluster/cmd.go
```diff
@@ -432,7 +432,7 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 		noProxySlice = strings.Split(noProxyInput, ",")
 	}
-	if httpProxy == nil && httpsProxy == nil && len(noProxySlice) > 0 {
+	if httpProxy == nil && httpsProxy == nil && len(noProxySlice) > 0 && cluster.Proxy() == nil {
```
zgalor, line 435: We better put this if logic in a func

PR #721, opened 2022-05-30
Reduce extra call to OCM when manipulating addons

tbrisker:
Looking up an addon installation triggers another call to OCM just to
get the cluster ID, but we already have it every time we fetch the
installation.


tbrisker:
@vkareh ptal

tbrisker:
hold with reviews, applying same change to the rest of the methods

PR #733, opened 2022-06-08
support creation of managed services with non-custom configurations

renan-campos:


renan-campos:
SDA-6175

ciaranRoche:
/lgtm

validated works well

PR #735, opened 2022-06-09
Ensure all flags passed during managed service creation are used.

renan-campos:


cmd/create/service/cmd.go
```diff
@@ -236,6 +249,23 @@ func run(cmd *cobra.Command, argv []string) {
 		})
 	}
 
+	// Ensure all flags were used.
+	var unusedFlags bool
+	cmd.Flags().VisitAll(func(flag *pflag.Flag) {
+		reporter.Infof("%q", flag.Name)
```
ciaranRoche, line 255: Not sure we want to keep this in here?

ciaranRoche (commented):
@renan-campos is there a ticket for this?

cmd/create/service/cmd.go
renan-campos: You're right, I was using that for debugging

renan-campos:
> @renan-campos is there a ticket for this?

https://issues.redhat.com/browse/SDE-2012

ciaranRoche:
Verified 
```
$ rosa create service --name=croche-test-1 --type=ocm-addon-test-operator --has-external-resources=true --machine-cidr='10.1.1.1/16' --noop=true
E: Cannot create managed service with the following unknown flags: ("has-external-resources", "noop")
```

ciaranRoche:
JIRA SDA-6220

PR #738, opened 2022-06-13
improving cluster-admins idp flow

zvikorn:

[SDA-6170](https://issues.redhat.com//browse/SDA-6170)

@pvasant

### DIFF
--- a/cmd/create/admin/cmd.go
+++ b/cmd/create/admin/cmd.go
@@ -31,10 +31,6 @@ import (
 	rprtr "github.com/openshift/rosa/pkg/reporter"
 )
 
-const (
-	idpName = "htpasswd-1"
-)
-
 var Cmd = &cobra.Command{
 	Use:   "admin",
 	Short: "Creates an admin user to login to the cluster",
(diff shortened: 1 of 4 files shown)


zvikorn:
@pvasant @zgalor PTAL

pvasant:
@zvikorn it should be cluster admin not cluster admins

cmd/create/admin/cmd.go
```diff
@@ -186,6 +182,8 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	reporter.Infof("Admin account has been added to cluster '%s'.", clusterKey)
+	reporter.Infof("An Identity Provider named 'Cluster-Admins' from type HTPasswdIdentityProvider, " +
```
pvasant, line 185: I would change it to I: An HTPasswd Identity Provider was created with the name 'Cluster-Admin' cc: @arendej

pvasant:
Also there is another issue we need to handle. 

during rosa create admin flow , If the htpasswd idp fails we just create the admin and not the htpasswd so that causes SDA 6106 as well. I think the proper fix is to move the logic to cluser service. rosa would call create admin in CS and CS will handle checking for idp and add it if it is not available.

arendej:
We need to take into account the issue in SDA-6170 as well, please

zvikorn:
@pvasant Is it a final decision to move the entire logic to CS and remove unnecessary code from ROSA?

pvasant:
> @pvasant Is it a final decision to move the entire logic to CS and remove unnecessary code from ROSA?

Yes we need to move the logic to CS  that way it is under same transaction and will be reverted if one of them fails 

`rosa would call create admin in CS and CS will handle checking for idp and add it if it is not available.`

zvikorn:
Closing this in favour of https://github.com/openshift/rosa/pull/769

PR #740, opened 2022-06-13
Initial implementation of runtime

tbrisker:
The runtime provides command run() methods with common tools such as
logger, reporter and various clients. This change will allow removing a
lot of boilerplate from the various commands and make refactoring them
simpler as extracting methods will require fewer parameters.

@vkareh @pvasant @oriAdler @zvikorn WDYT?
 I'd be happy for any suggestions on better naming, if we want this level of flexibility or if we should initialize all the clients for all commands anyways, and if you think this approach makes sense. If we agree on this direction i'll follow up with a cleanup of all the other commands (right now added just a couple to show the impact).
 
 [SDA-6207](https://issues.redhat.com//browse/SDA-6207)

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -40,11 +40,10 @@ import (
 	"github.com/openshift/rosa/pkg/aws"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/interactive/confirm"
-	"github.com/openshift/rosa/pkg/logging"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/output"
 	"github.com/openshift/rosa/pkg/properties"
-	rprtr "github.com/openshift/rosa/pkg/reporter"
+	"github.com/openshift/rosa/pkg/rosa"
 )
 
 //nolint
@@ -488,35 +487,19 @@ func networkTypeCompletion(cmd *cobra.Command, args []string, toComplete string)
 }
 
 func run(cmd *cobra.Command, _ []string) {
-	reporter := rprtr.CreateReporterOrExit()
-	logger := logging.NewLogger()
-	var err error
+	r := rosa.NewRuntime().WithOCM()
+	defer r.Cleanup()
 
-	// Create the client for the OCM API:
-	ocmClient, err := ocm.NewClient().
-		Logger(logger).
-		Build()
-	if err != nil {
-		reporter.Errorf("Failed to create OCM connection: %v", err)
-		os.Exit(1)
-	}
-	defer func() {
-		err = ocmClient.Close()
-		if err != nil {
-			reporter.Errorf("Failed to close OCM connection: %v", err)
-		}
-	}()
-
-	awsClient := aws.GetAWSClientForUserRegion(reporter, logger)
+	awsClient := aws.GetAWSClientForUserRegion(r.Reporter, r.Logger)
 
 	awsCreator, err := awsClient.GetCreator()
 	if err != nil {
-		reporter.Errorf("Unable to get IAM credentials: %v", err)
+		r.Reporter.Errorf("Unable to get IAM credentials: %v", err)
 		os.Exit(1)
 	}
 
 	if interactive.Enabled() {
-		reporter.Infof("Interactive mode enabled.\n" +
+		r.Reporter.Infof("Interactive mode enabled.\n" +
 			"Any optional fields can be left empty and a default will be selected.")
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/rosa/runtime.go
```diff
@@ -0,0 +1,71 @@
...
+		r = r.WithAWSClient()
+	}
+	var err error
+	r.Creator, err = r.AWSClient.GetCreator()
+	if err != nil {
+		r.Reporter.Errorf("Failed to get AWS creator: %v", err)
+		os.Exit(1)
+	}
+	return r
+}
```
vkareh, line 58: I would just replace these two with a single `WithAWS`. Getting the AWS Creator is a very low cost call, and removes the overhead of initializing yet another thing:
```
r := rosa.NewRuntime().WithOCM().WithAWS()
```
Is very easy to remember

vkareh, line 58: I would just replace these two with a single `WithAWS`. Getting the AWS Creator is a very low cost call, and removes the overhead of initializing yet another thing:
```
r := rosa.NewRuntime().WithOCM().WithAWS()
```
Is very easy to remember

pkg/rosa/runtime.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Adds all clients to the runtime. Requires a deferred call to `.Cleanup()` to close connections.
+func (r *Runtime) WithEverything() *Runtime {
+	return r.WithOCM().WithAWSClient().WithAWSCreator()
+}
```
vkareh, line 63: I would remove this. I think it risks folks using it even if unnecessary, especially if we add more components to the runtime.

Calling `rosa.NewRuntime().WithOCM().WithAWS()` is easy enough and explicit.

pkg/rosa/runtime.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (r *Runtime) Cleanup() {
+	if r.OCMClient != nil {
+		if err := r.OCMClient.Close(); err != nil {
+			r.Reporter.Errorf("Failed to close OCM connection: %v", err)
+		}
```
vkareh, line 69: Can this be done inside `WithOCM`?
```

func (r *Runtime) WithOCM() *Runtime {
	if r.OCMClient == nil {
		r.OCMClient = ocm.CreateNewClientOrExit(r.Logger, r.Reporter)
	}
	defer r.OCMClient.Close()
	return r
}
```

vkareh:
@tbrisker I like the `runtime` idea. I'm wondering if we can do lazy components though. For example:
```
r := rosa.NewRuntime()
cluster, err := r.OCM().GetCluster()
```
then inside the module
```
func (r *Runtime) OCM() *ocm.Client {
  if r.ocm == nil {
    r.ocm = ocm.CreateNewClientOrExit(r.Logger, r.Reporter)
    defer r.ocm.Close()
  }
  return r.ocm
}
```
Thoughts?

pkg/rosa/runtime.go
tbrisker: unfortunately not, as the defered method call will run when this function returns, not the caller

pkg/rosa/runtime.go
tbrisker: I wasn't sure, do we always have a creator when we initialize the aws client? it seems not all commands call both.

tbrisker:
@vkareh lazy init could work for aws but not for ocm client which requires the defered cleanup. Unfortunately go doesn't have a "destructor" concept. Is the initialization of the clients expensive enough that it's worth making it lazy?

pkg/rosa/runtime.go
vkareh: We don't always have/need a creator, but it's a very low cost call and it validates that the AWS client works.

vkareh:
Then let's avoid the lazy init and back to what you had originally :)

tbrisker:
rebased on top of https://github.com/openshift/rosa/pull/741 and fixed issues, other PR can go in first or this one can be merged and the other one closed.

tzvatot:
LGTM

vkareh:
@tbrisker you have a separate PR for the `Simplify logger package` commit - do you plan on removing it from this PR?

cmd/create/cluster/cmd.go
```diff
@@ -488,35 +487,19 @@ func networkTypeCompletion(cmd *cobra.Command, args []string, toComplete string)
...
-	}
-	defer func() {
-		err = ocmClient.Close()
-		if err != nil {
-			reporter.Errorf("Failed to close OCM connection: %v", err)
-		}
-	}()
-
-	awsClient := aws.GetAWSClientForUserRegion(reporter, logger)
+	awsClient := aws.GetAWSClientForUserRegion(r.Reporter, r.Logger)
```
tbrisker, line 493: didn't add this yet to the runtime, we can consider adding it as well.

tbrisker:
Added a few more commands and split between the runtime and the command migration. Would be good if we can get this in after 1.2.3 is released and then we can open separate PRs for migrating various commands instead of having one gigantic PR that migrates all of them and is impossible to review.

PR #743, opened 2022-06-15
Prompt the user to select multi or single AZ only in an interactive mood

oriAdler:
Bug example before changes:
 **rosa create machinepool -c oadler-test --instance-type m5.2xlarge --name workload --labels node-role.kubernetes.io/workload= --taints role=workload:NoSchedule --replicas 3 --yes
? Create multi-AZ machine pool: [? for help] (Y/n) 
E: Expected a valid value for create multi-AZ machine pool**


oriAdler:

PR #744, opened 2022-06-15
Fix bug - remove replicas constraint when editing single AZ machine pool

oriAdler:
Related: [SDA-6229](https://issues.redhat.com/browse/SDA-6229)


oriAdler:

PR #747, opened 2022-06-16
PolicyDocument refactoring

tbrisker:
Added a few more methods to PolicyDocument and refactored some of the code using them accordingly.

@vkareh PTAL


pkg/aws/policy_document.go
```diff
@@ -121,6 +121,81 @@ func (p *PolicyDocument) IsActionAllowed(wanted string) bool {
...
+	}
+
+	if len(failedActions) > 0 {
+		return false, fmt.Errorf("Actions not allowed with tested credentials: %v", failedActions)
+	}
+
+	return true, nil
+}
+
+func (p PolicyDocument) String() string {
```
tbrisker, line 191: This is not a pointer receiver on purpose - if you pass a `PolicyDocument` or `*PolicyDocument` to any formatted printer, this will print it properly while if this was a pointer receiver it would only print properly when passed a pointer.

pkg/aws/permissions.go
```diff
@@ -80,93 +77,3 @@ func (c *awsClient) ValidateSCP(target *string, policies map[string]string) (boo
...
-			input.ContextEntries = append(input.ContextEntries, &iam.ContextEntry{
-				ContextKeyName:   aws.String("aws:RequestedRegion"),
-				ContextKeyType:   aws.String("stringList"),
-				ContextKeyValues: []*string{aws.String(params.Region)},
-			})
-		}
-	}
-
-	// Either all actions are allowed and we'll return 'true', or it's a failure
-	allClear := true
```
tbrisker, line 116: No need for this flag, we can just check the length of the `failedActions` instead.

PR #750, opened 2022-06-20
Sts4

tbrisker:


vkareh:
LGTM

@tbrisker can you fix the conflict and I'll stamp it for merge?

cmd/install/addon/cmd.go
```diff
@@ -80,20 +84,65 @@ func run(cmd *cobra.Command, argv []string) {
...
+					r.Reporter.Errorf("%s", err)
+					os.Exit(1)
+				}
+			}
+			// TODO : verify the role has the right permissions
+
+			operatorRole, err := cmv1.NewOperatorIAMRole().
+				Name(cr.Name()).
+				Namespace(cr.Namespace()).
+				RoleARN(roleName).
```
vkareh, line 130: Shouldn't this be `roleArn`?

tbrisker: good catch, updated.

PR #751, opened 2022-06-20
Allow to select availability zones when creating a non-BYOVPC cluster

oriAdler:
Currently, CS is choosing the availability zone arbitrarily for the user,
this PR adds the option to select the availability zone.

1. Add `availability-zones` flag.
2. Add a step for the interactive mode to choose an availability zone.

Related: [SDA-6222](https://issues.redhat.com/browse/SDA-6222)

Example 1:
image
Example 2:
image

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -38,6 +38,7 @@ import (
 	installLogs "github.com/openshift/rosa/cmd/logs/install"
 	"github.com/openshift/rosa/pkg/arguments"
 	"github.com/openshift/rosa/pkg/aws"
+	"github.com/openshift/rosa/pkg/helper"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/interactive/confirm"
 	"github.com/openshift/rosa/pkg/ocm"
@@ -99,6 +100,9 @@ var args struct {
 	// SubnetIDs should come in pairs; two per availability zone, one private and one public,
 	// unless using PrivateLink, in which case it should only be one private per availability zone
 	subnetIDs []string
+
+	// Selecting availability zones for a non-BYOVPC cluster
+	availabilityZones []string
 
 	// Force STS mode for interactive and validation
 	sts bool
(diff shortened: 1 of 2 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -515,6 +527,14 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	isSubnetIDsSet := cmd.Flags().Changed("subnet-ids")
+	isAvailabilityZonesSet := cmd.Flags().Changed("availability-zones")
+	// Setting subnet IDs is choosing BYOVPC implicitly,
+	// and selecting availability zones is only allowed for non-BYOVPC clusters
+	if isSubnetIDsSet && isAvailabilityZonesSet {
```
oriAdler, line 534: I wasn't sure about this constraint. 
In CS we require both **subnets** and **availability zones** for BYOVPC cluster.

Here, if providing both, the alternative will be to validate that the subnets belonging to the availability zones provided.

If we won't allow providing both, we can prevent some user errors, as ROSA CLI is figuring out the availability zones for the subnets provided https://github.com/openshift/rosa/blob/master/cmd/create/cluster/cmd.go#L1277.

oriAdler:

zgalor:
LGTM

cmd/create/cluster/cmd.go
pvasant: is this a question ?

cmd/create/cluster/cmd.go
```diff
@@ -515,6 +527,14 @@ func run(cmd *cobra.Command, _ []string) {
...
+		reporter.Errorf("Setting both the `subnet-ids` flag and the `availability-zones` flag is prohibited")
```
pvasant, line 535: Setting both the `subnet-ids` flag and the `availability-zones` flag is not supported

cmd/create/cluster/cmd.go
pvasant, line 534: if isAvailabilityZonesSet  && isByoVPC {
"setting az is not supported for byo vpc. ROSA autodetects from subnet ids provided. 
}

oriAdler: Changed.

cmd/create/cluster/cmd.go
```diff
@@ -1175,7 +1196,8 @@ func run(cmd *cobra.Command, _ []string) {
 	subnetIDs := args.subnetIDs
 	subnetsProvided := len(subnetIDs) > 0
 	r.Reporter.Debugf("Received the following subnetIDs: %v", args.subnetIDs)
-	if !useExistingVPC && !subnetsProvided && interactive.Enabled() {
+	// If the user has set the availability zones (allowed for non-BYOVPC clusters), don't prompt the BYOVPC message
+	if !useExistingVPC && !subnetsProvided && !isAvailabilityZonesSet && interactive.Enabled() {
```
oriAdler, line 1200: @zgalor @pvasant I've added ` && !isAvailabilityZonesSet` to the **if** statement.

PR #752, opened 2022-06-21
The additional-trust-bundle-file can't be set via interactive mode if the cluster is not set proxy fields

zvikorn:

[SDA-6111](https://issues.redhat.com//browse/SDA-6111)


zvikorn:
@pvasant @zgalor please approve and review

PR #758, opened 2022-06-27
Drop unused CheckStackReadyForCreateCluster method

tbrisker:


tbrisker:
@vkareh can you PTAL? looks like you created this method

PR #765, opened 2022-06-29
Select a single subnet for a single AZ machine pool - BYOVPC clusters

oriAdler:
1. Add the `subnet` flag to the `create machine pool` command.
2. Add functions to fetch VPC's private subnets.

Possible flows:
1. BYOVPC single AZ cluster - the user is prompted with the option to select subnet.
2. BYOVPC multi-AZ cluster - if the user chooses a single AZ, they have the option
to provide a subnet or an availability zone.

Related: [SDA-6289](https://issues.redhat.com/browse/SDA-6289)

[select_subnet_for_single_az_machine_pool.webm](https://user-images.githubusercontent.com/57869309/176680592-b632f2f6-8fdb-44de-8d89-22a115eef6f6.webm)

PR #767, opened 2022-06-30
Refactor function for more general use

oriAdler:
1. Modify `GetSubnetsIDs` to recieve `DescribeSubnetsInput` for re-use.
2. Move `setSubnetOption` and `ParseSubnet` to `pkg/aws/helpers.go` for re-use.

Related: https://github.com/openshift/rosa/pull/765

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -1249,10 +1249,10 @@ func run(cmd *cobra.Command, _ []string) {
 			availabilityZone := awssdk.StringValue(subnet.AvailabilityZone)
 
 			// Create the options to prompt the user.
-			options[i] = setSubnetOption(subnetID, availabilityZone)
+			options[i] = aws.SetSubnetOption(subnetID, availabilityZone)
 			if subnetsProvided {
 				for _, subnetArg := range subnetIDs {
-					defaultOptions = append(defaultOptions, setSubnetOption(subnetArg, availabilityZone))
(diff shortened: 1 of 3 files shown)


oriAdler:
@zgalor @tbrisker PTAL

cmd/create/cluster/cmd.go
```diff
@@ -1217,7 +1218,7 @@ func run(cmd *cobra.Command, _ []string) {
 
 	var availabilityZones []string
 	if useExistingVPC || subnetsProvided {
-		subnets, err := awsClient.GetSubnetIDs()
+		subnets, err := awsClient.GetSubnetIDs(&ec2.DescribeSubnetsInput{})
```
zgalor, line 1221: 1. Where do you pass non-empty input?
2. I don't think it's nice to create aws specific object outside of the aws client and pass it to an aws client func

zgalor: If https://github.com/openshift/rosa/pull/765/files#diff-39260fbf46ea076702be0085fc20b3897395b779af4dbaee85902ce0b95d725dR363 is the only place you need the filter, which is internal to aws client I suggest the following.

Make `getSubnetIDs(&ec2.DescribeSubnetsInput{})` private and keep `GetSubnetIDs()` without the parameter
Then, from both `GetSubnetIDs()` and from `getVPCSubnets` call `getSubnetIDs`

oriAdler: Makes sense, changed, thanks a lot for this solution.

PR #768, opened 2022-06-30
Fix bug - set cluster key

oriAdler:
Set the cluster key global variable instead of the local variable.

Related: [SDA-6322](https://issues.redhat.com/browse/SDA-6322)


cmd/logs/install/cmd.go
```diff
@@ -82,11 +82,10 @@ func run(cmd *cobra.Command, argv []string) {
 
 	// Allow the command to be called programmatically
 	if len(argv) == 1 && !cmd.Flag("cluster").Changed {
-		clusterKey = argv[0]
+		ocm.SetClusterKey(argv[0])
 		watch = true
-	} else {
-		clusterKey = r.GetClusterKey()
 	}
+	clusterKey = r.GetClusterKey()
```
tbrisker, line 88: we can get rid of the `var clusterKey string` above here (and in the other file) as well

oriAdler: Removed.

PR #770, opened 2022-07-03
Fix error message - create non-BYOVPC - select availability zone

oriAdler:
Mention region name in the error message when the availability zone
doesn't belong to the region's availability zones.

Related: [SDA-6327](https://issues.redhat.com/browse/SDA-6327)


oriAdler:
@zgalor @zvikorn PTAL

zvikorn:
LGTM

PR #772, opened 2022-07-04
Fix bug - create a cluster with an invalid number of subnets

oriAdler:
Add subnets number validation to improve user experience.

Related: [SDA-6331](https://issues.redhat.com/browse/SDA-6331)

image

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -1280,6 +1280,12 @@ func run(cmd *cobra.Command, _ []string) {
 			}
 		}
 
+		err = validateSubnetsCount(multiAZ, subnetIDs)
+		if err != nil {
+			r.Reporter.Errorf("%s", err)
+			os.Exit(1)
+		}
+
 		for _, subnet := range subnetIDs {
 			az := mapSubnetToAZ[subnet]
 			if !mapAZCreated[az] {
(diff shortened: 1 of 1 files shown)


oriAdler:
@zgalor @ciaranRoche PTAL

oriAdler:
> @oriAdler should this kind of validation be handled in the API?
> 
> My argument here is this error is not specific to ROSA, and if a user just does a POST request with an invalid amount of subnets they will run into the same issue.
> 
> Also I think we should keep clients like ROSA as dumb as possible and leave validations and business logic to be decided as much as we can in the API itself.

I was thinking the same and wanted to implement it in CS in the beginning.
But we already have similar checks as part of the preflight in CS 
https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/cmd/clusters-service/service/aws/preflight_service.go#L376.

This bug is that **rosa** derives the availability zones from the subnets, and then we get this error from the handlers layer in CS:
_Failed to create cluster: Only a single availability zone can be provided to a single-availability-zone cluster, instead received 3_
https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/cmd/clusters-service/servecmd/apiserver/validation_helpers.go#L2102

I agree that we should keep ROSA CLI as dumb as possible, but the disadvantage is that the user will get the feedback only after sending the spec to CS, and this check is quite a basic one.

Alternatively, I can move this validation to the handlers layer in CS, let me know what you think is best.

vkareh:
@oriAdler Can this function be added as part of the input validators? This way it gets validated interactively as well, rather than killing the user flow.
```
			subnetIDs, err = interactive.GetMultipleOptions(interactive.Input{
				Question: "Subnet IDs",
				Help:     cmd.Flags().Lookup("subnet-ids").Usage,
				Required: false,
				Options:  options,
				Default:  defaultOptions,
----------------------------->  Validators: ...
			})
```

oriAdler:
> @oriAdler Can this function be added as part of the input validators? This way it gets validated interactively as well, rather than killing the user flow.
> 
> ```
> 			subnetIDs, err = interactive.GetMultipleOptions(interactive.Input{
> 				Question: "Subnet IDs",
> 				Help:     cmd.Flags().Lookup("subnet-ids").Usage,
> 				Required: false,
> 				Options:  options,
> 				Default:  defaultOptions,
> ----------------------------->  Validators: ...
> 			})
> ```

Sure, that makes sense, let me open a new PR to add it.

PR #778, opened 2022-07-05
Redact token output

vkareh:


yasun1:
@vkareh  I saw https://github.com/openshift/rosa/pull/779 is passed, maybe retest #778 would pass too.

hunterkepley (commented):
/lgtm

PR #780, opened 2022-07-06
Set `clusterKey` properly to support `r.FetchCluster`

oriAdler:
Use the `SetClusterKey` to assign the `cluster key` to the global variable.
In order to use it in the `r.FetchCluster` function.

Related: [SDA-6362](https://issues.redhat.com/browse/SDA-6362)

Follows up on https://github.com/openshift/rosa/pull/768

image


oriAdler:
@ciaranRoche @zgalor PTAL

PR #787, opened 2022-07-13
Refactored ROSA to create operator policies when running cluster creation

samira-barouti:
This MR moves the logic of operator policy creation to the cluster creation.
This is the pre-req for onboarding hypershift into ROSA: hypershift has different
set of policies. Therefore, the policy creation requires the cluster type, and we will have the
hosted-cp flag when creating a cluster.

### DIFF
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -22,7 +22,6 @@ import (
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws/arn"
-	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 
 	"github.com/openshift/rosa/cmd/login"
@@ -216,18 +215,13 @@ func run(cmd *cobra.Command, argv []string) {
 		r.Reporter.Errorf("Expected a valid role creation mode: %s", err)
 		os.Exit(1)
 	}
-	credRequests, err := r.OCMClient.GetCredRequests()
-	if err != nil {
-		r.Reporter.Errorf("Error getting operator credential request from OCM %s", err)
-		os.Exit(1)
-	}
 
 	switch mode {
 	case aws.ModeAuto:
 		r.Reporter.Infof("Creating roles using '%s'", r.Creator.ARN)
 
-		err = createRoles(r, prefix, permissionsBoundary, r.Creator.AccountID, env, policies,
-			defaultPolicyVersion, credRequests)
+		err = createRoles(r, prefix, permissionsBoundary, r.Creator.AccountID, env,
+			policies, defaultPolicyVersion)
 		if err != nil {
 			r.Reporter.Errorf("There was an error creating the account roles: %s", err)
 			if strings.Contains(err.Error(), "Throttling") {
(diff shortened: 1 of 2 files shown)


samira-barouti:
@vkareh @pvasant @oriAdler PTAL

cmd/create/accountroles/cmd.go
```diff
@@ -250,19 +244,7 @@ func run(cmd *cobra.Command, argv []string) {
 			ocm.Version:  defaultPolicyVersion,
 		})
 	case aws.ModeManual:
-		err = aws.GeneratePolicyFiles(r.Reporter, env, true, true, policies, credRequests)
```
vkareh, line 253: This function is still necessary to generate the account role policies (note the first `true` sets `generateAccountRolePolicies` in the called function)

pvasant:
lgtm will stamp once pipeline passes

cmd/create/accountroles/cmd.go
samira-barouti: How about moving this logic to operatorroles?

pvasant: ho yeah, great catch . we need it for manual mode policy creation for account role policies

samira-barouti: @vkareh I added this method back with some modifications. Please let me know how it looks like.

cmd/create/operatorroles/cmd.go
```diff
@@ -228,6 +234,18 @@ func run(cmd *cobra.Command, argv []string) {
...
+		if err != nil {
+			r.Reporter.Errorf("There was an error generating the policy files: %s", err)
+			r.OCMClient.LogEvent("ROSACreateOperatorRolesModeManual", map[string]string{
+				ocm.Response: ocm.Failure,
+			})
+			os.Exit(1)
+		}
+		if r.Reporter.IsTerminal() {
+			r.Reporter.Infof("All policy files saved to the current directory")
+			r.Reporter.Infof("Run the following commands to create the operator roles and policies:\n")
```
vkareh, line 247: This message is already shown below, maybe update that one instead (and remove this one, since it displays before the command is built)

cmd/create/operatorroles/cmd.go
```diff
@@ -228,6 +234,18 @@ func run(cmd *cobra.Command, argv []string) {
...
+		}
 		commands, err := buildCommands(r, prefix, permissionsBoundary, cluster, r.Creator.AccountID,
```
vkareh, line 249: The `buildCommands` here might need to be modified, as it assumes that policies already exist. You'll need to add commands to create those policies in the first place.

cmd/create/accountroles/cmd.go
```diff
@@ -307,30 +304,11 @@ func buildCommands(prefix string, permissionsBoundary string, accountID string,
...
-			"operator_namespace", operator.Namespace(),
-			"operator_name", operator.Name(),
-		)
-		createPolicy := fmt.Sprintf("aws iam create-policy \\\n"+
-			"\t--policy-name %s \\\n"+
-			"\t--policy-document file://openshift_%s_policy.json \\\n"+
-			"\t--tags %s",
-			name, credrequest, iamTags)
-		commands = append(commands, createPolicy)
-	}
```
vkareh, line 326: This block might need to be added to the `buildCommands` of the `create operator-roles` command.

vkareh: Ah, I see that this already exists there. Nevemrind.

cmd/create/accountroles/cmd.go
vkareh: Can you make it take a `nil` value instead? Allocating memory for an unused map doesn't seem right.

cmd/create/operatorroles/cmd.go
samira-barouti: Done.

cmd/create/accountroles/cmd.go
samira-barouti: Done.

samira-barouti (commented):
Will be fixed in the next patch.

vkareh:
LGTM

samira-barouti:
@vkareh I tested the account role creation as well as the cluster creation (in both auto and manual mode). All went well. Could you please merge this PR?

PR #795, opened 2022-07-21
fedramp: Update rosa-authenticator configuration

vkareh:


vkareh:
@mjlshen @dustman9000 @tonytheleg @markturansky - please take a look to make sure this works on stage/prod.

PR #799, opened 2022-07-25
Returning user-friendly error messages for managed services

renan-campos:
[SDA-6084](https://issues.redhat.com/browse/SDA-6084)

### DIFF
--- a/pkg/output/errors.go
+++ b/pkg/output/errors.go
@@ -1,0 +1,16 @@
+package output
+
+import (
+	sdkErr "github.com/openshift-online/ocm-sdk-go/errors"
+	"github.com/pkg/errors"
+)
+
+// Converts OCM errors to user-friendly strings
+func ErrorToString(err error) string {
+	switch err := errors.Cause(err).(type) {
+	case *sdkErr.Error:
+		return err.Reason()
+	default:
+		return "Unknown error"
+	}
+}
--- a/cmd/describe/service/cmd.go
+++ b/cmd/describe/service/cmd.go
@@ -23,6 +23,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/output"
 	"github.com/openshift/rosa/pkg/rosa"
 )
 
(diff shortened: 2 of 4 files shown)


renan-campos:
@ciaranRoche ptal, I'm probably missing some error messages

pkg/output/errors.go
```diff
@@ -0,0 +1,23 @@
...
+	sdkErr "github.com/openshift-online/ocm-sdk-go/errors"
+	"github.com/pkg/errors"
+)
+
+// Converts OCM errors to user-friendly strings
+func ErrorToString(err error) string {
+	switch err := errors.Cause(err).(type) {
+	case *sdkErr.Error:
+		switch err.Code() {
+		case "OCM-MS-7":
```
pvasant, line 13: i would refrain from doing it in cli if possible

renan-campos: Could you elaborate on the reason?

pvasant: i dont have a strong opinion but why is the SDK not returning the error message directly?

renan-campos: I think that processing the error within the cli allows us to return a more user-friendly error message back to the user instead of having to hope that the error the backend returns will print nicely. This work came about because the original message had redundant information: `E: Failed to get service with id "29elp8qrDer4dfSDW6NzibI7shG": Failed to get managed service with id 29elp8qrDer4dfSDW6NzibI7shG:: status is 500, identifier is '9', code is 'OCM-MS-9' and operation identifier is '29eo4veafWYIB6J6hNf4tflGUHq': Unable to get managed service`.

pvasant: Why is the backend returning a redundant message? The only issue is if we have this logic in cli, we cannot change it if needed immediately. I don't want to block this PR from merging but remember to not have business logic in cli as much as possible

pvasant:
@ciaranRoche PTAL and merge it

pkg/output/errors.go
renan-campos: I just thought of something: We could print the error's Reason(). That way we aren't processing the error in the cli. If the reason isn't clear we can enhance what the backend returns!

renan-campos: The error is clearer that way: `E: Failed to get service with id "foobar": Unable to get managed service: record not found`. Let me know what you think

pkg/output/errors.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	sdkErr "github.com/openshift-online/ocm-sdk-go/errors"
+	"github.com/pkg/errors"
+)
+
+// Converts OCM errors to user-friendly strings
+func ErrorToString(err error) string {
+	switch err := errors.Cause(err).(type) {
+	case *sdkErr.Error:
+		return err.Reason()
```
renan-campos, line 12: @pvasant

PR #800, opened 2022-07-27
Create cluster - for single AZ, only allow to select one AZ

oriAdler:
Use `interactive.GetOption` for a single AZ cluster.

Related: [SDA-6445](https://issues.redhat.com/browse/SDA-6445)

Bug:
image

Fix:
image


oriAdler:
@zgalor @pvasant PTAL

PR #806, opened 2022-08-09
support for path in iam roles and policies

ydayagi:
This is for account roles (operator, installer and such)
ocm role and user role will be handled in a separate PR
[SDA-6246](https://issues.redhat.com//browse/SDA-6246)

https://issues.redhat.com/browse/SDA-6246

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -118,6 +118,7 @@ var args struct {
 	// Operator IAM Roles
 	operatorIAMRoles                 []string
 	operatorRolesPrefix              string
+	operatorRolesPath                string
 	operatorRolesPermissionsBoundary string
 
 	// Proxy
@@ -225,6 +226,12 @@ func init() {
 		"",
 		"Prefix to use for all IAM roles used by the operators needed in the OpenShift installer. "+
 			"Leave empty to use an auto-generated one.",
+	)
+	flags.StringVar(
+		&args.operatorRolesPath,
+		"operator-roles-path",
+		"",
+		"Custom arn path for operator roles.",
 	)
 
 	flags.StringSliceVar(
@@ -961,6 +968,22 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
+	operatorRolePath := args.operatorRolesPath
+	if interactive.Enabled() {
+		operatorRolePath, err = interactive.GetString(interactive.Input{
+			Question: "Operator Role Path",
+			Help:     cmd.Flags().Lookup("operator-roles-path").Usage,
+			Default:  operatorRolePath,
+			Validators: []interactive.Validator{
+				aws.ARNPathValidator,
+			},
+		})
+		if err != nil {
+			r.Reporter.Errorf("Expected a valid path: %s", err)
+			os.Exit(1)
+		}
+	}
+
 	operatorIAMRoles := args.operatorIAMRoles
 	operatorIAMRoleList := []ocm.OperatorIAMRole{}
 	if isSTS {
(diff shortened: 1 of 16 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -1013,8 +1036,16 @@ func run(cmd *cobra.Command, _ []string) {
...
 		for _, role := range operatorIAMRoleList {
-			name := strings.SplitN(role.RoleARN, "/", 2)[1]
-			err := awsClient.ValidateRoleNameAvailable(name)
+			str, err := arn.Parse(role.RoleARN)
+			if err != nil {
+				r.Reporter.Errorf("Error validating role arn: %v", err)
+				os.Exit(1)
+			}
+			resource := str.Resource
+			m := strings.LastIndex(resource, "/")
```
oriAdler, line 1045: I would recommend extracting this logic to a common util function (like https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/pkg/aws/stshelper.go#L221) in one of the `helper` files.

Ideally, that function would be in https://gitlab.cee.redhat.com/service/ocm-common/, so both CS and ROSA CLI use the very same one, but I assume that can be added to `ocm-common` later.

cmd/upgrade/operatorroles/cmd.go
```diff
@@ -389,8 +422,9 @@ func buildMissingOperatorRoleCommand(missingRoles map[string]*cmv1.STSOperator,
 			"\t--role-name %s \\\n"+
 			"\t--assume-role-policy-document file://%s \\\n"+
 			"%s"+
-			"\t--tags %s",
-			roleName, filename, permBoundaryFlag, iamTags)
+			"\t--tags %s \\\n"+
+			"\t--path %s",
```
oriAdler, line 426: I think we might want to concatenate the `--path` only if `path!=""`, to avoid the case where the user doesn't provide a path (optional) so we won't print `rosa upgrade operator-roles ... --path`

Though from AWS's perspective providing an empty path or no path at all will be the same, WDYT?

pkg/aws/policies.go
```diff
@@ -233,13 +236,14 @@ func (c *awsClient) PutRolePolicy(roleName string, policyName string, policy str
...
 func (c *awsClient) EnsurePolicy(policyArn string, document string,
-	version string, tagList map[string]string) (string, error) {
+	version string, tagList map[string]string, path string) (string, error) {
 	output, err := c.IsPolicyExists(policyArn)
 	if err != nil {
 		if aerr, ok := err.(awserr.Error); ok {
 			switch aerr.Code() {
 			case iam.ErrCodeNoSuchEntityException:
-				return c.createPolicy(policyArn, document, tagList)
+				fmt.Printf("%s", "i mhererrrrrrrrr")
```
oriAdler, line 245: Probably we can remove this line.

pkg/ocm/machines.go
```diff
@@ -86,12 +86,14 @@ func (c *Client) GetAvailableMachineTypes() (MachineTypeList, error) {
 	if err != nil {
 		return nil, err
 	}
+	/*
```
oriAdler, line 89: We can probably remove this comment.

ydayagi: we have to :-) lint complains

cmd/create/cluster/cmd.go
ydayagi: do we have access to https://gitlab.cee.redhat.com/service/ocm-common/?

oriAdler: I've invited you to the repo.

ydayagi: What I meant was: does the github repo have access to redhat's gitlab? cause I can't access it without VPN.

cmd/upgrade/operatorroles/cmd.go
ydayagi: I wrapped with single quotes so there will always be a value for the path - at least empty

pkg/aws/policies.go
ydayagi: done

cmd/create/cluster/cmd.go
oriAdler: Ok, I understand, now that you mention it I'm not sure we import `ocm-common` in this project.

oriAdler:
LGTM

ydayagi:
> LGTM

bot works with "/lgtm" and "/approve"

PR #808, opened 2022-08-10
Refactor `GetCluster` function

oriAdler:
Related: [SDA-6551](https://issues.redhat.com/browse/SDA-6551)

### DIFF
--- a/vendor/modules.txt
+++ b/vendor/modules.txt
@@ -98,6 +98,7 @@ github.com/golang/protobuf/ptypes/any
 github.com/golang/protobuf/ptypes/duration
 github.com/golang/protobuf/ptypes/timestamp
 # github.com/google/uuid v1.3.0
+## explicit
 github.com/google/uuid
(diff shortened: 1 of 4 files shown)


vendor/modules.txt
```diff
@@ -98,6 +98,7 @@ github.com/golang/protobuf/ptypes/any
 github.com/golang/protobuf/ptypes/duration
 github.com/golang/protobuf/ptypes/timestamp
 # github.com/google/uuid v1.3.0
+## explicit
```
tzvatot, line 101: What do you mean by this comment?

oriAdler: TBH not sure, it was added automatically, let me check.

oriAdler: This is a related thread https://github.com/golang/go/issues/37946#issuecomment-601210447
I didn't understand the `explicit` comment completely, but I can see we already have a bunch of these comments in this file.

PR #810, opened 2022-08-11
[SDA-6439] Add red-hat-managed tag to roles and policies

hunterkepley:
[SDA-6439](https://issues.redhat.com/browse/SDA-6439)

### DIFF
--- a/pkg/aws/tags/tags.go
+++ b/pkg/aws/tags/tags.go
@@ -46,3 +46,6 @@ const Environment = prefix + "environment"
 
 // AdminRole tags the role as admin (true/false)
 const AdminRole = prefix + "admin_role"
+
+// RedHatManaged tags the role as red_hat_managed
+const RedHatManaged = "red-hat-managed"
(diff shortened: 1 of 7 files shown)


pkg/aws/tags/tags.go
```diff
@@ -46,3 +46,6 @@ const Environment = prefix + "environment"
 
 // AdminRole tags the role as admin (true/false)
 const AdminRole = prefix + "admin_role"
+
+// RedHatManaged tags the role as red_hat_managed
+const RedHatManaged = prefix + "red_hat_managed"
```
vkareh, line 51: This should be `red-hat-managed` (hyphens instead of underscores)

vkareh: Also, remove the `prefix`, since we want `red-hat-managed` to be the entire tag.

hunterkepley: @vkareh Made the changes you requested, please feel free to resolve this thread if it looks good!

vkareh:
This looks good, thanks!
Can you please squash/fixup these commits so that there's only a single atomic one?

hunterkepley:
> This looks good, thanks! Can you please squash/fixup these commits so that there's only a single atomic one?

Squashed them, should be one commit now!

hunterkepley:
> This looks good, thanks! Can you please squash/fixup these commits so that there's only a single atomic one?

@vkareh Squashed successfully, should be good to go now!

hunterkepley:
Finished testing right before the merge, everything seems to be working fine. STS roles are created with `red-hat-managed` set to True. Also checked policies

PR #811, opened 2022-08-15
properly handle "" no_proxy option in interactive mode

zvikorn:
…ractive mode

[SDA-6587](https://issues.redhat.com//browse/SDA-6587)

@zgalor @oriAdler @pvasant  @tbrisker @samira-barouti

### DIFF
--- a/cmd/edit/cluster/cmd.go
+++ b/cmd/edit/cluster/cmd.go
@@ -379,7 +379,7 @@ func run(cmd *cobra.Command, _ []string) {
 		noProxyInput, err := interactive.GetString(interactive.Input{
 			Question: "No proxy",
 			Help:     cmd.Flags().Lookup("no-proxy").Usage,
-			Default:  strings.Join(noProxySlice, ","),
+			Default:  cluster.Proxy().NoProxy(),
 			Validators: []interactive.Validator{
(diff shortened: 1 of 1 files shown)


cmd/edit/cluster/cmd.go
```diff
@@ -397,6 +397,10 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	if len(noProxySlice) > 0 {
+		if len(noProxySlice) == 1 && noProxySlice[0] == doubleQuotesToRemove {
+			noProxySlice[0] = ""
+		}
+
```
tbrisker, line 403: shouldn't we treat `""` the same as an empty slice?

tbrisker (commented):
nitpick - could you change the commit message to describe what changed (e.g. `properly handle "" no_proxy option in interactive mode`) rather than `fix - bug description`?

cmd/edit/cluster/cmd.go
zvikorn: no. we need to explicitly treat "" to remove the current value.

tbrisker: Ah, good point.

PR #814, opened 2022-08-17
compare arn path for existing policy/role

ydayagi:
EnsureRole will succeed if role name exists with different path
EnsurePolicy actually fails but returns success and wrong ARN
this change causes both functions to succeed only if the existing resource has same path as the requested resource

[SDA-6596](https://issues.redhat.com//browse/SDA-6596)
https://issues.redhat.com/browse/SDA-6596

### DIFF
--- a/pkg/aws/policies.go
+++ b/pkg/aws/policies.go
@@ -110,6 +110,12 @@ func (c *awsClient) EnsureRole(name string, policy string, permissionsBoundary s
 		}
 	}
 
+	// verify that the paths are the same
+	if aws.StringValue(output.Role.Path) != path {
+		return "", fmt.Errorf("Role with same name but different path exists. Existing role ARN: %s",
+			*output.Role.Arn)
+	}
+
 	if permissionsBoundary != "" {
 		_, err = c.iamClient.PutRolePermissionsBoundary(&iam.PutRolePermissionsBoundaryInput{
 			RoleName:            aws.String(name),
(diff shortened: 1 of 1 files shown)


pkg/aws/policies.go
```diff
@@ -241,7 +251,17 @@ func (c *awsClient) EnsurePolicy(policyArn string, document string,
 		if aerr, ok := err.(awserr.Error); ok {
 			switch aerr.Code() {
 			case iam.ErrCodeNoSuchEntityException:
-				return c.createPolicy(policyArn, document, tagList, path)
+				policyArn, err = c.createPolicy(policyArn, document, tagList, path)
+				if err != nil {
+					if aerr, ok := err.(awserr.Error); ok {
+						if aerr.Code() == iam.ErrCodeEntityAlreadyExistsException {
```
ciaranRoche, line 257: I am just wondering, how do we get to this exception? 
If we only trigger the create policy when no such entity exists, but yet we try handle entity already exists exception. Maybe I am missing something, so just looking for some clarification

pkg/aws/policies.go
```diff
@@ -110,6 +110,16 @@ func (c *awsClient) EnsureRole(name string, policy string, permissionsBoundary s
...
+	// verify that the paths are the same
+	existingRolePath := ""
+	if output.Role.Path != nil {
+		existingRolePath = *output.Role.Path
+	}
+	if existingRolePath != path {
+		return "", fmt.Errorf("Role with same name but different path exists. Existing role ARN: %s",
+			*output.Role.Arn)
+	}
+
```
ciaranRoche, line 122: I think we can shorten this to something like 
```
	if aws.StringValue(output.Role.Path) != path {
		return "", fmt.Errorf("role with same name but different path exists. Existing role ARN: %s",
			*output.Role.Arn)
	}
```

pkg/aws/policies.go
ydayagi: AWS allows you to retrieve a policy by providing a full ARN. We have the wrong path so we have a non-existing ARN. However, the policy name is the unique key. Creating a policy uses a different input method. It uses name and path. So 'create policy' will fail with 'already exists'. This is the indication that the ARN we have is wrong. Probably wrong path.

pkg/aws/policies.go
ydayagi: fixed

pkg/aws/policies.go
ciaranRoche: Damn that is annoying, that we can not search with the came criteria that we are creating with. Unless we fetched all policies and filtered them. Guess there is some overhead with that.

ciaranRoche:
Changes LGTM, will leave it to someone with more rosa context to add the label

pkg/aws/policies.go
ydayagi: indeed

ydayagi:
@samira-barouti @oriAdler @pvasant

PR #816, opened 2022-08-18
Run linter independet of locally installed linter

zgalor:
Copied from cluster-service makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,12 @@ fmt:
 
 .PHONY: lint
 lint:
-	golangci-lint run --timeout 5m0s
+	# Mirror of docker.io/golangci/golangci-lint, Docker Hub is blocked in ci-int.
+	podman run --rm --security-opt label=disable --volume="$(PWD):/app" --workdir=/app \
+	  -v "$(shell go env GOCACHE):/cache/go" -e "GOCACHE=/cache/go" -e "GOLANGCI_LINT_CACHE=/cache/go" \
+		-e "GIT_SSL_NO_VERIFY=1" -e "GOINSECURE=gitlab.cee.redhat.com"\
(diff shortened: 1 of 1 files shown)


zgalor:
@ciaranRoche @ydayagi @pvasant @oriAdler @tzvatot PTAL

Makefile
```diff
@@ -42,7 +42,12 @@ fmt:
 
 .PHONY: lint
 lint:
-	golangci-lint run --timeout 5m0s
+	# Mirror of docker.io/golangci/golangci-lint, Docker Hub is blocked in ci-int.
+	podman run --rm --security-opt label=disable --volume="$(PWD):/app" --workdir=/app \
+	  -v "$(shell go env GOCACHE):/cache/go" -e "GOCACHE=/cache/go" -e "GOLANGCI_LINT_CACHE=/cache/go" \
+		-e "GIT_SSL_NO_VERIFY=1" -e "GOINSECURE=gitlab.cee.redhat.com"\
+		"quay.io/app-sre/golangci-lint:v$(shell cat .golangciversion)" \
+		golangci-lint run -v
```
tzvatot, line 50: What about the timeout option?

zgalor: We are not using it in CS. Do we really need it?

tzvatot: I think it was added here for a reason. Might be related to github trigger issue, or might be that the rosa code is too slow to run linter on. Anyhow, I suggest that we preserve it.

ydayagi:
LGTM

tbrisker:
looks like prow doesn't have `podman` installed. Also, we don't include any modules from internal gitlab so that flag can be removed

tzvatot:
@zgalor can we close this?

PR #817, opened 2022-08-18
Allow setting billing model for addon installations

ciaranRoche:
JIRA - [SDA-6605](https://issues.redhat.com//browse/SDA-6605)

https://issues.redhat.com/browse/SDA-6605

Verification - https://gist.github.com/ciaranRoche/bdb19c81645e119dbd88a6fafea80612


ciaranRoche:
@tbrisker PTAL

cmd/install/addon/cmd.go
```diff
@@ -242,8 +268,51 @@ func run(cmd *cobra.Command, argv []string) {
...
 
+	billingModel := args.billingModel
+	billingModelAccountID := args.billingModelAccountID
+
+	if !cmd.Flags().Changed(billingModelFlag) && !interactive.Enabled() {
+		interactive.Enabled()
+		r.Reporter.Infof("Enabling interactive mode")
+	}
+
+	if interactive.Enabled() && !cmd.Flags().Changed("billing-model") {
```
renan-campos, line 279: Did you mean to leave "billing-model" hardcoded here instead of using billingModelFlag?

cmd/install/addon/cmd.go
```diff
@@ -242,8 +268,51 @@ func run(cmd *cobra.Command, argv []string) {
...
+			Default:  "standard",
+		})
+		if err != nil {
+			r.Reporter.Errorf("Expected a valid billing model: %s", err)
+			os.Exit(1)
+		}
+	}
+
+	if !cmd.Flags().Changed(billingModelAccountIDFlag) && !interactive.Enabled() {
+		interactive.Enabled()
```
renan-campos, line 292: I think this should be `interactive.Enable()`

cmd/install/addon/cmd.go
```diff
@@ -242,8 +268,51 @@ func run(cmd *cobra.Command, argv []string) {
 		})
 	}
 
+	billingModel := args.billingModel
+	billingModelAccountID := args.billingModelAccountID
+
+	if !cmd.Flags().Changed(billingModelFlag) && !interactive.Enabled() {
+		interactive.Enabled()
```
renan-campos, line 275: I think this should be `interactive.Enable()`

renan-campos (approved):
LGTM

cmd/install/addon/cmd.go
```diff
@@ -242,8 +268,51 @@ func run(cmd *cobra.Command, argv []string) {
...
+			os.Exit(1)
+		}
+	}
+
+	if !cmd.Flags().Changed(billingModelAccountIDFlag) && !interactive.Enabled() {
+		interactive.Enable()
+		r.Reporter.Infof("Enabling interactive mode")
+	}
+
+	if interactive.Enabled() && billingModel != string(v1.BillingModelStandard) &&
```
tbrisker, line 296: `interactive.Enabled()` will be `true` here if `!cmd.Flags().Changed(billingModelAccountIDFlag)`

cmd/install/addon/cmd.go
```diff
@@ -26,6 +26,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/iam"
+	v1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
```
tbrisker, line 29: should this be `amsv1`?

cmd/install/addon/cmd.go
```diff
@@ -242,8 +268,51 @@ func run(cmd *cobra.Command, argv []string) {
...
+	billingModel := args.billingModel
+	billingModelAccountID := args.billingModelAccountID
+
+	if !cmd.Flags().Changed(billingModelFlag) && !interactive.Enabled() {
+		interactive.Enable()
+		r.Reporter.Infof("Enabling interactive mode")
+	}
+
+	if interactive.Enabled() && !cmd.Flags().Changed(billingModelFlag) {
+		billingModel, err = interactive.GetString(interactive.Input{
```
tbrisker, line 280: could we use `Interactive.GetOption` here if we have a fixed set of supported options? Otherwise the user will have no idea what options are valid here.

cmd/install/addon/cmd.go
```diff
@@ -63,6 +74,21 @@ var Cmd = &cobra.Command{
 
 func init() {
 	flags := Cmd.Flags()
+
+	flags.StringVar(
+		&args.billingModel,
+		billingModelFlag,
+		"standard",
```
tbrisker, line 81: `amsv1.BillingModelStandard`?

cmd/install/addon/cmd.go
```diff
@@ -242,8 +268,51 @@ func run(cmd *cobra.Command, argv []string) {
...
+	if !cmd.Flags().Changed(billingModelFlag) && !interactive.Enabled() {
+		interactive.Enable()
+		r.Reporter.Infof("Enabling interactive mode")
+	}
+
+	if interactive.Enabled() && !cmd.Flags().Changed(billingModelFlag) {
+		billingModel, err = interactive.GetString(interactive.Input{
+			Question: "Billing Model",
+			Help:     cmd.Flags().Lookup(billingModelFlag).Usage,
+			Default:  "standard",
```
tbrisker, line 283: `amsv1.BillingModelStandard`?

cmd/install/addon/cmd.go
```diff
@@ -242,8 +268,51 @@ func run(cmd *cobra.Command, argv []string) {
...
+
+	if interactive.Enabled() && billingModel != string(v1.BillingModelStandard) &&
+		!cmd.Flags().Changed(billingModelAccountIDFlag) {
+		billingModelAccountID, err = interactive.GetString(interactive.Input{
+			Question: "Billing Account ID",
+			Help:     cmd.Flags().Lookup(billingModelAccountIDFlag).Usage,
+			Required: true,
+		})
+		if err != nil {
+			r.Reporter.Errorf("Expected a valid account id: %s", err)
```
tbrisker, line 304: Do/can we validate the account id in any way?

cmd/install/addon/cmd.go
```diff
@@ -242,8 +268,51 @@ func run(cmd *cobra.Command, argv []string) {
...
+		r.Reporter.Infof("Enabling interactive mode")
+	}
+
+	if interactive.Enabled() && billingModel != string(v1.BillingModelStandard) &&
+		!cmd.Flags().Changed(billingModelAccountIDFlag) {
+		billingModelAccountID, err = interactive.GetString(interactive.Input{
+			Question: "Billing Account ID",
+			Help:     cmd.Flags().Lookup(billingModelAccountIDFlag).Usage,
+			Required: true,
+		})
```
tbrisker, line 302: should we default to the user's current aws account? does it even make sense to ask for this in the ROSA context when there is already an AWS account defined?

pkg/ocm/addons.go
```diff
@@ -17,10 +17,22 @@ limitations under the License.
...
 
+type AddOnBilling struct {
+	BillingModel     string
+	BillingAccountID string
+}
+
+var BillingModels = map[string]cmv1.BillingModel{
+	string(amsv1.BillingModelMarketplace): cmv1.BillingModelMarketplace,
+	string(amsv1.BillingModelStandard):    cmv1.BillingModelStandard,
+}
```
tbrisker, line 34: Why do we need to translate these instead of using the ams constants directly?

cmd/install/addon/cmd.go
ciaranRoche: no because that is the current behavior, we inherit the billing model and struct from the cluster. But this is to allow addons be purchased from different accounts. So we can not assume the users aws account.

pkg/ocm/addons.go
ciaranRoche: For our SDK :see_no_evil: we have our own billing models defined, these consts are duplicated between cs and ams :see_no_evil: and ams has no way to fetch a list of valid billing models, so we are kind of stuck with this duplication.

cmd/install/addon/cmd.go
ciaranRoche: Not that I know of, personally, should leave this up to AMS and not do any validation here or in CS. Before the addon can be installed we will pass this information to AMS. If its not valid AMS will error when we try to consume the quota. 

Only validation i think we need here or in CS is if the billing model is not standard, then we need to pass an account id.

PR #820, opened 2022-08-25
[SDA-6610] account-role and operator-role policies are now deleted when roles are

hunterkepley:
[SDA-6610](https://issues.redhat.com/browse/SDA-6610)

### DIFF
--- a/cmd/dlt/accountroles/cmd.go
+++ b/cmd/dlt/accountroles/cmd.go
@@ -174,7 +174,7 @@ func run(cmd *cobra.Command, _ []string) {
 			}
 			err := r.AWSClient.DeleteAccountRole(role)
 			if err != nil {
-				r.Reporter.Errorf("There was an error deleting the account roles: %s", err)
+				r.Reporter.Warnf("There was an error deleting the account roles or policies: %s", err)
 				continue
 			}
 		}
@@ -189,7 +189,7 @@ func run(cmd *cobra.Command, _ []string) {
 		commands := buildCommand(finalRoleList, policyMap)
 
 		if r.Reporter.IsTerminal() {
-			r.Reporter.Infof("Run the following commands to delete the account roles:\n")
+			r.Reporter.Infof("Run the following commands to delete the account roles and policies:\n")
 		}
 		fmt.Println(commands)
 	default:
@@ -216,8 +216,10 @@ func buildCommand(roleNames []string, policyMap map[string][]aws.PolicyDetail) s
 		policyDetails := policyMap[roleName]
 		for _, policyDetail := range policyDetails {
 			if policyDetail.PolicType == aws.Attached && policyDetail.PolicyArn != "" {
-				deletePolicy := fmt.Sprintf("\taws iam detach-role-policy --role-name  %s  --policy-arn  %s",
+				detachPolicy := fmt.Sprintf("\taws iam detach-role-policy --role-name  %s  --policy-arn  %s",
 					roleName, policyDetail.PolicyArn)
+				commands = append(commands, detachPolicy)
+				deletePolicy := fmt.Sprintf("\taws iam delete-policy --policy-arn  %s", policyDetail.PolicyArn)
 				commands = append(commands, deletePolicy)
 			}
 			if policyDetail.PolicType == aws.Inline && policyDetail.PolicyName != "" {
--- a/cmd/dlt/operatorrole/cmd.go
+++ b/cmd/dlt/operatorrole/cmd.go
@@ -181,9 +181,10 @@ func run(cmd *cobra.Command, argv []string) {
 			if !confirm.Prompt(true, "Delete the operator roles  '%s'?", role) {
 				continue
 			}
-			err = r.AWSClient.DeleteOperatorRole(role)
+			err := r.AWSClient.DeleteOperatorRole(role)
+
 			if err != nil {
-				r.Reporter.Errorf("There was an error deleting the Operator Roles: %s", err)
+				r.Reporter.Warnf("There was an error deleting the Operator Roles or Policies: %s", err)
 				continue
 			}
 		}
@@ -197,7 +198,7 @@ func run(cmd *cobra.Command, argv []string) {
 		}
 		commands := buildCommand(roles, policyMap)
 		if r.Reporter.IsTerminal() {
-			r.Reporter.Infof("Run the following commands to delete the Operator roles:\n")
+			r.Reporter.Infof("Run the following commands to delete the Operator roles and policies:\n")
 		}
 		fmt.Println(commands)
 	default:
@@ -211,12 +212,14 @@ func buildCommand(roleNames []string, policyMap map[string][]string) string {
 	for _, roleName := range roleNames {
 		policyARN := policyMap[roleName]
 		detachPolicy := ""
+		deletePolicy := ""
 		if len(policyARN) > 0 {
 			detachPolicy = fmt.Sprintf("\taws iam detach-role-policy --role-name  %s --policy-arn  %s",
 				roleName, policyARN[0])
+			deletePolicy = fmt.Sprintf("\taws iam delete-policy --policy-arn  %s", policyARN[0])
 		}
 		deleteRole := fmt.Sprintf("\taws iam delete-role --role-name  %s", roleName)
-		commands = append(commands, detachPolicy, deleteRole)
+		commands = append(commands, detachPolicy, deleteRole, deletePolicy)
 	}
 	return strings.Join(commands, "\n")
 }
--- a/pkg/aws/policies.go
+++ b/pkg/aws/policies.go
@@ -729,18 +729,32 @@ func checkIfROSAOperatorRole(roleName *string, credRequest map[string]*cmv1.STSO
 
 func (c *awsClient) DeleteOperatorRole(roleName string) error {
 	role := aws.String(roleName)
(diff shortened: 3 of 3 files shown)


pkg/aws/policies.go
```diff
@@ -730,6 +734,10 @@ func (c *awsClient) DeleteOperatorRole(roleName string) error {
 		}
 		return err
 	}
+	err = c.deleteOperatorRolePolicies(policies)
```
vkareh, line 737: Operator role policies are shared across many clusters. Deleting a policy will likely fail as it will be attached to the operator roles of other clusters. _If_ it succeeds (it shouldn't) it would break those other clusters.

I think this function should check whether this role is attached to any other roles, and only delete them if no role is using it. Alternatively, this should only be called for `rosa delete account-roles`.

cmd/dlt/accountroles/cmd.go
```diff
@@ -172,8 +172,11 @@ func run(cmd *cobra.Command, _ []string) {
 			if !confirm.Prompt(true, "Delete the account role '%s'?", role) {
 				continue
 			}
-			err := r.AWSClient.DeleteAccountRole(role)
-			if err != nil {
+			policyARN, err := r.AWSClient.DeleteAccountRole(role)
+			if err != nil && strings.Contains(err.Error(), "DeleteConflict: Cannot delete a policy attached to entities.") {
```
vkareh, line 176: There _has_ to be a better way than comparing this. The AWS SDK should report the error type and you can compare against something like `aerr.Type() == awserr.ErrorCodeDeleteConflict`.

Worse case scenario, just do `string.Contains(err.Error(), "DeleteConflict")`

hunterkepley: I thought there would be something better, thanks for that, I'll work on this Monday

vkareh: sure thing! try something like this:
```go
if err != nil {
  aerr, ok := err.(awserr.Error)
  if ok && aerr.Code() == iam.ErrCodeDeleteConflictException {
    r.Reporter.Warnf("Error deleting policy %s. Cannot delete policy attached to other entities.", policyARN)
  } else {
    r.Reporter.Errorf("There was an error deleting the account roles: %s", err)
  }
  continue
}
```

pkg/aws/policies.go
```diff
@@ -738,9 +742,17 @@ func (c *awsClient) DeleteOperatorRole(roleName string) error {
 					roleName)
 			}
 		}
-		return err
+		return "", err
 	}
-	return c.DeleteRole(roleName, role)
+	policyARN, err := c.deleteOperatorRolePolicies(policies)
```
oriAdler, line 747: Do we need a similar function for deleting account role policies?
@pvasant @vkareh just double checking - we want to delete the account role policies after the account roles are removed, right?

pkg/aws/policies.go
```diff
@@ -848,18 +864,40 @@ func (c *awsClient) DeleteInlineRolePolicies(role string) error {
...
 	}
 
-	return nil
+	return "", nil
 }
+
+func (c *awsClient) deletePolicies(policies []string) (*iam.DeletePolicyOutput, string, error) {
+	var output *iam.DeletePolicyOutput
+	var err error
+	for i := range policies {
```
oriAdler, line 891: NP
```suggestion
	for _, policy := range policies {
```
And then we can remove `policyARN := policies[i]`

pkg/aws/policies.go
vkareh: This is a good point. The function that deletes the roles, detaches the policies but doesn't delete them. It only attempts deleting inline policies, but we're not using inline policies for this. The policies should be deleted after being detached.

pkg/aws/policies.go
hunterkepley: We cannot do this because we are taking the memory address of `policyARN` -- You cannot take the memory address of loop variables such as `policy` in your suggestion.

hunterkepley: @oriAdler

cmd/dlt/operatorrole/cmd.go
```diff
@@ -211,12 +212,15 @@ func buildCommand(roleNames []string, policyMap map[string][]string) string {
...
 		detachPolicy := ""
+		deletePolicy := ""
 		if len(policyARN) > 0 {
 			detachPolicy = fmt.Sprintf("\taws iam detach-role-policy --role-name  %s --policy-arn  %s",
 				roleName, policyARN[0])
+			deletePolicy = fmt.Sprintf("\taws iam delete-policy --policy-arn  %s", policyARN[0])
 		}
 		deleteRole := fmt.Sprintf("\taws iam delete-role --role-name  %s", roleName)
 		commands = append(commands, detachPolicy, deleteRole)
+		commands = append(commands, deletePolicy)
```
vkareh, line 223: This can be collapsed into a single row:
```
commands = append(commands, detachPolicy, deleteRole, deletePolicy)
```

pkg/aws/policies.go
```diff
@@ -729,18 +729,40 @@ func checkIfROSAOperatorRole(roleName *string, credRequest map[string]*cmv1.STSO
...
+	}
+	err = c.detachOperatorRolePolicies(role)
 	if err != nil {
 		if aerr, ok := err.(awserr.Error); ok {
 			switch aerr.Code() {
 			case iam.ErrCodeNoSuchEntityException:
 				fmt.Printf("Policies does not exists for role '%s'",
 					roleName)
+			case iam.ErrCodeDeleteConflictException:
+				err2 := c.DeleteRole(roleName, role)
```
vkareh, line 744: I'm not sure I like this pattern. It duplicates code. I think a better idea might be to explicitly ignore this error:
```go
err = c.detachOperatorRolePolicies(role)
if err != nil {
    switch aerr.Code() {
      fmt.Printf(...)
      err = nil
    case iam.ErrCodeDeleteConflictException:
      fmt.Printf(...)
      err = nil
    }
  }
  if err != nil {
    return err
  }
}
err = c.DeleteRole(roleName, role)
```
The idea being that these two error codes should not prevent ROSA from attempting to delete the role. The role deletion might still succeed for all we know, and these policy errors should either fail the entire thing, or just be logged and ignored.

hunterkepley: That's a much better way of handling this, implemented now

vkareh:
This looks good to me, thanks!
@oriAdler can you give it a final review and stamp it for merge if you agree with this approach?

pkg/aws/policies.go
hunterkepley: I have removed the variable policyARN anyway and now am only using one instance of `policies[i]`

pkg/aws/policies.go
```diff
@@ -909,6 +950,17 @@ func (c *awsClient) GetAttachedPolicy(role *string) ([]PolicyDetail, error) {
 	return policies, nil
 }
 
+func (c *awsClient) getAttachedOperatorRolePolicies(role *string) ([]*iam.AttachedPolicy, error) {
```
oriAdler, line 953: Maybe we can reuse the `GetPolicies` function instead of this new function,
And then we can remove `deleteOperatorRolePolicies` and use `deletePolicies` instead, WDYT?

oriAdler (commented):
@hunterkepley I've left a small comment, except that code LGTM!

pkg/aws/policies.go
hunterkepley: Simple enough of a change, thanks for the suggestion

hunterkepley:
@vkareh Could you give one final look at the small changes I made based on Ori's suggestion, and if it looks good merge? I don't think I have merge permissions on this project

pkg/aws/policies.go
```diff
@@ -926,6 +968,16 @@ func (c *awsClient) detachOperatorRolePolicies(role *string) error {
...
 
+/*func (c *awsClient) deleteOperatorRolePolicies(policies []*iam.AttachedPolicy) error {
+	for _, policy := range policies {
+		_, err := c.iamClient.DeletePolicy(&iam.DeletePolicyInput{PolicyArn: policy.PolicyArn})
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}*/
```
vkareh, line 979: Please remove commented out code (it will all be in the git history anyway, so no need to keep this stuff around).

pkg/aws/policies.go
```diff
@@ -848,18 +881,27 @@ func (c *awsClient) DeleteInlineRolePolicies(role string) error {
...
 }
 
-func (c *awsClient) deleteAccountRolePolicies(role *string) error {
-	err := c.detachAttachedRolePolicies(role)
-	if err != nil {
-		return err
-	}
-	err = c.DeleteInlineRolePolicies(aws.StringValue(role))
+func (c *awsClient) deleteAccountRoleInlinePolicies(role *string) error {
+	err := c.DeleteInlineRolePolicies(aws.StringValue(role))
```
vkareh, line 885: I think this function can be removed and just call `DeleteInlineRolePolicies` directly

hunterkepley: Ah yeah not sure why I made that an entire separate function in the first place

vkareh: the function was already there, you just reduced it since it had a lot of unnecessary stuff, but it ended up becoming just a no-op wrapper :stuck_out_tongue:

PR #821, opened 2022-08-26
policies: Ensure policy version succeeds

vkareh:
When adding a new policy version, there is a risk of hitting an AWS
limit. We therefore remove all but the default policy version and add
the new one to ensure success.

Fixes https://issues.redhat.com/browse/SDA-6659

### DIFF
--- a/pkg/aws/policies.go
+++ b/pkg/aws/policies.go
@@ -272,21 +272,29 @@ func (c *awsClient) EnsurePolicy(policyArn string, document string,
 	}
 
 	if !isCompatible {
+		// Since there is a limit to how many versions a policy can have, we delete all non-default
+		// policy versions from the list, thus making space for the new one.
+		policyVersionsOutput, err := c.iamClient.ListPolicyVersions(&iam.ListPolicyVersionsInput{
+			PolicyArn: aws.String(policyArn),
+		})
+		if err != nil {
+			return policyArn, err
+		}
+		for _, version := range policyVersionsOutput.Versions {
+			if aws.BoolValue(version.IsDefaultVersion) {
+				continue
+			}
+			c.iamClient.DeletePolicyVersion(&iam.DeletePolicyVersionInput{
+				PolicyArn: aws.String(policyArn),
+				VersionId: version.VersionId,
+			})
+		}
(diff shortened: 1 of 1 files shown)


pkg/aws/policies.go
```diff
@@ -272,21 +272,29 @@ func (c *awsClient) EnsurePolicy(policyArn string, document string,
...
+			PolicyArn: aws.String(policyArn),
+		})
+		if err != nil {
+			return policyArn, err
+		}
+		for _, version := range policyVersionsOutput.Versions {
+			if aws.BoolValue(version.IsDefaultVersion) {
+				continue
+			}
+			c.iamClient.DeletePolicyVersion(&iam.DeletePolicyVersionInput{
```
pvasant, line 287: @vkareh  Do you think we can leave the latest one before deleting it? Because if the upgrade fails or create policy version fails, we risk deleting everything. wdyt?

vkareh: The latest one (default) will remain, so after this function runs you will always have the current/default and the previous one.

pvasant:
@vkareh  Thanks, Victor. looks good to me. Do we need to document this, or is it an implementation detail that customers don't care about?

vkareh:
> @vkareh Thanks, Victor. looks good to me. Do we need to document this, or is it an implementation detail that customers don't care about?

Ah good point, I think this would be good to have documented.
/cc @arendej

tbrisker:
@vkareh is it possible the non-default policies are still in use be older clusters and deleting them will cause these clusters to break?

vkareh:
It shouldn't. Clusters only access the _roles_, and the role has the attached policy _without_ a version specified. AWS determines which policy version to use when assuming role based on this.

tbrisker:
@vkareh why do we use multiple versions at all then instead of deleting and recreating the policy when we update it?

vkareh:
@tbrisker because AWS does not allow you to delete a policy that is currently attached to a role. You would need to detach the policy (potentially incapacitating the cluster operator briefly), create the new version and re-attach it. Seems dangerous, and AWS provides the policy versioning as a way around that issue.

tbrisker:
/lgtm
thanks for the explanation @vkareh!

PR #824, opened 2022-09-01
move operator policies from account to operator commands

ydayagi:
[SDE-2115](https://issues.redhat.com//browse/SDE-2115)
https://issues.redhat.com/browse/SDE-2115

### DIFF
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -78,6 +78,7 @@ func init() {
 		"",
 		"The arn path for the account roles",
 	)
+	flags.MarkHidden("role-path")
 
 	flags.StringVar(
 		&args.policyPath,
@@ -85,6 +86,7 @@ func init() {
 		"",
 		"The arn path for the account policies",
 	)
+	flags.MarkHidden("policy-path")
 
 	aws.AddModeFlag(Cmd)
 
@@ -214,7 +216,7 @@ func run(cmd *cobra.Command, argv []string) {
 	}
 
 	rolePath := args.rolePath
-	if interactive.Enabled() {
+	if cmd.Flags().Changed("role-path") && interactive.Enabled() {
 		rolePath, err = interactive.GetString(interactive.Input{
 			Question: "Role Path",
 			Help:     cmd.Flags().Lookup("role-path").Usage,
@@ -230,7 +232,7 @@ func run(cmd *cobra.Command, argv []string) {
 	}
 
 	policyPath := args.policyPath
-	if interactive.Enabled() {
+	if cmd.Flags().Changed("policy-path") && interactive.Enabled() {
 		policyPath, err = interactive.GetString(interactive.Input{
 			Question: "Policy Path",
 			Help:     cmd.Flags().Lookup("policy-path").Usage,
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -24,7 +24,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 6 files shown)


cmd/create/accountroles/cmd.go
```diff
@@ -79,13 +78,15 @@ func init() {
 		"",
 		"The arn path for the account roles",
 	)
+	flags.MarkHidden("role-path")
```
vkareh, line 81: Why are hiding these role/policy paths? They are still used for the account roles and policies, so we need them both here _and_ in the operator roles/policies

ydayagi: the path feature is not yet implemented (not fully and has lots of bugs) and we do not want to release it yet. So after discussing with @pvasant we decided to hide it until the whole implementation is ready.

vkareh: Ah, that's a good idea. Can you do it in a separate commit please? It can be in the same PR, but in a separate commit in case one or the other needs to be reverted it doesn't drag the rest of the changes in.

ydayagi: done

cmd/create/accountroles/cmd.go
```diff
@@ -215,36 +216,7 @@ func run(cmd *cobra.Command, argv []string) {
...
-			Validators: []interactive.Validator{
-				aws.ARNPathValidator,
-			},
-		})
-		if err != nil {
-			r.Reporter.Errorf("Expected a valid path: %s", err)
-			os.Exit(1)
-		}
-	}
-
```
vkareh, line 232: I would honestly not remove these. How I've done it in the past is to only show the interactive mode for hidden flags if the user sets both the flag and interactive mode:
```go
if cmd.Flags().Changed("role-path") && interactive.Enabled() {
  ...
}
```
That way the code is still in place, but it only shows up if the user knows about the flags. It also means we can QE/test better to make sure the feature is ready. Then enabling it when it's ready becomes easier.

ydayagi:
Will do

cmd/upgrade/operatorroles/cmd.go
```diff
@@ -278,20 +285,19 @@ func handleModeFlag(cmd *cobra.Command, skipInteractive bool, mode string, err e
...
-	}
+func upgradeOperatorPolicies(isProgrammaticallyCalled bool, mode string, r *rosa.Runtime,
+	prefix string, isAccountRoleUpgradeNeed bool, policies map[string]string, env string, defaultPolicyVersion string,
+	credRequests map[string]*cmv1.STSOperator, cluster *cmv1.Cluster, policyPath string) error {
 	switch mode {
 	case aws.ModeAuto:
 		if !confirm.Prompt(true, "Upgrade the operator role policy to version %s?", defaultPolicyVersion) {
+			if isProgrammaticallyCalled {
+				return r.Reporter.Errorf("Operator roles need to be upgraded to proceed")
+			}
```
vkareh, line 296: Why is this here? Why are we blocking the user from upgrading the roles?

ydayagi: This was the behavior in account roles. When called from cluster upgrade (isProgramatically is true) then account roles upgrade prompts the usrr and returns error if user does not confirm

cmd/create/accountroles/cmd.go
ydayagi: done

cmd/create/operatorroles/cmd.go
```diff
@@ -78,8 +78,9 @@ func init() {
 		&args.policyPath,
 		"policy-path",
```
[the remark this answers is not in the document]
ydayagi: it is provided during 'create cluster'

oriAdler:
LGTM

PR #826, opened 2022-09-02
checking addon params

laurafitzgerald:
Removing WIP status

Addressed in this pr.
- Use GetOptions in cases where params have options listed - matches UI, the user is forced to select - set the first value in the array as the default.
- Check that val selected, or provided is contained in the options when there are options
- Move logic for install/edit get parameters section to function

### DIFF
--- a/cmd/edit/addon/cmd.go
+++ b/cmd/edit/addon/cmd.go
@@ -18,16 +18,15 @@ package addon
 
 import (
 	"fmt"
-	"net"
 	"os"
 	"regexp"
-	"strconv"
 	"strings"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 
 	"github.com/openshift/rosa/pkg/arguments"
+	"github.com/openshift/rosa/pkg/helper"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/rosa"
@@ -128,81 +127,53 @@ func run(cmd *cobra.Command, argv []string) {
 		}
 
 		var val string
-		var hasVal bool
+		var options []string
+		var values []string
+
+		parameterOptions, _ := param.GetOptions()
+
+		for _, opt := range parameterOptions {
+			options = append(options, opt.Name())
+			values = append(values, opt.Value())
+		}
+
 		// If value is already set in the CLI, ignore interactive prompt
 		flag := cmd.Flags().Lookup(param.ID())
 		if flag != nil {
 			val = flag.Value.String()
-			hasVal = true
-		} else if interactive.Enabled() {
-			// Set default value based on existing parameter, otherwise use parameter default
+		}
+		if interactive.Enabled() {
+			input := interactive.Input{}
+			input.Question = param.Name()
+			input.Help = fmt.Sprintf("%s: %s", param.ID(), param.Description())
+			input.Required = param.Required()
+			input.Options = options
+
 			dflt := param.DefaultValue()
 			if addOnInstallationParam != nil {
 				dflt = addOnInstallationParam.Value()
 			}
 
-			input := interactive.Input{
-				Question: param.Name(),
-				Help:     fmt.Sprintf("%s: %s", param.ID(), param.Description()),
-				Required: param.Required(),
-			}
-
-			// add a prompt to question name to indicate if the boolean param is required and check validation
-			if param.ValueType() == "boolean" && param.Validation() == "^true$" && param.Required() {
-				input.Question = fmt.Sprintf("%s (required)", param.Name())
-				input.Validators = []interactive.Validator{
-					interactive.RegExpBoolean(param.Validation()),
-				}
-			}
-
-			switch param.ValueType() {
-			case "boolean":
-				var boolVal bool
-				input.Default, _ = strconv.ParseBool(dflt)
-				boolVal, err = interactive.GetBool(input)
-				if boolVal {
-					val = "true"
-				} else {
-					val = "false"
-				}
-			case "cidr":
-				var cidrVal net.IPNet
-				if dflt != "" {
-					_, defaultIDR, _ := net.ParseCIDR(dflt)
-					input.Default = *defaultIDR
-				}
-				cidrVal, err = interactive.GetIPNet(input)
-				val = cidrVal.String()
-				if val == "<nil>" {
-					val = ""
-				}
-			case "number", "resource":
-				var numVal int
-				input.Default, _ = strconv.Atoi(dflt)
-				numVal, err = interactive.GetInt(input)
-				val = fmt.Sprintf("%d", numVal)
-			case "string":
-				input.Default = dflt
-				val, err = interactive.GetString(input)
-			}
+			val, err = interactive.GetAddonParameter(param, input, dflt)
 			if err != nil {
-				r.Reporter.Errorf("Expected a valid value for '%s': %v", param.ID(), err)
+				r.Reporter.Errorf("%s", err)
 				os.Exit(1)
 			}
-			hasVal = true
+		}
+		val = strings.Trim(val, " ")
+		if val != "" && param.Validation() != "" {
+			isValid, err := regexp.MatchString(param.Validation(), val)
+			if err != nil || !isValid {
+				r.Reporter.Errorf("Expected %v to match /%s/", val, param.Validation())
+				os.Exit(1)
+			}
 		}
 
-		if hasVal {
-			val = strings.Trim(val, " ")
-			if val != "" && param.Validation() != "" {
-				isValid, err := regexp.MatchString(param.Validation(), val)
-				if err != nil || !isValid {
-					r.Reporter.Errorf("Expected %v to match /%s/", val, param.Validation())
-					os.Exit(1)
-				}
-			}
-			params = append(params, ocm.AddOnParam{Key: param.ID(), Val: val})
+		if len(options) > 0 && !helper.Contains(values, val) {
+			r.Reporter.Errorf("Expected %v to match one of the options /%v/", val, values)
+			os.Exit(1)
 		}
+		params = append(params, ocm.AddOnParam{Key: param.ID(), Val: val})
 
 		return true
 	})
--- a/cmd/install/addon/cmd.go
+++ b/cmd/install/addon/cmd.go
@@ -18,10 +18,8 @@ package addon
 
 import (
 	"fmt"
-	"net"
 	"os"
 	"regexp"
-	"strconv"
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws/awserr"
@@ -34,6 +32,7 @@ import (
 	"github.com/openshift/rosa/pkg/arguments"
 	"github.com/openshift/rosa/pkg/aws"
 	"github.com/openshift/rosa/pkg/aws/tags"
+	"github.com/openshift/rosa/pkg/helper"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/interactive/confirm"
 	"github.com/openshift/rosa/pkg/ocm"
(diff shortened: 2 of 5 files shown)


cmd/install/addon/cmd.go
```diff
@@ -197,17 +197,25 @@ func run(cmd *cobra.Command, argv []string) {
...
 			if flag != nil {
 				val = flag.Value.String()
 				hasVal = true
 			} else if interactive.Enabled() {
-				input := interactive.Input{
-					Question: param.Name(),
-					Help:     fmt.Sprintf("%s: %s", param.ID(), param.Description()),
-					Required: param.Required(),
+				options, ok := param.GetOptions()
+				if ok {
```
tbrisker, line 212: Should we use a `interactive.GetOption` prompt when there are predefined options for the parameter?

cmd/install/addon/cmd.go
```diff
@@ -261,6 +269,12 @@ func run(cmd *cobra.Command, argv []string) {
 						os.Exit(1)
 					}
 				}
+
+				if len(input.Options) > 0 && !contains(input.Options, val) {
```
oriAdler, line 273: We can use the `Contains` function from `pkg/helper/helpers.go`
```suggestion
				if len(input.Options) > 0 && !helper.Contains(input.Options, val) {
```

austincunningham:
/lgtm
verified the functionality 
checked out this branch 
ran
```bash
make install
```
On a rosa cluster installed the addon
```bash
rosa install addon --cluster=<cluster-name> managed-api-service
```
Got access to a menu to select the quota

quota

The values for the quota don't reflect what the quota actually is `1==100k requests per day`, but we may address this elsewhere

cmd/edit/addon/cmd.go
```diff
@@ -201,9 +158,13 @@ func run(cmd *cobra.Command, argv []string) {
 					os.Exit(1)
 				}
 			}
+
+			if len(input.Options) > 0 && helper.Contains(input.Options, val) {
+				r.Reporter.Errorf("Expected %v to match one of the options /%v/", val, param.Options())
+				os.Exit(1)
+			}
```
tbrisker, line 165: This should probably be done in an input validator?

cmd/edit/addon/cmd.go
```diff
@@ -129,69 +126,29 @@ func run(cmd *cobra.Command, argv []string) {
 
 		var val string
 		var hasVal bool
+		input := &interactive.Input{
```
tbrisker, line 129: using an input outside of interactive mode only to get the options seems strange to me

pkg/interactive/interactive.go
```diff
@@ -46,7 +72,7 @@ func GetString(input Input) (a string, err error) {
 		dflt = ""
 	}
 	question := input.Question
-	if !input.Required && dflt == "" {
+	if !input.Required && dflt == "" && !strings.Contains(strings.ToLower(question), "optional") {
```
tbrisker, line 75: This looks like something that should be fixed in the parameter description, not here (and below). That's what the `required` attribute is for.

cmd/install/addon/cmd.go
```diff
@@ -197,61 +195,36 @@ func run(cmd *cobra.Command, argv []string) {
...
+					r.Reporter.Errorf("%s", err)
 					os.Exit(1)
 				}
 				hasVal = true
 			}
-
+			if !hasVal {
+				return false
+			}
 			if hasVal {
```
tbrisker, line 228: you can get rid of the second `if` if you return early

cmd/install/addon/cmd.go
```diff
@@ -197,61 +195,36 @@ func run(cmd *cobra.Command, argv []string) {
...
+				Help:     fmt.Sprintf("%s: %s", param.ID(), param.Description()),
+				Required: param.Required(),
+			}
+			options, ok := param.GetOptions()
+			if ok {
+				input.Options = []string{}
+				for _, opt := range options {
+					input.Options = append(input.Options, opt.Value())
+				}
+			}
```
tbrisker, line 209: is this part missing from the edit command on purpose?

cmd/install/addon/cmd.go
```diff
@@ -197,61 +195,36 @@ func run(cmd *cobra.Command, argv []string) {
...
 			var hasVal bool
+			input := &interactive.Input{
+				Question: param.Name(),
+				Help:     fmt.Sprintf("%s: %s", param.ID(), param.Description()),
+				Required: param.Required(),
+			}
+			options, ok := param.GetOptions()
+			if ok {
+				input.Options = []string{}
+				for _, opt := range options {
```
tbrisker, line 206: NP: if there are no options, `param.Options` will return a nil array, so you could use 
```suggestion
				for _, opt := range param.Options {
```
as a nil array will be a nop for range, and there is no need to initialize `input.Options` as it already has the zero value when the input is created.

pkg/interactive/interactive.go
```diff
@@ -38,6 +40,30 @@ type Input struct {
 	Validators []Validator
 }
 
+func (i *Input) LenOptions() int {
+	if i == nil {
+		return 0
+	}
+	if i.Options == nil {
+		return 0
+	}
```
tbrisker, line 49: No need for this check, the length of a nil array is 0

pkg/interactive/interactive.go
```diff
@@ -38,6 +40,30 @@ type Input struct {
 	Validators []Validator
 }
 
+func (i *Input) LenOptions() int {
+	if i == nil {
```
tbrisker, line 44: is this possible?

pkg/interactive/interactive.go
```diff
@@ -38,6 +40,30 @@ type Input struct {
...
+	if i.Options == nil {
+		return 0
+	}
+	return len(i.Options)
+}
+func (i *Input) OptionsContains(val string) bool {
+	if i == nil {
+		return false
+	}
+	if i.Options == nil {
```
tbrisker, line 56: again, no need for this check - range over nil array is no-op

pkg/interactive/interactive.go
```diff
@@ -38,6 +40,30 @@ type Input struct {
 	Validators []Validator
 }
 
+func (i *Input) LenOptions() int {
```
tbrisker, line 43: Not sure i understand why this method is needed instead of using `len(input.Options)` directly?

pkg/interactive/interactive.go
```diff
@@ -312,3 +338,64 @@ func PrintHelp(help Help) error {
...
+			input.Default = *defaultIDR
+		}
+		cidrVal, err = GetIPNet(*input)
+		value = cidrVal.String()
+		if value == "<nil>" {
+			value = ""
+		}
+	case "number", "resource":
+		if input.LenOptions() > 0 {
+			value, err = GetOption(*input)
```
tbrisker, line 382: if we provide options, shouldn't that take precedence over the value type check and make it redundant?

tbrisker (commented):
thanks @laurafitzgerald, added some comments/questions inline

pkg/interactive/interactive.go
laurafitzgerald: @tbrisker unfortunately we are caught between the OCM UI and the ROSA Cli. We have the optional and required in the string of the description. So in ROSA CLI it's included twice. If we remove it from the addon parameter configuration we lose it in the UI.

cmd/edit/addon/cmd.go
laurafitzgerald: We want to validate in case of interactive input an parameter passing. The user has a limited set of options that can be passed that the addon can handle.

cmd/edit/addon/cmd.go
laurafitzgerald: I'm declaring the object here in order to use it in the validation later. I can make a change where it just an empty object and put in the values inside the interactive block.

cmd/edit/addon/cmd.go
laurafitzgerald: I searched in ocm-sdk-go addon_parameter_type. I didn't see a ValidateOptions function there unfortunately.

cmd/install/addon/cmd.go
laurafitzgerald: No, not on purpose. I think I lost it when I refactored the function. Adding it in. 👍

pkg/interactive/interactive.go
laurafitzgerald: 😵‍💫  it's been a while since writing some code. lol. hope you forgive the silly mistake. 😄

pkg/interactive/interactive.go
```diff
@@ -312,3 +338,64 @@ func PrintHelp(help Help) error {
 	fmt.Fprint(terminal.NewAnsiStdout(os.Stdout), out)
 	return nil
 }
+
+func ProcessParamInteractive(addOnInstallationParam *cmv1.AddOnInstallationParameter, param *cmv1.AddOnParameter, input *Input) (error, string) {
```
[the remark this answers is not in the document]
laurafitzgerald: @tbrisker I can move it elsewhere. I don't see any existing addon code I can add it to. The Input object is used a lot and the other function inside this file which is why I chose this. It's a refactor of logic which exists in both `cmd/install/addon/cmd.go` and `cmd/edit/addon/cmd.go` Didn't seem right to put it in either one of those and there was no addon specific spot in `cmd` that I could see. If you have a preference elsewhere I'm happy to move it. First pr to this repo so I hope you can forgive the lack of knowledge.

pkg/interactive/interactive.go
laurafitzgerald: I think there can be options in the case of a number of the different types? Is the suggestion here to move this above the case statement and ignore the case if we have options?

laurafitzgerald:
@tbrisker thanks for the review. I've addressed some comments and answered others. If you have time to take another look that would be really appreciated. 👍

cmd/edit/addon/cmd.go
laurafitzgerald: I found AddOnParameterOptionList here -> https://github.com/openshift-online/ocm-sdk-go/blob/main/clustersmgmt/v1/add_on_parameter_option_type.go#L191 Potentially should we be using this instead of this -> https://github.com/openshift-online/ocm-sdk-go/blob/main/clustersmgmt/v1/add_on_parameter_type.go#L46 ?

pkg/interactive/interactive.go
tbrisker: yes, that's what i'm thinking - in case we have options we don't need to go through the case statement

pkg/interactive/interactive.go
laurafitzgerald: Actually I think this might be fixable from our side. I've opened an MR [1] so will remove the code here.
[1] https://gitlab.cee.redhat.com/service/managed-tenants/-/merge_requests/3036

laurafitzgerald: After merge of the above this is now working as expected so removed the code.

tbrisker: 🥳 glad you figured out a better way :)

pkg/interactive/interactive.go
```diff
@@ -18,13 +18,12 @@ package interactive
...
-	"net"
-	"os"
-	"strconv"
-
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/AlecAivazis/survey/v2/core"
 	"github.com/AlecAivazis/survey/v2/terminal"
+	"net"
+	"os"
+	"strconv"
```
tbrisker, line 26: This reorder is incorrect, i'm guessing some messed up editor config?

pkg/interactive/interactive.go
```diff
@@ -176,6 +175,9 @@ func GetOption(input Input) (a string, err error) {
 	if !input.Required && dflt == "" {
 		question = fmt.Sprintf("%s (optional)", question)
 	}
+	if len(input.Options) > 0 {
+		dflt = input.Options[0]
+	}
```
tbrisker, line 180: what if we have a default defined for the input that isn't the first option?

pkg/interactive/addon.go
```diff
@@ -0,0 +1,67 @@
+package interactive
+
+import (
+	"fmt"
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
```
tbrisker, line 5: NP: import order (this should be in a separate import section)

pkg/interactive/addon.go
```diff
@@ -0,0 +1,67 @@
...
+	"net"
+	"strconv"
+)
+
+func GetAddonParameter(addOnInstallationParam *cmv1.AddOnInstallationParameter, param *cmv1.AddOnParameter, input Input) (string, error) {
+	// Set default value based on existing parameter, otherwise use parameter default
+	dflt := param.DefaultValue()
+	if addOnInstallationParam != nil {
+		dflt = addOnInstallationParam.Value()
+	}
```
tbrisker, line 15: since we're only using `addOnInstallationParam` to override the default value when editing an addon installation, perhaps we could instead pass the default from the caller?

cmd/install/addon/cmd.go
```diff
@@ -197,72 +196,53 @@ func run(cmd *cobra.Command, argv []string) {
...
 			var val string
 			var hasVal bool
+			var optionStrings []string
+
+			options, ok := param.GetOptions()
+			if ok {
+				for _, opt := range options {
+					optionStrings = append(optionStrings, opt.Value())
+				}
+			}
```
tbrisker, line 206: NP: I think we can safely get rid of the ok check here
```suggestion
			for _, opt := range param.Options() {
				optionStrings = append(optionStrings, opt.Value())
			}
```

cmd/install/addon/cmd.go
```diff
@@ -197,72 +196,53 @@ func run(cmd *cobra.Command, argv []string) {
...
 				if err != nil {
-					r.Reporter.Errorf("Expected a valid value for '%s': %v", param.Name(), err)
+					r.Reporter.Errorf("%s", err)
 					os.Exit(1)
 				}
+
 				hasVal = true
 			}
+			if !hasVal {
+				return false
```
tbrisker, line 230: This will cause the parameter loop to stop iterating, i.e. if one parameter is missing a value, the rest of the parameters will be skipped. With `.Each` loops, `return false` is equal to `break` and `return true` to `continue`.

tbh i'm not sure we really need this `hasVal` flag, but that's outside the scope of this PR.

cmd/install/addon/cmd.go
```diff
@@ -18,10 +18,9 @@ package addon
 
 import (
 	"fmt"
-	"net"
+	"github.com/openshift/rosa/pkg/helper"
```
tbrisker, line 21: np: import order

cmd/edit/addon/cmd.go
```diff
@@ -18,19 +18,16 @@ package addon
...
-
 	"github.com/openshift/rosa/pkg/arguments"
+	"github.com/openshift/rosa/pkg/helper"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/rosa"
+	"github.com/spf13/cobra"
+	"os"
+	"regexp"
+	"strings"
```
tbrisker, line 30: np: import order

tbrisker (commented):
Great progress @laurafitzgerald ! please see a few small comments inline

cmd/install/addon/cmd.go
laurafitzgerald: If I remove this return now that the logic is refactored, there isn't any use of the `hasVal` flag. I think it could be removed and the value of `val set to `""` on instantiation. I'm happy to do this in this pr if you agree?

cmd/edit/addon/cmd.go
```diff
@@ -18,19 +18,17 @@ package addon
...
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
-	"github.com/spf13/cobra"
-
 	"github.com/openshift/rosa/pkg/arguments"
+	"github.com/openshift/rosa/pkg/helper"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/rosa"
+	"github.com/spf13/cobra"
```
ciaranRoche, line 31: np -> ordering of imports

Should be in this order
```
native imports

external imports

Internal imports
```

cmd/install/addon/cmd.go
```diff
@@ -197,72 +196,51 @@ func run(cmd *cobra.Command, argv []string) {
 		parameters.Each(func(param *cmv1.AddOnParameter) bool {
 			var val string
 			var hasVal bool
+			var optionStrings []string
+
+			options, _ := param.GetOptions()
+			for _, opt := range options {
+				optionStrings = append(optionStrings, opt.Value())
```
ciaranRoche, line 203: I think it would be better to create a list of `option.Name()`. The value is arbitrary, where as the name has some context behind it

cmd/install/addon/cmd.go
ciaranRoche: yes, please remove `hasVal` as it is not required any more

laurafitzgerald:
@ciaranRoche implemented your suggestions. thanks for the review.

ciaranRoche:
JIRA - SDA-6845

PR #828, opened 2022-09-06
[DRAFT] Add support for Hypershift cluster creation

samira-barouti:
* Added a new flag (--hosted-cp) to `create cluster`
* Modified GetCredRequests to take a new arg which indicates if the cluster is hypershift or not


cmd/create/cluster/cmd.go
```diff
@@ -508,6 +511,14 @@ func init() {
 			"roles in STS clusters.",
 	)
 
+	// Options releated to HyperShift:
+	flags.BoolVar(
+		&args.hostedClusterEnabled,
+		"hosted-cp",
+		false,
+		"Enable or disable use of Hosted Cluster mode.",
+	)
```
vkareh, line 520: Let's add a `flags.MarkHidden("hosted-cp")` here to not exposed non-supported features.

cmd/create/cluster/cmd.go
```diff
@@ -508,6 +511,14 @@ func init() {
 			"roles in STS clusters.",
 	)
 
+	// Options releated to HyperShift:
+	flags.BoolVar(
+		&args.hostedClusterEnabled,
+		"hosted-cp",
+		false,
+		"Enable or disable use of Hosted Cluster mode.",
```
vkareh, line 519: I think this should reflect better what the flag does from the customer point of view:

Maybe something like `"Enable the use of hosted control planes (HyperShift)."`

pkg/ocm/client.go
```diff
@@ -124,6 +124,7 @@ func (b *ClientBuilder) Build() (result *Client, err error) {
 	if b.cfg.URL != "" {
 		builder.URL(b.cfg.URL)
 	}
+	builder.AlternativeURL("/api/accounts_mgmt", "https://api.integration.openshift.com")
```
vkareh, line 127: Please remove this.

pkg/ocm/helpers.go
```diff
@@ -556,7 +556,8 @@ func (c *Client) GetPolicies(policyType string) (map[string]string, error) {
 	return m, nil
 }
 
-func (c *Client) GetCredRequests() (map[string]*cmv1.STSOperator, error) {
+// todo
+func (c *Client) GetCredRequests(isHypershift bool) (map[string]*cmv1.STSOperator, error) {
```
vkareh, line 560: I think the parameter should hide the purpose in this case. Maybe just send the cluster object and let this function decide what it needs from the cluster.
```
func (c *Client) GetCredRequests(cluster *cmv1.Cluster) (map[string]*cmv1.STSOperator, error) {
...
```

vkareh: Actually, you're not using `isHypershift` in this function, so maybe just remove it.

pkg/ocm/client.go
samira-barouti: The cluster creation does not work without this. That's why I opened the MR to ask for feedback on how to resolve this issue.

pkg/ocm/helpers.go
samira-barouti: The implementation is WIP. Still need to figure out how to use this boolean to get the right credential requests. Wanted to make sure that the structure is fine before moving forward to implement this.

pkg/ocm/client.go
vkareh: Maybe create a proxy in your localhost, or some /etc/hosts magic? I think some other folks that don't use `crc` use nginx for proxying. I've never done that, since `crc` handles all routing natively without changes in the code. Alternatively you can deploy AMS locally.

pkg/ocm/helpers.go
samira-barouti: @vkareh I tried to send the cluster object, but that won't work for the cluster creation; for the cluster creation we need the operatorroles slice, which is created in this block; therefore this piece of code cannot be moved to after the cluster creation.

cmd/create/cluster/cmd.go
```diff
@@ -508,6 +511,15 @@ func init() {
 			"roles in STS clusters.",
 	)
 
+	// Options releated to HyperShift:
+	flags.BoolVar(
```
pvasant, line 515: Do we need interactive mode for this

cmd/create/cluster/cmd.go
```diff
@@ -1327,7 +1339,7 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 
 		// Validate subnets in the case the user has provided them using the `args.subnets`
-		if subnetsProvided {
+		if subnetsProvided && !args.hostedClusterEnabled {
```
pvasant, line 1342: We still need to validate subnet right? whether it is just private?

igoihman: we should have interactive support

cmd/create/cluster/cmd.go
igoihman: we should have interactive support

cmd/create/cluster/cmd.go
pvasant: Also can you add the validation or enforce BYO VPC if it is a hosted cluster

cmd/create/cluster/cmd.go
samira-barouti: Good call. Will add the support for the interactive mode

cmd/create/cluster/cmd.go
samira-barouti: The current code in CS, only needs the private subnet id; it then fetches the VPC ID and we randomly chooses a public subnet for the default node pool. Had a chat with @vkareh and we decided this might cause surprises, and we need to modify this behavior. The next patch will have the fix for this, as well as the VPC validation.

pvasant: Why do we need a public subnet for a private link cluster? @vkareh @samira-barouti

samira-barouti: The OAuth server endpoint is always public (even for the private clusters)
Ref: https://docs.google.com/document/d/1egszbmAMkwI1_ZXgVGuAtCX5wEhGiorq9cMBoS9X08A/edit#heading=h.dxjx8ylhdkac

cmd/create/cluster/cmd.go
```diff
@@ -598,6 +611,20 @@ func run(cmd *cobra.Command, _ []string) {
 		isIAM = !isSTS
 	}
 
+	isHostedCluster := args.hostedClusterEnabled
+	if interactive.Enabled() {
```
tbrisker, line 615: Since we want to hide the flag until it's ready, we can use something like:
```suggestion
	if if cmd.Flags().Changed("hosted-cp") && interactive.Enabled() {
```
to enable showing it in interactive mode for testing and internal usage without it being presented to users yet.

tbrisker (commented):
Can you please add the issue number(s) to the commit message?
Also, if i'm not mistaken, this is blocked by [SDA-6728](https://issues.redhat.com/browse/SDA-6728)?

cmd/create/cluster/cmd.go
samira-barouti: Great catch! I almost forgot that this is a hidden flag. Will fix this.

samira-barouti:
@tbrisker I'll add the issue number. This PR will fix [SDA-6728](https://issues.redhat.com/browse/SDA-6728) as well. I added a new method which takes the `isHypershift` boolean, and returns the credRequest accordingly. All occurences of `aws.CredRequests` have been replaced with this method. Please let me know if this looks good.

tbrisker:
lgtm, @vkareh any other concerns?

cmd/create/cluster/cmd.go
```diff
@@ -598,6 +611,20 @@ func run(cmd *cobra.Command, _ []string) {
 		isIAM = !isSTS
 	}
 
+	isHostedCluster := args.hostedClusterEnabled
+	if interactive.Enabled() && cmd.Flags().Changed("hosted-cp") {
+		isHostedCluster, err = interactive.GetBool(interactive.Input{
+			Question: "Deploy a hosted cluster",
+			Help:     cmd.Flags().Lookup("hosted-cp").Usage,
+			Default:  false,
```
vkareh, line 619: The default value should be whatever the flag is currently set as, to ensure that the prompt does not override the user-set flag.

cmd/create/cluster/cmd.go
```diff
@@ -598,6 +611,20 @@ func run(cmd *cobra.Command, _ []string) {
 		isIAM = !isSTS
 	}
 
+	isHostedCluster := args.hostedClusterEnabled
+	if interactive.Enabled() && cmd.Flags().Changed("hosted-cp") {
+		isHostedCluster, err = interactive.GetBool(interactive.Input{
+			Question: "Deploy a hosted cluster",
```
vkareh, line 617: This should be "Deploy cluster with hosted control plane"

cmd/create/cluster/cmd.go
```diff
@@ -598,6 +611,20 @@ func run(cmd *cobra.Command, _ []string) {
 		isIAM = !isSTS
 	}
 
+	isHostedCluster := args.hostedClusterEnabled
+	if interactive.Enabled() && cmd.Flags().Changed("hosted-cp") {
+		isHostedCluster, err = interactive.GetBool(interactive.Input{
```
vkareh, line 616: /nitpick
Variable name should probably be `isHostedCP`

cmd/create/cluster/cmd.go
```diff
@@ -1322,9 +1356,8 @@ func run(cmd *cobra.Command, _ []string) {
 				subnetIDs[i] = aws.ParseSubnet(subnet)
 			}
 		}
-
 		// Validate subnets in the case the user has provided them using the `args.subnets`
-		if subnetsProvided {
+		if subnetsProvided && !isHostedCluster {
 			err = ocm.ValidateSubnetsCount(multiAZ, privateLink, len(subnetIDs))
```
vkareh, line 1361: Maybe this validation function can be modified to validate the number of subnet values based on the privacy values of the HC

pkg/ocm/clusters.go
```diff
@@ -114,6 +117,10 @@ type OperatorIAMRole struct {
 	Path      string
 }
 
+type Hypershift struct {
```
vkareh, line 120: /nitpick
Change struct name to HyperShift

pkg/ocm/helpers.go
vkareh: I'm wondering if the cluster spec should have a 'cluster.Provisioner' attribute with 'hive' or 'hypershift' as values, rather than checking 'cluster.HyperShift.Enabled' to determine that it's Hive after all...

samira-barouti: We had this conversation before, and we voted against that. We feel grouping anything related to hypershift in a struct will simplify adding further information in the future. As an example, we just added a new field (management cluster) in CS.

cmd/create/cluster/cmd.go
samira-barouti: I'd rather have that in a separate PR. Since M1 is for the internal preview and we explicitly ask the users to create subnets, we can skip this.

cmd/create/cluster/cmd.go
samira-barouti: done

PR #829, opened 2022-09-08
Ensure prerequisites for deleting operator and account role policies

oriAdler:
1. Check if the policy is attached to an entity.
2. Delete all policy versions.

Related: [SDA-6747](https://issues.redhat.com/browse/SDA-6747)

Delete a policy with multiple versions:
image

Skip policy deletion if it has attached entities:
image

### DIFF
--- a/pkg/aws/policies.go
+++ b/pkg/aws/policies.go
@@ -274,21 +274,11 @@ func (c *awsClient) EnsurePolicy(policyArn string, document string,
 	if !isCompatible {
 		// Since there is a limit to how many versions a policy can have, we delete all non-default
 		// policy versions from the list, thus making space for the new one.
-		policyVersionsOutput, err := c.iamClient.ListPolicyVersions(&iam.ListPolicyVersionsInput{
-			PolicyArn: aws.String(policyArn),
-		})
+		err = c.deletePolicyVersions(policyArn)
 		if err != nil {
 			return policyArn, err
 		}
-		for _, version := range policyVersionsOutput.Versions {
-			if aws.BoolValue(version.IsDefaultVersion) {
-				continue
-			}
-			c.iamClient.DeletePolicyVersion(&iam.DeletePolicyVersionInput{
-				PolicyArn: aws.String(policyArn),
-				VersionId: version.VersionId,
-			})
-		}
+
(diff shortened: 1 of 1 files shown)


oriAdler:
@tbrisker @ydayagi @hunterkepley PTAL

hunterkepley (commented):
lgtm

pkg/aws/policies.go
```diff
@@ -889,10 +879,37 @@ func (c *awsClient) DeleteInlineRolePolicies(role string) error {
...
+	if err != nil {
+		return false, err
+	}
+
+	isAttached := false
+	if *policyOutput.Policy.AttachmentCount > 0 {
+		isAttached = true
+	}
+
+	return isAttached, nil
```
tbrisker, line 893:
```suggestion
	if *policyOutput.Policy.AttachmentCount > 0 {
		return true, nil
	}

	return false, nil
```

pkg/aws/policies.go
```diff
@@ -889,10 +879,37 @@ func (c *awsClient) DeleteInlineRolePolicies(role string) error {
...
+}
+
 func (c *awsClient) deletePolicies(policies []string) (*iam.DeletePolicyOutput, error) {
 	var output *iam.DeletePolicyOutput
-	var err error
+
 	for i := range policies {
```
tbrisker, line 899: np, why not:
```suggestion
	for _, policy := range policies {
```

tbrisker (commented):
looks good in general, a couple of minor comments inline

pkg/aws/policies.go
oriAdler: Changed, thanks.

pkg/aws/policies.go
oriAdler: Changed.

tbrisker: ah, I guess that's why it was using `i`, sorry for the noise:
```
 pkg/aws/policies.go:912:76: G601: Implicit memory aliasing in for loop. (gosec)
		output, err = c.iamClient.DeletePolicy(&iam.DeletePolicyInput{PolicyArn: &policy})
		                                                                         ^
make: *** [lint] Error 1

pkg/aws/policies.go
oriAdler: Changed, thanks.

PR #830, opened 2022-09-11
[SDA-6405] feat: add version and channelGroup args to create/update acc. roles

gdbranco:
Allows version of policy to be chosen from available openshift versions

Before using command `rosa create account-roles` it autos to the latest default version, in this case 4.11
image

After using command `rosa create account-roles --version=4.10` the roles are created tagged with the chosen version. `--channel-group` could also be specified if a new version gets released in a specific channel, but it makes no difference for the example currently as the possible versions in the 'major.minor' format are the same.
image

### DIFF
--- a/pkg/ocm/helpers.go
+++ b/pkg/ocm/helpers.go
@@ -683,3 +683,77 @@ func ValidateAvailabilityZonesCount(multiAZ bool, availabilityZonesCount int) er
 
 	return nil
 }
+
+func (c *Client) GetVersion(userRequestedVersion string, channelGroup string) (string, error) {
+	versionList, err := c.GetVersionsList(channelGroup)
+	if err != nil {
+		err := fmt.Errorf("%v", err)
+		return userRequestedVersion, err
+	}
+
+	if userRequestedVersion == "" {
+		return versionList[0], nil
+	}
+
+	hasVersion := false
+	for _, vs := range versionList {
+		if vs == userRequestedVersion {
+			hasVersion = true
+			break
+		}
+	}
+
+	if !hasVersion {
+		versionSet := make(map[string]bool)
+		for _, vs := range versionList {
+			versionSet[vs] = true
+		}
+		err := fmt.Errorf("a valid version number must be specified\nValid versions: %v", KeysString(versionSet))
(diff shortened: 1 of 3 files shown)


gdbranco:
@vkareh @ciaranRoche @tbrisker PTAL

pkg/ocm/helpers.go
```diff
@@ -655,3 +656,77 @@ func ValidateAvailabilityZonesCount(multiAZ bool, availabilityZonesCount int) er
 
 	return nil
 }
+
+func (c *Client) GetVersion(userRequestedVersion string, channelGroup string) (string, error) {
```
pvasant, line 660: @gdbranco can we reuse the existing methods

pkg/ocm/helpers.go
```diff
@@ -655,3 +656,77 @@ func ValidateAvailabilityZonesCount(multiAZ bool, availabilityZonesCount int) er
...
+			break
+		}
+	}
+
+	if !hasVersion {
+		versionSet := make(map[string]bool)
+		for _, vs := range versionList {
+			versionSet[vs] = true
+		}
+		err := fmt.Errorf("a valid version number must be specified\nValid versions: %v", KeysString(versionSet))
```
pvasant, line 684: i dont know how many valid versions we will get to list . can i see the sample

gdbranco: After filtering everything the final list will show up as something like this:
`Valid versions: [4.10, 4.9, 4.8, 4.7, 4.11]`

As we are mapping versions to 'major.minor' and setting up a set of them, but the original response depends on the channel group. Nightly has a format like` "openshift-v4.11.0-0.nightly-2022-09-15-062234-nightly"` and around 40 entries. But stable has more than 100

pkg/ocm/helpers.go
gdbranco: Current methods return versions with a different format, for the policy version the create/upgrade acc. roles command was already filtering for 'major.minor' format through custom methods within each file, but I decided to unify those in the helpers.go file just to make it more concise.

gdbranco:
@pvasant @igoihman PTAL

PR #833, opened 2022-09-14
Create cluster - list region filtered by OCP version

oriAdler:
Use the `aws_inquiries` endpoint and pass the `version` to fetch the supported regions.

Related: [SDA-6765](https://issues.redhat.com/browse/SDA-6765)

image

### DIFF
--- a/pkg/ocm/regions.go
+++ b/pkg/ocm/regions.go
@@ -26,6 +26,50 @@ import (
 	"github.com/openshift/rosa/pkg/logging"
 )
 
+func (c *Client) GetFilteredRegionsByVersion(roleARN string, version string,
+	awsClient aws.Client, externalID string) (regions []*cmv1.CloudRegion, err error) {
+	cloudProviderDataBuilder, err := c.createCloudProviderDataBuilder(roleARN, awsClient, externalID)
+	if err != nil {
+		return []*cmv1.CloudRegion{}, err
+	}
+	if version != "" {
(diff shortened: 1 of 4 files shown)


oriAdler:
@zgalor @pvasant @ciaranRoche PTAL

pkg/ocm/regions.go
```diff
@@ -26,6 +26,46 @@ import (
...
+func (c *Client) GetFilteredRegionsByVersion(roleARN string, version string,
+	awsClient aws.Client, externalID string) (regions []*cmv1.CloudRegion, err error) {
+	cloudProviderDataBuilder, err := c.createCloudProviderDataBuilder(roleARN, awsClient, externalID)
+	if err != nil {
+		return []*cmv1.CloudRegion{}, err
+	}
+	cloudProviderData, err := cloudProviderDataBuilder.Version(cmv1.NewVersion().ID(version)).Build()
+	if err != nil {
+		return []*cmv1.CloudRegion{}, err
+	}
```
zgalor, line 38:
```
if version != "" {
	cloudProviderData, err := cloudProviderDataBuilder.Version(cmv1.NewVersion().ID(version)).Build()
	if err != nil {
		return []*cmv1.CloudRegion{}, err
	}
}
```

zgalor: I think this may not be required because version is never empty here

PR #834, opened 2022-09-14
Improve `EnsurePolicy` error message

oriAdler:
Related: [SDA-6606](https://issues.redhat.com/browse/SDA-6606)

image

### DIFF
--- a/pkg/aws/policies.go
+++ b/pkg/aws/policies.go
@@ -244,20 +244,21 @@ func (c *awsClient) EnsurePolicy(policyArn string, document string,
 	version string, tagList map[string]string, path string) (string, error) {
 	output, err := c.IsPolicyExists(policyArn)
 	if err != nil {
+		var policyArnLocal string
 		if aerr, ok := err.(awserr.Error); ok {
 			switch aerr.Code() {
 			case iam.ErrCodeNoSuchEntityException:
-				policyArn, err = c.createPolicy(policyArn, document, tagList, path)
+				policyArnLocal, err = c.createPolicy(policyArn, document, tagList, path)
 				if err != nil {
 					if aerr, ok := err.(awserr.Error); ok {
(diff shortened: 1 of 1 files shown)


oriAdler:
@pvasant @ydayagi PTAL

pkg/aws/policies.go
```diff
@@ -244,20 +244,21 @@ func (c *awsClient) EnsurePolicy(policyArn string, document string,
...
 			case iam.ErrCodeNoSuchEntityException:
-				policyArn, err = c.createPolicy(policyArn, document, tagList, path)
+				policyArnLocal, err = c.createPolicy(policyArn, document, tagList, path)
 				if err != nil {
 					if aerr, ok := err.(awserr.Error); ok {
 						if aerr.Code() == iam.ErrCodeEntityAlreadyExistsException {
-							return "", fmt.Errorf(
-								"Policy with same name but different ARN exists. Probably wrong path")
+							return "", errors.Wrapf(err,
+								fmt.Sprintf("Failed to create a policy with ARN %s", policyArn))
```
ydayagi, line 256: the original message is more informative and in reality alerts about the real problem. Providing the ARN in the error IMHO does not help the user.

ydayagi: I also don't see how this error relates to the message in the PR description.

oriAdler: The second part of the error message `EntityAlreadyExists` is from the AWS SDK.

My concern with `Probably wrong path` is that maybe there is another scenario that can lead to this error,
in case of a user is not aware of the `path` feature arriving here that might be confusing.

The user sees the policy name and can look it up in the AWS console, and the policy ARN they are trying to create.

oriAdler: How can we provide more info on this error message to make it more informative?

ydayagi: the caller to ensurePolcy already adds (or should add) more details. so policy name will be there.

PR #836, opened 2022-09-15
set mode only once in operator roles upgrade

ydayagi:
[sda-6745](https://issues.redhat.com//browse/sda-6745)

### DIFF
--- a/cmd/upgrade/operatorroles/cmd.go
+++ b/cmd/upgrade/operatorroles/cmd.go
@@ -214,12 +214,13 @@ func run(cmd *cobra.Command, argv []string) error {
 		os.Exit(1)
 	}
 
+	mode, err = handleModeFlag(cmd, skipInteractive, mode, err, r.Reporter)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/upgrade/operatorroles/cmd.go
```diff
@@ -214,12 +214,13 @@ func run(cmd *cobra.Command, argv []string) error {
 		os.Exit(1)
 	}
 
+	mode, err = handleModeFlag(cmd, skipInteractive, mode, err, r.Reporter)
```
oriAdler, line 217: Do we need to wrap it with the following condition?
`if isOperatorPolicyUpgradeNeeded || len(missingRolesInCS) > 0 {`

ydayagi: the condition is already checked earlier and we return if it is not satisfied

oriAdler: Oh right, missed that, thanks.

PR #837, opened 2022-09-15
[Hypershift] Modify `describe cluster` to differentiate between classic vs hosted-cp

samira-barouti:
* Relates [SDA-6797](https://issues.redhat.com//browse/SDA-6797)

### DIFF
--- a/cmd/describe/cluster/cmd.go
+++ b/cmd/describe/cluster/cmd.go
@@ -122,56 +122,6 @@ func run(cmd *cobra.Command, argv []string) {
 
 	detailsPage := getDetailsLink(r.OCMClient.GetConnectionURL())
 
-	// Display number of all worker nodes across the cluster
-	minNodes := 0
-	maxNodes := 0
-	var nodesStr string
-	machinePools, err := r.OCMClient.GetMachinePools(cluster.ID())
-	if err != nil {
-		r.Reporter.Errorf("Failed to get machine pools for cluster '%s': %v", clusterKey, err)
-		os.Exit(1)
-	}
-	// Accumulate all replicas across machine pools
-	for _, machinePool := range machinePools {
-		if machinePool.Autoscaling() != nil {
-			minNodes += machinePool.Autoscaling().MinReplicas()
-			maxNodes += machinePool.Autoscaling().MaxReplicas()
-		} else {
-			minNodes += machinePool.Replicas()
-			maxNodes += machinePool.Replicas()
-		}
-	}
-	// Add compute nodes as well
-	if cluster.Nodes().AutoscaleCompute() != nil {
-		minNodes += cluster.Nodes().AutoscaleCompute().MinReplicas()
-		maxNodes += cluster.Nodes().AutoscaleCompute().MaxReplicas()
-	} else {
-		minNodes += cluster.Nodes().Compute()
-		maxNodes += cluster.Nodes().Compute()
-	}
-	// Determine whether there is any auto-scaling in the cluster
-	if minNodes == maxNodes {
-		nodesStr = fmt.Sprintf(""+
-			"Nodes:\n"+
-			" - Control plane:           %d\n"+
-			" - Infra:                   %d\n"+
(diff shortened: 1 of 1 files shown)


samira-barouti:
@pvasant @vkareh @igoihman could you PTAL?

pkg/ocm/client.go
```diff
@@ -124,6 +124,10 @@ func (b *ClientBuilder) Build() (result *Client, err error) {
 	if b.cfg.URL != "" {
 		builder.URL(b.cfg.URL)
 	}
+	// when running ROSA against local CS, redirect AMS API calls to integration
+	if strings.Contains(b.cfg.URL, "localhost") {
+		builder.AlternativeURL("/api/accounts_mgmt", URLAliases["integration"])
```
pvasant, line 129: can we remove this . this is a dev config

cmd/describe/cluster/cmd.go
```diff
@@ -335,6 +286,74 @@ func run(cmd *cobra.Command, argv []string) {
...
+	return "Customer hosted"
+}
+
+func clusterInfraConfig(cluster *cmv1.Cluster, clusterKey string, r *rosa.Runtime) string {
+	var infraConfig string
+	if cluster.Hypershift().Enabled() {
+		infraConfig = fmt.Sprintf(""+
+			"Node Pools:\n"+
+			" - Compute:                 %d\n",
+			cluster.Nodes().Compute(),
```
pvasant, line 302: I think we still have to do everything below just don't display the master and infra. we need autoscaling and replicas same as regular rosa sts

samira-barouti: CS does not persist the nodepool info in the database; there is no reconciliation on NodePool CR to determine the replica size as well. For the first iteration, we can only show the number of worker nodes (as it is stored in the db). The second iteration will add more info to this field.

pkg/ocm/client.go
samira-barouti: I put this on purpose; w/o this line, running the local ROSA against the local CS is not possible.

pvasant: I think we can document and take it out from the repo . wdyt?

samira-barouti: That makes sense. I'll remove it.

cmd/describe/cluster/cmd.go
vkareh: we need to persist the nodepool info, otherwise this becomes a problem for tracking quota consumption in AMS and any cluster reshape will drift the cluster authorizations.

cmd/describe/cluster/cmd.go
```diff
@@ -335,6 +286,74 @@ func run(cmd *cobra.Command, argv []string) {
...
+		} else {
+			minNodes += cluster.Nodes().Compute()
+			maxNodes += cluster.Nodes().Compute()
+		}
+		// Determine whether there is any auto-scaling in the cluster
+		if minNodes == maxNodes {
+			infraConfig = fmt.Sprintf(""+
+				"Nodes:\n"+
+				" - Control plane:           %d\n"+
+				" - Infra:                   %d\n"+
```
vkareh, line 336: I think these components can be hidden, but everything else should still be in place (but instead of `GetMachinePools` you should call `GetNodePools` or the equivalent function). No need for a large `if hypershift {...} else {...}`

cmd/describe/cluster/cmd.go
samira-barouti: Right, we have two cards to address this: one to persist nodepool in db (that I am working on); the other is to consume NodePool status to fetch the autoscaling info. For the first iteration, we just show the node count.

cmd/describe/cluster/cmd.go
samira-barouti: The key is also different; there is no `Nodes` but `Node Pools` instead. We need to refactor this once the nodepool is in db, and nodepool controller is in place.

hunterkepley:
LGTM

PR #840, opened 2022-09-18
Upgrade cluster - manual mode - prompt the aws operator role upgrade commands

oriAdler:
Related: [SDA-6744](https://issues.redhat.com/browse/SDA-6744)


oriAdler:
@pvasant @zgalor @ydayagi PTAL

The `upgrade account roles` command is prompted from the `upgrade cluster`.
If we exit in `upgrade operator roles` instead of returning to the `upgrade cluster`,
we miss the call to `upgrade operator roles`.

PR #841, opened 2022-09-18
logging: Disable quoting of values - makes request/response dumps readable

cben:
This stops the output from being machine-parsable :-( but makes `rosa --debug` multi-line JSON dumps way more readable by (1) actually printing them multi-line instead of `\n` (2) letting string quotes be `"` not `\"`. Makes is similar to `ocm --debug` :man_dancing: 

Before:
```
time="2022-09-18T15:49:55+03:00" level=debug msg="Response body follows"
time="2022-09-18T15:49:55+03:00" level=debug msg="{\n  \"kind\": \"MachineTypeList\",\n  \"page\": 1,\n  \"size\": 100,\n  \"total\": 241,\n  \"items\": [\n    {\n      \"kind\": \"MachineType\",\n      \"name\": \"p3.16xlarge - Accelerated Computing (8 GPUs)\",\n      \"category\": \"accelerated_computing\",\n      \"size\": \"xxxlarge\",\n      \"id\": \"p3.16xlarge\",\n      \"href\": \"/api/clusters_mgmt/v1/machine_types/p3.16xlarge\",\n      \"memory\": {\n        \"value\": 523986010112,\n        \"unit\": \"B\"\n      },\n
...
```

After:
```
time=2022-09-18T15:47:52+03:00 level=debug msg=Response body follows
time=2022-09-18T15:47:52+03:00 level=debug msg={
  "kind": "MachineTypeList",
  "page": 1,
  "size": 100,
  "total": 241,
  "items": [
    {
      "kind": "MachineType",
      "name": "p3.16xlarge - Accelerated Computing (8 GPUs)",
...
```


cben:
Is it OK to just do this unconditionally, or does this need some config knob?

(AFAICT all messages only use time, level, msg fields, so it's not like you are benefitting much from the promise of "structured logging" being machine-parsable.  If you do want a parsable mode, consider JsonFormatter?)

cben:
/remove-lifecycle stale

I still believe this improves the usability of --debug.  @oriAdler @zgalor WDYT?

oriAdler:
@cben I agree it improves the output readability.
I'm not sure if we should provide a config knob to enable `machine-parsable` output.
cc @vkareh

vkareh:
I think this should be fine to do unconditionally. Do we use this logging for anything other than `--debug`? If so, are there any edge cases (I think there are none).

cben:
/remove-lifecycle stale

So, what's the verdict?  Can someone merge this?

I'm not aware of anybody relying on the log format, I just asking hypothetically in case the original format was a conscious choice...  aka overthinking it :laughing:

PR #842, opened 2022-09-19
rosa list instance-types: Avoid division by zero

cben:
Zero cost is the norm for ROSA quotas, and int/int division by zero would cause a Go panic :scream_cat:.  

* Float division is not an option.  x/0.0 de-facto returns +Inf but formally [is implemetation defined](https://github.com/golang/go/issues/43577); also 0.0/0.0 returns NaN, and that's what we'll frequently see dealing with cost 0. And anyway we want int/int division for rounding down.  So kept int/int with a guard condition.

I guess in practice this hasn't exploded because this whole computation was _only_ done on GPU types (skipped if `Category() != AcceleratedComputing`, and we ask backend only for quota_cost where `quota_id~='gpu'`)?  
UPDATE: this did explode on recent addition of cost=0 GPU quota (https://issues.redhat.com/browse/SDA-8154) :boom: 

[This is a step towards being able to drop this special-casing of GPU :pray: but out of scope here.]

### DIFF
--- a/pkg/ocm/machines.go
+++ b/pkg/ocm/machines.go
@@ -19,6 +19,7 @@ package ocm
 import (
 	"errors"
 	"fmt"
+	"math"
 	"strings"
 
 	amsv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
@@ -106,13 +107,13 @@ func getDefaultNodes(multiAZ bool) int {
 }
 
 type MachineType struct {
-	MachineType    *cmv1.MachineType
-	Available      bool
-	availableQuota int
+	MachineType *cmv1.MachineType
+	Available   bool
+	nodeQuota   float64 // may be +Inf when Cost == 0.
 }
 
 func (mt MachineType) HasQuota(multiAZ bool) bool {
-	return mt.MachineType.Category() != AcceleratedComputing || mt.availableQuota > getDefaultNodes(multiAZ)
+	return mt.MachineType.Category() != AcceleratedComputing || mt.nodeQuota > float64(getDefaultNodes(multiAZ))
 }
 
 // GetAvailableMachineTypesInRegion get the supported machine type in the region.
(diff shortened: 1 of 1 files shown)


pkg/ocm/machines.go
```diff
@@ -224,14 +225,17 @@ func (mtl *MachineTypeList) UpdateAvailableQuota(quotaCosts *amsv1.QuotaCostList
...
 			for _, relatedResource := range quotaCost.RelatedResources() {
 				if machineType.MachineType.GenericName() == relatedResource.ResourceName() && isCompatible(relatedResource) {
-					availableQuota := (quotaCost.Allowed() - quotaCost.Consumed()) / relatedResource.Cost()
-					machineType.Available = availableQuota > 1
-					machineType.availableQuota = availableQuota
+					if relatedResource.Cost() == 0 {
+						machineType.nodeQuota = math.Inf(+1)
+					} else {
+						machineType.nodeQuota += float64((quotaCost.Allowed() - quotaCost.Consumed()) / relatedResource.Cost())
+					}
```
tbrisker, line 232: Wouldn't it be better to avoid the float math and keep it as ints? since we can't have 1.4 nodes anyways, and i'm worried about getting something like 1.999999999 due to quirky float arithmetic edge cases.
```suggestion
					if relatedResource.Cost() == 0 {
						machineType.nodeQuota = math.MaxInt
					} else {
						machineType.nodeQuota += (quotaCost.Allowed() - quotaCost.Consumed()) / relatedResource.Cost()
					}
```

cben: Note that the division itself remained int/int, rounding down; it's cast to a float afterwards.

I'm using floats for their ability to represent infinity, which fits checks like `nodeQuota > getDefaultNodes(multiAZ)` very naturally.  
(I've used a similar +Inf representation widely in the UI and consider it a success.)

Hmm yes, `MaxInt` could play a similar role in practice.  I find it less "elegant" but it's workable.  Do you want me to change it?

gdbranco:
Instead of doing all that would it be possible to add an extra condition of resource cost 0 as we check category `Category() != AcceleratedComputing`

cben:
@tbrisker @gdbranco pushed a version using integers / MaxInt, and special-casing cost=0 which I _think_ works.  
And added a lot of TODOs for other issues.

I'm not really happy with it.  See, there are 2 levels of aggregation:
* `RelatedResources` are alternatives — normally at most one will match, but technically if several match we should pick the minimum cost => max quota.
* separate `QuotaCostItems` are additive — we sum them with `+=`.

A "real" float Infinity had all the right properties for such aggregations: max(Inf, x) = Inf;  Inf + 1 = Inf + Inf = Inf.  
Whereas MaxInt is fragile: MaxInt + 1 already overflows to a negative number.  So have to take care to bail out of the loop to avoid `+=` after finding a cost=0 match :worried: 

----

I also realized the `.Available` boolean is ill-defined.  I guess the intent of its `quota > 1` computation is `>= 2` because 2 is the lowest minimumNodes for _some_ cluster.  

I'll add some cleanup commits removing it (but the special-casing of GPU types will remain, out of scope for this PR).

PR #843, opened 2022-09-20
Removing unnecessary hypershift check for managed services.

renan-campos:
https://issues.redhat.com/browse/SDA-6835


renan-campos:
@ciaranRoche ptal

PR #845, opened 2022-09-21
Fetch all regions for non-interactive mode

oriAdler:
For non-interactive mode, the CLI doesn't display the regions to the user, 
therefore fetching all regions to get the correct error message eventually.

Related: [SDA-6837](https://issues.redhat.com/browse/SDA-6837)

image


oriAdler:
@hunterkepley @zgalor PTAL

PR #846, opened 2022-09-21
OHSS-14855: Making sure extracting the current operator roles prefix does not crash when upgrading the roles

Nikokolas3270:
A default prefix (based on the cluster name) will be created if no prefix can be extracted from the existing operator roles.

This change has been discussed here:
https://coreos.slack.com/archives/CB53T9ZHQ/p1663701584121839?thread_ts=1662139356.619269&cid=CB53T9ZHQ

- If the existing operator roles are prefixed: the code is still re-using their prefix for the operator role(s) to create during upgrade.
- If the existing operator roles are not prefixed:
  - Current code is crashing, this is the issue we have in [OHSS-14855](https://issues.redhat.com/browse/OHSS-14855).
  - With the new proposed code:
    - Command `rosa upgrade operator-roles` will crash no more.
    - The new operator roles to create will be created with a default prefix.
    - So the operator roles will then not all follow the same naming convention (as the old roles will be unprefixed)... but that's NOT an issue as the only place where this naming convention is used & reversed engineered (which is a bad practice) is precisely when the `rosa upgrade operator-roles` command is trying to extract the prefix.

### DIFF
--- a/cmd/upgrade/operatorroles/cmd.go
+++ b/cmd/upgrade/operatorroles/cmd.go
@@ -442,16 +442,17 @@ func upgradeMissingOperatorRole(missingRoles map[string]*cmv1.STSOperator, clust
 
 func getRoleName(cluster *cmv1.Cluster, missingOperator *cmv1.STSOperator) string {
 	operatorIAMRoles := cluster.AWS().STS().OperatorIAMRoles()
-	rolePrefix := ""
-	if len(operatorIAMRoles) > 0 {
-		roleARN := operatorIAMRoles[0].RoleARN()
+	rolePrefix := ocm.ComputeDefaultOperatorRolesPrefix(cluster.Name())
+	for _, operatorIAMRole := range operatorIAMRoles {
+		roleARN := operatorIAMRole.RoleARN()
 		roleName, err := aws.GetResourceIdFromARN(roleARN)
-		if err != nil {
-			return ""
+		if err == nil {
+			m := strings.LastIndex(roleName, "-openshift")
+			if m != -1 {
+				rolePrefix = roleName[0:m]
+				break
+			}
 		}
-
-		m := strings.LastIndex(roleName, "-openshift")
-		rolePrefix = roleName[0:m]
 	}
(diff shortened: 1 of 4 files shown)


Nikokolas3270:
Forgot to say that there is not really a decent workaround to this problem.
We could:
1. Temporarily prefix the first role (when the name part of the ARN actually) handled by the `operator_iam_roles` attribute of the OCM cluster object
2. Then run the `rosa upgrade operator-roles` command which will reuse the introduced prefix for the new roles
3. And then revert the change done in step 1 (to not introduce a discrepancy between the roles in `operator_iam_roles` attribute and the real AWS roles)

... but this is a bit of a far-fetched workaround.

rafael-azevedo:
@jharrington22 anything we can do to get this merged and hold removed ?

cmd/upgrade/operatorroles/cmd.go
```diff
@@ -442,16 +442,16 @@ func upgradeMissingOperatorRole(missingRoles map[string]*cmv1.STSOperator, clust
 
 func getRoleName(cluster *cmv1.Cluster, missingOperator *cmv1.STSOperator) string {
 	operatorIAMRoles := cluster.AWS().STS().OperatorIAMRoles()
-	rolePrefix := ""
+	rolePrefix := ocm.ComputeDefaultOperatorRolesPrefix(cluster.Name())
```
pvasant, line 445: We cannot do this

Nikokolas3270: @pvasant: Could you please detail a bit more why we cannot do this?
We are only defining the default prefix here.
The nominal case is still to extract the prefix from the operator role name in the code that follows. If the prefix cannot be extracted, the code is currently ending in exception while with my change the code will continue to run normally and create the new operator roles shipped in the upgrade with the default prefix.

gdbranco:
@Nikokolas3270 Could you close this PR?

Nikokolas3270:
Hello,

This PR has been replaced by the following PR: https://github.com/openshift/rosa/pull/897

A new command has been added to upgrade both cluster roles and operator roles at once:
`rosa upgrade roles`

PR #847, opened 2022-09-21
[SDA-6780] fix: compatibility issue older rosa cli

gdbranco:
# What
Creating an admin with rosa cli 1.1.5 then deleting it with 1.2.6 was not deleting correctly and newly created admin with 1.2.6 would cause 500 response codes on `oc login`.
Related issue: https://issues.redhat.com/browse/SDA-6780

# Why
This causes frustration for clients as they needed to have a second idp with admin access to delete the resources through `oc`.

# Cause
The entire identity provider was being deleted when created from older clis, but a condition was still trying to delete it a second time. I'm not too familiar as to why the second call didn't give back an error or why it caused the 500 response codes, maybe we should look into the delete route of idp for ocm.

# Steps before
`rosa115 create admin -c htpwd-test`
`rosa126 delete admin -c htpwd-test`
`rosa126 create admin -c htpwd-test`
`oc login <url> --username cluster-admin --password <pwd>`
```
Error from server (InternalError): Internal error occurred: unexpected response: 500
```

# Steps after
`rosa115 create admin -c htpwd-test`
`rosaWithChanges delete admin -c htpwd-test`
`rosaWithChanges create admin -c htpwd-test`
`oc login <url> --username cluster-admin --password <pwd>`
```
Login successful.
```

I was also able to delete from the error state into creating a new admin and login just fine.

### DIFF
--- a/cmd/dlt/admin/cmd.go
+++ b/cmd/dlt/admin/cmd.go
@@ -45,85 +45,91 @@ func run(cmd *cobra.Command, _ []string) {
 	r := rosa.NewRuntime().WithAWS().WithOCM()
 	defer r.Cleanup()
 
-	clusterKey := r.GetClusterKey()
-
 	cluster := r.FetchCluster()
 	if cluster.State() != cmv1.ClusterStateReady {
-		r.Reporter.Errorf("Cluster '%s' is not yet ready", clusterKey)
+		r.Reporter.Errorf("Cluster '%s' is not yet ready", r.ClusterKey)
 		os.Exit(1)
 	}
 
 	// Try to find the htpasswd identity provider:
 	r.Reporter.Debugf("Loading HTPasswd identity provider")
-	idps, err := r.OCMClient.GetIdentityProviders(cluster.ID())
+	clusterID := cluster.ID()
+	idps, err := r.OCMClient.GetIdentityProviders(clusterID)
 	if err != nil {
-		r.Reporter.Errorf("Failed to get HTPasswd identity provider for cluster '%s': %v", clusterKey, err)
+		r.Reporter.Errorf("Failed to get HTPasswd identity provider for cluster '%s': %v", r.ClusterKey, err)
 		os.Exit(1)
 	}
 
-	var htpasswdIDP *cmv1.IdentityProvider
+	var identityProvider *cmv1.IdentityProvider
 	for _, item := range idps {
 		if ocm.IdentityProviderType(item) == ocm.HTPasswdIDPType {
-			htpasswdIDP = item
+			identityProvider = item
 		}
 	}
-	if htpasswdIDP == nil {
-		r.Reporter.Errorf("Cluster '%s' does not have an admin user", clusterKey)
+	if identityProvider == nil {
+		r.Reporter.Errorf("Cluster '%s' does not have an admin user", r.ClusterKey)
 		os.Exit(1)
 	}
 
-	if confirm.Confirm("delete %s user on cluster %s", idp.ClusterAdminUsername, clusterKey) {
+	if confirm.Confirm("delete %s user on cluster %s", idp.ClusterAdminUsername, r.ClusterKey) {
 		// delete `cluster-admin` user from the HTPasswd IDP
-		r.Reporter.Debugf("Deleting user '%s' identity provider on cluster '%s'", idp.ClusterAdminUsername, clusterKey)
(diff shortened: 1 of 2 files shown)


gdbranco:
@pvasant @zvikorn @zgalor @oriAdler @samira-barouti 

Hey you guys, I'm tagging you as some have changed the files/reviewed original issues for this

cmd/dlt/admin/cmd.go
```diff
@@ -45,85 +45,93 @@ func run(cmd *cobra.Command, _ []string) {
...
-			r.Reporter.Errorf("Failed to get htpasswd idp of cluster '%s': %s",
-				clusterKey, err)
+			r.Reporter.Errorf("Failed to get htpasswd idp for cluster '%s'", clusterID)
 			os.Exit(1)
 		}
-
-		if users.Len() == 0 && htpasswdIdentityProvider.Username() == "" {
-			//delete the idp as users list is empty
-			err := r.OCMClient.DeleteIdentityProvider(cluster.ID(), htpasswdIDP.ID())
+		if htpasswdIdp.Username() == "cluster-admin" {
```
gdbranco, line 90: Not liking this if else, will take a look at a strategy pattern in golang to set this up

cmd/dlt/admin/cmd.go
```diff
@@ -45,85 +45,93 @@ func run(cmd *cobra.Command, _ []string) {
...
 	if cluster.State() != cmv1.ClusterStateReady {
-		r.Reporter.Errorf("Cluster '%s' is not yet ready", clusterKey)
+		r.Reporter.Errorf("Cluster '%s' is not yet ready", r.ClusterKey)
 		os.Exit(1)
 	}
 
 	// Try to find the htpasswd identity provider:
 	r.Reporter.Debugf("Loading HTPasswd identity provider")
-	idps, err := r.OCMClient.GetIdentityProviders(cluster.ID())
+	clusterID := cluster.ID()
```
zgalor, line 56: NP: just wondering if it's not an optimization the compiler makes anyway

cmd/dlt/admin/cmd.go
zgalor: That logic L85-90 better be encapsulated inside a func `wasAdminCreatedUsingOldRosa()`
That condition inside it might change, but shouldn't impact how we handle it

cmd/dlt/admin/cmd.go
gdbranco: hmm, probably. But it was more to concentrate it in a single place in case we ever change something

cmd/dlt/admin/cmd.go
```diff
@@ -45,85 +99,58 @@ func run(cmd *cobra.Command, _ []string) {
...
-		}
 
-		users, err = r.OCMClient.GetHTPasswdUserList(cluster.ID(), htpasswdIDP.ID())
-		if err != nil {
-			r.Reporter.Errorf("Failed to list htpasswd idp users of cluster '%s': %s",
-				clusterKey, err)
-			os.Exit(1)
+		var deletionStrategy DeleteAdminUserStrategy = &DeleteUserAdminFromIDP{}
+		if wasAdminCreatedUsingOldROSA(r, identityProvider) {
+			deletionStrategy = &DeleteAdminIDP{}
```
samira-barouti, line 141: I suggest extracting lines 139-141 to a separate method, and name it `adminUserDeletionStrategy()`

cmd/dlt/admin/cmd.go
```diff
@@ -45,85 +99,58 @@ func run(cmd *cobra.Command, _ []string) {
...
-			}
-		}
-		r.Reporter.Infof("Admin user '%s' has been deleted from cluster '%s'", idp.ClusterAdminUsername, clusterKey)
+func wasAdminCreatedUsingOldROSA(r *rosa.Runtime, identityProvider *cmv1.IdentityProvider) bool {
+	htpasswdIdp, ok := identityProvider.GetHtpasswd()
+	if !ok {
+		r.Reporter.Errorf("Failed to get htpasswd idp for cluster '%s'", r.Cluster.ID())
+		os.Exit(1)
 	}
+	return htpasswdIdp.Username() == "cluster-admin"
```
samira-barouti, line 155: do we define the literal `cluster-admin` somewhere? we should probably use that.

PR #850, opened 2022-09-26
check quota available before presenting options

laurafitzgerald:
https://issues.redhat.com/browse/SDA-6854

Changes include a function to get the quota_cost by addonname aka `addon|addon-<addononname>`
Check the orgs quota is not 0 before allowing install to progress
Check the quota when param is of type resource, only offer options that there is quota for 
on edit this includes the value of the installation being edited.

### DIFF
--- a/cmd/edit/addon/cmd.go
+++ b/cmd/edit/addon/cmd.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strconv"
 	"strings"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
@@ -75,6 +76,12 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
+	quotaCost, err := r.OCMClient.GetAddonParameterQuotaCost(addOnID)
+	if err != nil {
+		r.Reporter.Warnf("%s", err)
+		os.Exit(0)
+	}
+
 	parameters, err := r.OCMClient.GetAddOnParameters(cluster.ID(), addOnID)
 	if err != nil {
 		r.Reporter.Errorf("Failed to get add-on '%s' parameters: %v", addOnID, err)
@@ -86,7 +93,6 @@ func run(cmd *cobra.Command, argv []string) {
 		r.Reporter.Errorf("Failed to get add-on '%s' installation: %v", addOnID, err)
 		os.Exit(1)
 	}
-
 	if parameters.Len() == 0 {
 		r.Reporter.Errorf("Add-on '%s' has no parameters to edit", addOnID)
 		os.Exit(1)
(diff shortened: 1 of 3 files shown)


cmd/edit/addon/cmd.go
```diff
@@ -75,6 +76,12 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
+	quotaCost, err := r.OCMClient.GetAddonParameterQuotaCost(addOnID)
+	if err != nil {
+		r.Reporter.Warnf("%s", err)
+		os.Exit(0)
```
vkareh, line 82: If you're exiting due to an error it should be
```
		r.Reporter.Errorf("%s", err)
		os.Exit(1)
```

cmd/install/addon/cmd.go
```diff
@@ -117,6 +118,17 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
+	quotaCost, err := r.OCMClient.GetAddonParameterQuotaCost(addOnID)
+	if err != nil {
+		r.Reporter.Warnf("%s", err)
+		os.Exit(0)
```
vkareh, line 124: Same here

cmd/edit/addon/cmd.go
```diff
@@ -125,16 +131,35 @@ func run(cmd *cobra.Command, argv []string) {
...
+			availableQuota := quotaCost.Allowed() - quotaCost.Consumed() + paramValueInt
+			valueInt, err := strconv.Atoi(opt.Value())
+			if err != nil {
+				r.Reporter.Errorf("%s", err)
+				os.Exit(1)
+			}
+			if valueInt <= availableQuota {
+				options = append(options, opt.Name())
+				values = append(values, opt.Value())
+			}
```
ciaranRoche, line 162: The more I am thinking about this, we are assuming the quota does not change between when we fetch the quota and the params, versus when the user actually sends the request. 

I think we should maintain the current logic, give the user all options and allow them to make the request, and let AMS make the decision to reject or accept it. 

WDYT

ciaranRoche (commented):
Lets set up a call and have a chat about this ask from QE?

cmd/install/addon/cmd.go
```diff
@@ -117,6 +118,17 @@ func run(cmd *cobra.Command, argv []string) {
...
+	}
+	availableQuota := quotaCost.Allowed() - quotaCost.Consumed()
+	if availableQuota < 1 {
+		r.Reporter.Warnf("Your organisation account does not have any quota to install the add on %s", addOnID)
+		os.Exit(0)
+	}
```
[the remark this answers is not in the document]
vkareh: At the very least, this decision should be done inside `pkg/ocm` in a "HasAddOnQuota" function, not in the `cmd/`, but I agree, maybe we need a "has quota" endpoint in AMS.

@ciaranRoche do you think we can we use the addon inquiries endpoint for this?

PR #853, opened 2022-09-27
[SDA-4764] Hide region arg in account roles commands

gdbranco:
Related issue: [SDA-4764](https://issues.redhat.com/browse/SDA-4764)

# What
Account roles region arg should be hidden

# Why
Account roles commands don't need to specify region as they are global

# Before changes
```
Global Flags:
      --color string     Surround certain characters with escape sequences to display them in color on the terminal. Allowed options are [auto never always] (default "auto")
      --debug            Enable debug mode.
      --profile string   Use a specific AWS profile from your credential file.
      --region string    Use a specific AWS region, overriding the AWS_REGION environment variable.
  -y, --yes              Automatically answer yes to confirm operation.
```

# After changes
```
Global Flags:
      --color string     Surround certain characters with escape sequences to display them in color on the terminal. Allowed options are [auto never always] (default "auto")
      --debug            Enable debug mode.
      --profile string   Use a specific AWS profile from your credential file.
  -y, --yes              Automatically answer yes to confirm operation.
```

### DIFF
--- a/cmd/create/cmd.go
+++ b/cmd/create/cmd.go
@@ -59,4 +59,9 @@ func init() {
 	arguments.AddProfileFlag(flags)
 	arguments.AddRegionFlag(flags)
 	confirm.AddFlag(flags)
+
+	accountroles.Cmd.SetHelpFunc(func(command *cobra.Command, strings []string) {
+		arguments.MarkGlobalFlagsHidden(Cmd, "region")
+		command.Parent().HelpFunc()(command, strings)
+	})
 }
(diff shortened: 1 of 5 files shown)


cmd/create/cmd.go
```diff
@@ -59,4 +59,9 @@ func init() {
 	arguments.AddProfileFlag(flags)
 	arguments.AddRegionFlag(flags)
 	confirm.AddFlag(flags)
+
+	accountroles.Cmd.SetHelpFunc(func(command *cobra.Command, strings []string) {
```
pvasant, line 63: i dont see region in account roles create . why do you want to do it in this file anyways

pvasant: AAH I see this is help

PR #857, opened 2022-09-28
[SDA-5894] fix: adding trim spaces and tabs when validating cluster name

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-5894
# What
Interactive mode of input is still not trimming 'invisible characters'

# Why
Facilitates user input

### DIFF
--- a/pkg/ocm/helpers.go
+++ b/pkg/ocm/helpers.go
@@ -83,6 +83,7 @@ func IsValidClusterName(clusterName string) bool {
 
 func ClusterNameValidator(name interface{}) error {
 	if str, ok := name.(string); ok {
+		str := strings.Trim(str, " \t")
 		if !IsValidClusterName(str) {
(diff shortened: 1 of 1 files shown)


pkg/ocm/helpers.go
```diff
@@ -83,6 +83,7 @@ func IsValidClusterName(clusterName string) bool {
 
 func ClusterNameValidator(name interface{}) error {
 	if str, ok := name.(string); ok {
+		str := strings.Trim(str, " \t")
```
gdbranco, line 86: Ended up doing in this manner, I looked into the survey dependency and they have a TransformString option which would have been useful, but it only applies transform after validation, so we would either have to allow ' \t' on the regex or trim the internal string anyway

gdbranco: @vkareh PTAL

zgalor: strings.Trim is fine

PR #860, opened 2022-09-28
[SDA-6320] feat: set interactive enabled if local flags are unchanged, except for cluster flag

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-6320

# What
When no flags are passed to `rosa edit ingress` a request without changes was being sent

# Why
Make sure user will input changes if for some reason he forgets to specify arg changes

### DIFF
--- a/cmd/edit/ingress/cmd.go
+++ b/cmd/edit/ingress/cmd.go
@@ -24,7 +24,9 @@ import (
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 
+	"github.com/openshift/rosa/pkg/helper"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/rosa"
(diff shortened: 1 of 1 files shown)


cmd/edit/ingress/cmd.go
```diff
@@ -63,6 +65,20 @@ var Cmd = &cobra.Command{
 	},
 }
 
+func areLocalFlagsUnchanged(cmd *cobra.Command) bool {
+	parentFlagSet := cmd.Parent().Flags()
+	localFlagsExcluded := []string{"cluster"}
```
gdbranco, line 70: Tried to make sure that if we add more local flags to this command this would still work. Open to suggestions

cmd/edit/ingress/cmd.go
```diff
@@ -205,6 +226,10 @@ func run(cmd *cobra.Command, argv []string) {
 	}
 
 	r.Reporter.Debugf("Updating ingress '%s' on cluster '%s'", ingress.ID(), clusterKey)
+	if private == nil || len(routeSelectors) == 0 {
```
gdbranco, line 229: Not sure how to set this up for any given moment
Thought of doing a function that would take DTO like object as input, but we would need to change that too if we ever decide for more conditions

Open to better suggestions

zgalor: L228 should be inside that `if` block

cmd/edit/ingress/cmd.go
zgalor: I am not sure the logic is correct in general.
We would like to go to interactive if the minimal set of flags wasn't specified for this command.
I think using a generic function would be difficult to implement.
Take a look at what we do on other commands, where we don't enable interactive mode globaly but ask for it per feature set. for example create/machine_pool/cmd.go all the replicas and autoscaling flags

gdbranco: Oh no, this is not general in the sense that any command should use it.
It is just for this one atm, given that cluster has it's own check and is needed to know which cluster the ingress id is from anyway.
The other parameters are 'optional' in the sense that they create some sort of change when specified. So for the current moment we can use such an approach, but maybe at some point we'll have to change the contents of this function

PR #862, opened 2022-09-29
[SDA-6784] fix: enable path arg visibility

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-6784

# What
path arg visibility for account, ocm and user roles was hidden

# Why
Allows clients to see the help for path arg


gdbranco:
@pvasant PTAL

PR #864, opened 2022-09-29
[SDA-6939] feat: adding message about operator roles and policies path

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-6939

# What
There should be a message informing the client, the message should be made available in create cluster and create operator-roles

# Why
The client is not currently informed of this behavior, which might cause confusion.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -1013,6 +1013,9 @@ func run(cmd *cobra.Command, _ []string) {
 
 	credRequests, err := r.OCMClient.GetCredRequests(isHostedCP)
 	if isSTS {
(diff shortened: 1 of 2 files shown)


pvasant:
@arendej can you please take a look at the message 

I: Path '/ptest/' detected, this path will be used for subsequent created operator roles and policies

arendej:
> @arendej can you please take a look at the message
> 
> I: Path '/ptest/' detected, this path will be used for subsequent created operator roles and policies

Please change it to as below and also retrieve the scoped account-role set prefix so customers know which account-roles will have the common set of ARN paths.

`I: ARN path '\ptest\' detected. This ARN path will be used for subsequent created operator roles and policies, for the account roles with prefix '\<account-roles-prefix>\'.`

PR #872, opened 2022-10-05
[SDA-6965] fix: adding arn path validator to create account roles --path arg

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-6965
# What
Path was not being validate outside interactive mode

# Why
Clients may setup a wrong path and won't be able to continue with their flow

# After changes
`./rosa create account-roles --mode manual -y --prefix xueli --path /xueli/test/t1`
```
I: AWS credentials are valid!
I: Validating AWS quota...
I: Creating account roles
E: The specified value for path is invalid. It must begin and end with / and contain only alphanumeric characters and/or / characters.
```

### DIFF
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -241,6 +241,11 @@ func run(cmd *cobra.Command, argv []string) {
 		}
 	}
 
+	if !aws.ARNPath.MatchString(path) {
+		r.Reporter.Errorf("It must begin and end with / and contain only alphanumeric characters and/or / characters.")
+		os.Exit(1)
+	}
+
 	if interactive.Enabled() {
 		mode, err = interactive.GetOption(interactive.Input{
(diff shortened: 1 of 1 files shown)


cmd/create/accountroles/cmd.go
```diff
@@ -241,6 +241,12 @@ func run(cmd *cobra.Command, argv []string) {
 		}
 	}
 
+	if !aws.ARNPath.MatchString(path) {
+		r.Reporter.Errorf("The specified value for path is invalid. " +
+			"It must begin and end with / and contain only alphanumeric characters and/or / characters.")
```
zgalor, line 246: np: use `'/'`
to separate between the character and the `and/or`

gdbranco, line 246:
```suggestion
			"It must begin and end with '/' and contain only alphanumeric characters and/or '/' characters.")
```

PR #876, opened 2022-10-06
[SDE-1816] Retrieving roles from cluster instead of using prefix from args when called from upgrade command

gdbranco:
Related issues: 
https://issues.redhat.com/browse/SDA-6407
https://issues.redhat.com/browse/SDE-1816
# What
Providing a way to retrieve account role information from the cluster id instead of using the prefix to reconstruct everything

# Why
Easier and less prone to error, also fixes clients using acc roles from older not standardized role prefixes.

Prefix kept as optional for someone who wants to upgrade only acc without having them linked to a cluster.
Cluster added as optional for someone who wants to upgrade acc roles linked too a cluster.. Also solves the problem with acc roles that don't match expected prefix

### DIFF
--- a/cmd/upgrade/accountroles/cmd.go
+++ b/cmd/upgrade/accountroles/cmd.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/briandowns/spinner"
+	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 
 	"github.com/openshift/rosa/pkg/aws"
@@ -95,16 +96,10 @@ func run(cmd *cobra.Command, argv []string) error {
 
 	isInvokedFromClusterUpgrade := false
 	skipInteractive := false
+	var cluster *v1.Cluster
 	if len(argv) >= 2 && !cmd.Flag("prefix").Changed {
 		ocm.SetClusterKey(argv[1])
-		clusterKey := r.GetClusterKey()
-		cluster := r.FetchCluster()
-		prefix, err := aws.GetPrefixFromAccountRole(cluster)
-		if err != nil {
-			r.Reporter.Errorf("Could not get role prefix for cluster '%s' : %v", clusterKey, err)
-			os.Exit(1)
-		}
-		args.prefix = prefix
+		cluster = r.FetchCluster()
 		aws.SetModeKey(argv[0])
 		if argv[1] != "" {
 			skipInteractive = true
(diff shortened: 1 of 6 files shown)


cmd/upgrade/accountroles/cmd.go
```diff
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/briandowns/spinner"
+	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
```
pvasant, line 26: np: sort imports

cmd/upgrade/accountroles/cmd.go
```diff
@@ -95,15 +97,14 @@ func run(cmd *cobra.Command, argv []string) error {
 
 	isInvokedFromClusterUpgrade := false
 	skipInteractive := false
-	if len(argv) >= 2 && !cmd.Flag("prefix").Changed {
-		args.prefix = argv[0]
-		aws.SetModeKey(argv[1])
+	var cluster *v1.Cluster
+	if len(argv) == 2 && !cmd.Flag("prefix").Changed {
```
pvasant, line 101: This will not work if the user does 

rosa upgrade accoount-roles --cluster-id

cmd/upgrade/accountroles/cmd.go
```diff
@@ -251,6 +295,144 @@ func LogError(key string, ocmClient *ocm.Client, defaultPolicyVersion string, er
...
+	reporter *rprtr.Object,
+	awsClient aws.Client,
+	cluster *v1.Cluster,
+	accountID string,
+	policies map[string]string,
+	policyVersion string,
+	policyPath string,
+	isVersionChosen bool,
+) error {
+	for file, role := range aws.AccountRoles {
```
pvasant, line 308: This we cannot do it for the older formats since they dont follow our standards

cmd/upgrade/accountroles/cmd.go
gdbranco: This is only when calling from within `upgrade cluster`

cmd/upgrade/accountroles/cmd.go
gdbranco: We can probably go over this, but I'm using those as a way to setup a map to the roles defined within the cluster

pvasant: Yes, please. We need not use the prefix at all. The flow should be simple for cluster upgrades. Just fetch the roles from CS DB and update. So we need to take out all the prefix logic .

PR #877, opened 2022-10-06
[SDA-6974] add managed services path support

gdbranco:
Related issues: https://issues.redhat.com/browse/SDA-6974
# What
Adding support for path in acc roles ARN

# Why
Adds feature parity to ms even though clients might not be using it

### DIFF
--- a/cmd/create/service/cmd.go
+++ b/cmd/create/service/cmd.go
@@ -398,10 +398,6 @@ func run(cmd *cobra.Command, argv []string) {
 		r.Reporter.Errorf("Expected a valid path for  '%s': %v", roleARN, err)
 		os.Exit(1)
 	}
-	if path != "" {
-		r.Reporter.Errorf("Managed Services currently does not support path in ARN")
-		os.Exit(1)
-	}
 
 	// operator role logic.
 	operatorRolesPrefix := getRolePrefix(args.ClusterName)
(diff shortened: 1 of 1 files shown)


gdbranco:
@pvasant @renan-campos PTAL

cmd/create/service/cmd.go
```diff
@@ -419,7 +429,7 @@ func run(cmd *cobra.Command, argv []string) {
 		operatorIAMRoleList = append(operatorIAMRoleList, ocm.OperatorIAMRole{
 			Name:      operator.Name(),
 			Namespace: operator.Namespace(),
-			RoleARN:   getOperatorRoleArn(operatorRolesPrefix, operator, r.Creator),
+			RoleARN:   getOperatorRoleArn(operatorRolesPrefix, operator, r.Creator, path),
```
gdbranco, line 432: Feel like this would solve the problem of creating the operator roles

renan-campos: Looking at the other managed service operations, you're absolutely right! I don't think the current clients of managed services will use this feature, but since it doesn't affect the old way and helps maintain feature parity I think we can add support for it. Do you want to change the title of the pr to add support instead of blocking?

gdbranco: Sure, let's test that then

renan-campos:
LGTM

PR #878, opened 2022-10-07
[SDA-6965] Add validation to path ocm/user roles

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-6965
# What
A validation to the path was done to acc roles, but not for ocm/user roles

# Why
Making sure client knows what is happening if he inputs a wrong path


andreadecorte:
lgtm

PR #880, opened 2022-10-09
[SDA-6986] Consider empty path valid creating acc roles

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-6986
# What
Validation was being applied when path is empty

# Why
Empty path is a valid path


andreadecorte:
lgtm

PR #884, opened 2022-10-10
[SDA-6984] Better not clarify channel group when validating OCP version for HyperShift

jordigilh:
Removes the channel group from the error message when the OCP cluster version is not compatible with HyperShift.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -2156,11 +2156,7 @@ func validateVersion(version string, versionList []string, channelGroup string, 
 				return "", fmt.Errorf("error while parsing OCP version '%s': %v", version, err)
 			}
 			if !valid {
-				v := version
-				if len(channelGroup) > 0 {
-					v = fmt.Sprintf("%s-%s", version, channelGroup)
-				}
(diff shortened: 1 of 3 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -2156,11 +2156,7 @@ func validateVersion(version string, versionList []string, channelGroup string,
 				return "", fmt.Errorf("error while parsing OCP version '%s': %v", version, err)
 			}
 			if !valid {
-				v := version
-				if len(channelGroup) > 0 {
-					v = fmt.Sprintf("%s-%s", version, channelGroup)
-				}
-				return "", fmt.Errorf("version '%s' is not suported for HyperShift clusters", v)
+				return "", fmt.Errorf("version '%s' is not supported for HyperShift clusters", version)
```
igoihman, line 2159: we are using `hosted clusters` as official terminology

jordigilh: Updated

igoihman:
CI failed:
```cmd/create/cluster/cmd_test.go:27: line is 124 characters (lll)
		It("KO: Fails to validate a cluster for a hosted cluster when the user provides an invalid or malformed version", func() {```

igoihman:
LGTM

PR #892, opened 2022-10-13
[SDA-7021] --hosted-cp flag now forces byo vpc prompt like privatelink does

hunterkepley:
[SDA-7021](https://issues.redhat.com/browse/SDA-7021)

Now, if you include the `--hosted-cp` flag when creating a cluster:

`rosa create cluster --sts --mode auto --hosted-cp`

Then once you get to the `Would you like to install on an existing VPC?`  step, it automatically responds Yes and asks you which subnets to use.Before, it always prompted Would you like to install on an existing VPC? even with the `--hosted-cp`  flag

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -1262,7 +1262,7 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
-	if privateLink {
+	if privateLink || isHostedCP {
 		useExistingVPC = true
 	}
 


cmd/create/cluster/cmd.go
```diff
@@ -1262,7 +1262,7 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
-	if privateLink {
+	if privateLink || cmd.Flags().Changed("hosted-cp") {
```
gdbranco, line 1265: Not so sure as I'm no so familiar with this flow, but I feel like you can probably use this
https://github.com/openshift/rosa/blob/2a4b109a7080b54e2a4ce9cc68797487b57c99d7/cmd/create/cluster/cmd.go#L1207 instead of doing flags changed

andreadecorte: by the way privateLink should already include the isPrivateHostedCP I think
https://github.com/openshift/rosa/blob/2a4b109a7080b54e2a4ce9cc68797487b57c99d7/cmd/create/cluster/cmd.go#L1208

hunterkepley: @gdbranco isHostedCP only returns true when you answer `Yes` to the "? Deploy cluster with hosted control plane: [? for help] (Y/n)" question

When discussing this story with the assigner (Priya) she said that it should be as long as the `--hosted-cp` flag is used

gdbranco: https://github.com/openshift/rosa/blob/d09fd9f9cb0ab4e865a60942bb418ac99b91d260/cmd/create/cluster/cmd.go#L600

Based on that I believe so, at least it is the answer to the prompt and it is also where the arg value is held as a variable

hunterkepley:
@gdbranco PTAL! Added the isHostedCP bool instead after talking more with Priya

PR #895, opened 2022-10-17
[SDA-6984] Add support for nightly builds for Hosted CP clusters

jordigilh:
Adds support for nightly builds of OCP as viable versions for Hosted Clusters. Previously only 4.11.6-candidate and newer were supported, but not nightly builds as they are tagged with `<major>.<minor>.0-nightly-<timestamp>-nightly`, example:

```
  4.11.0-0.nightly-2022-10-17-040259-nightly
  4.10.0-0.nightly-2022-10-17-082154-nightly
  4.10.0-0.nightly-2022-10-14-023020-nightly
  4.10.0-0.nightly-2022-10-13-011121-nightly
  4.10.0-0.nightly-2022-10-12-184111-nightly
  4.8.0-0.nightly-2022-10-14-021947-nightly
  4.8.0-0.nightly-2021-05-26-071911-nightly
```

@pvasant @jakub-dzon  PTAL

### DIFF
--- a/pkg/ocm/versions.go
+++ b/pkg/ocm/versions.go
@@ -27,9 +27,10 @@ import (
 
 const (
 	DefaultChannelGroup   = "stable"
+	NightlyChannelGroup   = "nightly"
 	LowestSTSSupport      = "4.7.11"
 	LowestSTSMinor        = "4.7"
-	LowestHostedCPSupport = "4.11.6-candidate"
+	LowestHostedCPSupport = "4.11.6"
 )
 
 func (c *Client) GetVersions(channelGroup string) (versions []*cmv1.Version, err error) {
@@ -78,7 +79,7 @@ func (c *Client) GetVersions(channelGroup string) (versions []*cmv1.Version, err
 }
 
 func HasSTSSupport(rawID string, channelGroup string) bool {
-	if channelGroup == "nightly" {
+	if channelGroup == NightlyChannelGroup {
 		return true
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/ocm/versions.go
```diff
@@ -29,7 +29,7 @@ const (
 	DefaultChannelGroup   = "stable"
 	LowestSTSSupport      = "4.7.11"
 	LowestSTSMinor        = "4.7"
-	LowestHostedCPSupport = "4.11.6-candidate"
+	LowestHostedCPSupport = "4.11.6"
```
pvasant, line 32: i dont think nightly have patch. For eg. 4.11.0-0.nightly-2022-10-17-040259

pvasant:
@gdbranco  PTAL

pkg/ocm/versions.go
jordigilh: it will still work, the code looks for `-0.nightly` as prefix of the Prerelease field (whatever is after the semantic versioning):
https://github.com/openshift/rosa/pull/895/files#diff-83f815beabf111da90b662b338ca5594199acfab18880dd9ed5c2ee4dc100d12R115

gdbranco:
lgtm

pkg/ocm/versions.go
pvasant: can we check the channleGroup of the version instead of parsing the version format

jordigilh: yeah, much better. thanks for the suggestion!. Code updated :smile:

PR #896, opened 2022-10-17
[SDA-6643] STS is now default mode for cluster creation, added flags for mint mode/non-sts mode

hunterkepley:
[SDA-6643](https://issues.redhat.com/browse/SDA-6643)

Adds `mint-mode` [alias] and `non-sts` flags to ROSA CLI, and sets STS as the default mode with mint mode being an optional mode now. Also only prompts if they want to use STS when mint-mode and interactive are enabled.

Flow:

When creating a new cluster, just using the default `rosa create cluster` will now use STS mode. If you use the new flags (either one) `rosa create cluster --non-sts` or `rosa create cluster --mint-mode` -- it will use the legacy (IAM) mode for creating clusters.

If you are in interactive mode (by not providing a cluster name for example) and are using `--non-sts` or `--mint-mode` , you will be prompted whether or not you'd like to use STS mode instead. Otherwise, the flow of using `rosa create cluster` or `rosa create cluster --sts` will remain the same as before sans the change to the default mode.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -112,6 +112,9 @@ var args struct {
 	// Force STS mode for interactive and validation
 	sts bool
 
+	// Force IAM mode (mint mode) for interactive
+	nonSts bool
+
 	// Account IAM Roles
 	roleARN             string
 	externalID          string
@@ -174,7 +177,21 @@ func init() {
 		&args.sts,
 		"sts",
 		false,
-		"Use AWS Security Token Service (STS) instead of IAM credentials to deploy your cluster.",
+		"Use AWS Security Token Service (STS) instead of IAM credentials to deploy your cluster. "+
+			"This flag is no longer needed as STS is the default mode for cluster creation.",
+	)
+	flags.MarkDeprecated("sts", "STS is now the default mode of cluster creation.")
+	flags.BoolVar(
+		&args.nonSts,
+		"non-sts",
+		false,
+		"Use legacy way of creating clusters (IAM mode).",
+	)
+	flags.BoolVar(
+		&args.nonSts,
+		"mint-mode",
+		false,
+		"Use legacy way of creating clusters (IAM mode). This is an alias for --non-sts.",
 	)
 	flags.StringVar(
 		&args.roleARN,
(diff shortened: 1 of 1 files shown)


hunterkepley:
@pvasant This change causes `? Deploy cluster using AWS STS: Yes` to show when creating a cluster. Should I make it skip this? Or is that fine?

pvasant:
@arendej  PTAL

cmd/create/cluster/cmd.go
```diff
@@ -176,6 +180,18 @@ func init() {
 		false,
 		"Use AWS Security Token Service (STS) instead of IAM credentials to deploy your cluster.",
 	)
+	flags.BoolVar(
+		&args.nonSts,
+		"non-sts",
+		false,
+		"Use legacy way of creating clusters (IAM mode).",
+	)
+	flags.BoolVar(
```
gdbranco, line 189: You can use the same arg to bind it to make an 'alias' so you don't end up with two args

cmd/create/cluster/cmd.go
```diff
@@ -563,6 +579,11 @@ func run(cmd *cobra.Command, _ []string) {
 			"Any optional fields can be left empty and a default will be selected.")
 	}
 
+	// warn if sts flag is used
+	if cmd.Flags().Changed("sts") {
+		r.Reporter.Warnf("Cluster creation is with STS by default, and the --sts flag is no longer necessary.")
```
arendej, line 584: Please change:
Cluster creation is with STS by default and the --sts flag is no longer necessary.

arendej (approved):
Reviewed, Thanks

cmd/create/cluster/cmd.go
```diff
@@ -174,7 +177,20 @@ func init() {
 		&args.sts,
 		"sts",
 		false,
-		"Use AWS Security Token Service (STS) instead of IAM credentials to deploy your cluster.",
+		"Use AWS Security Token Service (STS) instead of IAM credentials to deploy your cluster. "+
```
gdbranco, line 180: Maybe mark it deprecated?

pvasant:
@hunterkepley please add the flow in the description to help visualize the changs

hunterkepley:
Added flow to description

cmd/create/cluster/cmd.go
hunterkepley: Marked!

hunterkepley:
> @hunterkepley I just noticed that if the user selected the `non-sts` flag, it doesn't appear on the `To create this cluster again...` output: Can you please add this flag to the printed command?

Guilherme has worked on a reversion for this MR and revised it with new needs from the team. @gdbranco is this something you would want to do for your work on this?

gdbranco:
> > @hunterkepley I just noticed that if the user selected the `non-sts` flag, it doesn't appear on the `To create this cluster again...` output: Can you please add this flag to the printed command?
> 
> Guilherme has worked on a reversion for this MR and revised it with new needs from the team. @gdbranco is this something you would want to do for your work on this?

I think this will be needed when we move forward with sts default again, but not atm

PR #901, opened 2022-10-23
removed --channel-group  from --help options.

olisakov:
@tzvatot PTAL
[SDA-7066](https://issues.redhat.com//browse/SDA-7066)

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -249,6 +249,7 @@ func init() {
 		ocm.DefaultChannelGroup,
 		"Channel group is the name of the group where this image belongs, for example \"stable\" or \"fast\".",
 	)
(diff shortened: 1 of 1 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -1686,7 +1687,6 @@ func run(cmd *cobra.Command, _ []string) {
 		Region:                    region,
 		MultiAZ:                   multiAZ,
 		Version:                   version,
-		ChannelGroup:              channelGroup,
```
tzvatot, line 1689: I think we still need this otherwise it won't work.

oriAdler:
@olisakov why do we need `openshift-client-linux.tar.gz` in this PR?
Except for that LGTM.

PR #902, opened 2022-10-24
[SDA-6920] Aws command builder

gdbranco:
Related issues: https://issues.redhat.com/browse/SDA-6920
# What
Share behavior for constructing manual aws commands.

# Why
From time to time we change the manual commands, adding or removing parameter. This can cause differences on how they behave throughout each rosa command.

# After Changes example output
`rosa create account-roles --mode manual --prefix=gdb-path --path /gdbsts/ --version=4.9 -y`
```
I: AWS credentials are valid!
I: Validating AWS quota...
I: Creating account roles
I: All policy files saved to the current directory
I: Run the following commands to create the account roles and policies:

aws iam create-role \
	--assume-role-policy-document file://sts_installer_trust_policy.json \
	--path /gdbsts/ \
	--role-name gdb-path-Installer-Role \
	--tags Key=red-hat-managed,Value=true Key=rosa_openshift_version,Value=4.9 Key=rosa_role_prefix,Value=gdb-path Key=rosa_role_type,Value=installer

aws iam create-policy \
	--path /gdbsts/ \
	--policy-document file://sts_installer_permission_policy.json \
	--policy-name gdb-path-Installer-Role-Policy \
	--tags Key=red-hat-managed,Value=true Key=rosa_openshift_version,Value=4.9 Key=rosa_role_prefix,Value=gdb-path Key=rosa_role_type,Value=installer

aws iam attach-role-policy \
	--policy-arn arn:aws:iam::765374464689:policy/gdbsts/gdb-path-Installer-Role-Policy \
	--role-name gdb-path-Installer-Role

aws iam create-role \
	--assume-role-policy-document file://sts_instance_controlplane_trust_policy.json \
	--path /gdbsts/ \
	--role-name gdb-path-ControlPlane-Role \
	--tags Key=red-hat-managed,Value=true Key=rosa_openshift_version,Value=4.9 Key=rosa_role_prefix,Value=gdb-path Key=rosa_role_type,Value=instance_controlplane

aws iam create-policy \
	--path /gdbsts/ \
	--policy-document file://sts_instance_controlplane_permission_policy.json \
	--policy-name gdb-path-ControlPlane-Role-Policy \
	--tags Key=red-hat-managed,Value=true Key=rosa_openshift_version,Value=4.9 Key=rosa_role_prefix,Value=gdb-path Key=rosa_role_type,Value=instance_controlplane

aws iam attach-role-policy \
	--policy-arn arn:aws:iam::765374464689:policy/gdbsts/gdb-path-ControlPlane-Role-Policy \
	--role-name gdb-path-ControlPlane-Role

aws iam create-role \
	--assume-role-policy-document file://sts_instance_worker_trust_policy.json \
	--path /gdbsts/ \
	--role-name gdb-path-Worker-Role \
	--tags Key=red-hat-managed,Value=true Key=rosa_openshift_version,Value=4.9 Key=rosa_role_prefix,Value=gdb-path Key=rosa_role_type,Value=instance_worker

aws iam create-policy \
	--path /gdbsts/ \
	--policy-document file://sts_instance_worker_permission_policy.json \
	--policy-name gdb-path-Worker-Role-Policy \
	--tags Key=red-hat-managed,Value=true Key=rosa_openshift_version,Value=4.9 Key=rosa_role_prefix,Value=gdb-path Key=rosa_role_type,Value=instance_worker

aws iam attach-role-policy \
	--policy-arn arn:aws:iam::765374464689:policy/gdbsts/gdb-path-Worker-Role-Policy \
	--role-name gdb-path-Worker-Role

aws iam create-role \
	--assume-role-policy-document file://sts_support_trust_policy.json \
	--path /gdbsts/ \
	--role-name gdb-path-Support-Role \
	--tags Key=red-hat-managed,Value=true Key=rosa_openshift_version,Value=4.9 Key=rosa_role_prefix,Value=gdb-path Key=rosa_role_type,Value=support

aws iam create-policy \
	--path /gdbsts/ \
	--policy-document file://sts_support_permission_policy.json \
	--policy-name gdb-path-Support-Role-Policy \
	--tags Key=red-hat-managed,Value=true Key=rosa_openshift_version,Value=4.9 Key=rosa_role_prefix,Value=gdb-path Key=rosa_role_type,Value=support

aws iam attach-role-policy \
	--policy-arn arn:aws:iam::765374464689:policy/gdbsts/gdb-path-Support-Role-Policy \
	--role-name gdb-path-Support-Role
```

### DIFF
--- a/cmd/create/oidcprovider/cmd.go
+++ b/cmd/create/oidcprovider/cmd.go
@@ -31,6 +31,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/openshift/rosa/pkg/aws"
+	awscb "github.com/openshift/rosa/pkg/aws/commandbuilder"
 	"github.com/openshift/rosa/pkg/aws/tags"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/interactive/confirm"
@@ -203,18 +204,22 @@ func buildCommands(r *rosa.Runtime, cluster *cmv1.Cluster) (string, error) {
 	}
 	r.Reporter.Debugf("Using thumbprint '%s'", thumbprint)
 
-	tag := fmt.Sprintf(
-		"Key=%s,Value=%s",
-		tags.ClusterID, cluster.ID())
-	createOpenIDConnectProvider := fmt.Sprintf("aws iam create-open-id-connect-provider \\\n"+
-		"\t--url %s \\\n"+
-		"\t--client-id-list %s %s \\\n"+
-		"\t--thumbprint-list %s \\\n"+
-		"\t --tag %s",
-		oidcEndpointURL, aws.OIDCClientIDOpenShift, aws.OIDCClientIDSTSAWS, thumbprint, tag)
+	tag := map[string]string{
+		tags.ClusterID: cluster.ID(),
+	}
+
+	clientIdList := strings.Join([]string{aws.OIDCClientIDOpenShift, aws.OIDCClientIDSTSAWS}, " ")
+
+	createOpenIDConnectProvider := awscb.NewIAMCommandBuilder().
+		SetCommand(awscb.CreateOpenIdConnectProvider).
+		AddParam(awscb.Url, oidcEndpointURL).
+		AddParam(awscb.ClientIdList, clientIdList).
+		AddParam(awscb.ThumbprintList, thumbprint).
+		AddTags(tag).
+		Build()
 	commands = append(commands, createOpenIDConnectProvider)
 
-	return strings.Join(commands, "\n\n"), nil
+	return awscb.JoinCommands(commands), nil
 }
 
 func getThumbprint(oidcEndpointURL string) (string, error) {
--- a/cmd/create/userrole/cmd.go
+++ b/cmd/create/userrole/cmd.go
@@ -19,13 +19,13 @@ package userrole
 import (
 	"fmt"
 	"os"
-	"strings"
 
 	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/spf13/cobra"
 
 	linkuser "github.com/openshift/rosa/cmd/link/userrole"
 	"github.com/openshift/rosa/pkg/aws"
+	awscb "github.com/openshift/rosa/pkg/aws/commandbuilder"
 	"github.com/openshift/rosa/pkg/aws/tags"
 	"github.com/openshift/rosa/pkg/helper"
 	"github.com/openshift/rosa/pkg/interactive"
@@ -258,29 +258,24 @@ func buildCommands(prefix string, path string, userName string,
 	roleName := aws.GetUserRoleName(prefix, aws.OCMUserRole, userName)
 
 	roleARN := aws.GetRoleARN(accountID, roleName, path)
-	iamTags := fmt.Sprintf(
-		"Key=%s,Value=%s Key=%s,Value=%s Key=%s,Value=%s Key=%s,Value=%s",
-		tags.RolePrefix, prefix,
-		tags.RoleType, aws.OCMUserRole,
-		tags.Environment, env,
-		tags.RedHatManaged, "true",
-	)
-	permBoundaryFlag := ""
-	if permissionsBoundary != "" {
-		permBoundaryFlag = fmt.Sprintf("\t--permissions-boundary %s \\\n", permissionsBoundary)
+	iamTags := map[string]string{
+		tags.RolePrefix:    prefix,
+		tags.RoleType:      aws.OCMUserRole,
+		tags.Environment:   env,
+		tags.RedHatManaged: "true",
 	}
-	createRole := fmt.Sprintf("aws iam create-role \\\n"+
-		"\t--role-name %s \\\n"+
-		"\t--assume-role-policy-document file://sts_%s_trust_policy.json \\\n"+
-		"%s"+
-		"\t--tags %s",
-		roleName, aws.OCMUserRolePolicyFile, permBoundaryFlag, iamTags)
-	if path != "" {
-		createRole = fmt.Sprintf(createRole+" \\\n\t--path %s", path)
-	}
+	createRole := awscb.NewIAMCommandBuilder().
+		SetCommand(awscb.CreateRole).
+		AddParam(awscb.RoleName, roleName).
+		AddParam(awscb.AssumeRolePolicyDocument,
+			fmt.Sprintf("file://sts_%s_trust_policy.json", aws.OCMUserRolePolicyFile)).
+		AddParam(awscb.PermissionsBoundary, permissionsBoundary).
+		AddTags(iamTags).
+		AddParam(awscb.Path, path).
+		Build()
 	linkRole := fmt.Sprintf("rosa link user-role --role-arn %s", roleARN)
 	commands = append(commands, createRole, linkRole)
-	return strings.Join(commands, "\n\n")
+	return awscb.JoinCommands(commands)
 }
 
 func createRoles(r *rosa.Runtime,
--- a/cmd/dlt/userrole/cmd.go
+++ b/cmd/dlt/userrole/cmd.go
@@ -19,13 +19,13 @@ package userrole
 import (
 	"fmt"
 	"os"
-	"strings"
 
 	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/spf13/cobra"
 
 	unlinkuserrole "github.com/openshift/rosa/cmd/unlink/userrole"
 	"github.com/openshift/rosa/pkg/aws"
+	awscb "github.com/openshift/rosa/pkg/aws/commandbuilder"
 	"github.com/openshift/rosa/pkg/helper"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/interactive/confirm"
@@ -209,7 +209,6 @@ func run(cmd *cobra.Command, argv []string) {
 
 func buildCommands(roleName string, roleARN string, isLinked bool, awsClient aws.Client) (string, error) {
 	var commands []string
-
 	if isLinked {
 		unlinkRole := fmt.Sprintf("rosa unlink user-role \\\n"+
 			"\t--role-arn %s", roleARN)
@@ -221,10 +220,11 @@ func buildCommands(roleName string, roleARN string, isLinked bool, awsClient aws
 		return "", err
 	}
 	for _, policy := range policies {
-		detachPolicy := fmt.Sprintf("aws iam detach-role-policy \\\n"+
-			"\t--role-name %s \\\n"+
-			"\t--policy-arn %s",
-			roleName, policy.PolicyArn)
+		detachPolicy := awscb.NewIAMCommandBuilder().
+			SetCommand(awscb.DetachRolePolicy).
+			AddParam(awscb.RoleName, roleName).
+			AddParam(awscb.PolicyArn, policy.PolicyArn).
+			Build()
 		commands = append(commands, detachPolicy)
 	}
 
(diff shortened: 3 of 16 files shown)


tbrisker:
@gdbranco i like the general approach! I would however consider using string consts instead of iota + switch cases.

pvasant:
@vkareh @tbrisker  PTAL

cmd/create/oidcprovider/cmd.go
```diff
@@ -203,15 +203,21 @@ func buildCommands(r *rosa.Runtime, cluster *cmv1.Cluster) (string, error) {
...
-		"Key=%s,Value=%s",
-		tags.ClusterID, cluster.ID())
-	createOpenIDConnectProvider := fmt.Sprintf("aws iam create-open-id-connect-provider \\\n"+
-		"\t--url %s \\\n"+
-		"\t--client-id-list %s %s \\\n"+
-		"\t--thumbprint-list %s \\\n"+
-		"\t --tag %s",
-		oidcEndpointURL, aws.OIDCClientIDOpenShift, aws.OIDCClientIDSTSAWS, thumbprint, tag)
+	tag := map[string]string{
+		tags.ClusterID: cluster.ID(),
```
gdbranco, line 207: Shouldn't this have red hat managed?

cmd/create/userrole/cmd.go
```diff
@@ -258,26 +258,21 @@ func buildCommands(prefix string, path string, userName string,
...
+		tags.RoleType:      aws.OCMUserRole,
+		tags.Environment:   env,
+		tags.RedHatManaged: "true",
 	}
+	builder := &aws.CommandBuilder{}
+	builder.SetService(aws.IAM)
+	builder.SetCommand(aws.CREATE_ROLE)
+	builder.AddParam(aws.ROLE_NAME, roleName)
+	builder.AddParam(aws.ASSUME_ROLE_POLICY_DOCUMENT, aws.OCMUserRolePolicyFile)
+	builder.AddParam(aws.PERMISSIONS_BOUNDARY, fmt.Sprintf("file://sts_%s_trust_policy.json", permissionsBoundary))
```
gdbranco, line 272: Invert with assume role policy

tbrisker (changes requested):

pkg/aws/commandbuilder.go
```diff
@@ -0,0 +1,126 @@
...
+	b.service = fmt.Sprint(awsService)
+}
+
+func (b *CommandBuilder) SetCommand(awsCommand Command) {
+	b.command = fmt.Sprint(awsCommand)
+}
+
+func (b *CommandBuilder) AddParam(awsParam Param, value interface{}) {
+	if b.params == nil {
+		b.params = make([]string, 1)
```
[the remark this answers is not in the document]
gdbranco: That was supposed to be 0,1.
Will check being nil and remove if not needed

pkg/aws/commandbuilder.go
```diff
@@ -0,0 +1,126 @@
...
+	SET_AS_DEFAULT               Param = "set-as-default"
+)
+
+type ICommandBuilder interface {
+	SetService(awsService Service)
+	SetCommand(awsCommand Command)
+	AddParam(awsParam Param, value interface{})
+	ResetParams()
+	Build() string
+}
```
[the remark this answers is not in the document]
gdbranco: Not atm, but if we ever want to change other commands that are strings, like rosa commands for example I would use the interface to setup the concrete rosa command builder

pkg/aws/commandbuilder.go
```diff
@@ -0,0 +1,126 @@
...
+	"bytes"
+	"fmt"
+	"sort"
+	"strings"
+)
+
+type Service string
+
+const (
+	UNDEFINED_SERVICE Service = "undefined service"
```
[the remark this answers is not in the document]
gdbranco: Not anymore, since we don't have the switch anymore. Will be removing it

pkg/aws/commandbuilder.go
```diff
@@ -0,0 +1,126 @@
...
+	b.command = fmt.Sprint(awsCommand)
+}
+
+func (b *CommandBuilder) AddParam(awsParam Param, value interface{}) {
+	if b.params == nil {
+		b.params = make([]string, 1)
+	}
+	stringValue := ""
+	typeOf := fmt.Sprintf("%T", value)
+	switch typeOf {
```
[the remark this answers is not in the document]
gdbranco: Yeah, I was thinking of restructuring this indeed!

pkg/aws/commandbuilder.go
```diff
@@ -0,0 +1,126 @@
...
+func (b *CommandBuilder) AddBoolParam(awsParam Param) {
+	if b.params == nil {
+		b.params = make([]string, 1)
+	}
+	b.params = append(b.params, fmt.Sprintf("\t--%s", awsParam))
+}
+
+func (b *CommandBuilder) ResetParams() {
+	b.params = nil
+}
```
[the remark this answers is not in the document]
gdbranco: Agree

pkg/aws/commandbuilder.go
```diff
@@ -0,0 +1,126 @@
...
+	keys := make([]string, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, k := range keys {
+		fmt.Fprintf(b, "Key=%s,Value=%s ", k, m[k])
+	}
+	bString := b.String()
+	return bString[0 : len(bString)-1]
```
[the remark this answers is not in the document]
gdbranco: Since I'm 'joining' with a space in the end it will end up having an extra space when adding the ' \\\n'

pkg/aws/commandbuilder.go
```diff
@@ -0,0 +1,126 @@
...
+	return fmt.Sprintf("aws %s %s%s", b.service, b.command, strings.Join(b.params, " \\\n"))
+}
+
+func createKeyValuePairs(m map[string]string) string {
+	b := new(bytes.Buffer)
+	keys := make([]string, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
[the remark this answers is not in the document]
gdbranco: Not really, but since order does not matter it changes depending on the execution. Ordering makes sure we always get the same output, so we can test it later more easily

pkg/aws/commandbuilder.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	stringValue := ""
+	typeOf := fmt.Sprintf("%T", value)
+	switch typeOf {
+	case "string":
+		stringValue = value.(string)
+	case "map[string]string":
+		stringValue = createKeyValuePairs(value.(map[string]string))
+	default:
+		stringValue = ""
```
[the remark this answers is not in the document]
gdbranco: We could just error out too as the default, type not supported or something.
The point is to make sure someone would need to implement it as we don't have this specific case yet

pkg/aws/commandbuilder.go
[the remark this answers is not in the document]
gdbranco: We can use this now, only checking StringMap, but we still need a different method though. The point of the BoolParam was more of a param without value, renamed it

pkg/aws/commandbuilder.go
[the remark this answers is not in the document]
gdbranco: Ended up getting the string value with %v, this will cover most of the base cases.
Is there a guideline to have some form of try catch like, didn't want to error as would need to check everytime we add a param

pkg/aws/commandbuilder.go
```diff
@@ -0,0 +1,114 @@
...
+	b.service = awsService
+}
+
+func (b *CommandBuilder) SetCommand(awsCommand Command) {
+	b.command = awsCommand
+}
+
+type StringMap map[string]string
+
+func (b *CommandBuilder) AddParam(awsParam Param, value interface{}) {
```
tbrisker, line 76: how about splitting this to two methods, like so:
```suggestion
func (b *CommandBuilder) AddParam(awsParam Param, value string) {...
func (b *CommandBuilder) AddKVParam(awsParam Param, value map[string]string) {...
```
Or even, if i'm not mistaken and these are always used for `--tags` parameter:
```suggestion
func (b *CommandBuilder) AddParam(awsParam Param, value string) {...
func (b *CommandBuilder) AddTags(tags map[string]string) {...
```
and add a `tags []string` to the builder struct?
That will allow us to use the compile time type safety, make the methods clearer and allow conditionally adding more tags (e.g. for the `isAdmin` case)

pkg/aws/commandbuilder.go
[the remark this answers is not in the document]
tbrisker: Lets not make a premature interface then. If we ever have multiple implementations for a certain interface we could always create that interface as needed - it might even not include all the current methods.

tbrisker (changes requested):

pkg/aws/commandbuilder.go
gdbranco: Coming from overloading I just wanted to have a single entry point to params, but this is ok by me as well. Will update it following this setup

pkg/aws/commandbuilder.go
```diff
@@ -0,0 +1,126 @@
...
+func (b *CommandBuilder) AddParamNoValue(awsParam Param) *CommandBuilder {
+	b.params = append(b.params, fmt.Sprintf("\t--%s", awsParam))
+	return b
+}
+
+func (b *CommandBuilder) Build() string {
+	if len(b.tags) != 0 {
+		b.AddParam(TAGS, createTags(b.tags))
+	}
+	sort.Strings(b.params)
```
gdbranco, line 91: No particular reason, but I found it easier to read

tbrisker (commented):
Looking really good! one nitpick inline, feel free to ignore it. 
One more suggestion, what do you think about moving the builder to a separate package? since it isn't really related to the rest of the aws package

cmd/create/oidcprovider/cmd.go
```diff
@@ -203,15 +204,20 @@ func buildCommands(r *rosa.Runtime, cluster *cmv1.Cluster) (string, error) {
...
+	}
+
+	clientIdList := strings.Join([]string{aws.OIDCClientIDOpenShift, aws.OIDCClientIDSTSAWS}, " ")
+	clientIdList = clientIdList[0 : len(clientIdList)-1]
```
tbrisker, line 212: This isn't needed, `strings.join` won't add a space at the end, you'll be dropping the last char of `aws.OIDCClientIDSTSAWS`.

cmd/dlt/userrole/cmd.go
```diff
@@ -233,15 +234,18 @@ func buildCommands(roleName string, roleARN string, isLinked bool, awsClient aws
...
-	deleteRole := fmt.Sprintf("aws iam delete-role \\\n"+
-		"\t--role-name %s", roleName)
+	deleteRole := awscb.NewIAMCommandBuilder().
+		SetCommand(awscb.DELETE_ROLE).
+		AddParam(awscb.ROLE_NAME, roleName).
+		Build()
 	commands = append(commands, deleteRole)
 
-	return strings.Join(commands, "\n"), nil
+	return strings.Join(commands, "\n\n"), nil
```
tbrisker, line 250: why the extra newline between commands in this case?

cmd/upgrade/accountroles/cmd.go
```diff
@@ -310,56 +311,67 @@ func buildCommands(prefix string, accountID string, isUpgradeNeedForAccountRoleP
...
-			tagRole := fmt.Sprintf("aws iam tag-role \\\n"+
-				"\t--tags %s \\\n"+
-				"\t--role-name %s",
-				iamRoleTags, name)
+			policyTags := map[string]string{
+				tags.OpenShiftVersion: defaultPolicyVersion,
+			}
+			iamRoleTags := map[string]string{
+				tags.OpenShiftVersion: defaultPolicyVersion,
+			}
```
tbrisker, line 319: NP: these ares the same tags, could we use the same variable in both cases?

pkg/aws/commandbuilder/commandbuilder.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (b *CommandBuilder) AddParam(awsParam Param, value string) *CommandBuilder {
+	if value != "" {
+		b.params = append(b.params, createParamString(awsParam, value))
+	}
+	return b
+}
+
+func (b *CommandBuilder) AddTags(value map[string]string) *CommandBuilder {
+	b.tags = mergeMaps(b.tags, value)
```
tbrisker, line 78: NP: considering we don't need to combine multiple maps, and the target map is always the same one, you could inline the simpler case here:
```suggestion
	for k, v := range value {
		b.tags[k] = v
	}
```

pkg/aws/commandbuilder/commandbuilder.go
```diff
@@ -0,0 +1,126 @@
+package commandbuilder
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
+const PARAM_LINE_SEPARATOR = " \\\n"
```
tbrisker, line 9: NP: sorry for not noticing this earlier, but the convention we have in this project is not using `ALL_CAPS` for constant names. (also for the other consts below)

tbrisker (changes requested):
Noticed one small bug and a few nitpicks, sorry for not picking them up in the previous review cycle.

cmd/dlt/userrole/cmd.go
gdbranco: All other commands had this, I guess it is to space them out. Just wanted to make it standard in relation to others, most commands have two new lines, e.g creates and updates. For some reason deletes only have a single, maybe should add that so all are standardized.

cmd/upgrade/accountroles/cmd.go
gdbranco: Yeah I noticed that, not sure why it was like that already. But we can

tbrisker:
/lgtm
great work @gdbranco, thanks !

PR #912, opened 2022-10-27
[SDA-6857] Labels & taints strings containing only quotes ("""") are set to empty string

hunterkepley:
This change was brought about because you cannot set labels to an empty string (which is supposed to mean no labels) in interactive mode. When  you set the labels to `""` in interactive mode, the CLI thinks you are entering `""""` and does not work as expected.

This change just sets a string containing two quotes to an empty string to fix this problem.

### DIFF
--- a/pkg/interactive/interactive.go
+++ b/pkg/interactive/interactive.go
@@ -40,6 +40,7 @@ type Input struct {
 
 // Gets string input from the command line
 func GetString(input Input) (a string, err error) {
+	transformer := survey.TransformString(toEmptyString)
 	core.DisableColor = !color.UseColor()
 	dflt, ok := input.Default.(string)
 	if !ok {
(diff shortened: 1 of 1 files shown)


cmd/edit/machinepool/cmd.go
```diff
@@ -236,6 +236,11 @@ func run(cmd *cobra.Command, argv []string) {
 			os.Exit(1)
 		}
 	}
+
+	if labels == "\"\"" {
```
gdbranco, line 240: What if we apply a transformer on the survey input instead? Not sure if we ever want an escaped empty string when prompting anything interactively

https://github.com/AlecAivazis/survey/blob/master/transform_test.go

https://github.com/openshift/rosa/blob/ab250b83b4bafceaa1c91cba89d2eef2ce29e536/pkg/interactive/interactive.go#L60

hunterkepley: Good point, could someone let me know if this is the best path forward? I will make this change if so.

PR #922, opened 2022-11-04
FIPS: Unhide flag

vkareh:
By unhiding the FIPS flag we effectively expose it for all customers to provision FIPS-enabled clusters.

Related: https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/4803

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -301,7 +301,6 @@ func init() {
 		false,
 		"Create cluster that uses FIPS Validated / Modules in Process cryptographic libraries.",
 	)
-	flags.MarkHidden("fips")
 
 	flags.StringVar(
 		&args.httpProxy,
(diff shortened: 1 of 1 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -1759,14 +1758,14 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	fips := args.fips || fedramp.Enabled()
-	if interactive.Enabled() && fips && !fedramp.Enabled() {
```
wgordon17, line 1762: @vkareh what does this statement do? If we're removing the `fips` check, does that mean we _always_ enable FIPS? It should still be a use option, with the default to be False. It's _entirely_ possible and likely I'm just missing some other context ¯\_(ツ)_/¯

vkareh: This extra `fips` check was put in place so that the interactive "Do you want FIPS?" question showed up _only_ if the user manually specified the `--fips` flag. I removed it so that now any time the user is in interactive mode, the question will show up.

It doesn't change the actual value of the flag, just whether the interactive prompt shows.

wgordon17: Oh, perfect! That was the context I was missing, thanks!

wgordon17 (approved):
LGTM 👍

PR #924, opened 2022-11-07
[SDA-6443] Message deprecation non sts (legacy iam creds)

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-6643

# Changes
`rosa create cluster --sts --mode auto`
```
I: Enabling interactive mode
? Cluster name: test
W: --sts flag won't be necessary.
```

`rosa create cluster --mode auto`
```
I: Enabling interactive mode
? Cluster name: test
? Deploy cluster using AWS STS: No
W: --non-sts/--mint-mode flag will be necessary in a future release if you do not wish to use STS.
```

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -680,7 +680,7 @@ func run(cmd *cobra.Command, _ []string) {
 		r.Reporter.Warnf("--sts flag won't be necessary.")
 	}
 
-	if r.Reporter.IsTerminal() && (cmd.Flags().Changed("non-sts") || cmd.Flags().Changed("mint-mode")) {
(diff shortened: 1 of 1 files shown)


oriAdler:
Should we include `--non-sts` and `--mint-mode` flags in this MR?
As they don't take effect at the moment.

Maybe it is worth bringing back the logic for these flags if we add them,
or alternatively, hide the flags using `flags.MarkHidden()`

arendej:
> Should we include `--non-sts` and `--mint-mode` flags in this MR? As they don't take effect at the moment.
> 
> Maybe it is worth bringing back the logic for these flags if we add them, or alternatively, hide the flags using `flags.MarkHidden()`

We want to add those new flags as preparation for the future actual change of the default.

PR #927, opened 2022-11-08
[SDA-6987] Streamline cluster acc roles arns selection

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-6987
# What
Streamlines acc roles arns selection by bundling them up with the prefix. This way the client either selects a possible bundle that has all acc role arns and is compatible with chosen version or goes with custom selection over all arns.

# Why
There are lots of steps, conditions selecting the possible acc roles arns for sts cluster creation. Although custom arns still have a lot of code, the prefixed bundles are more streamlined

# Changes
`./rosa create cluster --sts --mode auto -y `
```
I: Enabling interactive mode
? Cluster name: gdb
? OpenShift version: 4.11.20
I: Fetching acc roles for the aws account: xxx
W: More than one account role bundle found
? Account role bundle prefix:  [Use arrows to move, type to filter]
  cse2etests
  dario-dev
  lponce-local
> managedopenshift
  oadler-policies
.
.
.
? Account role bundle prefix: managedopenshift
I: Using 'arn:aws:iam::xxx:role/ManagedOpenShift-Installer-Role' role as Installer role
I: Using 'arn:aws:iam::xxx:role/ManagedOpenShift-ControlPlane-Role' role as ControlPlane role
I: Using 'arn:aws:iam::xxx:role/ManagedOpenShift-Worker-Role' role as Worker role
I: Using 'arn:aws:iam::xxx:role/ManagedOpenShift-Support-Role' role as Support role
```


pvasant:
What is the impact of this new parameter in the old rosa cli ?

pvasant:
I want to ensure that none of the automation flows (just with prefixes, with all the arns in the args) are broken and that older cli is not broken.

gdbranco:
> I want to ensure that none of the automation flows (just with prefixes, with all the arns in the args) are broken and that older cli is not broken.

Don't think `create cluster` has a `--prefix` param, but the params for individual arns or for the operator prefix have not been touched.

hunterkepley (commented):
/lgtm

hunterkepley:
I have no suggestions, i like that you added a spinner to this. Hopefully someone else will have suggestions on what could be changed

cmd/create/cluster/cmd.go
```diff
@@ -764,105 +766,71 @@ func run(cmd *cobra.Command, _ []string) {
...
 		minor := ocm.GetVersionMinor(version)
-		role := aws.AccountRoles[aws.InstallerAccountRole]
 
+		var spin *spinner.Spinner
+		if r.Reporter.IsTerminal() {
+			spin = spinner.New(spinner.CharSets[9], 100*time.Millisecond)
+		}
+		if spin != nil {
+			if !output.HasFlag() && r.Reporter.IsTerminal() {
+				r.Reporter.Infof("Fetching acc roles for the aws account: %s", awsCreator.AccountID)
```
oriAdler, line 776: Since this is going to be displayed to the users:
```suggestion
				r.Reporter.Infof("Fetching account roles for the aws account: %s", awsCreator.AccountID)
```

pkg/aws/helpers.go
```diff
@@ -27,7 +27,23 @@ import (
 	rprtr "github.com/openshift/rosa/pkg/reporter"
 )
 
-var RoleNameRE = regexp.MustCompile(`^[\w+=,.@-]+$`)
+const RoleNameREString = `[\w+=,.@-]+`
+
+var RoleNameRE = regexp.MustCompile(fmt.Sprintf("^%s$", RoleNameREString))
+
+func getREOfAllAccRolesSuffixes() string {
```
oriAdler, line 34: What does `REO` stands for?

pkg/aws/sts.go
```diff
@@ -202,3 +204,73 @@ func BuildOperatorRoleCommands(prefix string, accountID string, awsClient Client
 	}
 	return commands
 }
+
+func (c *awsClient) FindAccRoleArnBundles(version string) (map[string]map[string]string, error) {
```
oriAdler, line 208: Can we break this function into smaller functions?
To make it easier to follow the logic :)

pkg/aws/helpers.go
gdbranco: Regular Expression Of, the acronym is actually just `RE`. I think I saw this being used somewhere else in the code, that is why

pkg/aws/sts.go
gdbranco: Sure, let me look into that

cmd/create/cluster/cmd.go
```diff
@@ -764,105 +766,71 @@ func run(cmd *cobra.Command, _ []string) {
...
-		}
-
-		if roleARN != "" {
-			// Get role prefix
-			rolePrefix, err := getAccountRolePrefix(roleARN, role)
-			if err != nil {
-				r.Reporter.Errorf("Failed to find prefix from %s account role", role.Name)
-				os.Exit(1)
+		} else {
+			possibleBundles := make([]string, 0, 1)
```
oriAdler, line 795: Why do we need the `possibleBundles` slice?
Can we sort the `currentArnBundles` and choose the first one to prompt to the user?

cmd/create/service/cmd.go
```diff
@@ -315,78 +316,38 @@ func run(cmd *cobra.Command, argv []string) {
...
-		roleARN = defaultRoleARN
-	} else if len(roleARNs) == 1 {
-		if !output.HasFlag() || r.Reporter.IsTerminal() {
-			r.Reporter.Infof("Using %q for the %s role", roleARNs[0], role.Name)
-		}
-		roleARN = roleARNs[0]
-	} else {
-		r.Reporter.Errorf("No account roles found. " +
+	if len(currentArnBundles) == 0 {
+		r.Reporter.Warnf("No account roles found. " +
```
oriAdler, line 327: Should we specify bundles?
Because the user might have plenty of account roles.
```suggestion
		r.Reporter.Warnf("No account roles bundles found. " +
```

pkg/aws/sts.go
```diff
@@ -202,3 +204,99 @@ func BuildOperatorRoleCommands(prefix string, accountID string, awsClient Client
...
+func (c *awsClient) FindAccountRoleArnBundles(version string) (map[string]map[string]string, error) {
+
+	roles, err := c.ListRoles()
+	if err != nil {
+		return map[string]map[string]string{}, err
+	}
+	prefixesCountMap := make(map[string]int)
+	prefixesARNMap := make(map[string]map[string]string)
+	for _, role := range roles {
+		foundPrefix, foundType, ok := isMatchWithAccountRolePrefix(role)
```
oriAdler, line 217: NP
```suggestion
		rolePrefix, roleType, ok := isMatchWithAccountRolePrefix(role)
```

pkg/aws/sts.go
```diff
@@ -202,3 +204,99 @@ func BuildOperatorRoleCommands(prefix string, accountID string, awsClient Client
...
+	foundType = matches[typeIndex]
+	return foundPrefix, foundType, true
+
+}
+
+func extractCompleteAccountBundles(prefixesCountMap map[string]int,
+	prefixesArnMap map[string]map[string]string) map[string]map[string]string {
+	bundles := make(map[string]map[string]string)
+	for key, value := range prefixesCountMap {
+		isBundleComplete := value == len(AccountRoles)*2
```
oriAdler, line 292: Why do we have to multiply by two?
A complete bundle contains four account role

gdbranco: The first checks if it finds the prefix 4 times one for each type of role. The second is that it checks the compatibility with the version supplied and fetches the actual arn. It is a hacky solution to only using one map. It can definitely be refactored. If you think it is worth it I'll change it as soon as I get through some other higher priorities

oriAdler: Ok, it's your call.
Maybe it is worth adding a comment about the calculation, just for other team members in the future.

cmd/create/cluster/cmd.go
gdbranco: The possible bundles only show the prefix key so the user can take a look at the possibilities when choosing from the interactive menu

cmd/create/service/cmd.go
gdbranco: Makes sense, will do

PR #930, opened 2022-11-10
[SDA-7003] Block --watch when on manual mode for creating cluster

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-7003

# What
Block watch param when on manual mode for creating cluster

# Why
User need to input outputted commands so we can't watch

# Changes
`rosa create cluster --sts --mode=manual --yes --watch`
```
I: Enabling interactive mode
? Cluster name: asd
W: --sts flag won't be necessary.
? OpenShift version: 4.11.12
E: Cannot watch in mode 'manual' since the user will need to run outputted commands.
```

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -807,8 +807,16 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	if args.watch && isSTS && mode == aws.ModeAuto && !confirm.Yes() {
-		r.Reporter.Errorf("Cannot watch for STS cluster installation logs in mode 'auto'." +
+		r.Reporter.Errorf("Cannot watch for STS cluster installation logs in mode 'auto' " +
+			"without also supplying '--yes' option." +
 			"To watch your cluster installation logs, run 'rosa logs install' instead after the cluster has began creating.")
+		os.Exit(1)
+	}
+
+	if args.watch && isSTS && mode == aws.ModeManual {
+		r.Reporter.Errorf("Cannot watch for STS cluster installation logs in mode 'manual'." +
+			"It requires manual commands to be performed as part of the process." +
+			"To watch your cluster installation logs, run 'rosa logs install' after the cluster has began creating.")
 		os.Exit(1)
 	}
 


cmd/create/cluster/cmd.go
```diff
@@ -742,6 +746,11 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	if args.watch && isSTS && mode == aws.ModeManual {
+		r.Reporter.Errorf("Cannot watch in mode 'manual' since the user will need to run outputted commands.")
```
gdbranco, line 750: Maybe yielded instead of outputted?

tbrisker: how about `Cannot watch in mode 'manual' as you will need to perform manual commands as part of the process`? but probably best to get feedback on exact phrasing from the reporter/bu

cmd/create/cluster/cmd.go
```diff
@@ -807,11 +807,19 @@ func run(cmd *cobra.Command, _ []string) {
...
+		r.Reporter.Errorf("Cannot watch for STS cluster installation logs in mode 'auto' " +
+			"without also supplying '--yes' option." +
 			"To watch your cluster installation logs, run 'rosa logs install' instead after the cluster has began creating.")
 		os.Exit(1)
 	}
 
+	if args.watch && isSTS && mode == aws.ModeManual {
+		r.Reporter.Errorf("Cannot watch for STS cluster installation logs in mode 'manual'." +
+			"It requires manual commands to be performed as part of the process." +
+			"To watch your cluster installation logs, run 'rosa longs install' after the cluster has began creating.")
```
oriAdler, line 819:
```suggestion
			"To watch your cluster installation logs, run 'rosa logs install' after the cluster has began creating.")
```
Would it be better to provide here the full command with `-c <cluster_name> --watch`?

gdbranco: The one above does not provide the full commands, so I'd rather keep this as it is the 'standard'

PR #933, opened 2022-11-16
Create a machine pool with a different region configured in the AWS CLI

oriAdler:
Initiate an AWS client with the cluster's region to fetch the region's subnets.

Related: [SDA-7285](https://issues.redhat.com//browse/SDA-7285)

image


oriAdler:

PR #934, opened 2022-11-16
Create machine pool - display spinner when fetching instance types

oriAdler:


oriAdler:

hunterkepley (approved):
LGTM, no suggestions on my side

gdbranco:
Not for this one, but I think it makes sense to have a helper that receives a func as param and adds a spinner to it. Just because we do this a few times and in some instances we are forgetting to add the spinner, specially in fetchs

PR #938, opened 2022-11-17
[SDA-7342] Fixing wrongful behaviors on upgrade roles

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-7342
* Remove os exit when roles are up to date
* Sort prefixes to ensure consistency when they are the same rank
* Order of messages
* Does not os exit when called from other command

### DIFF
--- a/cmd/upgrade/roles/cmd.go
+++ b/cmd/upgrade/roles/cmd.go
@@ -288,7 +288,7 @@ func run(cmd *cobra.Command, argv []string) error {
 	}
 
 	//OPERATOR ROLES
-	if !isInvokedFromClusterUpgrade {
+	if !args.isInvokedFromClusterUpgrade {
 		reporter.Infof("Ensuring operator role policies compatibility for upgrade")
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/upgrade/roles/cmd.go
```diff
@@ -338,18 +338,19 @@ func run(cmd *cobra.Command, argv []string) error {
...
+	if spin != nil {
+		spin.Stop()
+	}
+
 	if len(missingRolesInCS) <= 0 && !isOperatorPolicyUpgradeNeeded {
+		r.Reporter.Infof(
+			"Operator roles/policies associated with the cluster '%s' are already up-to-date.",
+			cluster.ID(),
+		)
 		if !args.isInvokedFromClusterUpgrade {
```
oriAdler, line 350: NP 
Not directly related to this MR, but can simplify:
```suggestion
		if args.isInvokedFromClusterUpgrade {
		    return nil
		}
		os.Exit(0)
```
WDYT?

pkg/helper/helpers.go
```diff
@@ -62,6 +62,13 @@ func RankMapStringInt(values map[string]int) []string {
 	for i, kv := range ss {
 		ranked[i] = kv.Key
 	}
+	sort.Slice(ranked, func(i, j int) bool {
```
oriAdler, line 65: To be honest I don't fully understand the `RankMapStringInt` function.
Could you please add a small example to the function description? or some explanation.

From my understanding, we are trying to find the most common prefix.
Would a `findMax` function will work here? to find the entry in the map with the maximal value?

gdbranco: I think this should help understand https://github.com/openshift/rosa/pull/935/files#diff-fb37be3d2646b1c87d4ba50f69ca7ee651bb3c4a0dec17820c622915d8975241R32

Finding a single entry with maximum value requires sorting to keep consistency of which element we will retrieve in case there are two entries with the same maximal value

oriAdler: Thanks for the explanation.
I would recommend adding some comments to **RankMapStringInt** or breaking it into smaller functions as we discussed.

PR #941, opened 2022-11-23
[SDA-7201] Always show warning, don't go into interactive if non sts mode is enabled

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-7201

# What
This PR changes behavior for mint-mode/non-sts flags so that it does not prompt choice of sts mode if those have been supplied. It also always show the warning message for future release use of STS mode.

# Why
User was being asked for STS mode and the default was Yes. This behavior should only happen if no flags for sts mode were supplied.

# Changes
`./rosa create cluster -c yuwan-ksr1 --mint-mode --sts -i`
```
I: Interactive mode enabled.
? Cluster name: yuwan-ksr1
E: Can't use both STS and mint mode at the same time.
```

`./rosa create cluster -c yuwan-ksr1 --sts -i`
```
I: Interactive mode enabled.
? Cluster name: yuwan-ksr1
```

`./rosa create cluster -c yuwan-ksr1 --mint-mode -i`
```
I: Interactive mode enabled.
? Cluster name: yuwan-ksr1
```

`./rosa create cluster -c yuwan-ksr1 --non-sts -i`
```
I: Interactive mode enabled.
? Cluster name: yuwan-ksr1
```

`./rosa create cluster -c yuwan-ksr1 -i`
```
I: Interactive mode enabled.
? Cluster name: yuwan-ksr1
? Deploy cluster using AWS STS: Yes
```

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -628,7 +628,12 @@ func run(cmd *cobra.Command, _ []string) {
 
 	// all hosted clusters are sts
 	isSTS := args.sts || args.roleARN != "" || fedramp.Enabled() || isHostedCP
-	isIAM := cmd.Flags().Changed("sts") && !isSTS
+	isIAM := (cmd.Flags().Changed("sts") && !isSTS) || args.nonSts
+
+	if isSTS && isIAM {
+		r.Reporter.Errorf("Can't use both STS and mint mode at the same time.")
+		os.Exit(1)
+	}
 
 	if interactive.Enabled() && (!isSTS && !isIAM) {
 		isSTS, err = interactive.GetBool(interactive.Input{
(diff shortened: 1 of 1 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -644,6 +649,14 @@ func run(cmd *cobra.Command, _ []string) {
 		isIAM = !isSTS
 	}
 
+	isSTS = isSTS || awsCreator.IsSTS
```
oriAdler, line 652: NP
Maybe we can append `|| awsCreator.IsSTS` to line `630`?
To have all the scenarios in one place.

cmd/create/cluster/cmd.go
```diff
@@ -644,6 +649,14 @@ func run(cmd *cobra.Command, _ []string) {
...
+
+	if r.Reporter.IsTerminal() {
+		r.Reporter.Warnf("In a future release STS will be the default mode.")
+		r.Reporter.Warnf("--sts flag won't be necessary if you wish to use STS.")
```
oriAdler, line 656: I've noticed that previously we printed the flag messages conditionally.
Should we preserve it with `if` statements?

gdbranco: I thought of removing it to bring more attention to the user. I'd users are using --sts, so they need to be alerted about the other option as well if they ever want to use it. Since the other option is a new param they would probably never see the message unless someone decides to check --help

cmd/create/cluster/cmd.go
gdbranco: I wasn't sure why this was after actually, maybe there's a reason. But if you are okay with it as well I'll gladly change it

oriAdler: Oh, I thought you've added it here.
Now I saw the deletion, never mind.

PR #942, opened 2022-11-23
[SDA-7382] Checking undefined aws region on init and create cluster

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-7382

# What
Non existing region should be validated

# Why
The command is stuck and returns only after 20 minutes return only after 20 minutes with not user-friendly error. 

# Changes
`./rosa create cluster -c region-check --region="undefined"`
```
E: Unsupported region 'undefined', available regions: [ap-east-1, eu-west-1, eu-west-2, eu-west-3, sa-east-1, us-east-1, us-east-2, us-west-1, us-west-2, af-south-1, ap-south-1, eu-north-1, eu-south-1, me-south-1, ca-central-1, eu-central-1, us-gov-east-1, us-gov-west-1, ap-northeast-1, ap-northeast-2, ap-northeast-3, ap-southeast-1, ap-southeast-2, ap-southeast-3]
```

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -551,7 +551,11 @@ func run(cmd *cobra.Command, _ []string) {
 	r := rosa.NewRuntime().WithOCM()
 	defer r.Cleanup()
 
-	awsClient := aws.GetAWSClientForUserRegion(r.Reporter, r.Logger)
+	supportedRegions, err := r.OCMClient.GetDatabaseRegionList()
+	if err != nil {
+		r.Reporter.Errorf("Unable to retrieve supported regions: %v", err)
+	}
+	awsClient := aws.GetAWSClientForUserRegion(r.Reporter, r.Logger, supportedRegions)
 
 	awsCreator, err := awsClient.GetCreator()
 	if err != nil {
@@ -1209,13 +1213,7 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	} else {
 		if isHostedCP && !interactive.Enabled() {
-			found := false
-			for i := range regionList {
-				if regionList[i] == region {
-					found = true
-					break
-				}
-			}
+			found := helper.Contains(regionList, region)
 			if !found {
 				r.Reporter.Errorf("Region '%s' not currently available for Hosted Cluster. "+
 					"Use --region to specify another region, and rosa list regions to see what's available.", region)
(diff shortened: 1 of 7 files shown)


pkg/aws/helpers.go
```diff
@@ -106,6 +108,17 @@ func GetRegion(region string) (string, error) {
 
 		region = *defaultSession.Config.Region
 	}
+	VPCRegions := awsenum.VPCRegion.Values("")
```
gdbranco, line 111: Not sure if using this is the best path https://github.com/aws/aws-sdk-go-v2/blob/main/service/route53/types/enums.go#L535

But it covers all current regions, so it is better than not validating or having a specific set of regions to validate against maintained on the CLI

What do you think about it?

pkg/aws/region/flag.go
```diff
@@ -36,11 +37,11 @@ func AddFlag(flags *pflag.FlagSet) {
 
 // Region returns a string with the name of the AWS region being used.
 func Region() string {
-	if region != "" {
+	if helper.HandleEscapedEmptyString(region) != "" {
 		return region
 	}
 	awsRegion := os.Getenv("AWS_REGION")
-	if awsRegion != "" {
+	if helper.HandleEscapedEmptyString(region) != "" {
```
oriAdler, line 44:
```suggestion
	if helper.HandleEscapedEmptyString(awsRegion) != "" {
```

pkg/aws/helpers.go
oriAdler: We have an endpoint in OCM to get all the available supported region per user account:
`ocm post /api/clusters_mgmt/v1/aws_inquiries/regions`
We use it right after calling this function `GetRegion`.

I'm not sure if we support all regions retrieved from `awsenum.VPCRegion.Values("")`.

cmd/create/cluster/cmd.go
```diff
@@ -551,7 +551,11 @@ func run(cmd *cobra.Command, _ []string) {
 	r := rosa.NewRuntime().WithOCM()
 	defer r.Cleanup()
 
-	awsClient := aws.GetAWSClientForUserRegion(r.Reporter, r.Logger)
+	supportedRegions, err := r.OCMClient.GetDatabaseRegionList()
```
oriAdler, line 554: NP
I wonder if we should encapsulate this into `GetAWSClientForUserRegion`, WDYT?

oriAdler (commented):
Left a small comment,
other than that LGTM

cmd/create/cluster/cmd.go
gdbranco: Cyclic import if not mistaken

gdbranco: That would be cyclic import, so we would need to refactor some more

PR #945, opened 2022-11-24
Move HostedCP region supports check to the backend side

andreadecorte:
Reduces permissions needed for the check and simplify code
Related: [SDA-7081](https://issues.redhat.com//browse/SDA-7081)


gdbranco:
LGTM

PR #950, opened 2022-11-28
Add new user to the owners list

oriAdler:


oriAdler:
@zgalor @gdbranco PTAL

OWNERS
```diff
@@ -9,6 +9,7 @@ reviewers:
 - oriAdler
 - gdbranco
 - renan-campos
+- tzvatot
```
gdbranco, line 12: I think we actually need to extra line

vkareh:
@tzvatot

> User is not a member of the org. User is not a collaborator. Satisfy at least one of these conditions to make the user trusted.

You need to join the OpenShift org.

PR #952, opened 2022-11-28
Check HostedCP version support also in interactive mode

andreadecorte:
Align version check with the backend to fail early
Related: [SDA-7087](https://issues.redhat.com//browse/SDA-7087)

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -693,7 +693,7 @@ func run(cmd *cobra.Command, _ []string) {
 	// OpenShift version:
 	version := args.version
 	channelGroup := args.channelGroup
-	versionList, err := getVersionList(r, channelGroup, isSTS)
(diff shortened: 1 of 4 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -712,11 +712,12 @@ func run(cmd *cobra.Command, _ []string) {
 			r.Reporter.Errorf("Expected a valid OpenShift version: %s", err)
 			os.Exit(1)
 		}
-	}
-	version, err = validateVersion(version, versionList, channelGroup, isSTS, isHostedCP)
-	if err != nil {
-		r.Reporter.Errorf("Expected a valid OpenShift version: %s", err)
-		os.Exit(1)
+	} else {
```
gdbranco, line 715: Any specific reason for the else?

andreadecorte: there is some duplication in the validation, but after double checking we can't just skip the second round. I removed it from the PR, it can be improved separately

PR #954, opened 2022-11-30
[SDA-7436] Ensure operator roles have attached policies

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-7436

# What
Ensures operator roles have expected attached policy

# Why
Upgrading of cluster could happen even though operator role didn't have attached policy

# Concerns
If the user used non standard names for policies we can't auto identify the correct policy it will fail and from my point of view requires user to manually check on aws to solve it


hunterkepley (commented):
/lgtm

PR #959, opened 2022-12-01
Release v1.2.10

oriAdler:
- Check HostedCP version support also in interactive mode + align versions
- fix: going through all policies to check upgrade


oriAdler:
@gdbranco @hunterkepley PTAL

gdbranco:
/lgtm

Since it's on hold I'll do this now until we get confirmation

PR #961, opened 2022-12-05
[SDA-7474] Upgrading from pre release would fail to validate version

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-7474

# What
Upgrading from a channel group other than stable would fail to validate version

# Why
Channel group was not being taken into account for some validations

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -2175,38 +2175,37 @@ func getAccountRolePrefix(roleARN string, role aws.AccountRole) (string, error) 
 // Validate OpenShift versions
 func validateVersion(version string, versionList []string, channelGroup string, isSTS,
 	isHostedCP bool) (string, error) {
-	if version != "" {
-		// Check and set the cluster version
-		hasVersion := false
-		for _, v := range versionList {
-			if v == version {
-				hasVersion = true
-			}
+	if version == "" {
+		return version, nil
+	}
+	// Check and set the cluster version
+	hasVersion := false
+	for _, v := range versionList {
+		if v == version {
+			hasVersion = true
 		}
-		if !hasVersion {
-			allVersions := strings.Join(versionList, " ")
-			err := fmt.Errorf("A valid version number must be specified\nValid versions: %s", allVersions)
-			return version, err
-		}
-
-		if isSTS && !ocm.HasSTSSupport(version, channelGroup) {
-			err := fmt.Errorf("Version '%s' is not supported for STS clusters", version)
-			return version, err
-		}
-		if isHostedCP {
-			valid, err := ocm.HasHostedCPSupport(version)
-			if err != nil {
-				return "", fmt.Errorf("error while parsing OCP version '%s': %v", version, err)
-			}
-			if !valid {
-				return "", fmt.Errorf("version '%s' is not supported for hosted clusters", version)
-			}
-		}
-
-		version = "openshift-v" + version
+	}
+	if !hasVersion {
+		allVersions := strings.Join(versionList, " ")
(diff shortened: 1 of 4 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -2175,38 +2175,37 @@ func getAccountRolePrefix(roleARN string, role aws.AccountRole) (string, error)
...
-			}
+		if !valid {
+			return "", fmt.Errorf("version '%s' is not supported for hosted clusters", version)
 		}
-
-		version = "openshift-v" + version
 	}
 
-	return version, nil
+	return ocm.CreateVersionID(version, channelGroup), nil
```
gdbranco, line 2208: CreateVersionID does the prefix of "openshift-v" and adds the suffix for channel group if it is different than stable

cmd/upgrade/cluster/cmd.go
```diff
@@ -201,7 +201,7 @@ func run(cmd *cobra.Command, _ []string) {
 	if isSTS {
 		r.Reporter.Infof("Ensuring account and operator role policies for cluster '%s'"+
 			" are compatible with upgrade.", cluster.ID())
-		err = roles.Cmd.RunE(roles.Cmd, []string{mode, cluster.ID(), version})
+		err = roles.Cmd.RunE(roles.Cmd, []string{mode, cluster.ID(), version, cluster.Version().ChannelGroup()})
```
gdbranco, line 204: Now takes into consideration channel group

cmd/upgrade/roles/cmd.go
```diff
@@ -109,9 +109,8 @@ func run(cmd *cobra.Command, argv []string) error {
 		aws.SetModeKey(argv[0])
 		ocm.SetClusterKey(argv[1])
 		skipInteractive = true
-		if len(argv) > 2 && argv[2] != "" {
-			args.clusterUpgradeVersion = argv[2]
-		}
+		args.clusterUpgradeVersion = argv[2]
```
gdbranco, line 112: Coming from cluster upgrade it should always have all the argvs, checking didn't make sense

pkg/ocm/versions.go
```diff
@@ -248,9 +248,12 @@ func IsValidVersion(userRequestedVersion string, supportedVersion string, cluste
 	if err != nil {
 		return false, err
 	}
+
+	isPreRelease := a.Prerelease() != "" && b.Prerelease() != ""
+
 	versionSplit := a.Segments64()
 	//If user has specified patch we check directly and return the result
-	if len(versionSplit) > 2 && versionSplit[2] > 0 {
+	if len(versionSplit) > 2 && versionSplit[2] > 0 || isPreRelease {
```
gdbranco, line 256: If the user has specified patch or is a pre release version we can compare directly

pkg/ocm/versions.go
```diff
@@ -248,9 +248,13 @@ func IsValidVersion(userRequestedVersion string, supportedVersion string, cluste
...
 	}
+
+	isPreRelease := a.Prerelease() != "" && b.Prerelease() != ""
+
 	versionSplit := a.Segments64()
-	//If user has specified patch we check directly and return the result
-	if len(versionSplit) > 2 && versionSplit[2] > 0 {
+	// If user has specified patch or not specified patch but is a preRelease
+	// Check directly
+	if len(versionSplit) > 2 && versionSplit[2] > 0 || (versionSplit[2] == 0 && isPreRelease) {
```
gdbranco, line 257: If it was a preRelease but patch was not 0 it was already taking into account. But if it is a pre release and patch is 0 we still need to check directly

hunterkepley (commented):
/lgtm

PR #962, opened 2022-12-07
Refactoring duplicated code

gdbranco:
* Removing lots of duplicate code
* Unifies missing operator functionality
* Adds deprecate warning for `upgrade operator-roles`


pkg/helper/roles/helpers.go
```diff
@@ -19,16 +18,17 @@ const (
 
 func GetOperatorRoleName(cluster *cmv1.Cluster, missingOperator *cmv1.STSOperator) string {
 	operatorIAMRoles := cluster.AWS().STS().OperatorIAMRoles()
-	rolePrefix := ""
-	if len(operatorIAMRoles) > 0 {
-		roleARN := operatorIAMRoles[0].RoleARN()
+	rolePrefix := aws.GenerateOperatorRolePrefix(cluster.Name())
```
gdbranco, line 21: This is actually pretty much the change from #846 and it is still needed

gdbranco: Dropped in favor of doing from the backend

cmd/create/service/cmd.go
```diff
@@ -401,7 +400,8 @@ func run(cmd *cobra.Command, argv []string) {
 	}
 
 	// operator role logic.
-	operatorRolesPrefix := getRolePrefix(args.ClusterName)
+	clusterName := strings.Trim(args.ClusterName, " \t")
```
oriAdler, line 403: Why do we need to trim the cluster name?

cmd/upgrade/operatorroles/cmd.go
```diff
@@ -204,37 +206,38 @@ func run(cmd *cobra.Command, argv []string) error {
 	}
 
 	if isOperatorPolicyUpgradeNeeded {
-		err = upgradeOperatorPolicies(mode, r, prefix, isAccountRoleUpgradeNeed,
-			policies, env, defaultPolicyVersion, credRequests, cluster, policyPath)
+		err = upgradeOperatorPolicies(
```
oriAdler, line 209: Should we stick to the current convention in the project and have all the parameters in the same line? 
Like the convention in CS.

cmd/create/service/cmd.go
gdbranco: This was already done on create/cluster but not on create/service. It is just to keep things closer since they do really close flows

PR #963, opened 2022-12-07
Log a warning if the user's organization doesn't have the needed capability

andreadecorte:
Fixes: [SDA-7486](https://issues.redhat.com//browse/SDA-7486)

### DIFF
--- a/cmd/list/region/cmd.go
+++ b/cmd/list/region/cmd.go
@@ -93,6 +93,10 @@ func run(cmd *cobra.Command, _ []string) {
 		r.Reporter.Errorf("%s", err)
 		os.Exit(1)
 	}
+	if !hypershiftEnabled && cmd.Flags().Changed("hosted-cp") {
+		r.Reporter.Errorf("'%s' not set for current organization", ocm.HypershiftCapability)
+		os.Exit(1)
+	}
 
 	// Filter out unwanted regions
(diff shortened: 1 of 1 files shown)


cmd/list/region/cmd.go
```diff
@@ -127,6 +127,9 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(0)
 	}
 
+	if !hypershiftEnabled && cmd.Flags().Changed("hosted-cp") {
+		r.Reporter.Warnf("'%s' not set for current organization", ocm.HypershiftCapability)
+	}
```
vkareh, line 132: Can this check be moved above to right after `IsCapabilityEnabled` is called? Otherwise we're doing a bunch of region support checks to just error out in the end.

andreadecorte: so far it was just a warning and not exiting, so it was good to have it there.
Changed to error with exit and moved up.

andreadecorte:
Output example
```
rosa list regions --hosted-cp
E: 'capability.organization.hypershift' not set for current organization
```

cmd/list/region/cmd.go
vkareh: Yes, I think this makes sense - otherwise the warning might just confuse users.

PR #969, opened 2022-12-12
Using lower case before comparing expected acc role arns

gdbranco:
# What
Uses lower case before comparison

# Why
AWS IAM names are case insensitive

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -815,9 +815,17 @@ func run(cmd *cobra.Command, _ []string) {
 					os.Exit(1)
 				}
 				selectedARN := ""
+				expectedResourceIDForAccRole := strings.ToLower(fmt.Sprintf("%s-%s-Role", rolePrefix, role.Name))
 				for _, rARN := range roleARNs {
-					if strings.Contains(rARN, fmt.Sprintf("%s-%s-Role", rolePrefix, role.Name)) {
+					resourceId, err := aws.GetResourceIdFromARN(rARN)
+					if err != nil {
+						r.Reporter.Errorf("Failed to get resource ID from arn. %s", err)
+						os.Exit(1)
+					}
+					lowerCaseResourceIdToCheck := strings.ToLower(resourceId)
+					if lowerCaseResourceIdToCheck == expectedResourceIDForAccRole {
 						selectedARN = rARN
+						break
 					}
 				}
 				if selectedARN == "" {


cmd/create/cluster/cmd.go
```diff
@@ -816,7 +816,7 @@ func run(cmd *cobra.Command, _ []string) {
 				}
 				selectedARN := ""
 				for _, rARN := range roleARNs {
-					if strings.Contains(rARN, fmt.Sprintf("%s-%s-Role", rolePrefix, role.Name)) {
+					if strings.Contains(strings.ToLower(rARN), strings.ToLower(fmt.Sprintf("%s-%s-Role", rolePrefix, role.Name))) {
 						selectedARN = rARN
```
tbrisker, line 820: np: we can break once we find a match

cmd/create/cluster/cmd.go
```diff
@@ -816,7 +816,7 @@ func run(cmd *cobra.Command, _ []string) {
 				}
 				selectedARN := ""
 				for _, rARN := range roleARNs {
-					if strings.Contains(rARN, fmt.Sprintf("%s-%s-Role", rolePrefix, role.Name)) {
+					if strings.Contains(strings.ToLower(rARN), strings.ToLower(fmt.Sprintf("%s-%s-Role", rolePrefix, role.Name))) {
```
tbrisker, line 819: np: we could move the expected name outside the inner loop instead of rebuilding it for every arn checked. Also, should this be `strings.HasSuffix` instead of `Contains`?
And not directly related, but happens if the role name is truncated due to hitting the maximum length limit?

gdbranco: For the acc roles in particular there is a validation to limit it to 32 characters so it would always have the suffix

tbrisker: the complexity isn't too much of a consideration in this case, any implementation would work fast enough that it isn't noticeable. The comment was more regarding expressing the intent more clearly - iiuc, we expect the role arn to have a certain suffix, so we should use a method that expresses this expectation.

gdbranco: We can also do this, what do you think?
https://github.com/openshift/rosa/pull/969/files#diff-943c0ab57b82c7028652df55e1d49e4f4e4dc41fa094dc4a9b3f2c35887d44b5R826

tbrisker: even better!

PR #974, opened 2022-12-15
Allow editing default machine pool labels

zgalor:
Related: [SDA-7375](https://issues.redhat.com//browse/SDA-7375)

Also fixed some issues with interactive mode, in the additional machine pool. Before the fix, if the user sets only labels or taints, they we still be asked to set replicas, after the fix the user can set labels, taints (not for default machine pool), or replicas/autoscaling params.

```
$ ./rosa edit machinepool -c zg-dec18  Default
? Labels: a=b
I: Updated machine pool 'Default' on cluster 'zg-dec18'
```

If the user set at least one of those they will not be prompted for anything more
```
$ ./rosa edit machinepool -c zg-dec18  --labels "a=f,b=w" Default
I: Updated machine pool 'Default' on cluster 'zg-dec18'
```

Taints are forbidden for default mp:
```
$ ./rosa edit machinepool -c zg-dec18  --taints "a=f,b=w" Default
E: Taints are not supported on the Default machine pool
```

### DIFF
--- a/cmd/edit/machinepool/machinepool.go
+++ b/cmd/edit/machinepool/machinepool.go
@@ -26,43 +26,62 @@ func editMachinePool(cmd *cobra.Command, machinePoolID string, clusterKey string
 		os.Exit(1)
 	}
 
+	isMinReplicasSet := cmd.Flags().Changed("min-replicas")
+	isMaxReplicasSet := cmd.Flags().Changed("max-replicas")
+	isReplicasSet := cmd.Flags().Changed("replicas")
+	isAutoscalingSet := cmd.Flags().Changed("enable-autoscaling")
+	isLabelsSet := cmd.Flags().Changed("labels")
+	isTaintsSet := cmd.Flags().Changed("taints")
+
+	// if no value set enter interactive mode
+	if !(isMinReplicasSet || isMaxReplicasSet || isReplicasSet || isAutoscalingSet || isLabelsSet || isTaintsSet) {
+		interactive.Enable()
+	}
+
 	// Editing the default machine pool is a different process
 	if machinePoolID == "Default" {
-		if cmd.Flags().Changed("labels") {
-			r.Reporter.Errorf("Labels cannot be updated on the Default machine pool")
-			os.Exit(1)
-		}
-		if cmd.Flags().Changed("taints") {
+		if isTaintsSet {
 			r.Reporter.Errorf("Taints are not supported on the Default machine pool")
 			os.Exit(1)
 		}
 
-		autoscaling, replicas, minReplicas, maxReplicas := getMachinePoolReplicas(cmd, r.Reporter, machinePoolID,
-			cluster.Nodes().Compute(), cluster.Nodes().AutoscaleCompute())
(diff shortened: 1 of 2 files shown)


zgalor:
@oriAdler @gdbranco @hunterkepley @tzvatot PTAL.
Even if LGTY, please don't merge before the weekend, as this also impacts the existing edit machine pool flow

cmd/edit/machinepool/machinepool.go
```diff
@@ -26,43 +26,62 @@ func editMachinePool(cmd *cobra.Command, machinePoolID string, clusterKey string
...
+					(autoscaling && isMinReplicasSet && minReplicas < 3) {
+					r.Reporter.Errorf("Default machine pool for AZ cluster requires at least 3 compute nodes")
+					os.Exit(1)
+				}
 
-		if cluster.MultiAZ() {
-			if !autoscaling && replicas < 3 ||
-				(autoscaling && cmd.Flags().Changed("min-replicas") && minReplicas < 3) {
-				r.Reporter.Errorf("Default machine pool for AZ cluster requires at least 3 compute nodes")
+				if !autoscaling && replicas%3 != 0 ||
```
oriAdler, line 62: I think this condition applies only to multi-AZ clusters.

cmd/edit/machinepool/machinepool.go
```diff
@@ -96,56 +115,32 @@ func editMachinePool(cmd *cobra.Command, machinePoolID string, clusterKey string
...
 
-	autoscaling, replicas, minReplicas, maxReplicas := getMachinePoolReplicas(cmd, r.Reporter, machinePoolID,
-		machinePool.Replicas(), machinePool.Autoscaling())
-
-	if !autoscaling && replicas < 0 ||
-		(autoscaling && cmd.Flags().Changed("min-replicas") && minReplicas < 0) {
-		r.Reporter.Errorf("The number of machine pool replicas needs to be a non-negative integer")
-		os.Exit(1)
+	// if no value set enter interactive mode
+	if !(isMinReplicasSet || isMaxReplicasSet || isReplicasSet || isAutoscalingSet || isLabelsSet || isTaintsSet) {
```
oriAdler, line 119: Should we extract this check to line `35`?
Then we don't have to check the condition for default machine pool and non-default.

cmd/edit/machinepool/machinepool.go
zgalor: That is why it's under the `if` on line 55

cmd/edit/machinepool/machinepool.go
zgalor: I see your point, but I want to make sure we won't miss anything there.

Let's see:
If the user doesn't set anything - enable interactive.
If the user sets taints for default they will get an error.

Yes, I think that would work.
Changing

zgalor:
@oriAdler rePTAL

cmd/edit/machinepool/machinepool.go
zgalor: fixed

PR #976, opened 2022-12-15
fix: Incorrect OIDC Provider Sometimes Targeted for Deletion

jjulien:
Fixes a bug where the wrong OIDC Provider may be targeted for removal when more than one OIDC Provider exists and a request to delete an OIDC Provider by Cluster Name is issued, and the Cluster Name specified either does not exist, or belongs to a cluster in a ready state.

### DIFF
--- a/cmd/dlt/oidcprovider/cmd.go
+++ b/cmd/dlt/oidcprovider/cmd.go
@@ -78,16 +78,19 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
-	clusterID := clusterKey
 	if sub != nil {
-		clusterID = sub.ClusterID()
+		clusterKey = sub.ClusterID()
 	}
-	c, err := r.OCMClient.GetClusterByID(clusterID, r.Creator)
+	c, err := r.OCMClient.GetCluster(clusterKey, r.Creator)
 	if err != nil {
 		if errors.GetType(err) != errors.NotFound {
 			r.Reporter.Errorf("Error validating cluster '%s': %v", clusterKey, err)
 			os.Exit(1)
+		} else if sub == nil {
+			r.Reporter.Errorf("Failed to get cluster '%s': %v", r.ClusterKey, err)
+			os.Exit(1)
 		}
+
 	}
 	if c != nil && c.ID() != "" {
 		r.Reporter.Errorf("Cluster '%s' is in '%s' state. OIDC provider can be deleted only for the "+


cmd/dlt/oidcprovider/cmd.go
```diff
@@ -78,16 +78,19 @@ func run(cmd *cobra.Command, argv []string) {
...
-		clusterID = sub.ClusterID()
+		clusterKey = sub.ClusterID()
 	}
-	c, err := r.OCMClient.GetClusterByID(clusterID, r.Creator)
+	c, err := r.OCMClient.GetCluster(clusterKey, r.Creator)
 	if err != nil {
 		if errors.GetType(err) != errors.NotFound {
 			r.Reporter.Errorf("Error validating cluster '%s': %v", clusterKey, err)
 			os.Exit(1)
+		} else if sub == nil {
```
gdbranco, line 89: Hey there sorry for taking a while to get to this and commenting.

Not sure if I follow this check, could you provide some more info in why this is needed?

cmd/dlt/oidcprovider/cmd.go
```diff
@@ -78,16 +78,19 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
-	clusterID := clusterKey
 	if sub != nil {
-		clusterID = sub.ClusterID()
+		clusterKey = sub.ClusterID()
 	}
-	c, err := r.OCMClient.GetClusterByID(clusterID, r.Creator)
+	c, err := r.OCMClient.GetCluster(clusterKey, r.Creator)
```
gdbranco, line 84: I was looking into this and it seems that this function already looks up the subscription as well. I'm thinking we can simplify this whole flow doing only GetCluster instead of also doing lines from 70 through 79 too. What do you think?

cmd/dlt/oidcprovider/cmd.go
jjulien: Hey @gdbranco , no worries, figured end of year is a pretty busy time for all.

The `if sub == nil` check here is to identify if the cluster ever existed at all.  `GetCluster` and `GetClusterByID` will return `errors.NotFound` for a cluster that has been deleted. The assumption is that `errors.NotFound` means the cluster has been deprovisioned and is thus in a state where the OIDC Provider can be deleted, and that `sub` can be used to find the value of the cluster ID.

This is not an accurate assumption for a cluster that never existed at all.  If `GetCluster` cannot find the cluster and `sub == nil`, then the cluster never existed, and we should not attempt to delete an OIDC Provider.

Without the `sub == nil` check, ` r.AWSClient.GetOpenIDConnectProvider(sub.ClusterID())` is called, which results in an empty string being sent, causing the first OIDC provider found by the AWS API to be returned and later deleted.

You can see that behavior by attempting to delete the OIDC provider for a cluster that never existed, in an account where at least one OIDC provider does exist.
```
rosa delete oidc-provider -m manual -c cluster-that-never-existed-at-all
I: Run the following commands to delete the OIDC provider:

aws iam delete-open-id-connect-provider \
	--open-id-connect-provider-arn arn:aws:iam::1111222233334444:oidc-provider/rh-oidc.s3.us-east-1.amazonaws.com/xxxxxxxxxxxxxx
```

cmd/dlt/oidcprovider/cmd.go
jjulien: This was actually my first thought as well.  I even opened a PR for it, and then closed it after I discovered why the sub lookup is important.

`GetCluster` will not return a cluster that has been deprovisioned, so you cannot use it to lookup the ClusterId of a cluster that no longer exists.  Only the subscription can tell you what the ID was for a deprovisioned cluster.

We can't rely on just the subscription returned from `GetClusterUsingSubscription` either, because this function will only return subscriptions where `status = 'Deprovisioned'`, it will not return active clusters.

I thought about changing the logic of `GetClusterUsingSubscription` to not filter by `status = 'Deprovisioned'`, but since this is also used by the `rosa delete operator-roles` command, and that command is behaving properly, I thought it better to provide the logic fixes to `rosa delete oidc-provider` instead.

Not filtering by `status = 'Deprovisioned'` would also change current behavior of the `rosa delete oidc-provider` command.  With the current behavior, if you have 1 active cluster, and 1 deprovisioned cluster, both using the same name, the `rosa delete oidc-provider` command will find and delete the correct OIDC provider.   If `GetClusterUsingSubscription` were modified to no longer filter by `status = 'Deprovisioned'`, then an active and deprovisioned cluster existing with the same name would now result in `errors.Conflict` being thrown.

This would cause the command to fail in a scenario, where it was previously successful.

gdbranco:
/assign @vkareh

Hey Victor, I think it looks good to me and I've talked to @pvasant as well and she agrees with solution. Could you take a look and see if you approve?

cmd/dlt/oidcprovider/cmd.go
vkareh: Would it then make sense to fail early when fetching the sub? I mean in line 70 above:
```
sub, err := r.OCMClient.GetClusterUsingSubscription(clusterKey, r.Creator)
if sub == nil {
  r.Reporter.Errorf("Failed to get cluster '%s': Subscription not found", r.ClusterKey)
  os.Exit(1)
}
```

jjulien: @vkareh Did my explanation make sense?  I think as the PR stands currently, it's the least amount of code, touching only the parts of the logic that are currently broken.

If you'd rather take a step back and look at a broader change to the logic to make it more clear, let me know and I can work on that too.  Even just refactoring `GetClusterUsingSubscription` to `GetDeprovisionedClusterUsingSubscription` would make it a bit more clear.

vkareh: @jjulien yes, this makes perfect sense. Thanks for the clarification!

PR #977, opened 2022-12-16
[SDA-7580] Adjust NodePool headers

lucasponce:
Discussed not using "NodePools" wording for "MachinePool for hosted clusters".


lucasponce:
cc @gdbranco

PR #980, opened 2022-12-20
[SDA-7565] Use default/in-place value for addon param first

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-7565
# What
Populates in place/default value of param before checking values from arguments

# Why
Params not specified in the args were being sent as empty or nil


renan-campos:
LGTM

PR #981, opened 2022-12-20
Add default machine pool labels validations

zgalor:
Related: [SDA-7377](https://issues.redhat.com//browse/SDA-7377)

1. Not hypershift cluster
2. No label key duplications

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -640,6 +640,11 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
+	if isHostedCP && cmd.Flags().Changed("default-mp-labels") {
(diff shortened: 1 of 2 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -1682,8 +1682,13 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	// Default machine pool labels
+	if isHostedCP && cmd.Flags().Changed("default-mp-labels") {
```
oriAdler, line 1685: NP
I think we can block the user earlier when we determine the value of `isHostedCP` in line `629`.

oriAdler (commented):

cmd/create/cluster/cmd.go
zgalor: fixed

PR #982, opened 2022-12-20
[SDA-7593] Disabled `Replicas` question when autoscaling is enabled

hunterkepley:
[SDA-7593](https://issues.redhat.com//browse/SDA-7593) is a critical bug where ROSA incorrectly prompts users how many replicas they'd like to set when autoscaling is enabled.

### DIFF
--- a/cmd/edit/machinepool/machinepool.go
+++ b/cmd/edit/machinepool/machinepool.go
@@ -294,6 +294,7 @@ func getMachinePoolReplicas(cmd *cobra.Command,
 			}
 		}
 		scalingUpdated = true
+	} else {
 		if !isReplicasSet {
 			replicas = existingReplicas
 		}


oriAdler:
LGTM

cmd/edit/machinepool/machinepool.go
```diff
@@ -294,6 +294,7 @@ func getMachinePoolReplicas(cmd *cobra.Command,
 			}
 		}
 		scalingUpdated = true
+	} else {
```
lucasponce, line 297: Can we check if this fix is needed in the hostedcluster logic?

https://github.com/openshift/rosa/blob/master/cmd/edit/machinepool/nodepool.go

gdbranco: https://github.com/openshift/rosa/blob/master/cmd/edit/machinepool/nodepool.go#L48

Doesn't look like so, what do you think?

gdbranco: Btw, this seems to be the exact same code for both
https://github.com/openshift/rosa/blob/master/cmd/edit/machinepool/nodepool.go#L67

We should probably add a separate file for things like these

lucasponce: Yes, but once we consolidate all the flows.
I'd keep it separate as machinepools for hosted clusters don't share 100% features for regular clusters.

And for hosted clusters, we are adding features in an incremental way, so when all features are consolidated, I'm 100% in refactoring and organizing common params.

But the motivation of this split is to be sure that the classic machinepools are untouched while working on the hosted clusters area (with the cost of a little well known duplication)

hunterkepley: @lucasponce Anything I need to do or does this LGTY?

gdbranco: @hunterkepley Just to keep consistency when we decide to separate common code please change it to 
`else if interactive.Enabled() || !isReplicasSet {` as it is in nodepools :)

hunterkepley: @gdbranco Sounds good, doing it now

hunterkepley:
Solved here https://github.com/openshift/rosa/pull/990

PR #986, opened 2022-12-26
Refactor `GetPolicies` function

oriAdler:
Refactor the function to return a map with `AWSSTSPolicy` values, 
in order to use the `ARN` attribute for managed policies in the future.

Related: [SDA-7609](https://issues.redhat.com/browse/SDA-7609)

### DIFF
--- a/pkg/aws/helpers.go
+++ b/pkg/aws/helpers.go
@@ -570,12 +570,13 @@ func GetInstallerAccountRoleName(cluster *cmv1.Cluster) (string, error) {
 }
 
 func GeneratePolicyFiles(reporter *rprtr.Object, env string, generateAccountRolePolicies bool,
-	generateOperatorRolePolicies bool, policies map[string]string, credRequests map[string]*cmv1.STSOperator) error {
+	generateOperatorRolePolicies bool, policies map[string]*cmv1.AWSSTSPolicy,
+	credRequests map[string]*cmv1.STSOperator) error {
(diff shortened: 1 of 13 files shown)


pkg/aws/helpers.go
```diff
@@ -762,3 +762,12 @@ func FindOperatorRoleBySTSOperator(operatorRoles []*cmv1.OperatorIAMRole, operat
 	}
 	return ""
 }
+
+func GetSTSPolicyDetails(policies map[string]*cmv1.AWSSTSPolicy, key string) string {
```
oriAdler, line 766: This function preserves the current behavior, as we used a map of `[string]string` for policies.
If a key isn't in the map, the value retrieved from the map is `""`.

pkg/ocm/helpers.go
```diff
@@ -518,7 +518,7 @@ func (c *Client) GetPolicies(policyType string) (map[string]string, error) {
 		return m, handleErr(accountRolePoliciesResponse.Error(), err)
 	}
 	accountRolePoliciesResponse.Items().Each(func(awsPolicy *cmv1.AWSSTSPolicy) bool {
-		m[awsPolicy.ID()] = awsPolicy.Details()
+		m[awsPolicy.ID()] = awsPolicy
```
oriAdler, line 521: This is a preparation for managed policies, where we need the `ARN` attribute as well, 
therefore returning the whole struct instead of only the details field.

PR #988, opened 2022-12-27
Create account roles with existing policies

oriAdler:
This is a preparation for managed policies.

1. Validate the environment supports the `managed` flag (not production).
2. Don't output the `create policy` commands in manual mode.
3. Validate there aren't existing unmanaged roles with the same names.
4. Add a new role tag `rosa_managed_policies` to managed roles.

**Note:** In order to use this command one has to create policies in the AWS account
with these ARNs https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/pkg/aws/loader/aws_sts_managed_policies.go#L6.

Related: [SDA-7656](https://issues.redhat.com/browse/SDA-7656)

Validate environment:
image

Don't display `create policy` commands in the manual mode:
image

Check if there are exiting unmanaged roles with the same names:
image

### DIFF
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -43,6 +43,7 @@ var args struct {
 	path                string
 	version             string
 	channelGroup        string
+	managed             bool
 }
 
 var Cmd = &cobra.Command{
@@ -98,6 +99,15 @@ func init() {
 	)
 	flags.MarkHidden("channel-group")
 
+	// TODO: change to `unmanaged` once AWS managed policies are in place (managed will be the default)
+	flags.BoolVar(
+		&args.managed,
+		"managed",
+		false,
+		"Attach AWS managed policies to the account roles",
+	)
+	flags.MarkHidden("managed")
+
 	aws.AddModeFlag(Cmd)
 
 	confirm.AddFlag(flags)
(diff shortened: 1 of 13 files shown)


oriAdler:
@gdbranco @zgalor PTAL

cmd/create/accountroles/cmd.go
```diff
@@ -130,6 +140,14 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
+	// Determine if managed policies are enabled
+	isManagedSet := cmd.Flags().Changed("managed")
+	if isManagedSet && env == "production" {
```
zgalor, line 145: np: const?

cmd/create/accountroles/cmd.go
```diff
@@ -382,32 +405,38 @@ func createRoles(r *rosa.Runtime, prefix, permissionsBoundary, accountID, env st
...
-			}, path)
-		if err != nil {
-			return err
+		if managedPolicies {
+			policyARN, err = aws.GetSTSPolicyARN(policies, filename)
+			if err != nil {
+				return err
+			}
+		} else {
+			policyDetail = aws.GetSTSPolicyDetails(policies, filename)
```
zgalor, line 430: this var should be used inside this block only

cmd/create/accountroles/cmd.go
oriAdler: Changed.

cmd/create/accountroles/cmd.go
oriAdler: Fixed.

pkg/aws/helpers.go
```diff
@@ -581,25 +581,27 @@ func GeneratePolicyFiles(reporter *rprtr.Object, env string, generateAccountRole
...
+				if policyDetail == "" {
+					continue
+				}
+				//Check and save it as json file
+				filename = GetFormattedFileName(filename)
+				reporter.Debugf("Saving '%s' to the current directory", filename)
+				err = helper.SaveDocument(policyDetail, filename)
+				if err != nil {
+					return err
+				}
```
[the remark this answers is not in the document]
oriAdler: Done.

gdbranco:
I think it looks good to me. But I'm not being too fond of adding more params to all those functions.

It is nitpicking but I think next time we need to add anything we actually need to change it to receive some structure with all the data instead, and also work into removing conditions in favor of factory alongside strategy pattern, just cause this is getting harder to read.

PR #990, opened 2022-12-28
Fix scaling bug and improve interactive mode

zgalor:
Reltaed: [SDA-7651](https://issues.redhat.com//browse/SDA-7651)

Also, don't prompt for scaling params if minimal update cluster params were set


zgalor:
@tzvatot @oriAdler @gdbranco PTAL

PR #991, opened 2022-12-29
Fix bug create account roles with existing policies

oriAdler:
Related: https://issues.redhat.com/browse/SDA-7656

Create managed account roles:
image

Create unmanaged account roles:
image

### DIFF
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -322,8 +322,12 @@ func run(cmd *cobra.Command, argv []string) {
 			})
 			os.Exit(1)
 		}
-		commands := buildCommands(prefix, permissionsBoundary, r.Creator.AccountID, policyVersion,
-			path, managedPolicies)
+		commands, err := buildCommands(prefix, permissionsBoundary, r.Creator.AccountID, policyVersion,
+			path, managedPolicies, policies)
+		if err != nil {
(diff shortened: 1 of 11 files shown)


oriAdler:
@zgalor @gdbranco PTAL

cmd/create/accountroles/cmd.go
```diff
@@ -398,7 +414,7 @@ func createRoles(r *rosa.Runtime, prefix, permissionsBoundary, accountID, env st
 		}
 
 		filename := fmt.Sprintf("sts_%s_trust_policy", file)
-		policyDetail := aws.GetSTSPolicyDetails(policies, filename)
+		policyDetail := aws.GetManagedPolicyDetails(policies, filename)
```
zgalor, line 417: Always get the managed policy? Even if the bool is false?

oriAdler: Good catch, lousy naming, updated to `GetPolicyDetails`.
`GetPolicyDetails` applies both for unmanaged and managed policies, the function was added for this change https://github.com/openshift/rosa/pull/986/files#diff-97477d2275d7797e234adfc6164c02beabfdbdb0eb516faed787f52d1cfb30c1R521.

The other function `GetManagedPolicyARN` applies only for managed policies.

PR #994, opened 2023-01-02
Refactor sort strings helper

gdbranco:
# What
Extracts sorting string with respects to length to it's own function

# Why
Cleaning up duplication


gdbranco:
@oriAdler PTAL

PR #995, opened 2023-01-02
[SDA-4764] Hide region from other globally available commands

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-4764
# What
Hides region in globally available commands, basically what deals with aws iam service

# Why
Globally available commands should have region hidden in help


gdbranco:
@oriAdler PTAL refactoring #853 to include all iam commands and easier to follow flow

PR #1001, opened 2023-01-04
Upgrade account roles with managed policies

oriAdler:
If the roles have the tag `rosa_managed_policies:true`, verify that the managed policies are attached and exit. 
Policies upgrade isn't needed for managed policies.

Related: [SDA-7678](https://issues.redhat.com/browse/SDA-7678)

Examples:
1. User provides invalid prefix:
image
2.  User upgrades roles with managed policies successfully:
image
3. One of the roles is missing the attached managed policy:
image
4. Managed policies are disabled on production:
image

### DIFF
--- a/cmd/create/operatorroles/cmd.go
+++ b/cmd/create/operatorroles/cmd.go
@@ -207,7 +207,7 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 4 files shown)


hunterkepley (commented):
This LGTM but i would like a second opinion if possible, please someone else add /lgtm if it does look fine

PR #1005, opened 2023-01-06
Fix OAuth callback URL logic for HyperShift clusters

andreadecorte:
The implementation leverages ExternalDNS which will be the default for all the clusters. 
No change for the non HyperShift clusters

Fixes: [SDA-7648](https://issues.redhat.com//browse/SDA-7648)


cmd/create/cluster/cmd_test.go
```diff
@@ -27,7 +27,7 @@ var _ = Describe("Validates OCP version", func() {
 			v, err := validateVersion("4.12.0-0.nightly-2022-11-25-185455-nightly",
 				[]string{"4.12.0-0.nightly-2022-11-25-185455-nightly"}, nightly, false, true)
 			Expect(err).NotTo(HaveOccurred())
-			Expect(v).To(Equal("openshift-v4.12.0-0.nightly-2022-11-25-185455-nightly"))
+			Expect(v).To(Equal("openshift-v4.12.0-0.nightly-2022-11-25-185455-nightly-nightly"))
```
gdbranco, line 30: Thanks for noticing this

PR #1006, opened 2023-01-06
[SDA-6753] Now updating stsBuilder.AutoMode properly

hunterkepley:
[SDA-6753](https://issues.redhat.com//browse/SDA-6753) was a bug where in ROSA when you created a cluster using auto mode (`rosa create cluster --sts --mode auto` for example), the "credential_configuration_mode" Tracking property was set to `manual`. It is now updated properly by updating `stsBuilder.AutoMode` with the correct value depending on if it's auto mode or manual mode.

This change was tested using Segment and creating a cluster.

### DIFF
--- a/pkg/ocm/clusters.go
+++ b/pkg/ocm/clusters.go
@@ -36,6 +36,7 @@ import (
 )
 
 var NetworkTypes = []string{"OpenShiftSDN", "OVNKubernetes"}
+var modeAuto = "auto"
 
 // Spec is the configuration for a cluster spec.
 type Spec struct {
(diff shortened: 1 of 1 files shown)


gdbranco:
Nitpicking: We should have auto as a const somewhere at some point

hunterkepley:
> Nitpicking: We should have auto as a const somewhere at some point

I wholeheartedly agree, let me see if i can find a good place for it right now

hunterkepley:
> Nitpicking: We should have auto as a const somewhere at some point

I wholeheartedly agree, let me see if i can find a good place for it right now @gdbranco

gdbranco:
> > Nitpicking: We should have auto as a const somewhere at some point
> 
> I wholeheartedly agree, let me see if i can find a good place for it right now @gdbranco

Is it possible to use this one?
https://github.com/openshift/rosa/blob/49ef038958eef794bea1d6e3f9065f9367b6e318/pkg/aws/flag.go#L28

hunterkepley:
> > > Nitpicking: We should have auto as a const somewhere at some point
> > 
> > 
> > I wholeheartedly agree, let me see if i can find a good place for it right now @gdbranco
> 
> Is it possible to use this one?
> 
> https://github.com/openshift/rosa/blob/49ef038958eef794bea1d6e3f9065f9367b6e318/pkg/aws/flag.go#L28

Using it now :) Please review when you get the chance

PR #1007, opened 2023-01-06
reprompt user when passed invalid addon argument

renan-campos:
The interactive mode was terminating if the user passed an invalid string argument when installing an addon.
This set of commits modifies the behavior of _interactive.GetAddonArgument()_ to continue prompting the user.

[SDA-6852](https://issues.redhat.com/browse/SDA-6852)


renan-campos:
@laurafitzgerald PTAL

ciaranRoche (commented):
From testing this it hangs if you repeatedly enter the wrong value

renan-campos:
> From testing this it hangs if you repeatedly enter the wrong value

pushed a change addressing this issue and the problem with the default value.

PR #1008, opened 2023-01-08
Upgrade operator roles with managed policies

oriAdler:
1. Check that managed policies are attached to the roles.
2. If the user specified the upgrade version, check for missing roles.

Related: [SDA-7679](https://issues.redhat.com/browse/SDA-7679)

image
image

### DIFF
--- a/cmd/upgrade/roles/cmd.go
+++ b/cmd/upgrade/roles/cmd.go
@@ -984,6 +984,8 @@ func createOperatorRole(
 			policies,
 			unifiedPath,
 			operatorRolePolicyPrefix,
+			// TODO: pass the actual managed policies value when the command is updated
+			false,
 		)
 		if err != nil {
 			return err
--- a/pkg/aws/commandbuilder/helper/roles/roles.go
+++ b/pkg/aws/commandbuilder/helper/roles/roles.go
@@ -16,6 +16,7 @@ type ManualCommandsForMissingOperatorRolesInput struct {
 	Filename                 string
 	RolePath                 string
 	PolicyARN                string
+	ManagedPolicies          bool
 }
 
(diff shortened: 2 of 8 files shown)


oriAdler:

pkg/aws/commandbuilder/helper/roles/roles.go
```diff
@@ -27,6 +28,9 @@ func ManualCommandsForMissingOperatorRole(input ManualCommandsForMissingOperator
 		tags.OperatorName:      input.Operator.Name(),
 		tags.RedHatManaged:     "true",
 	}
+	if input.ManagedPolicies {
+		iamTags[tags.ManagedPolicies] = "true"
```
zgalor, line 32: I would love replace the `if` something like 
```
iamTags[tags.ManagedPolicies] = fmt.Sprintf("%v", input.ManagedPolicies)
```

But we probably don't want the tag if it's false, right?

oriAdler: Yes, since all existing roles are non-managed and don't have the tag.
To be consistent with existing roles.
Otherwise, we'll have to check later if the tags exist and their values, currently, if the tag exists it's managed.

cmd/upgrade/roles/cmd.go
```diff
@@ -984,6 +984,7 @@ func createOperatorRole(
 			policies,
 			unifiedPath,
 			operatorRolePolicyPrefix,
+			false,
```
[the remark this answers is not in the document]
oriAdler: Added a TODO item above this line.
The next command I'm working on is `rosa upgrade roles`, and I'm going to update this line.

PR #1009, opened 2023-01-09
[SDA-7684] Encrypt private key in oidc config

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-7684

# What
Encrypt private key for bound service account signing key during creation of oidc config

# Why
Adds a layer of security to user flow


cmd/create/oidcconfig/cmd.go
```diff
@@ -110,23 +120,86 @@ func run(cmd *cobra.Command, argv []string) {
 			r.Reporter.Errorf("Expected a valid OIDC provider creation mode: %s", err)
 			os.Exit(1)
 		}
+
+		kmsArn, err = interactive.GetString(
+			interactive.Input{
+				Question: "KMS ARN",
+				Help:     cmd.Flags().Lookup("kms-arn").Usage,
+				Required: false,
```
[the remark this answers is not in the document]
gdbranco: That can be an option indeed, not sure if we should though. Also not sure how to get the default one, would need to check. Is it a static arn for the account?

cmd/create/oidcconfig/cmd.go
```diff
@@ -110,23 +120,86 @@ func run(cmd *cobra.Command, argv []string) {
...
+		if err != nil {
+			r.Reporter.Errorf("Expected a valid KMS ARN: %s", err)
+			os.Exit(1)
+		}
+	}
+
+	if kmsArn != "" {
+		_, err := arn.Parse(kmsArn)
+		if err != nil {
+			r.Reporter.Errorf("Expected a valid policy ARN for permissions boundary: %s", err)
```
oriAdler, line 140: Should the error message be related to the `kmsARN`?

gdbranco: Thanks! fixed

gdbranco:
/hold

Holding until we discuss it on Jan 12th 2023

PR #1012, opened 2023-01-11
Upgrade roles command - handle managed policies

oriAdler:
1. Check that the account roles have managed policies attached.
2. Check that the operator roles have managed policies attached.
2. Check for missing operator roles by OCP version.

Related: [SDA-7712](https://issues.redhat.com/browse/SDA-7712)

image

### DIFF
--- a/cmd/upgrade/operatorroles/cmd.go
+++ b/cmd/upgrade/operatorroles/cmd.go
@@ -20,21 +20,17 @@ import (
 	"fmt"
 	"os"
 	"strings"
-	"time"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
-	"github.com/openshift/rosa/pkg/helper"
 	"github.com/openshift/rosa/pkg/helper/roles"
 	"github.com/openshift/rosa/pkg/rosa"
 	"github.com/spf13/cobra"
 
 	"github.com/openshift/rosa/pkg/aws"
 	awscb "github.com/openshift/rosa/pkg/aws/commandbuilder"
-	"github.com/openshift/rosa/pkg/aws/tags"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/interactive/confirm"
 	"github.com/openshift/rosa/pkg/ocm"
-	rprtr "github.com/openshift/rosa/pkg/reporter"
 )
 
 var args struct {
@@ -162,7 +158,13 @@ func run(cmd *cobra.Command, argv []string) error {
 	}
 
 	if managedPolicies {
-		err = validateManagedPolicies(cmd, r, cluster, credRequests, policies, mode, prefix, unifiedPath,
+		mode, err = handleModeFlag(cmd, mode)
+		if err != nil {
+			r.Reporter.Errorf("%s", err)
+			os.Exit(1)
+		}
+
+		err = roles.ValidateOperatorRolesManagedPolicies(r, cluster, credRequests, policies, mode, prefix, unifiedPath,
 			args.upgradeVersion)
 		if err != nil {
 			r.Reporter.Errorf("Failed while validating managed policies: %v", err)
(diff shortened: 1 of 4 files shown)


oriAdler:
@gdbranco @hunterkepley @zgalor PTAL

pkg/helper/roles/helpers.go
```diff
@@ -83,3 +90,169 @@ func BuildMissingOperatorRoleCommand(
...
+	operatorRoles map[string]*cmv1.STSOperator, policies map[string]*cmv1.AWSSTSPolicy, mode string, prefix string,
+	unifiedPath string, upgradeVersion string) error {
+	if upgradeVersion != "" {
+		missingRolesInCS, err := r.OCMClient.FindMissingOperatorRolesForUpgrade(cluster, upgradeVersion)
+		if err != nil {
+			return err
+		}
+		if len(missingRolesInCS) > 0 {
+			r.Reporter.Infof("Starting to upgrade the operator IAM roles")
+			mode, err = HandleModeFlag(cmd, mode, err, r.Reporter)
```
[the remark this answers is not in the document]
oriAdler: This was the previous behavior in the `upgrade operator roles` command.
The motivation is to check the mode only if action is needed by the user.
So if all roles are in place, we don't even prompt the mode question.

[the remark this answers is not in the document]
gdbranco: I see it was used here keeping behavior
https://github.com/openshift/rosa/pull/1012/files#diff-3802c119f0c619f9d349b65a5fa516ef3f7bde25defcd998d71bbe117dbb7e01R216

I'm saying that also checking within the actual validation might not be needed, but expected that is has been dealt prior to calling validation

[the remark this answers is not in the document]
oriAdler: Good check.
Fixed, thanks.

cmd/upgrade/operatorroles/cmd.go
```diff
@@ -340,108 +291,26 @@ func upgradeOperatorPolicies(mode string, r *rosa.Runtime,
...
-			return err
-		}
-		if r.Reporter.IsTerminal() {
-			r.Reporter.Infof("Run the following commands to create the operator roles:\n")
-		}
-		fmt.Println(commands)
-	default:
-		r.Reporter.Errorf("Invalid mode. Allowed values are %s", aws.Modes)
-		os.Exit(1)
+func handleModeFlag(cmd *cobra.Command, mode string) (string, error) {
```
gdbranco, line 294: Is this the same function from upgrade roles?

cmd/upgrade/roles/cmd.go
```diff
@@ -150,42 +141,95 @@ func run(cmd *cobra.Command, argv []string) error {
...
-	if interactive.Enabled() && !skipInteractive {
-		mode, err = interactive.GetOption(interactive.Input{
-			Question: "Roles upgrade mode",
-			Help:     cmd.Flags().Lookup("mode").Usage,
-			Default:  aws.ModeAuto,
-			Options:  aws.Modes,
-			Required: true,
-		})
+	if managedPolicies {
+		mode, err = handleModeFlag(cmd, mode, skipInteractive)
```
gdbranco, line 174: This can be done before here and reuse value for both managed policies and default flow in my opinion

cmd/upgrade/operatorroles/cmd.go
oriAdler: Functions are similar, but not the same.

cmd/upgrade/roles/cmd.go
oriAdler: Done

cmd/upgrade/operatorroles/cmd.go
oriAdler: Removed the `handleModeFlag` function from `upgrade roles` since we're going to call it once.
Now it is like it was previously.

PR #1015, opened 2023-01-12
[SDA-7756] create config with secrets manager

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-7756

What
Stores private key in secrets manager for bound service account signing key during creation of oidc config

Why
Adds a layer of security to user flow


cmd/create/oidcconfig/cmd.go
```diff
@@ -112,21 +113,58 @@ func run(cmd *cobra.Command, argv []string) {
...
 		r.Reporter.Errorf("%s", err)
 		os.Exit(1)
 	}
-	bucketUrl, privateKeyFilename := oidcConfigStrategy.execute(r)
+	oidcConfigStrategy.execute(r, oidcConfigInput)
 	if r.Reporter.IsTerminal() {
 		r.Reporter.Infof("To create a cluster with this oidc config, please run:\n"+
 			"rosa create cluster --sts --oidc-endpoint-url %s "+
-			"--bound-service-account-signing-key-path ./%s", bucketUrl, privateKeyFilename)
+			"--bound-service-account-signing-key-secret-name %s",
```
vkareh, line 126: This should be `--bound-service-account-signing-key-secret-arn`, also why is it templated to `oidcConfigInput.BucketName`?

Also, do you think there's a better name for this flag? Like `--oidc-private-key`, for example?
cc @jharrington22

cmd/create/oidcconfig/cmd.go
```diff
@@ -112,21 +113,58 @@ func run(cmd *cobra.Command, argv []string) {
...
+			oidcConfigInput.BucketUrl, oidcConfigInput.BucketName)
+	}
+}
+
+type OidcConfigInput struct {
+	BucketName         string
+	BucketUrl          string
+	PrivateKey         []byte
+	PrivateKeyFilename string
```
vkareh, line 135: Why do we need a file name if we're using the secrets manager? I think you should generate the secret and keep it in memory until you send it to AWSSM. No need to store in a file.

cmd/create/oidcconfig/cmd.go
```diff
@@ -136,55 +174,62 @@ const (
...
+func (s *CreateOidcConfigManualStrategy) execute(r *rosa.Runtime, input OidcConfigInput) {
 	commands := []string{}
-	randomLabel := helper.RandomLabel(DEFAULT_LENGTH_RANDOM_LABEL)
-	bucketName := fmt.Sprintf("oidc-%s", randomLabel)
-	bucketUrl := fmt.Sprintf("https://%s.s3.%s.amazonaws.com", bucketName, args.region)
+	bucketName := input.BucketName
+	discoveryDocument := input.DiscoveryDocument
+	jwks := input.Jwks
+	privateKey := input.PrivateKey
+	privateKeyFilename := input.PrivateKeyFilename
```
vkareh, line 215: I see... You're storing the file for manual mode? Could be useful, but I wouldn't invest too much in this for the time being. That said, if it all works well then keep the code in place.

cmd/create/oidcconfig/cmd.go
```diff
@@ -112,21 +113,58 @@ func run(cmd *cobra.Command, argv []string) {
...
+	BucketUrl          string
+	PrivateKey         []byte
+	PrivateKeyFilename string
+	DiscoveryDocument  string
+	Jwks               []byte
+}
+
+func buildOidcConfigInput(r *rosa.Runtime) OidcConfigInput {
+	randomLabel := helper.RandomLabel(DEFAULT_LENGTH_RANDOM_LABEL)
+	bucketName := fmt.Sprintf("oidc-%s", randomLabel)
```
jharrington22, line 142: This will result in a confusing SSM secret name, should we create a separate field for the SSM secret name? What do you think about naming this something like `rosa-oidc-private-key-%s`

cmd/create/oidcconfig/cmd.go
```diff
@@ -112,21 +113,58 @@ func run(cmd *cobra.Command, argv []string) {
...
+	bucketUrl := fmt.Sprintf("https://%s.s3.%s.amazonaws.com", bucketName, args.region)
+	privateKey, publicKey, err := createKeyPair()
+	if err != nil {
+		r.Reporter.Errorf("There was a problem generating key pair: %s", err)
+		os.Exit(1)
+	}
+	privateKeyFilename := fmt.Sprintf("private-key-%s.key", bucketName)
```
jharrington22, line 149: This should be updated if we change the SSM secret name

cmd/create/oidcconfig/cmd.go
gdbranco: Sure, will do!

cmd/create/oidcconfig/cmd.go
gdbranco: We need to keep it for manual as create secret requires a file, I mean not requires but it is safer than output the private key in the screen to run the command. It would also generate a big command which is difficult to see where it ends

Within the listing of commands I also add commands to remove the files that have been generated

cmd/create/oidcconfig/cmd.go
gdbranco: Sure, that works out

cmd/create/oidcconfig/cmd.go
gdbranco: Yes

cmd/create/oidcconfig/cmd.go
gdbranco: Oh this is leftover, this time I can't have this output actually. I can do it for auto mode, however in manual the user will have to run the commands which will then output the correct ARN. And to keep things consistent I'd rather not output this

cmd/create/oidcconfig/cmd.go
```diff
@@ -58,9 +59,10 @@ var args struct {
...
 	Aliases: []string{"oidcconfig"},
-	Short:   "Create OIDC config for an STS cluster.",
-	Long: "Create OIDC config in an S3 bucket for the " +
-		"client AWS account and populates it to be compliant with OIDC protocol.",
+	Short:   "Create OIDC config compliant with OIDC protocol.",
+	Long: "Create OIDC config in a S3 bucket for the " +
+		"client AWS account and populates it to be compliant with OIDC protocol. " +
+		"It also creates a Secret in Secrets Manager containing the private key.",
 	Example: `  # Create OIDC config rosa create oidc-config`,
 	Run:     run,
```
vkareh, line 67: Please add `Hidden: true,` to this command

cmd/create/oidcconfig/cmd.go
```diff
@@ -195,18 +249,6 @@ func (s *CreateOidcConfigManualStrategy) execute(r *rosa.Runtime) (string, strin
 		AddParam(awscb.CreateBucketConfiguration, createBucketConfig).
```
vkareh, line 249: I think we should remove the `createS3BucketCommand` from the manual mode. Just output the discovery document and the private key, and let the customer host it wherever they want.

gdbranco: I see your point, but I'd say to keep consistency we should keep it. And maybe add a --raw or something like that as a facilitator for people who don't want to deal with setting up the documents, but still want to host elsewhere

gdbranco: Decided to set it up

cmd/create/oidcconfig/cmd.go
```diff
@@ -112,79 +128,172 @@ func run(cmd *cobra.Command, argv []string) {
 		}
 	}
 
-	oidcConfigStrategy, err := getOidcConfigStrategy(mode)
+	if !args.rawFiles {
+		r.Reporter.Infof("This command will create a S3 bucket populating it with documents " +
+			"to be compliant with OIDC protocol. It will also create a Secret in Secrets Manager containing the private key.")
+	}
+
+	if args.rawFiles && mode != "" {
```
oriAdler, line 136: Maybe it is worth checking this before line `117` where we ask the user to choose the mode.
Then ask the user to choose a mode, only if `rawFiles` isn't supplied.

cmd/create/oidcconfig/cmd.go
```diff
@@ -94,10 +110,15 @@ func run(cmd *cobra.Command, argv []string) {
...
+	if !interactive.Enabled() && !cmd.Flags().Changed("mode") && !cmd.Flags().Changed(rawFilesFlag) {
 		interactive.Enable()
 	}
 
+	if args.rawFiles && mode != "" {
+		r.Reporter.Warnf("--raw-files param is not supported alongside --mode param.")
+		os.Exit(1)
+	}
+
 	if interactive.Enabled() {
```
oriAdler, line 122: Should we enter here only if `args.rawFiles == false`?

gdbranco: I see what you mean, if the user specifies -i he would get there alongside raw-files.

PR #1016, opened 2023-01-13
Delete correct operator roles and support hypershift

gdbranco:
# What
Adds support for hypershift credential requests coming from subscription

# Why
Hypershift roles were not being deleted and there was a chance to identify the wrong cluster when deleting operator roles


lucasponce:
LGTM thanks

hunterkepley:
LGTM

PR #1017, opened 2023-01-13
[SDA-7758] Accept pre release version during upgrade

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-7758

# What
Check if it is a pre release version

# Why
Version was defaulting to the first available upgrade in slice

### DIFF
--- a/pkg/ocm/versions.go
+++ b/pkg/ocm/versions.go
@@ -322,8 +322,9 @@ func CheckAndParseVersion(availableUpgrades []string, version string) (string, e
 	if err != nil {
 		return "", err
 	}
+	isPreRelease := a.Prerelease() != ""
 	versionSplit := a.Segments64()
-	if versionSplit[2] > 0 {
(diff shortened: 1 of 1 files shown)


pkg/ocm/versions.go
```diff
@@ -322,8 +322,9 @@ func CheckAndParseVersion(availableUpgrades []string, version string) (string, e
 	if err != nil {
 		return "", err
 	}
+	isPreRelease := a.Prerelease() != ""
 	versionSplit := a.Segments64()
-	if versionSplit[2] > 0 {
+	if len(versionSplit) > 2 && versionSplit[2] > 0 || (versionSplit[2] == 0 && isPreRelease) {
```
hunterkepley, line 327: I'm not sure I fully understand -- is a pre release version just `X.X.0`?

gdbranco: They have a -rc which gets caught under the isPrerelease. I've taken a look at other channel groups (fast, nightly and candidate) only fast has versions where z from x.y.z are different than 0. So that should be okay

PR #1018, opened 2023-01-13
[SDA-7757] byo OIDC secret arn support

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-7757


vkareh:
> /label tide/merge-method-squash

Do we need auto-squashing? It messes with commit logs, but your commits are already nice and clean. Anyway, up to you, it's all personal preference :stuck_out_tongue:

cmd/create/cluster/cmd.go
```diff
@@ -250,6 +255,20 @@ func init() {
...
+		"oidc-endpoint-url",
+		"",
+		"Endpoint url for BYO OIDC config",
+	)
+
+	flags.StringVar(
+		&args.boundServiceAccountSigningKeySecretArn,
+		"bound-service-account-signing-key-secret-arn",
+		"",
+		"AWS Secrets Manager ARN for retrieval of bound service account signing key",
```
vkareh, line 269: Is using "bound service account signing key" exposing too much of the STS implementation detail than necessary? Something like "oidc-private-key" might be sufficient.

cc @jharrington22

cmd/create/cluster/cmd.go
```diff
@@ -250,6 +255,20 @@ func init() {
 			"Leave empty to use an auto-generated one.",
 	)
 
+	flags.StringVar(
+		&args.oidcEndpointUrl,
+		"oidc-endpoint-url",
+		"",
+		"Endpoint url for BYO OIDC config",
+	)
```
vkareh, line 263: You should add `MarkHidden` to these new flags for now.

cmd/create/cluster/cmd.go
```diff
@@ -1145,6 +1164,74 @@ func run(cmd *cobra.Command, _ []string) {
 			}
 		}
 	}
+	oidcEndpointUrl := args.oidcEndpointUrl
+	boundServiceAccountSigningKeySecretArn := args.boundServiceAccountSigningKeySecretArn
+	isByoOidcSet := false
+	if isSTS {
+		// Support byo oidc config
+		if interactive.Enabled() {
```
vkareh, line 1172: Since these should be hidden for now, can you do the same pattern we've used before for hidden flags?
```
if oidcEndpointUrl != "" && interactive.Enabled()
```
That way only users who are aware of the flag _and_ intend on using it get the interactive prompt.

cmd/create/cluster/cmd.go
jharrington22: +1 --oidc-private-key-arn. Probably a good idea to end with the -arn suffix so its clear to the user its not a path or plain text key.

cmd/create/oidcprovider/cmd.go
```diff
@@ -50,10 +50,21 @@ var Cmd = &cobra.Command{
...
 
+var args struct {
+	oidcEndpointUrl string
+}
+
 func init() {
 	flags := Cmd.Flags()
 
-	ocm.AddClusterFlag(Cmd)
+	flags.StringVar(
```
jharrington22, line 60: This should be hidden too since users should only be passing this if they're aware of the other hidden flags during cluster creation

cmd/create/oidcprovider/cmd.go
```diff
@@ -66,16 +77,18 @@ func run(cmd *cobra.Command, argv []string) {
 
 	// Allow the command to be called programmatically
 	skipInteractive := false
-	if len(argv) == 2 && !cmd.Flag("cluster").Changed {
+	if len(argv) == 3 && !cmd.Flag("cluster").Changed {
```
jharrington22, line 80: I think we want to leave this as 2 and have another check to disable interactive mode if there are 3 arguments otherwise this will break the existing flow?

cmd/create/cluster/cmd.go
gdbranco: At first I thought saying private key was worse haha. Will change this

cmd/create/cluster/cmd.go
gdbranco: Sure

cmd/create/oidcprovider/cmd.go
gdbranco: I fixed the current flow so both would send 3 arguments, this is for when it is called progmmatically from the `create cluster` command

cmd/create/cluster/cmd.go
```diff
@@ -1145,6 +1172,74 @@ func run(cmd *cobra.Command, _ []string) {
...
+			if len(attributesForByoOidc) > 0 {
+				r.Reporter.Errorf("Missing attributes for byo oidc '%s'", helper.SliceToSortedString(attributesForByoOidc))
+				os.Exit(1)
+			}
+			err = interactive.IsURL(oidcEndpointUrl)
+			if err != nil {
+				r.Reporter.Errorf("%s", err)
+				os.Exit(1)
+			}
+			parsedURI, _ := url.ParseRequestURI(oidcEndpointUrl)
```
vkareh, line 1225: This `interactive.IsURL` function just calls `url.ParseRequestURI` internally, so no need to do it twice, I would remove the `IsURL` call, as it's meant for the interactive validation only.

cmd/create/cluster/cmd.go
```diff
@@ -1145,6 +1172,74 @@ func run(cmd *cobra.Command, _ []string) {
...
+			err := interactive.IsURLReachable(fmt.Sprintf("%s:%s", parsedURI.Host, parsedURI.Scheme))
```
vkareh, line 1226: I don't think this function belongs in the `interactive` package, as we're not actually using it there. I would just put it in a random helper function elsewhere.

cmd/create/cluster/cmd.go
```diff
@@ -2398,6 +2504,11 @@ func buildCommand(spec ocm.Spec, operatorRolesPrefix string,
 	if operatorRolesPrefix != "" {
 		command += fmt.Sprintf(" --operator-roles-prefix %s", operatorRolesPrefix)
 	}
+	if spec.OidcEndpointUrl != "" && spec.BoundServiceAccountSigningKeySecretArn != "" {
+		command += fmt.Sprintf(" --oidc-endpoint-url %s", spec.OidcEndpointUrl)
+		command += fmt.Sprintf(" --bound-service-account-signing-key-secret-arn %s",
```
vkareh, line 2509: The flag name needs to be updated here.

gdbranco: Nice, missed it

cmd/create/cluster/cmd.go
```diff
@@ -1145,6 +1172,69 @@ func run(cmd *cobra.Command, _ []string) {
...
+					})
+				if err != nil {
+					r.Reporter.Errorf("Expected a valid ARN to the secret containing the private key: %s", err)
+					os.Exit(1)
+				}
+			}
+		}
+		isByoOidcSet = oidcEndpointUrl != "" || oidcPrivateKeySecretArn != ""
+		if isByoOidcSet {
+			attributesForByoOidc := []string{}
```
oriAdler, line 1209: NP
Maybe worth emphasizing this
```suggestion
			missingAttributesForByoOidc := []string{}
```

cmd/create/cluster/cmd.go
```diff
@@ -1145,6 +1172,69 @@ func run(cmd *cobra.Command, _ []string) {
 			}
 		}
 	}
+	oidcEndpointUrl := args.oidcEndpointUrl
+	oidcPrivateKeySecretArn := args.oidcPrivateKeySecretArn
+	isByoOidcSet := false
+	if isSTS {
+		// Support byo oidc config
```
oriAdler, line 1179: We can extract lines `1179-1236` to a new function like **handleByoOidcConfig**.

PR #1019, opened 2023-01-15
SDA-7767 Let user define machinepool name on hosted cluster

lucasponce:
Follow-up work for [SDA-7767](https://issues.redhat.com//browse/SDA-7767), is needed when the backend MR is merged.

Summary of changes:
- Removing the "INSTANCE-PREFIX" column (as now the machinepool is properly identify by the ID).
- Allow to use "--name" parameter for specifying the name of a MachinePool for hosted clusters.

Use cases, test scenarios covered with this work:

```
# List existing machinepools for a cluster
$ rosa list machinepools -c lponce-local-02
ID       AUTOSCALING  REPLICAS  INSTANCE TYPE  AVAILABILITY ZONE  SUBNET                    
workers  No           2         m5.xlarge      us-west-2a         subnet-0e3a4046c1c2f1078  
test1    No           2         m5.xlarge      us-west-2a         subnet-0e3a4046c1c2f1078  
test3    No           3         m5.xlarge      us-west-2a         subnet-0e3a4046c1c2f1078  
test4    No           2         m5.xlarge      us-west-2a         subnet-0e3a4046c1c2f1078  

# Create a new machinepool using --name param
uhc-clusters-service(nodepool-name) L  $ rosa create machinepool -c lponce-local-02 --name other-1
? Replicas: 2
I: Fetching instance types
I: Machine pool 'other-1' created successfully on hosted cluster 'lponce-local-02'
I: To view all machine pools, run 'rosa list machinepools -c lponce-local-02'

# Validate a name is already used for a cluster
uhc-clusters-service(nodepool-name) L  $ rosa create machinepool -c lponce-local-02
I: Enabling interactive mode
? Machine pool name: other-1
X Sorry, your reply was invalid: Value is required
? Replicas: 2
I: Fetching instance types
? Instance type: m5.xlarge
E: Failed to add machine pool to hosted cluster 'lponce-local-02': NodePool 'other-1' exists for cluster '21a1olhmroh6ds6scsh69bf10be92om6'
uhc-clusters-service(nodepool-name) L  $ rosa create machinepool -c lponce-local-02
I: Enabling interactive mode
? Machine pool name: other-2
? Replicas: 2
I: Fetching instance types
? Instance type: m5.xlarge
I: Machine pool 'other-2' created successfully on hosted cluster 'lponce-local-02'
I: To view all machine pools, run 'rosa list machinepools -c lponce-local-02'

# Final list, INSTANCE-PREFIX is outdated and now it doesn't add value so, saving space in the table for state info
uhc-clusters-service(nodepool-name) L  $ rosa list machinepools -c lponce-local-02
ID       AUTOSCALING  REPLICAS  INSTANCE TYPE  AVAILABILITY ZONE  SUBNET                    
workers  No           2         m5.xlarge      us-west-2a         subnet-0e3a4046c1c2f1078  
test1    No           2         m5.xlarge      us-west-2a         subnet-0e3a4046c1c2f1078  
test3    No           3         m5.xlarge      us-west-2a         subnet-0e3a4046c1c2f1078  
test4    No           2         m5.xlarge      us-west-2a         subnet-0e3a4046c1c2f1078  
other-1  No           2         m5.xlarge      us-west-2a         subnet-0e3a4046c1c2f1078  
other-2  No           2         m5.xlarge      us-west-2a         subnet-0e3a4046c1c2f1078  

```

### DIFF
--- a/cmd/create/machinepool/nodepool.go
+++ b/cmd/create/machinepool/nodepool.go
@@ -3,6 +3,7 @@ package machinepool
 import (
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/briandowns/spinner"
@@ -31,9 +32,29 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
 		os.Exit(1)
 	}
 
-	// Hosted clusters create identifiers for NodePools, users don't interact directly with these resources
-	if cmd.Flags().Changed("name") {
-		r.Reporter.Errorf("Setting the `name` is not supported for hosted clusters")
+	// Machine pool name:
+	name := strings.Trim(args.name, " \t")
+	if name == "" && !interactive.Enabled() {
+		interactive.Enable()
+		r.Reporter.Infof("Enabling interactive mode")
+	}
+	if name == "" || interactive.Enabled() {
+		name, err = interactive.GetString(interactive.Input{
+			Question: "Machine pool name",
+			Default:  name,
+			Required: true,
+			Validators: []interactive.Validator{
+				interactive.RegExp(machinePoolKeyRE.String()),
+			},
+		})
+		if err != nil {
+			r.Reporter.Errorf("Expected a valid name for the machine pool: %s", err)
+			os.Exit(1)
+		}
+	}
+	name = strings.Trim(name, " \t")
+	if !machinePoolKeyRE.MatchString(name) {
+		r.Reporter.Errorf("Expected a valid name for the machine pool")
 		os.Exit(1)
 	}
 
@@ -114,6 +135,7 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
 	}
 
 	npBuilder := cmv1.NewNodePool()
+	npBuilder.ID(name)
 
 	if autoscaling {
 		npBuilder = npBuilder.Autoscaling(
--- a/cmd/list/machinepool/nodepool.go
+++ b/cmd/list/machinepool/nodepool.go
@@ -21,16 +21,15 @@ func listNodePools(r *rosa.Runtime, clusterKey string, cluster *cmv1.Cluster) {
 	// Create the writer that will be used to print the tabulated results:
 	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
 
-	fmt.Fprintf(writer, "ID\tAUTOSCALING\tREPLICAS\tINSTANCE TYPE\tAVAILABILITY ZONE\tSUBNET\tINSTANCE PREFIX\t\n")
+	fmt.Fprintf(writer, "ID\tAUTOSCALING\tREPLICAS\tINSTANCE TYPE\tAVAILABILITY ZONE\tSUBNET\t\n")
 	for _, nodePool := range nodePools {
-		fmt.Fprintf(writer, "%s\t%s\t%s\t%s\t%s\t%s\t%s\t\n",
+		fmt.Fprintf(writer, "%s\t%s\t%s\t%s\t%s\t%s\t\n",
 			nodePool.ID(),
 			printNodePoolAutoscaling(nodePool.Autoscaling()),
 			printNodePoolReplicas(nodePool.Autoscaling(), nodePool.Replicas()),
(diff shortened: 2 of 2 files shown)


cmd/create/machinepool/nodepool.go
```diff
@@ -31,9 +32,29 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
 		os.Exit(1)
 	}
 
-	// Hosted clusters create identifiers for NodePools, users don't interact directly with these resources
-	if cmd.Flags().Changed("name") {
-		r.Reporter.Errorf("Setting the `name` is not supported for hosted clusters")
+	// Machine pool name:
+	name := strings.Trim(args.name, " \t")
```
vkareh, line 36: We have a `hypershift.CleanName` already, maybe that function can be updated to replace `\t` as well and reused across nodepools.

vkareh: Ah, the function is in the CS code. But might be worth looking into how to works so that we're consistent between API and CLI

cmd/create/machinepool/nodepool.go
```diff
@@ -31,9 +32,29 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
...
+	if name == "" && !interactive.Enabled() {
+		interactive.Enable()
+		r.Reporter.Infof("Enabling interactive mode")
+	}
+	if name == "" || interactive.Enabled() {
```
vkareh, line 41: this `name == ""` is unnecessary, as you're forcing interactive mode above. I would remove it.

cmd/create/machinepool/nodepool.go
```diff
@@ -31,9 +32,29 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
...
+		name, err = interactive.GetString(interactive.Input{
+			Question: "Machine pool name",
+			Default:  name,
+			Required: true,
```
vkareh, line 45: Is there a benefit to making this required vs optional? We could easily generate `$cluster.name-$randomhash` as a default instead of forcing the user.

cmd/create/machinepool/nodepool.go
lucasponce: Note that I'm mimic-ing here the same behaviour done in the classic rosa for machinepool.
I.e. https://github.com/openshift/rosa/blob/master/cmd/create/machinepool/machinepool.go#L59

I mean, I tried to be consistent both and this logic should be extracted for both cases (and changed in both cases if it's not correct).

cmd/create/machinepool/nodepool.go
lucasponce: Yeah, I didn't realize.
I just reuse the same code used in the machinepool for classic rosa.

https://github.com/openshift/rosa/blob/master/cmd/create/machinepool/machinepool.go#L65

I prefer to add a little duplication here before jumping in a refactor for the common parameter (in favor to make this PR smaller).

cmd/create/machinepool/nodepool.go
lucasponce: No idea, at least the requeriment was to make the NodePools to operate as the classic machinepool is working today.

https://github.com/openshift/rosa/blob/master/cmd/create/machinepool/machinepool.go#L69

Yeah, I like that suggestion; perhaps when consolidated, we could propose it for all cases? (classic rosa and hosted clusters)

cmd/create/machinepool/nodepool.go
andreadecorte: AFAIK this becomes the name of a NodePool resource no? So it means it must be a valid kube resource name (for example, no underscore)

lucasponce: It's part of the name in the kube resource.
There is already a regex that checks if the name is correct.

I think it's good for kube (unless I missed something)

```
// Regular expression to used to make sure that the identifier given by the
// user is safe and that it there is no risk of SQL injection:
var machinePoolKeyRE = regexp.MustCompile(`^[a-z]([-a-z0-9]*[a-z0-9])?$`)
```

gdbranco:
LGTM

lucasponce:
cc @wgordon17

PR #1020, opened 2023-01-16
NodePools status in ROSA

lucasponce:
Proposed changes at "rosa list machinepools" for hosted clusters use cases:
- "REPLICAS" updated to "DESIRED REPLICAS"
- Added new "CURRENT REPLICAS" field
- Added new "MESSAGE" field that would indicate extra information (coming from NodePool) when Current Replicas are not ready

```
$ rosa list machinepools -c lponce-local-01 
ID       AUTOSCALING  DESIRED REPLICAS  CURRENT REPLICAS  INSTANCE TYPE  AVAILABILITY ZONE  SUBNET                    MESSAGE  
workers  No           2                 2                 m5.xlarge      us-west-2a         subnet-0e3a4046c1c2f1078    
```

Proposed changes at "rosa describe cluster" for hosted clusters::
- Updated title "Compute (desired)."
- Added "Computed (current)" to define the current replicas of a cluster

```
$ rosa describe cluster -c lponce-local-01 

Name:                       lponce-local-01
ID:                         21bji08f0085qg11rdf2t3acd2kf9gsi
External ID:                67454bd9-8837-4631-92c4-1d2c4bc998d8
Control Plane:              Red Hat hosted
OpenShift Version:          
Channel Group:              candidate
DNS:                        lponce-local-01.pfza.hypershift.sdev.devshift.net
API URL:                    https://aa65a4d8c25c34d1cae391195ad161ea-67b36684e201bf00.elb.us-west-2.amazonaws.com:6443
Console URL:                https://console-openshift-console.apps.lponce-local-01.pfza.hypershift.sdev.devshift.net
Region:                     us-west-2
Nodes:
 - Compute (desired):           2
 - Compute (current):           2

```


cmd/describe/cluster/cmd.go
```diff
@@ -324,11 +324,47 @@ func controlPlaneConfig(cluster *cmv1.Cluster) string {
 func clusterInfraConfig(cluster *cmv1.Cluster, clusterKey string, r *rosa.Runtime) string {
 	var infraConfig string
 	if cluster.Hypershift().Enabled() {
-		infraConfig = fmt.Sprintf(""+
-			"Nodes:\n"+
-			" - Compute (workers):           %d\n",
-			cluster.Nodes().Compute(),
-		)
+		nodePools, err := r.OCMClient.GetNodePools(cluster.ID())
```
vkareh, line 327: I hate that we have two API endpoints for essentially the same stuff: nodepools vs machinepools. The original proposal when this project started was to abstract those differences from the API to avoid this type of duplication... not anyone's fault (apart from maybe mine), just ranting :stuck_out_tongue:

cmd/list/machinepool/nodepool.go
```diff
@@ -21,16 +21,18 @@ func listNodePools(r *rosa.Runtime, clusterKey string, cluster *cmv1.Cluster) {
...
+		fmt.Fprintf(writer, "%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\t\n",
 			nodePool.ID(),
 			printNodePoolAutoscaling(nodePool.Autoscaling()),
 			printNodePoolReplicas(nodePool.Autoscaling(), nodePool.Replicas()),
+			printNodePoolCurrentReplicas(nodePool.Status()),
 			printNodePoolInstanceType(nodePool.AWSNodePool()),
 			nodePool.AvailabilityZone(),
 			nodePool.Subnet(),
 			printNodePoolName(nodePool.AWSNodePool()),
+			printNodePoolMessage(nodePool.Status()),
```
vkareh, line 35: Can we leave the `message` out of the table and output it only on `--output=[json|yaml]`? Otherwise it's a very long table.

cmd/describe/cluster/cmd.go
lucasponce: I know :). There was some back-and-forth on this.
The tradeoff was to keep the "nodepools"  API in OCM as an implementation detail and consolidate the "machinepool" or "hosted machinepools" at ROSA CLI level.

Trying to use a single "machinepool" API from OCM for both scenarios would trigger an interesting refactoring both in classic and hypershift scenarios.

This was a tradeoff to keep progressing until we can progress, but I see what you mean (I hope that at least I can share the "rationale" about the "why").

cmd/list/machinepool/nodepool.go
lucasponce: Note that when the https://github.com/openshift/rosa/pull/1019 is consolidated it will eliminate the "instance prefix" column and it will look pretty similar as the typical controller list in "kubectl"

lucasponce:
cc @wgordon17 @gdbranco

cmd/list/machinepool/nodepool.go
wgordon17: I think `message` is still important to show at the table level, even if it does increase table width

PR #1021, opened 2023-01-16
Ensure console URL is available before offering it to the user

andreadecorte:
As it may be not the case in HyperShift context during cluster creation

### DIFF
--- a/cmd/create/idp/cmd.go
+++ b/cmd/create/idp/cmd.go
@@ -80,8 +80,8 @@ var args struct {
 	htpasswdPassword string
 }
 
-var validIdps []string = []string{"github", "gitlab", "google", "htpasswd", "ldap", "openid"}
-var validMappingMethods []string = []string{"add", "claim", "generate", "lookup"}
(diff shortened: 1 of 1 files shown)


cmd/create/idp/cmd.go
```diff
@@ -435,10 +435,13 @@ func doCreateIDP(
...
 		"Identity Provider '%s' has been created.\n"+
 			"   It may take several minutes for this access to become active.\n"+
-			"   To add cluster administrators, see 'rosa grant user --help'.\n"+
-			"   To login into the console, open %s and click on %s.",
-		idpName, cluster.Console().URL(), idpName,
-	)
+			"   To add cluster administrators, see 'rosa grant user --help'.\n", idpName)
+	// Console may not be available yet
+	clusterConsole := cluster.Console().URL()
+	if clusterConsole != "" {
```
zgalor, line 441: I think `cluster.Console()` can also return nil so I suggest
```
if cluster.Console() != nil && cluster.Console().URL() != "" {
	clusterConsole := cluster.Console().URL()
...
}
```

andreadecorte: fixed, thanks

PR #1026, opened 2023-01-18
Skip region check if we use shard pinning

andreadecorte:
This is consistent with the behaviour on backend side, which will also check the authorization of using it

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -572,6 +572,14 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	shardPinningEnabled := false
+	for _, value := range args.properties {
+		if strings.HasPrefix(value, properties.ProvisionShardId) {
+			shardPinningEnabled = true
+			break
+		}
+	}
+
 	isBYOVPC := cmd.Flags().Changed("subnet-ids")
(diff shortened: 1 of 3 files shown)


pkg/properties/properties.go
```diff
@@ -32,3 +32,5 @@ const FakeCluster = "fake_cluster"
 
 // nolint:gosec // Linter thinks there are hardcoded credentials here...
 const UseLocalCredentials = "use_local_credentials"
+
+const ProvisionShardId = "provision_shard_id"
```
lucasponce, line 36: This property is for developer scenarios, right?
I mean, do we expect that users require to add the provision_shard_id ?

andreadecorte: it is optional (providing properties is an hidden flag). This PR doesn't change anything if we are not providing something like
--properties provision_shard_id:xxx in input

cmd/create/cluster/cmd.go
```diff
@@ -1237,7 +1245,7 @@ func run(cmd *cobra.Command, _ []string) {
 		r.Reporter.Errorf("Expected a valid AWS region")
 		os.Exit(1)
 	} else {
-		if isHostedCP && !interactive.Enabled() {
+		if isHostedCP && !interactive.Enabled() && !shardPinningEnabled {
```
tiwillia, line 1248: I think this might be incorrect. Why is this region check being done only when shad pinning is _not_ enabled?

PR #1028, opened 2023-01-18
Permissions boundary shouldn't be asked if supposed to skip interactive

gdbranco:
# What
Asks for permissions boundaries for operator roles during the cluster creation. If supplied consider we can skip interactive on creation of operator roles

# Why
Some automations may get interactive question during creation operator roles

### DIFF
--- a/cmd/create/operatorroles/cmd.go
+++ b/cmd/create/operatorroles/cmd.go
@@ -139,7 +139,7 @@ func run(cmd *cobra.Command, argv []string) {
 	}
 
 	permissionsBoundary := args.permissionsBoundary
-	if interactive.Enabled() {
+	if interactive.Enabled() && !skipInteractive {
 		permissionsBoundary, err = interactive.GetString(interactive.Input{
 			Question: "Permissions boundary ARN",
 			Help:     cmd.Flags().Lookup("permissions-boundary").Usage,


gdbranco:
@vkareh PTAL

vkareh (changes requested):
I don't like this. If this breaks automation, just add the `&& !skipInteractive` condition. but do not duplicate the interactive prompt. We already capture the `--permissions-boundary` flag and forward it. That should be sufficient for automation. In fact, if the interactive prompt shows during cluster creation, automation has already been broken.

gdbranco:
> I don't like this. If this breaks automation, just add the `&& !skipInteractive` condition. but do not duplicate the interactive prompt. We already capture the `--permissions-boundary` flag and forward it. That should be sufficient for automation. In fact, if the interactive prompt shows during cluster creation, automation has already been broken.

skipinteractive is indeed needed for the automation. Adding it as part of cluster creation just adds an optional param though, would breaking the automation be due to a user supplying entries individually during interactive mode instead of using the param? Not sure if I followed

vkareh:
My point is that we already allow the flag to be set AND we also have interactive flow for the flag in the second command.

We shouldn't duplicate the interactive prompt, so either you set the skipInteractive flag or you don't, and the prompt is shown based on that.

Does that make sense? Also, I may have the flow wrong, to be honest :P

gdbranco:
> My point is that we already allow the flag to be set AND we also have interactive flow for the flag in the second command. We shouldn't duplicate the interactive prompt, so either you set the skipInteractive flag or you don't, and the prompt is shown based on that. Does that make sense? Also, I may have the flow wrong, to be honest :P
> […](#)
> On January 20, 2023 6:34:01 p.m. EST, Guilherme Branco ***@***.***> wrote: > I don't like this. If this breaks automation, just add the `&& !skipInteractive` condition. but do not duplicate the interactive prompt. We already capture the `--permissions-boundary` flag and forward it. That should be sufficient for automation. In fact, if the interactive prompt shows during cluster creation, automation has already been broken. skipinteractive is indeed needed for the automation. Adding it as part of cluster creation just adds an optional param though, would breaking the automation be due to a user supplying entries individually during interactive mode instead of using the param? Not sure if I followed

Makes sense, changed so that we only skip question in create operator roles if is called from another command

PR #1029, opened 2023-01-18
[SDA-7898] Add force param to forcefully create policies

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-7898
# What
When creating account roles this param "--force-policy-creation" or "-f" skips compatibility checks ensuring policy will be created.

# Why
Current version might receive extra permissions during it's lifetime and needs to be updated

# Changes
`./rosa create account-roles --mode auto -y -f `
```
I: AWS credentials are valid!
I: Validating AWS quota...
I: Creating account roles
I: Creating roles using 'arn:aws:iam::xxx:user/gbranco'
I: Created role 'ManagedOpenShift-Worker-Role' with ARN 'arn:aws:iam::xxx:role/ManagedOpenShift-Worker-Role'
I: Created role 'ManagedOpenShift-Support-Role' with ARN 'arn:aws:iam::xxx:role/ManagedOpenShift-Support-Role'
I: Created role 'ManagedOpenShift-Installer-Role' with ARN 'arn:aws:iam::xxx:role/ManagedOpenShift-Installer-Role'
I: Created role 'ManagedOpenShift-ControlPlane-Role' with ARN 'arn:aws:iam::xxx:role/ManagedOpenShift-ControlPlane-Role'
I: To create a cluster with these roles, run the following command:
rosa create cluster --sts

```

### DIFF
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -44,6 +44,7 @@ var args struct {
 	version             string
 	channelGroup        string
 	managed             bool
+	forcePolicyCreation bool
 }
 
 var Cmd = &cobra.Command{
@@ -107,6 +108,14 @@ func init() {
 		"Attach AWS managed policies to the account roles",
 	)
 	flags.MarkHidden("managed")
+
+	flags.BoolVarP(
+		&args.forcePolicyCreation,
+		"force-policy-creation",
+		"f",
+		false,
+		"Forces creation of policies skipping compatibility check",
+	)
 
 	aws.AddModeFlag(Cmd)
 
@@ -281,7 +290,13 @@ func run(cmd *cobra.Command, argv []string) {
 			os.Exit(1)
 		}
 	}
-	policies, err := r.OCMClient.GetPolicies("")
+
+	if args.forcePolicyCreation && mode != aws.ModeAuto {
+		r.Reporter.Warnf("Forcing creation of policies only works in auto mode")
+		os.Exit(1)
+	}
+
+	policies, err := r.OCMClient.GetPolicies("AccountRole")
 	if err != nil {
 		r.Reporter.Errorf("Expected a valid role creation mode: %s", err)
 		os.Exit(1)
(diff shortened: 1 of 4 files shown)


pkg/aws/policies.go
```diff
@@ -373,9 +373,6 @@ func (c *awsClient) HasCompatibleVersionTags(iamTags []*iam.Tag, version string)
 	}
 	for _, tag := range iamTags {
 		if aws.StringValue(tag.Key) == tags.OpenShiftVersion {
-			if version == aws.StringValue(tag.Value) {
-				return true, nil
-			}
```
jharrington22, line 378: This will recreate policies every time you create a cluster which isn't the intention of have a set version policies. The intention is to send the least amount of IAM calls to AWS.

gdbranco: Yes, I had setup a check for each action before. However, @pvasant considered that approach overkill

jharrington22: Maybe we could implement a new flag to allow a user to force new policies?

gdbranco: Sure!

cmd/create/accountroles/cmd.go
```diff
@@ -281,6 +290,12 @@ func run(cmd *cobra.Command, argv []string) {
 			os.Exit(1)
 		}
 	}
+
+	if args.forcePolicyCreation && mode != aws.ModeAuto {
+		r.Reporter.Warnf("Forcing creation of policies only work in auto mode")
```
hunterkepley, line 295: Nitpick: "Forcing creation of policies only work**s** in auto mode"

hunterkepley (commented):
/lgtm

gdbranco:
Holding until we a second lgtm

cmd/create/operatorroles/cmd.go
```diff
@@ -126,15 +135,15 @@ func run(cmd *cobra.Command, argv []string) {
 	}
 
 	if len(missingRoles) == 0 &&
-		cluster.State() != cmv1.ClusterStateWaiting && cluster.State() != cmv1.ClusterStatePending {
+		cluster.State() != cmv1.ClusterStateWaiting && cluster.State() != cmv1.ClusterStatePending &&
+		!args.forcePolicyCreation {
```
oriAdler, line 139: Why do we need to check the force flag here?

pkg/aws/policies.go
```diff
@@ -289,7 +299,7 @@ func (c *awsClient) EnsurePolicy(policyArn string, document string,
 		return policyArn, err
 	}
 
-	if !isCompatible {
+	if !isCompatible || force {
```
oriAdler, line 302: In the case of `force==true` can we skip the compatible check on line `297`?
As we are going to ignore the result anyway.

cmd/create/operatorroles/cmd.go
gdbranco: They might have all roles but not up to date on the policies

pkg/aws/policies.go
gdbranco: Yes, just thought that this way would change the code less. But I can do that so we have less calls

gdbranco: fixed

pkg/aws/policies.go
```diff
@@ -284,9 +294,12 @@ func (c *awsClient) EnsurePolicy(policyArn string, document string,
 
 	policyArn = aws.StringValue(output.Policy.Arn)
 
-	isCompatible, err := c.IsPolicyCompatible(policyArn, version)
-	if err != nil {
-		return policyArn, err
+	isCompatible := false
```
oriAdler, line 297: Should this be true for `force==true`?

gdbranco: No we are always considering incompatible when forced so it will go through recreating policies

PR #1032, opened 2023-01-23
Fix managed policies cluster upgrade path

oriAdler:
In case of success in the upgrade roles command, return nil instead of exit. 
This change is required since we trigger the upgrade roles command from the upgrade cluster command.

Related: [SDA-7761](https://issues.redhat.com/browse/SDA-7761)

image


oriAdler:

PR #1033, opened 2023-01-23
Delete roles with managed policies

oriAdler:
Skip the policy deletion for managed policies.

Related: [SDA-7728](https://issues.redhat.com/browse/SDA-7728)

image

image

### DIFF
--- a/cmd/dlt/accountroles/cmd.go
+++ b/cmd/dlt/accountroles/cmd.go
@@ -153,6 +153,17 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
+	managedPolicies, err := r.AWSClient.HasManagedPolicies(roles[0].RoleARN)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


oriAdler:
@gdbranco @hunterkepley @zgalor PTAL

gdbranco:
I feel like we are adding managed policies bool to a lot of functions. Personally, I would rather separate these into two as they are slightly different. We might want to refactor at some point

But I'm okay if you would like to merge as is

oriAdler:
> I feel like we are adding managed policies bool to a lot of functions. Personally, I would rather separate these into two as they are slightly different. We might want to refactor at some point
> 
> But I'm okay if you would like to merge as is

I agree we can take advantage of go design patterns to differentiate the cases.
As for now, I'm satisfied with the approach of passing a boolean.
We can do more refactoring in the future, once functionality is in place.

PR #1034, opened 2023-01-24
Create OCM role with managed policies

oriAdler:
Attach existing policy instead of creating a new one.

Related: [SDA-7864](https://issues.redhat.com/browse/SDA-7864)

image

### DIFF
--- a/cmd/create/ocmrole/cmd.go
+++ b/cmd/create/ocmrole/cmd.go
@@ -40,6 +40,7 @@ var args struct {
 	permissionsBoundary string
 	admin               bool
 	path                string
+	managed             bool
 }
 
 var Cmd = &cobra.Command{
@@ -85,6 +86,15 @@ func init() {
 		"The arn path for the ocm role and policies",
 	)
 
+	// TODO: change to `unmanaged` once AWS managed policies are in place (managed will be the default)
+	flags.BoolVar(
+		&args.managed,
+		"managed",
+		false,
+		"Attach AWS managed policies to the account roles",
+	)
+	flags.MarkHidden("managed")
+
 	aws.AddModeFlag(Cmd)
 
 	confirm.AddFlag(flags)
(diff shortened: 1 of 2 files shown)


oriAdler:

cmd/create/ocmrole/cmd.go
```diff
@@ -326,52 +356,95 @@ func buildCommands(prefix string, roleName string, rolePath string, permissionsB
...
+				AddParam(awscb.PolicyName, policyName).
+				AddParam(awscb.PolicyDocument, fmt.Sprintf("file://sts_%s_permission_policy.json", aws.OCMAdminRolePolicyFile)).
+				AddTags(adminTags).
+				AddParam(awscb.Path, rolePath).
+				Build()
+		}
 
+		if managedPolicies {
+			policyARN, err = aws.GetManagedPolicyARN(policies,
+				fmt.Sprintf("sts_%s_permission_policy", aws.OCMRolePolicyFile))
```
gdbranco, line 407: Is there an OCMAdmin Policy for managed policies?

cmd/create/ocmrole/cmd.go
```diff
@@ -326,52 +356,95 @@ func buildCommands(prefix string, roleName string, rolePath string, permissionsB
...
+			if err != nil {
+				return "", err
+			}
+		} else {
+			policyARN = aws.GetPolicyARN(accountID, roleName, rolePath)
```
gdbranco, line 412: GetAdminPolicyARN

cmd/create/ocmrole/cmd.go
oriAdler: Good catch!
Fixed, thanks.

cmd/create/ocmrole/cmd.go
oriAdler: Missed that, fixed it.
Thanks.

PR #1035, opened 2023-01-24
Delete OCM role with managed policies

oriAdler:
Skip the deletion of the attached managed policies.

Related: [SDA-7878](https://issues.redhat.com/browse/SDA-7878)

image

### DIFF
--- a/cmd/dlt/ocmrole/cmd.go
+++ b/cmd/dlt/ocmrole/cmd.go
@@ -124,6 +124,12 @@ func run(cmd *cobra.Command, argv []string) error {
 		os.Exit(1)
 	}
 
+	managedPolicies, err := r.AWSClient.HasManagedPolicies(roleARN)
+	if err != nil {
+		r.Reporter.Errorf("Failed to determine if cluster has managed policies: %v", err)
+		os.Exit(1)
+	}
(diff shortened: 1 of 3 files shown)


oriAdler:
@hunterkepley @gdbranco @zgalor PTAL

cmd/dlt/ocmrole/cmd.go
```diff
@@ -124,6 +125,22 @@ func run(cmd *cobra.Command, argv []string) error {
...
+		os.Exit(1)
+	}
+	env, err := ocm.GetEnv()
+	if err != nil {
+		r.Reporter.Errorf("Failed to determine OCM environment: %v", err)
+		os.Exit(1)
+	}
+	// TODO: remove once AWS managed policies are in place
+	if managedPolicies && env == ocm.Production {
+		r.Reporter.Errorf("Managed policies are not supported in this environment")
```
zgalor, line 140: This is probably obsolete in deletion, no? 
I mean, the user won't be able to create the role with managed policies, so no need to check it in deletion

oriAdler: Makes sense, removed.

PR #1040, opened 2023-01-26
Edit machinepool min replicas

oriAdler:
Check if the min/max replicas value was updated, to differentiate the zero default value of the flags from the user input.

**Note:** Zero value for min/max replicas is a valid value.

Related: [SDA-7905](https://issues.redhat.com/browse/SDA-7905)

image

### DIFF
--- a/cmd/edit/machinepool/machinepool.go
+++ b/cmd/edit/machinepool/machinepool.go
@@ -47,7 +47,7 @@ func editMachinePool(cmd *cobra.Command, machinePoolID string, clusterKey string
 
 		clusterConfig := ocm.Spec{}
 
-		autoscaling, replicas, minReplicas, maxReplicas, scalingUpdated :=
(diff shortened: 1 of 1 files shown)


oriAdler:
@zgalor @lucasponce @hunterkepley PTAL

zgalor:
Thinking out loud if it wouldn't be simpler to set the default to -1?

oriAdler:
> Thinking out loud if it wouldn't be simpler to set the default to -1?

That might work, but I'm not sure if we should set the default value to an invalid value (also to be consistent with other commands we usually set the default value to zero/false/empty string).
With that being said I don't mind taking this approach if you think it simplifies the flow.

zgalor:
It simplifies, but less clear, so let's keep it.

zgalor:
LGTM

PR #1042, opened 2023-01-26
Add port for OAuth Callback URI in OpenID

andreadecorte:
Improve also the UX by showing the callback URI
at the end of creation when not in interactive mode.

Fixes: [SDA-7914](https://issues.redhat.com//browse/SDA-7914)


pkg/ocm/idps.go
```diff
@@ -152,22 +152,29 @@ func getIDPListWithoutAuthURLSupport() []string {
 }
 
 // BuildOAuthURL builds the correct OAuthURL depending on the cluster type
-func BuildOAuthURL(cluster *cmv1.Cluster) (string, error) {
+func BuildOAuthURL(cluster *cmv1.Cluster, includePort bool) (string, error) {
```
gdbranco, line 155: I see most types don't need to include port. What do you think of a check function kind like we have HasAuthURLSupport, so we don't need to include this bool

andreadecorte: good idea as it is more explicit, changed that

PR #1043, opened 2023-01-26
IDP related minor cleanup

andreadecorte:
* Use SDK enums which are now correct
* Remove a (broken) usage of a global variable to align with the rest


gdbranco:
I'm okay with this but the fixme label does say we needed to update API too, has that been fixed before already?

andreadecorte:
> I'm okay with this but the fixme label does say we needed to update API too, has that been fixed before already?

yes, sorry, forgot to link the related fix merged back in April 2022 https://github.com/openshift-online/ocm-api-model/pull/538

PR #1045, opened 2023-01-29
Create cluster - validate account roles have managed policies attached

oriAdler:
Error out in case of missing managed policies for the account roles.

Related: [SDA-7904](https://issues.redhat.com/browse/SDA-7904)

Control plane role is missing the attached policy:
image

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -31,6 +31,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/arn"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/openshift/rosa/cmd/create/machinepool"
+	"github.com/openshift/rosa/pkg/helper/roles"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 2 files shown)


oriAdler:

cmd/create/cluster/cmd.go
```diff
@@ -2629,3 +2641,38 @@ func buildCommand(spec ocm.Spec, operatorRolesPrefix string,
 func getRolePrefix(clusterName string) string {
 	return fmt.Sprintf("%s-%s", clusterName, helper.RandomLabel(4))
 }
+
+func validateManagedAccountRoles(r *rosa.Runtime, prefix string, awsClient aws.Client) error {
```
zgalor, line 2645: My only though is if these 2 funcs belong in the `create cluster cmd` or maybe can also be used for other commands, and should be placed in some common pkg?

oriAdler: Good point.
Reused `ValidateAccountRolesManagedPolicies` from the `roles` package.
Moved `ValidateUnmanagedAccountRoles` to the `roles` package as well, though didn't find yet a place to reuse the function.

PR #1046, opened 2023-01-31
Refactor `create account roles command` to use interfaces

oriAdler:
Use interface for role creation with implementation for `managed` and `unmanaged` policies, to simplify code.
This is a preparation to accommodate more changes for the managed policies flow.

Related: [SDA-7941](https://issues.redhat.com/browse/SDA-7941)

### DIFF
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -27,14 +27,10 @@ import (
 	"github.com/openshift/rosa/cmd/verify/oc"
 	"github.com/openshift/rosa/cmd/verify/quota"
 	"github.com/openshift/rosa/pkg/aws"
-	awscb "github.com/openshift/rosa/pkg/aws/commandbuilder"
-	"github.com/openshift/rosa/pkg/aws/tags"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/interactive/confirm"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/rosa"
-
-	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 )
 
 var args struct {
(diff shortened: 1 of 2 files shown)


cmd/create/accountroles/cmd.go
```diff
@@ -358,120 +360,144 @@ func run(cmd *cobra.Command, argv []string) {
...
+		permissionsBoundary:  permissionsBoundary,
+		accountID:            accountID,
+		env:                  env,
+		policies:             policies,
+		defaultPolicyVersion: defaultPolicyVersion,
+		path:                 path,
+	}
+}
+
+type rolesCreationInput struct {
```
gdbranco, line 391: I know this is only used here and that it is internal, but what do you think about naming it AccountRolesCreationInput? Just so we don't mix it up with other structs that will show up from refactors

cmd/create/accountroles/cmd.go
```diff
@@ -358,120 +360,144 @@ func run(cmd *cobra.Command, argv []string) {
...
-	commands := []string{}
+type creator interface {
+	createRoles(*rosa.Runtime, rolesCreationInput) error
+	getRoleTags(string, rolesCreationInput) map[string]string
+	buildCommands(rolesCreationInput) (string, error)
+}
+
+func initCreator(managedPolicies bool) creator {
+	if managedPolicies {
+		return managedPoliciesCreator{}
```
zgalor, line 371: should be `&managedPoliciesCreator`.

An interface is a pointer so implementing structs should also be pointers

cmd/create/accountroles/cmd.go
```diff
@@ -358,120 +360,144 @@ func run(cmd *cobra.Command, argv []string) {
...
+	accountID            string
+	env                  string
+	policies             map[string]*cmv1.AWSSTSPolicy
+	defaultPolicyVersion string
+	path                 string
+}
+
+type managedPoliciesCreator struct{}
+
+func (mp managedPoliciesCreator) createRoles(r *rosa.Runtime, input rolesCreationInput) error {
```
zgalor, line 403: `mp *managedPoliciesCreator`

zgalor: `input *rolesCreationInput` no need to copy by value the entire struct

cmd/create/accountroles/cmd.go
```diff
@@ -358,120 +360,144 @@ func run(cmd *cobra.Command, argv []string) {
...
-	policies map[string]*cmv1.AWSSTSPolicy, defaultPolicyVersion string,
-	path string, managedPolicies bool) error {
+func (mp managedPoliciesCreator) getRoleTags(roleType string, input rolesCreationInput) map[string]string {
+	tagsList := getBaseRoleTags(roleType, input)
+	tagsList[tags.ManagedPolicies] = "true"
+
+	return tagsList
+}
+
+type unmanagedPoliciesCreator struct{}
```
zgalor, line 466: It would be cleaner if the creators are in a separate go file

cmd/create/accountroles/cmd.go
oriAdler: Changed.

cmd/create/accountroles/cmd.go
oriAdler: Makes sense, fixed, thanks.

cmd/create/accountroles/cmd.go
oriAdler: Fixed.

cmd/create/accountroles/cmd.go
oriAdler: Moved to a new file `creators.go`.

gdbranco:
LGTM

I think you will run into similar refactors for other role types

PR #1048, opened 2023-01-31
SDA-7895 Enable editing subnet in machinepools for hosted clusters

lucasponce:
ROSA CLI UX for
https://issues.redhat.com/browse/SDA-7895

It requires backend MR 
https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/5154

it will enable using "subnet" parameter for creating a machine-pool for a hosted cluster.
Additionally, it will allow the user to resolve the "subnet" from an "availability zone."

ROSA CLI UX:

```

## Non interactive tests

# It will work; "subnet-06023544bf8e5b199" is PRIVATE, and it belongs to the VPC where the cluster was created
rosa create machinepool -c lponce-local-02 --name mp-2c --subnet subnet-06023544bf8e5b199 --replicas 2

# It will fail; "subnet-0fd5b560cd0518bb9" is PUBLIC
rosa create machinepool -c lponce-local-02 --name mp-2c-public --subnet subnet-0fd5b560cd0518bb9 --replicas 2

# It will fail; "subnet-bad" doesn't exist
rosa create machinepool -c lponce-local-02 --name mp-2c --subnet subnet-bad --replicas 2

# It will work; there is a private subnet under the "us-west-2b" AZ in the VPC that cluster was created
rosa create machinepool -c lponce-local-02 --name mp-2b --availability-zone us-west-2b

# It will fail; there is no a private subnet under the "bad-az" AZ
rosa create machinepool -c lponce-local-02 --name mp-2b --availability-zone bad-az --replicas 2

# It will move to interactive mode where there are several subnets for a given availability zone
rosa create machinepool -c lponce-local-03 --name mp-2d --availability-zone us-west-2d
I: There are several subnets for availability zone 'us-west-2d'
? Select subnet for a hosted machine pool (optional): Yes
? Subnet ID: subnet-00afcc1145d04b345 (us-west-2d)
? Replicas: 2
I: Fetching instance types
? Instance type: m5.xlarge
I: Machine pool 'mp-2d' created successfully on hosted cluster 'lponce-local-03'
I: To view all machine pools, run 'rosa list machinepools -c lponce-local-03'



## Interactive tests:

# Select subnet
 rosa create machinepool -c lponce-local-02 --name mp-2b --interactive
? Machine pool name: mp-2b
? Select subnet for a single AZ machine pool (optional): Yes
? Subnet ID:  [Use arrows to move, type to filter, ? for more help]
> subnet-06023544bf8e5b199 (us-west-2c)
  subnet-0310ceb7c0094e92a (us-west-2b)
  subnet-0d65e2465258986db (us-west-2a)
  subnet-00afcc1145d04b345 (us-west-2d)

# Select AZ
$ rosa create machinepool -c lponce-local-02 --name mp-2b --interactive
? Machine pool name: mp-2b
? Select subnet for a single AZ machine pool (optional): No
? AWS availability zone:  [Use arrows to move, type to filter, ? for more help]
  us-west-2c
  us-west-2b
> us-west-2a
  us-west-2d
```

Please @gdbranco @wgordon17 @pvasant take a look if the UX is addressed as expected.
Thanks

### DIFF
--- a/cmd/create/machinepool/helper.go
+++ b/cmd/create/machinepool/helper.go
@@ -1,0 +1,81 @@
+package machinepool
+
+import (
+	"os"
+
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/openshift/rosa/pkg/aws"
+	"github.com/openshift/rosa/pkg/interactive"
+	"github.com/openshift/rosa/pkg/rosa"
+	"github.com/spf13/cobra"
+)
+
+func getSubnetFromUser(cmd *cobra.Command, r *rosa.Runtime, isSubnetSet bool,
+	cluster *cmv1.Cluster) string {
+	var selectSubnet bool
+	var subnet string
+	var err error
+
+	question := "Select subnet for a single AZ machine pool"
+	questionError := "Expected a valid value for subnet for a single AZ machine pool"
+
+	if cluster.Hypershift().Enabled() {
+		question = "Select subnet for a hosted machine pool"
+		questionError = "Expected a valid value for subnet for a hosted machine pool"
+	}
+
+	if !isSubnetSet && interactive.Enabled() {
+		selectSubnet, err = interactive.GetBool(interactive.Input{
+			Question: question,
+			Help:     cmd.Flags().Lookup("subnet").Usage,
+			Default:  false,
+			Required: false,
+		})
+		if err != nil {
+			r.Reporter.Errorf(questionError)
+			os.Exit(1)
+		}
+	} else {
+		subnet = args.subnet
+	}
+
+	if selectSubnet {
+		subnetOptions, err := getSubnetOptions(r, cluster)
+		if err != nil {
+			r.Reporter.Errorf("%s", err)
+			os.Exit(1)
+		}
+
+		subnetOption, err := interactive.GetOption(interactive.Input{
+			Question: "Subnet ID",
+			Help:     cmd.Flags().Lookup("subnet").Usage,
+			Options:  subnetOptions,
+			Default:  subnetOptions[0],
+			Required: true,
+		})
+		if err != nil {
+			r.Reporter.Errorf("Expected a valid AWS subnet: %s", err)
+			os.Exit(1)
+		}
+		subnet = aws.ParseSubnet(subnetOption)
+	}
+
+	return subnet
+}
+
+// getSubnetOptions gets one of the cluster subnets and returns a slice of formatted VPC's private subnets.
+func getSubnetOptions(r *rosa.Runtime, cluster *cmv1.Cluster) ([]string, error) {
+	// Fetch VPC's subnets
+	privateSubnets, err := r.AWSClient.GetVPCPrivateSubnets(cluster.AWS().SubnetIDs()[0])
+	if err != nil {
+		return nil, err
+	}
+
+	// Format subnet options
+	var subnetOptions []string
+	for _, subnet := range privateSubnets {
+		subnetOptions = append(subnetOptions, aws.SetSubnetOption(*subnet.SubnetId, *subnet.AvailabilityZone))
+	}
+
+	return subnetOptions, nil
+}
--- a/cmd/create/machinepool/nodepool.go
+++ b/cmd/create/machinepool/nodepool.go
@@ -17,19 +17,11 @@ import (
 func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r *rosa.Runtime) {
 	var err error
 
-	// TODO NodePool commands don't support (yet) some of the machinepool flags
-	if cmd.Flags().Changed("multi-availability-zone") {
-		r.Reporter.Errorf("Setting the `multi-availability-zone` flag is not yet supported for hosted clusters")
-		os.Exit(1)
-	}
-
-	if cmd.Flags().Changed("availability-zone") {
-		r.Reporter.Errorf("Setting the `availability-zone` flag is not yet supported for hosted clusters")
-		os.Exit(1)
-	}
-
-	if cmd.Flags().Changed("subnet") {
-		r.Reporter.Errorf("Setting the `subnet` flag is not yet supported for hosted clusters")
+	isAvailabilityZoneSet := cmd.Flags().Changed("availability-zone")
+	isSubnetSet := cmd.Flags().Changed("subnet")
+	if isSubnetSet && isAvailabilityZoneSet {
+		r.Reporter.Errorf("Setting both `subnet` and `availability-zone` flag is not supported." +
+			" Please select `subnet` or `availability-zone` to create a single availability zone machine pool")
 		os.Exit(1)
 	}
 
@@ -57,6 +49,18 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
 	if !machinePoolKeyRE.MatchString(name) {
 		r.Reporter.Errorf("Expected a valid name for the machine pool")
 		os.Exit(1)
+	}
+
+	// Allow the user to select subnet for a single AZ BYOVPC cluster
+	subnet := getSubnetFromUser(cmd, r, isSubnetSet, cluster)
+
+	// Select availability zone if the user didn't select subnet
+	if subnet == "" {
+		subnet, err = getSubnetFromAvailabilityZone(cmd, r, isAvailabilityZoneSet, cluster)
+		if err != nil {
+			r.Reporter.Errorf("%s", err)
+			os.Exit(1)
+		}
 	}
 
 	isMinReplicasSet := cmd.Flags().Changed("min-replicas")
@@ -147,6 +151,10 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
 		npBuilder = npBuilder.Replicas(replicas)
 	}
 
+	if subnet != "" {
+		npBuilder.Subnet(subnet)
+	}
+
 	// Machine pool instance type:
 	// NodePools don't support MultiAZ yet, so the availabilityZonesFilters is calculated from the cluster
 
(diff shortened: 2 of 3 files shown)


pvasant:
if the user specifies the az and there are more than one subnet in the AZ do we drop to interactive mode, cc: @oriAdler

lucasponce:
> 
if the user specifies the az and there are more than one subnet in the AZ do we drop to interactive mode, cc: @oriAdler

Good point. No, that case is missing on this PR, I'll add it

cmd/create/machinepool/nodepool.go
```diff
@@ -59,6 +57,21 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
 		os.Exit(1)
 	}
 
+	// Allow the user to select subnet for a single AZ BYOVPC cluster
+	var subnet string
+	if isBYOVPC(cluster) {
```
oriAdler, line 62: Do we need this check or all Hypershift clusters are going to be BYOVPC? 
Does the user always provide subnets on Hypershift cluster creation - day 1?

cmd/create/machinepool/nodepool.go
```diff
@@ -59,6 +57,21 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
 		os.Exit(1)
 	}
 
+	// Allow the user to select subnet for a single AZ BYOVPC cluster
```
oriAdler, line 60: Does this feature is supported only for single AZ clusters?

cmd/create/machinepool/nodepool.go
```diff
@@ -59,6 +57,21 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
...
+	var subnet string
+	if isBYOVPC(cluster) {
+		subnet = getSubnetFromUser(cmd, r, isSubnetSet, cluster)
```
oriAdler, line 63: NP
Since this function is now used both for `machinepool` and `nodepool` I wonder if we should move the implementation from `machinepool.go` to a helper file in this package.

cmd/create/machinepool/nodepool.go
```diff
@@ -222,3 +239,44 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
...
+			availabilityZones = append(availabilityZones, *privateSubnet.AvailabilityZone)
+			subnetsMap[*privateSubnet.AvailabilityZone] = *privateSubnet.SubnetId
+		}
+	}
+	availabilityZone := cluster.Nodes().AvailabilityZones()[0]
+	if !isAvailabilityZoneSet && interactive.Enabled() {
+		availabilityZone, err = interactive.GetOption(interactive.Input{
+			Question: "AWS availability zone",
+			Help:     cmd.Flags().Lookup("availability-zone").Usage,
+			Options:  availabilityZones,
```
oriAdler, line 265: Just to simplify it for us and the users I would suggest following the same pattern as classic ROSA.
1. If the user chooses AZ, we go for the cluster's subnet belonging to the AZ, which was provided on day 1.
2. Otherwise, the user can select a subnet that belongs to the VPC, choosing AZ implicitly.

I think it covers all cases, please let me know if I missed anything.

lucasponce:
> > if the user specifies the az and there are more than one subnet in the AZ do we drop to interactive mode, cc: @oriAdler
> 
> In classic ROSA if the user specifies AZ we pass the AZ to the backend, and CS will choose the subnet belonging to this AZ that was provided on day 1. Moreover, we allow the user to select AZ from the clusters AZs https://github.com/openshift/rosa/blob/master/cmd/create/machinepool/machinepool.go#L130.
> 
> If the user decides not to select from the cluster's AZs, they could choose the subnet previously, because when they are choosing a subnet they are choosing also AZ implicitly.
> 
> @lucasponce what are the requirements for the node pool?

Requeriments are close but the same.
NodePools doesn't have the concept of MultiAZ, basically a NodePool is linked with a subnet (and an AZ).

And I guess this can be more dynamic in Day 2, as they can create a new NP in a new subnet that doesn't exist in Day 1.

So, then it makes sense @pvasant suggestion, if users select a specific AZ, then we can ask for the specific subnet if there are more than one.

cmd/create/machinepool/nodepool.go
lucasponce: All hosted clusters are BYOVPC, so perhaps we can assume this validation.
On Day 1 they can provide the subnets in the similar flow as rosa classic clusters (2 subnets for singleaz, 6 per multiaz) but only private subnets are used.

But I guess on Day 2 user may have new subnets available and they can create new NodePools on those new subnets.

cmd/create/machinepool/nodepool.go
lucasponce: I'd say yes, as single AZ, multi-AZ is somehow deprecated for hosted clusters. As soon as subnets are in the  VPC, customers can create a new NodePool on them.

Hosted ControlPlane is always in HA (MultiAZ) mode.

cmd/create/machinepool/nodepool.go
lucasponce: +1 I can add a helper, it makes sense.

cmd/create/machinepool/nodepool.go
lucasponce: But I guess we will hit again the scenario where users in day 2 add a new subnet to that AZ, then perhaps @pvasant makes sense to detect that there is more than one and fall into the subnet selection.

oriAdler: I agree.

What I meant is that if you fall into a subnet selection, it was possible to select a subnet in the first place.
For example:
1. ? Select subnet for a single AZ machine pool (optional): No
2. ? AWS availability zone:  [Use arrows to move, type to filter, ? for more help]
  us-west-2c
  us-west-2b
> us-west-2a
3. ? Subnet ID:  [Use arrows to move, type to filter, ? for more help]
> subnet-06023544bf8e5b199 (us-west-2c)

lucasponce: Ah, I think that if the subnet is selected in the first place, it shouldn't ask you to select AZ.
But yes, if you select AZ, and it detects more than one Subnet, it will ask again for a Subnet.

oriAdler: I discussed this with @pvasant and she said it's ok to limit the AZ selection to the cluster's AZs.
So if the user wants to use another AZ they have to provision a subnet there and select it.

It's more of UX experience than functionality.
I would recommend sticking with the way it's done in classic ROSA, but it's your call.
Let me know if I missed anything.

lucasponce: > So if the user wants to use another AZ they have to provision a subnet there and select it.

So, what's the usual flow in rosa classic? users edit the cluster to add an additional subnet there?

lucasponce: But, do users need to add any new subnet in the cluster even if that subnet exists in the same AZ?

Perhaps these two flows are not 100% clear to me.

lucasponce: Anyway, I'm ok with the suggestions @oriAdler @pvasant I have a pending backend work for Day 1 that will enable the pending updates in the cluster to check this.

wgordon17: > limit the AZ selection to the cluster's AZs

There _are_ no cluster AZs, so this doesn't really make sense ¯\_(ツ)_/¯ 

IMO, we should take whatever input is provided by the customer, and then validate, e.g.,

- `rosa create machinepool --subnet subnet-1234` -> `subnet-1234` exists and is private -> OK
- `rosa create machinepool --subnet subnet-bad` -> `subnet-bad` doesn't exist -> ERROR
- `rosa create machinepool --availability-zone us-east-1a` -> `us-east-1a` has more than 1 private subnet -> interactive mode
- `rosa create machinepool --availability-zone us-east-1b` -> `us-east-1b` has 1 private subnet -> OK
- `rosa create machinepool --availability-zone us-east-1c` -> `us-east-1c` has 0 private subnets -> ERROR

There shouldn't need to be any additional consideration or validation on subnets/AZs that were provided on day-1.

This is a restriction on _classic_ ROSA only because of the Control Plane node placement, but that restriction doesn't exist in HCP

wgordon17:
@lucasponce My only minor nit is

> subnet for a single AZ machine pool

Can we remove "single AZ"? It implies that it would be possible to create a "multi AZ" machine pool and I don't want to cause confusion or open ourselves up to questions.

lucasponce:
@wgordon17 @pvasant @oriAdler 
I've tried to address all comments.

I've tested the scenarios described in the description/first comment:
https://github.com/openshift/rosa/pull/1048#issue-1564836204

Please, let me know if this looks good to consolidate.
Thanks

cmd/create/machinepool/helper.go
```diff
@@ -0,0 +1,81 @@
+package machinepool
+
+import (
+	"os"
+
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/openshift/rosa/pkg/aws"
+	"github.com/openshift/rosa/pkg/interactive"
+	"github.com/openshift/rosa/pkg/rosa"
+	"github.com/spf13/cobra"
```
pvasant, line 10: np: import order

cmd/create/machinepool/helper.go
```diff
@@ -0,0 +1,81 @@
...
+	var selectSubnet bool
+	var subnet string
+	var err error
+
+	question := "Select subnet for a single AZ machine pool"
+	questionError := "Expected a valid value for select subnet for a single AZ machine pool"
+
+	if cluster.Hypershift().Enabled() {
+		question = "Select subnet for a hosted machine pool"
+		questionError = "Expected a valid value for select subnet for a hosted machine pool"
```
pvasant, line 24: "Expected a valid value for select subnet for a hosted machine pool" 

shouldnt this be 

""Expected a valid value for subnet for a hosted machine pool

cmd/create/machinepool/nodepool.go
pvasant: My only concern right now is we should then allow giving subnet and az together in the command else automation will not work for these scenarios. 

rosa create machinepool --availability-zone us-east-1a -> us-east-1a has more than 1 private subnet -> interactive mode

pvasant: @lucasponce i think we need to support giving giving both subnet and AZ for HS

lucasponce: @pvasant +1 I can update it that way.

lucasponce: Or perhaps, what about

rosa create machinepool --availablility-zone us-east-1a -> Move to interactive only if --interactive is set, otherwise, to fail ?

cmd/create/machinepool/helper.go
pvasant: np: 

```
"os"

"github.com/spf13/cobra"

	"github.com/openshift/rosa/pkg/aws"
```

cmd/create/machinepool/nodepool.go
pvasant: I m not sure if i understand . requirement is there should  be a way to run without interactive mode. they pass subnet then az is not mandatory since we can derive if they pass az subnet is mandatory to run in non interactive mode right

lucasponce: Ok, perhaps I was overthinking these two cases:

- When subnet and AZ are both passed, if subnet doesn't belong to AZ, should it fail? or should ignore the AZ parameter?
- When AZ is passed, if there are more than one subnet and it's not in interactive mode should fail, right?

oriAdler:
LGTM

cmd/create/machinepool/nodepool.go
wgordon17: IMO, I think the two (AZ and subnet) should be mutually exclusive, and provide a helpful error message when both provided.

If you want to use a specific _AZ_, then you should provide that.

If you want to use a specific _subnet_, then you should provide that.

Providing both, seems redundant and leads to confusion as illustrated ¯\_(ツ)_/¯

lucasponce: Thanks, @wgordon17 @pvasant @oriAdler; I think we are very close to addressing this one.

So, then if the user specifies an AZ, and this AZ has more than one subnet:
- If it's in interactive mode, it should ask to select a subnet.
- If not, it should fail.

Am I right?

oriAdler: Not sure about the failure, should we fall into interactive mode in this case?

lucasponce: The current tests & scenarios that these PR implements are described https://github.com/openshift/rosa/pull/1048#issue-1564836204

I also suggest that perhaps we can progress with those (if there are no blockers), and we can refine them in iterative steps if that looks oks

PR #1050, opened 2023-02-02
[SDA-8006] Add user prefix to oidc configuration

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8006

### DIFF
--- a/cmd/create/oidcconfig/cmd.go
+++ b/cmd/create/oidcconfig/cmd.go
@@ -49,12 +49,14 @@ import (
 )
 
 const (
-	DEFAULT_LENGTH_RANDOM_LABEL = 4
+	defaultLengthRandomLabel = 4
+	maxLengthUserPrefix      = 15
 )
 
 var args struct {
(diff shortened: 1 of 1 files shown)


cmd/create/oidcconfig/cmd.go
```diff
@@ -138,6 +150,27 @@ func run(cmd *cobra.Command, argv []string) {
...
+		})
+		if err != nil {
+			r.Reporter.Errorf("Expected a valid prefix for the configuration: %s", err)
+			os.Exit(1)
+		}
+		args.userPrefix = prefix
+	}
+	args.userPrefix = strings.Trim(args.userPrefix, " \t")
+
+	if len([]rune(args.userPrefix)) > maxLengthUserPrefix {
```
hunterkepley, line 168: Why are we making the string a run slice here?

hunterkepley: gotcha, didn't know this :)

PR #1051, opened 2023-02-03
[SDA-8022] Forcing creation should only work for unmanaged policies

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8022
# What
When forcing creation of policies for updating actions on current version should validate if they are using managed policies first

# Why
Managed policies are always up to date and don't require forcing update

# Changes
`rosa create account-roles --mode auto -y -f --managed`
```
.
.
.
W: Forcing creation of policies only works for unmanaged policies
```

### DIFF
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -153,6 +153,11 @@ func run(cmd *cobra.Command, argv []string) {
 	}
 	managedPolicies := args.managed
 
+	if args.forcePolicyCreation && managedPolicies {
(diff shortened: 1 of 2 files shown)


cmd/create/accountroles/cmd.go
```diff
@@ -292,6 +292,11 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
+	if args.forcePolicyCreation && managedPolicies {
```
oriAdler, line 295: NP
Should we move this check to line `154`? right after determining if the user has chosen managed policies.

oriAdler (commented):

PR #1057, opened 2023-02-07
[SDA-8059] Add command delete oidc-config and minor fixes

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8059
# What
ROSA needs to support deleting oidc configs

# Why
Currently the client needs to go to aws to remove the configs that were created within ROSA

### DIFF
--- a/cmd/dlt/oidcconfig/cmd.go
+++ b/cmd/dlt/oidcconfig/cmd.go
@@ -1,0 +1,230 @@
+/*
+Copyright (c) 2023 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package oidcconfig
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/briandowns/spinner"
+	"github.com/spf13/cobra"
+	"github.com/zgalor/weberr"
+
+	"github.com/openshift/rosa/pkg/arguments"
+	"github.com/openshift/rosa/pkg/aws"
+	awscb "github.com/openshift/rosa/pkg/aws/commandbuilder"
+	"github.com/openshift/rosa/pkg/interactive"
+	"github.com/openshift/rosa/pkg/interactive/confirm"
+	"github.com/openshift/rosa/pkg/rosa"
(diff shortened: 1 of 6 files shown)


cmd/dlt/oidcconfig/cmd.go
```diff
@@ -0,0 +1,230 @@
...
+	confirm.AddFlag(flags)
+}
+
+func run(cmd *cobra.Command, argv []string) {
+	r := rosa.NewRuntime().WithAWS()
+	defer r.Cleanup()
+
+	mode, err := aws.GetMode()
+	if err != nil {
+		r.Reporter.Errorf("%s", err)
```
hunterkepley, line 83: should this be `%w`?

cmd/dlt/oidcconfig/cmd.go
```diff
@@ -0,0 +1,230 @@
...
+		if err != nil {
+			r.Reporter.Errorf("Expected a valid OIDC provider creation mode: %s", err)
+			os.Exit(1)
+		}
+	}
+
+	oidcConfigInput := buildOidcConfigInput(r)
+	oidcConfigStrategy, err := getOidcConfigStrategy(mode, oidcConfigInput)
+	if err != nil {
+		r.Reporter.Errorf("%s", err)
```
hunterkepley, line 117: Same for this, `%w`?

pkg/aws/client.go
```diff
@@ -956,6 +958,37 @@ func (c *awsClient) CreateS3Bucket(bucketName string, region string) error {
 	return nil
 }
 
+func (c *awsClient) DeleteS3Bucket(bucketName string) error {
+	c.emptyS3Bucket(bucketName)
+	bucketInput := &s3.DeleteBucketInput{
+		Bucket: aws.String(bucketName),
+	}
+	_, err := c.s3Client.DeleteBucket(bucketInput)
```
hunterkepley, line 966: Is the `emptyS3Bucket` operation a quick one? Do we need to wait in case the bucket has large items in it that take some time to empty? Or is that not a worry here due to what the bucket *should* contain?

hunterkepley (commented):
Looks great overall, added some small comments

pkg/aws/client.go
gdbranco: It should only contain two small files, if it has anything else the client used it wrongfully as far as we need to know

cmd/dlt/oidcconfig/cmd.go
gdbranco: Makes sense, but we need to check. Today we don't almost don't use `%w` so I think we should refactor alongside everything else at some other point

cmd/dlt/oidcconfig/cmd.go
gdbranco: Same as https://github.com/openshift/rosa/pull/1057#discussion_r1099081924

cmd/dlt/oidcconfig/cmd.go
hunterkepley: I noticed that as well, we should be using this when possible for error wrapping. Maybe this can be tackled at another point in time when we can refactor every error message

PR #1059, opened 2023-02-08
[SDA-8091] Improve error messages for deleting oidc-config

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8091
# What
Improving some error messages when deleting oidc-config

# Why
Clearer steps for the client

### DIFF
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -52,6 +52,7 @@ import (
 	"github.com/openshift/rosa/pkg/fedramp"
 	"github.com/openshift/rosa/pkg/reporter"
 	"github.com/sirupsen/logrus"
+	"github.com/zgalor/weberr"
 
 	"github.com/openshift/rosa/pkg/aws/profile"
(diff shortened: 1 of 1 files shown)


pkg/aws/client.go
```diff
@@ -1021,7 +1037,13 @@ func (c *awsClient) CreateSecretInSecretsManager(name string, secret string) (st
 }
 
 func (c *awsClient) DeleteSecretInSecretsManager(secretArn string) error {
-	_, err := c.smClient.DeleteSecret(
+	_, err := c.smClient.DescribeSecret(&secretsmanager.DescribeSecretInput{
```
oriAdler, line 1040: Are we aiming to determine if the `arn` is valid here?
If so, can we use the `arn.Parse` function and save an API call to AWS?

gdbranco: No, this aims at checking if secret exists in aws account

PR #1061, opened 2023-02-08
[SDA-8092] Adding validation to bucket name

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8092
# What
Validate bucket names before moving forward into creation

# Why
Some names are invalid for aws s3 bucket names

### DIFF
--- a/cmd/create/oidcconfig/cmd.go
+++ b/cmd/create/oidcconfig/cmd.go
@@ -31,6 +31,7 @@ import (
 	"encoding/pem"
 	"fmt"
 	"os"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cmd/create/oidcconfig/cmd.go
```diff
@@ -390,6 +399,18 @@ func (s *CreateOidcConfigManualStrategy) execute(r *rosa.Runtime) {
 		Build()
 	commands = append(commands, putJwksCommand)
 	commands = append(commands, fmt.Sprintf("rm %s", jwksFilename))
+	createSecretCommand := awscb.NewSecretsManagerCommandBuilder().
```
gdbranco, line 402: Changing so that if the bucket name is still not valid for some reason we don't create the secret

cmd/create/oidcconfig/cmd.go
```diff
@@ -310,6 +326,11 @@ func (s *CreateOidcConfigAutoStrategy) execute(r *rosa.Runtime) {
 			"to S3 bucket '%s': %s", bucketName, err)
 		os.Exit(1)
 	}
+	secretsARN, err := r.AWSClient.CreateSecretInSecretsManager(privateKeySecretName, string(privateKey[:]))
```
gdbranco, line 329: Changing so that if the bucket name is still not valid for some reason we don't create the secret

PR #1064, opened 2023-02-09
aws: Ensure ARNs have the correct partition

vkareh:
Not all ARNs belong to the 'aws' partition based on their region. This ensures that all ARNs have the correct one.


vkareh:
/hold
until this is tested

oriAdler:
LGTM

PR #1066, opened 2023-02-09
[SDA-7716] Add possibility to reuse operator roles

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-7716
# What
Allows reusing operator roles by specifying each operator role or operator role prefix as long as the provided OIDC matches the the trust relationship to the role

# Why
Client does not need to created operator roles for each cluster

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -1136,8 +1136,11 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
-	credRequests, err := r.OCMClient.GetCredRequests(isHostedCP)
+	isByoOidcSet := false
+	oidcEndpointUrl := ""
+	oidcPrivateKeySecretArn := ""
 	if isSTS {
+		credRequests, err := r.OCMClient.GetCredRequests(isHostedCP)
 		if err != nil {
 			r.Reporter.Errorf("Error getting operator credential request from OCM %s", err)
 			os.Exit(1)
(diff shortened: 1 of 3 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -1194,22 +1198,21 @@ func run(cmd *cobra.Command, _ []string) {
...
-			if err != nil {
-				r.Reporter.Errorf("Error validating role arn. Arn: %v. Error: %v", role.RoleARN, err)
-				os.Exit(1)
-			}
-
-			err = awsClient.ValidateRoleNameAvailable(name)
-			if err != nil {
-				r.Reporter.Errorf("Error validating role: %v", err)
+		isByoOidcSet, oidcEndpointUrl, oidcPrivateKeySecretArn = handleByoOidcOptions(r, cmd, isSTS)
+		err = validateOperatorRolesAvailabilityUnderUserAwsAccount(awsClient, operatorIAMRoleList)
```
oriAdler, line 1202: If it's a `ByoOidc` flow, do we still have to check that the operator role's name is available?

cmd/create/cluster/cmd.go
```diff
@@ -2251,6 +2257,42 @@ func run(cmd *cobra.Command, _ []string) {
...
+func validateOperatorRolesMatchOidcProvider(awsClient aws.Client,
+	operatorIAMRoleList []ocm.OperatorIAMRole, oidcEndpointUrl string) error {
+	operatorIAMRoles := operatorIAMRoleList
+	for _, operatorIAMRole := range operatorIAMRoles {
+		roleARN := operatorIAMRole.RoleARN
+		roleObject, err := awsClient.GetRoleByARN(roleARN)
+		if err != nil {
+			return err
+		}
+		parsedUrl, err := url.Parse(oidcEndpointUrl)
```
oriAdler, line 2285: Can we extract the parsing outside of the loop?

cmd/create/operatorroles/cmd.go
```diff
@@ -542,3 +561,22 @@ func validateOperatorRoles(r *rosa.Runtime, cluster *cmv1.Cluster) ([]string, er
 	}
 	return missingRoles, nil
 }
+
+func validateOperatorRolesMatchOidcProvider(r *rosa.Runtime, cluster *cmv1.Cluster) error {
```
oriAdler, line 565: This function is very similar to `validateOperatorRolesMatchOidcProvider` in `cmd/create/cluster/cmd.go`.
Can we extract the function to a common file and reuse it?

cmd/create/cluster/cmd.go
gdbranco: Not really, it can not exist or be in use and it should pick it up either way without any issues. At some point in the future we'll allow to actually create them using the prefix prior to creating the cluster. By then we might want to enforce they pre exist.

cmd/create/cluster/cmd.go
gdbranco: Sure, I feel like I did this lol. Maybe it was in CS or it got left out in some stash

cmd/create/operatorroles/cmd.go
gdbranco: Sort of, had to adapt it as the operator roles types are different. But it does work

PR #1069, opened 2023-02-13
Add a `AWS managed` column to list role commands

oriAdler:
Display a new column with info about the role's policies.

Related: [SDA-7969](https://issues.redhat.com/browse/SDA-7969)

image

image


oriAdler:

PR #1070, opened 2023-02-13
Add etcd encyprtion kms arn support

zgalor:
Related: SDA-7675


zgalor:
@oriAdler PTAL

cmd/create/cluster/cmd.go
```diff
@@ -1871,7 +1880,24 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	etcdEncryption := args.etcdEncryption
-	if interactive.Enabled() && !fips {
+	etcdEncryptionKmsARN := args.etcdEncryptionKmsARN
+
+	if etcdEncryptionKmsARN != "" {
+		// validate hosted cp
+		if !isHostedCP {
```
oriAdler, line 1887: NP
We can check that earlier after line `677` where `isHostedCP` is set, to fail the user earlier.
Maybe something like:
```
if isHostedCP && args.etcdEncryptionKmsARN != "" {
		r.Reporter.Errorf("etcd encryption kms arn is only allowed for hosted cp")
		os.Exit(1)
	}
```

cmd/create/cluster/cmd.go
```diff
@@ -1891,6 +1917,27 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
+	if etcdEncryption && isHostedCP && (etcdEncryptionKmsARN == "" || interactive.Enabled()) {
+		etcdEncryptionKmsARN, err = interactive.GetString(interactive.Input{
+			Question: "Etcd encryption KMS ARN",
+			Help:     cmd.Flags().Lookup("etcd-encryption-kms-arn").Usage,
+			Default:  etcdEncryptionKmsARN,
+			Required: etcdEncryption && isHostedCP,
```
oriAdler, line 1925: NP
If we entered the `if` statement in line `1920`, is it always going to equal `true`?

oriAdler (commented):
Left two small comments, other than that LGTM

cmd/create/cluster/cmd.go
zgalor: Fixed

PR #1071, opened 2023-02-13
Release v1.2.15

oriAdler:
- fix: improve error messages for deleting oidc-config
- fix: adding some validations to bucket name
- fix: allow empty label match editing ingress interactive
- aws: Ensure ARNs have the correct partition
- Attach three policies to the installer role - managed policies
- to fix empty DNS domain when DNS not ready: [SDA-7418](https://issues.redhat.com//browse/SDA-7418)
- fix: add a few more validations to bucket/folder name
- Add a `AWS managed` column to list role commands
- Add etcd encyprtion kms arn support

### DIFF
--- a/CHANGES.adoc
+++ b/CHANGES.adoc
@@ -2,6 +2,18 @@
 
 This document describes the relevant changes between releases of the `rosa` command line tool.
 
+== 1.2.15 Feb 13 2023
+
+- fix: improve error messages for deleting oidc-config
+- fix: adding some validations to bucket name
+- fix: allow empty label match editing ingress interactive
+- aws: Ensure ARNs have the correct partition
+- Attach three policies to the installer role - managed policies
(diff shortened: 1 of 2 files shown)


hack/commit-release.sh
```diff
@@ -19,7 +19,8 @@
 
 git fetch --tags
 tagcommit=$(git rev-list --tags --max-count=1)
-current=$(git describe --tags $tagcommit | sed 's/v//')
+#current=$(git describe --tags $tagcommit | sed 's/v//')
+current="1.2.14"
```
gdbranco, line 23: This shouldn't be needed
You usually fix it by `git fetch upstream` to get latest tags locally

oriAdler: Good catch!
Fixed, thanks.

PR #1072, opened 2023-02-14
Fix etcd encryption default enforecment

zgalor:
Default should be false unless fips or kms arn provided
Related: [SDA-7675](https://issues.redhat.com//browse/SDA-7675)


andreadecorte:
lgtm

PR #1076, opened 2023-02-15
[SDA-8074] Add description when available for error state in describe cluster

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8074
# What
Add description when available for error state in describe cluster

# Why
Better info for client/SRE to understand

### DIFF
--- a/cmd/describe/cluster/cmd.go
+++ b/cmd/describe/cluster/cmd.go
@@ -99,16 +99,8 @@ func run(cmd *cobra.Command, argv []string) {
 	switch cluster.State() {
 	case cmv1.ClusterStateWaiting:
 		phase = "(Waiting for user action)"
-		status, _ := r.OCMClient.GetClusterStatus(cluster.ID())
-		if status.Description() != "" {
-			phase = fmt.Sprintf("(%s)", status.Description())
-		}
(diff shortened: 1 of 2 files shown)


cmd/describe/cluster/cmd.go
```diff
@@ -120,6 +120,11 @@ func run(cmd *cobra.Command, argv []string) {
 			}
 			phase = "(" + errorCode + "Install is taking longer than expected)"
 		}
+	case cmv1.ClusterStateError:
+		status, _ := r.OCMClient.GetClusterStatus(cluster.ID())
```
tbrisker, line 124: You don't need to get the cluster status, the description should already be present in cluster.Status() (but perhaps we need to update the sdk if it isn't there). That's probably also true for other places where we currently use `GetClusterStatus`.

gdbranco: Please take a second look, tested with error state and it did work

gdbranco: Just tested with waiting state and seems fine, waiting for install to complete

gdbranco: Seems fine, let me know what you think later

PR #1077, opened 2023-02-15
[WIP] Add hosted-cp flag to account roles

jharrington22:
This PR adds the hidden flag `--hosted-cp` to the account-roles command. It should do two things

1. By default (not supplying the flag) add both the regular installer permission policy and the new `sts_installer_hosted_cp_permission_policy` policy to the installer role
2. Allow attaching just the `sts_installer_hosted_cp_permission_policy` policy to the install role when the flag is passed ie `rosa create account-roles --hosted-cp`

The new permission policy will need to be added to OCM for this to work

### DIFF
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -41,6 +41,9 @@ var args struct {
 	channelGroup        string
 	managed             bool
 	forcePolicyCreation bool
+
+	// Hypershift options:
+	hostedCPEnabled bool
 }
 
 var Cmd = &cobra.Command{
@@ -113,6 +116,16 @@ func init() {
 		"Forces creation of policies skipping compatibility check",
 	)
 
+	// Options releated to HyperShift:
+	flags.BoolVar(
+		&args.hostedCPEnabled,
+		"hosted-cp",
+		false,
+		"Enable the use of hosted control planes (HyperShift)",
+	)
+
+	flags.MarkHidden("hosted-cp")
+
 	aws.AddModeFlag(Cmd)
 
 	confirm.AddFlag(flags)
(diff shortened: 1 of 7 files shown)


cmd/create/accountroles/cmd.go
```diff
@@ -145,6 +158,12 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
+	// Set HostedCP for Runtime if we are creating HostedCP roles and policies
+	isHostedCP := cmd.Flags().Changed("hosted-cp")
+	if isHostedCP {
+		r.IsHostedCP()
```
oriAdler, line 164: Not sure if this attribute should be part of the `Runtime` or elsewhere.
Maybe we should have this field added here **accountRolesCreationInput** https://github.com/openshift/rosa/blob/master/cmd/create/accountroles/creators.go#L28 for this command.
@tbrisker tagging you here since you presented the `Runtime` struct.

cmd/create/accountroles/creators.go
```diff
@@ -118,6 +118,13 @@ type unmanagedPoliciesCreator struct{}
 
 func (up *unmanagedPoliciesCreator) createRoles(r *rosa.Runtime, input *accountRolesCreationInput) error {
 	for file, role := range aws.AccountRoles {
+		// If hosted-cp flag was passed, create only the hosted CP installer policy
+		if r.HostedCP != nil {
+			if *r.HostedCP && file == aws.InstallerAccountRole {
```
oriAdler, line 123: Here we skip also the role creation in line `137` **r.AWSClient.EnsureRole**.
We have another function to create roles with managed policies https://github.com/openshift/rosa/blob/master/cmd/create/accountroles/creators.go#L54.

oriAdler (commented):
Are the Hive Installer policy permissions going to be a subset of the Hypershift Installer policy permissions?

cmd/create/accountroles/cmd.go
tbrisker: I agree, this doesn't feel like it belongs as part of the `Runtime`

PR #1078, opened 2023-02-15
[OHSS-18020] Improve logging so that it's more obvious what is wrong

mjlshen:
Customers can get this log when attempting to upgrade account and operator roles, but it's not immediately obvious what the problem is from the existing logs:

```
I: Ensuring account and operator role policies for cluster 'REDACTED' are compatible with upgrade.
E: expected a valid IAM role ARN: arn: invalid prefix
```

[OHSS-18020](https://issues.redhat.com//browse/OHSS-18020)


jharrington22 (approved):
/lgtm

PR #1081, opened 2023-02-17
[SDA-8196] Set byo oidc enabled when specifying byo oidc attributes

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8196
# What
Set BYO OIDC enabled when user specifies BYO OIDC attributes, survey question for interactive mode will come as part of exposing BYO OIDC commands

# Why
Allows to search for BYO OIDC clusters more easily


ciaranRoche (approved):
/lgtm

PR #1086, opened 2023-02-23
Use cluster attribute `ManagedPolicies` to identify policy type

oriAdler:
Since we are now persisting this information upon cluster creation, use the attribute to reduce API calls to AWS.

Related: [SDA-8278](https://issues.redhat.com/browse/SDA-8278)


oriAdler:
@zgalor @davidleerh PTAL

zgalor:
@oriAdler is the code that populates this field in CS already in prod?

PR #1091, opened 2023-02-28
SDA-8325 Add subnets field for Default Worker

lucasponce:
A minimal PR to fix https://issues.redhat.com/browse/SDA-8325
As the "Default" machinepool for Hive clusters is built in the rosa side, it requires logic to populate the subnets from the cluster.

Please @oriAdler @zgalor @gdbranco take a look; feel free to suggest if there is a better way to patch this. 
I don't usually work on hive clusters, and I may miss some context.

Thanks

### DIFF
--- a/cmd/list/machinepool/machinepool.go
+++ b/cmd/list/machinepool/machinepool.go
@@ -33,6 +33,12 @@ func listMachinePools(r *rosa.Runtime, clusterKey string, cluster *cmv1.Cluster)
 				MaxReplicas(cluster.Nodes().AutoscaleCompute().MaxReplicas()),
 		)
 	}
+
+	// In case of AWS clusters we can query the subnbets
+	if cluster.AWS() != nil && len(cluster.AWS().SubnetIDs()) > 0 {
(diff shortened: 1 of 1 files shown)


cmd/list/machinepool/machinepool.go
```diff
@@ -33,6 +33,12 @@ func listMachinePools(r *rosa.Runtime, clusterKey string, cluster *cmv1.Cluster)
 				MaxReplicas(cluster.Nodes().AutoscaleCompute().MaxReplicas()),
 		)
 	}
+
+	// In case of AWS clusters we can query the subnbets
+	if cluster.AWS() != nil && len(cluster.AWS().SubnetIDs()) > 0 {
```
gdbranco, line 38: I'm okay considering that's how we are dealing with default subnet already. I wonder why we don't return it back within the machine pool endpoint

lucasponce: I'm afraid it's by design; in hive clusters, the "Default" machinepool is built in rosa from the cluster itself.
The bug I try to fix is to add the subnets that are missing in the list.

PR #1096, opened 2023-03-03
[SDA-7992] Add option to remove oidc provider from oidc endpoint url

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-7992
# What
Adds support for deleting oidc providers created for/from BYO OIDC. Also adds optional step to create OIDC provider after creating OIDC configuration, equivalent flow for deleting OIDC config.

# Why
Facilitates flow for customers using BYO OIDC as they don't need to go to aws console/cli.


cmd/dlt/oidcprovider/cmd.go
```diff
@@ -53,67 +71,110 @@ func run(cmd *cobra.Command, argv []string) {
...
-	if err != nil {
-		r.Reporter.Errorf("Failed to get the OIDC provider for cluster '%s'.", clusterKey)
-		os.Exit(1)
-	}
-	if providerARN == "" {
-		r.Reporter.Infof("Cluster '%s' doesn't have OIDC provider associated with it.", clusterKey)
-		return
-	}
+		}
+		if c != nil && c.ID() != "" {
```
oriAdler, line 128: NP
```suggestion
		if cluster != nil && cluster.ID() != "" {
```

pkg/ocm/clusters.go
```diff
@@ -396,6 +396,22 @@ func (c *Client) HasAClusterUsingOidcConfig(bucketName string) (bool, error) {
 	return false, nil
 }
 
+func (c *Client) HasAClusterUsingOidcProvider(oidcEndpointURL string) (bool, error) {
```
oriAdler, line 399: Maybe it is worth returning the cluster's ID?
So we can display the ID in the error message, to provide more info to the user.

gdbranco: We currently don't have any statistics into how many clusters client's will use in a single oidc. So I thought about leaving as is for now as having an idea of how many pages we'll need to fetch before adding this UX feature

PR #1097, opened 2023-03-03
[SDA-8387] Add possibility to delete operator roles from prefix

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8387
# What
Operator roles created from/for BYO OIDC clusters do not have cluster ID tag, so it is required to provide a different way of deleting them

# Why
Facilitates flow for customers so they don't need to deal directly with aws console/cli

### DIFF
--- a/cmd/dlt/operatorrole/cmd.go
+++ b/cmd/dlt/operatorrole/cmd.go
@@ -34,8 +34,12 @@ import (
 	"github.com/openshift/rosa/pkg/rosa"
 )
 
+const (
+	PrefixFlag = "prefix"
+)
+
 var args struct {
-	clusterKey string
+	prefix string
 }
 
 var Cmd = &cobra.Command{
@@ -51,13 +55,15 @@ var Cmd = &cobra.Command{
 func init() {
 	flags := Cmd.Flags()
 
-	flags.StringVarP(
-		&args.clusterKey,
-		"cluster",
-		"c",
+	flags.StringVar(
+		&args.prefix,
+		PrefixFlag,
 		"",
-		"ID or Name of the cluster (deleted/archived) to delete the operator roles from (required).",
+		"Operator role prefix, this flag needs to be used in case of BYO OIDC",
 	)
+	flags.MarkHidden(PrefixFlag)
+
+	ocm.AddOptionalClusterFlag(Cmd)
 	aws.AddModeFlag(Cmd)
 	confirm.AddFlag(flags)
 }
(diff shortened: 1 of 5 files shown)


cmd/dlt/operatorrole/cmd.go
```diff
@@ -154,43 +118,109 @@ func run(cmd *cobra.Command, argv []string) {
...
+			} else if sub == nil {
+				r.Reporter.Errorf("Failed to get cluster '%s': %v", r.ClusterKey, err)
+				os.Exit(1)
+			}
+		}
 
-	isHypershift := false
-	if c != nil {
-		isHypershift = c.Hypershift().Enabled()
+		if c != nil && c.ID() != "" {
```
oriAdler, line 156: NP
```suggestion
		if cluster != nil && cluster.ID() != "" {
```

cmd/dlt/operatorrole/cmd.go
```diff
@@ -154,43 +118,109 @@ func run(cmd *cobra.Command, argv []string) {
...
+			r.Reporter.Errorf("Cluster '%s' is in '%s' state. Operator roles can be deleted only for the "+
+				"uninstalled clusters", c.ID(), c.State())
+			os.Exit(1)
+		}
+		isHypershift := false
+		if c != nil {
+			isHypershift = c.Hypershift().Enabled()
```
oriAdler, line 163: Why do we enable `Hypershift` if the cluster is not `nil`?

cmd/dlt/operatorrole/cmd.go
```diff
@@ -154,43 +118,109 @@ func run(cmd *cobra.Command, argv []string) {
...
+		if err != nil {
+			r.Reporter.Errorf("Error getting operator credential request from OCM %s", err)
+			os.Exit(1)
+		}
+		foundOperatorRoles, _ = r.AWSClient.GetOperatorRolesFromAccountByPrefix(args.prefix, credRequests)
+		if len(foundOperatorRoles) != 0 {
+			if spin != nil {
+				spin.Stop()
+			}
+			if !confirm.Prompt(true, "You are running delete operation for '%s' prefix."+
```
oriAdler, line 195: For BYO OIDC, do we have to check that the operator roles are not re-used by another cluster?

cmd/dlt/operatorrole/cmd.go
gdbranco: It retrieves the value that the cluster is using, not necessarily enabled it

cmd/dlt/operatorrole/cmd.go
gdbranco: Would be better, however I couldn't come up with a search query for this just yet. The operator roles prefix attribute is not stored in database, so I'd need to rely on the operator_iam_roles and those are jsonb and the search is not working properly. So after talking to @pvasant we decided to go with a warning for now and work on proper checking later

PR #1099, opened 2023-03-05
Update k8s.io/apimachinery to v0.26.2

mjlshen:
This remediates [CVE-2022-41723](https://github.com/advisories/GHSA-vvpx-j8f3-3w6h)

Ref: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.26.md#dependencies


go.mod
```diff
@@ -74,14 +71,14 @@ require (
...
-	golang.org/x/term v0.3.0 // indirect
-	golang.org/x/text v0.5.0 // indirect
+	golang.org/x/net v0.7.0 // indirect
+	golang.org/x/sys v0.5.0 // indirect
+	golang.org/x/term v0.5.0 // indirect
+	golang.org/x/text v0.7.0 // indirect
 	google.golang.org/protobuf v1.28.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
-	k8s.io/apimachinery v0.26.1
```
mjlshen, line 84: This was previously an explicit requirement, but was somehow placed down here with the other indirect dependencies.

go.mod
```diff
@@ -26,12 +26,7 @@ require (
 	github.com/zgalor/weberr v0.6.0
 	gitlab.com/c0b/go-ordered-json v0.0.0-20171130231205-49bbdab258c2
 	gopkg.in/square/go-jose.v2 v2.6.0
-)
-
-require (
-	github.com/go-logr/logr v1.2.3 // indirect
```
mjlshen, line 32: These three aren't really being deleted, but combined with the other indirect dependencies below.

`k8s.io/apimachinery` is being upgraded as an explicit dependency that was previously misplaced below.

vkareh:
/lgtm
Builds and works well

PR #1102, opened 2023-03-07
SDA-8224 Deprecate MultiAZ validations for hosted clusters

lucasponce:
Context, hosted cluster will deprecate the existing "multi-az" parameters and it will allow to create hosted clusters with:
- Public hosted clusters: at least one public subnet per VPC and at least one private subnet per VPC.
- Private hosted clusters: at least one private subnet per VPC.

When more than one private subnet is provided, the replicas will be distributed across machinepools, i.e. 4 replicas on 2 private subnets will mean creation of 2 machine pools with 2 replicas.

Also, classic format should be accepted not to break backward compatibility:

```
# 1 public subnet, 2 private subnets; public hosted cluster:
rosa create cluster --cluster-name lponce-local-06 --hosted-cp --machine-cidr 10.0.0.0/16 --subnet-ids subnet-0cbc29215162f7548,subnet-0d65e2465258986db,subnet-0310ceb7c0094e92a --mode auto --channel-group candidate --version 4.12.6 --region us-west-2 --replicas 4
```

```
# 1 private subnet; private hosted cluster:
rosa create cluster --cluster-name lponce-local-07 --hosted-cp --machine-cidr 10.0.0.0/16 --subnet-ids subnet-0d65e2465258986db --mode auto --channel-group candidate --version 4.12.6 --region us-west-2 --replicas 4
```

This PR is still in draft as it has a strong dependency with backend MR:
https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/5395

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -1265,7 +1265,7 @@ func run(cmd *cobra.Command, _ []string) {
 
 	// Multi-AZ:
 	multiAZ := args.multiAZ
-	if interactive.Enabled() {
+	if interactive.Enabled() && !isHostedCP {
 		multiAZ, err = interactive.GetBool(interactive.Input{
 			Question: "Multiple availability zones",
 			Help:     cmd.Flags().Lookup("multi-az").Usage,
@@ -1274,6 +1274,16 @@ func run(cmd *cobra.Command, _ []string) {
 		if err != nil {
 			r.Reporter.Errorf("Expected a valid multi-AZ value: %s", err)
 			os.Exit(1)
+		}
+	}
+
+	// Hosted clusters will be multiAZ by definition
+	if isHostedCP {
+		multiAZ = true
+		if cmd.Flags().Changed("multi-az") {
+			r.Reporter.Warnf("Hosted clusters deprecate the --multi-az flag. " +
+				"The hosted control plane will be MultiAZ, machinepools will be created in the different private " +
+				"subnets provided under --subnet-ids flag.")
 		}
 	}
 
@@ -1450,6 +1460,9 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	// For hosted cluster we will need the number of the private subnets the users has selected
+	privateSubnetsCount := 0
+
 	var availabilityZones []string
 	if useExistingVPC || subnetsProvided {
 		subnets, err := awsClient.GetSubnetIDs()
@@ -1516,9 +1529,17 @@ func run(cmd *cobra.Command, _ []string) {
 				subnetIDs[i] = aws.ParseSubnet(subnet)
 			}
 		}
+
 		// Validate subnets in the case the user has provided them using the `args.subnets`
 		if subnetsProvided {
-			err = ocm.ValidateSubnetsCount(multiAZ, privateLink, len(subnetIDs))
+			if isHostedCP {
+				// Hosted cluster should validate that
+				// - Public hosted clusters have at least one public subnet
+				// - Private hosted clusters have all subnets private
+				privateSubnetsCount, err = ocm.ValidateHostedClusterSubnets(awsClient, privateLink, subnetIDs)
+			} else {
+				err = ocm.ValidateSubnetsCount(multiAZ, privateLink, len(subnetIDs))
+			}
 			if err != nil {
 				r.Reporter.Errorf("%s", err)
 				os.Exit(1)
@@ -1698,7 +1719,7 @@ func run(cmd *cobra.Command, _ []string) {
 				Default:  minReplicas,
 				Required: true,
 				Validators: []interactive.Validator{
-					minReplicaValidator(multiAZ),
+					minReplicaValidator(multiAZ, isHostedCP, privateSubnetsCount),
 				},
 			})
 			if err != nil {
@@ -1706,7 +1727,7 @@ func run(cmd *cobra.Command, _ []string) {
 				os.Exit(1)
 			}
 		}
-		err = minReplicaValidator(multiAZ)(minReplicas)
+		err = minReplicaValidator(multiAZ, isHostedCP, privateSubnetsCount)(minReplicas)
 		if err != nil {
 			r.Reporter.Errorf("%s", err)
 			os.Exit(1)
@@ -1719,7 +1740,7 @@ func run(cmd *cobra.Command, _ []string) {
 				Default:  maxReplicas,
 				Required: true,
 				Validators: []interactive.Validator{
-					maxReplicaValidator(multiAZ, minReplicas),
+					maxReplicaValidator(multiAZ, minReplicas, isHostedCP, privateSubnetsCount),
 				},
 			})
 			if err != nil {
@@ -1727,7 +1748,7 @@ func run(cmd *cobra.Command, _ []string) {
 				os.Exit(1)
 			}
 		}
-		err = maxReplicaValidator(multiAZ, minReplicas)(maxReplicas)
+		err = maxReplicaValidator(multiAZ, minReplicas, isHostedCP, privateSubnetsCount)(maxReplicas)
 		if err != nil {
 			r.Reporter.Errorf("%s", err)
 			os.Exit(1)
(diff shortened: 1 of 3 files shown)


pkg/ocm/helpers.go
```diff
@@ -636,6 +638,49 @@ func ValidateSubnetsCount(multiAZ bool, privateLink bool, subnetsInputCount int)
...
+	if privateSubnetsErr != nil {
+		return 0, privateSubnetsErr
+	}
+
+	privateSubnetCount := len(privateSubnets)
+	publicSubnetsCount := len(subnets) - privateSubnetCount
+
+	if isPrivate {
+		if publicSubnetsCount > 0 {
+			return 0, fmt.Errorf("The number of public subnets for a private hosted cluster should be zero")
```
cben, line 673: Would be nice to list the offending subnet(s).  
But to implement that you'd probably want to change `FilterVPCsPrivateSubnets` to partition into 2 arrays.

cmd/create/cluster/cmd.go
```diff
@@ -2345,12 +2366,20 @@ func handleByoOidcOptions(r *rosa.Runtime, cmd *cobra.Command, isSTS bool) (bool
...
+func minReplicaValidator(multiAZ bool, isHostedCP bool, privateSubnetsCount int) interactive.Validator {
 	return func(val interface{}) error {
 		minReplicas, err := strconv.Atoi(fmt.Sprintf("%v", val))
 		if err != nil {
 			return err
 		}
+		if isHostedCP {
+			if minReplicas%privateSubnetsCount != 0 {
+				return fmt.Errorf("Hosted clusters require that the number of compute nodes be a multiple of "+
+					"the number of private subnets %d, instead received: %d", privateSubnetsCount, minReplicas)
```
cben, line 2378: Q: is `privateSubnetsCount` used here as a proxy for the number of AZs, assuming one private subnet per AZ?  
Or can HostedCP clusters have multiple private subnets in same AZ, and you meant this literally?

pkg/ocm/helpers.go
```diff
@@ -636,6 +638,49 @@ func ValidateSubnetsCount(multiAZ bool, privateLink bool, subnetsInputCount int)
...
+			}
+		}
+	}
+
+	privateSubnets, privateSubnetsErr := awsClient.FilterVPCsPrivateSubnets(subnets)
+	if privateSubnetsErr != nil {
+		return 0, privateSubnetsErr
+	}
+
+	privateSubnetCount := len(privateSubnets)
```
cben, line 668: consistency nit: singular "Subnet" here but plural in `privateSubnetsCount` as same value is called outside, as well as in `publicSubnetsCount`.

pkg/ocm/helpers.go
```diff
@@ -636,6 +638,49 @@ func ValidateSubnetsCount(multiAZ bool, privateLink bool, subnetsInputCount int)
 	return nil
 }
 
+func ValidateHostedClusterSubnets(awsClient aws.Client, isPrivate bool, subnetIDs []string) (int, error) {
+	if isPrivate && len(subnetIDs) < 1 {
+		return 0, fmt.Errorf("The number of subnets for a private hosted cluster should be at least one")
+	}
+	if !isPrivate && len(subnetIDs) < 2 {
+		return 0, fmt.Errorf("The number of subnets for a public hosted cluster should be at least two")
```
cben, line 646: Q: Is this because it needs at least one private + at least one public?  
However, I don't see privateSubnetCount >= 1 enforced anywhere.

These quick len() checks avoid the need for API calls, but perhaps the returned error is less informative than user could get if you proceeded anyway? :thinking:

cmd/create/cluster/cmd.go
```diff
@@ -2345,12 +2366,20 @@ func handleByoOidcOptions(r *rosa.Runtime, cmd *cobra.Command, isSTS bool) (bool
...
 
-func minReplicaValidator(multiAZ bool) interactive.Validator {
+func minReplicaValidator(multiAZ bool, isHostedCP bool, privateSubnetsCount int) interactive.Validator {
 	return func(val interface{}) error {
 		minReplicas, err := strconv.Atoi(fmt.Sprintf("%v", val))
 		if err != nil {
 			return err
 		}
+		if isHostedCP {
+			if minReplicas%privateSubnetsCount != 0 {
```
cben, line 2376: Does this risk a panic if privateSubnetsCount == 0?

cben:
I wonder if the user experience / error messages could be clearer if user were to provide separate `--private-subnet-ids` / `--public-subnet-ids` lists :grey_question:

lucasponce:
cc @andreadecorte @gdbranco this is the counter part of the backend work

cmd/create/cluster/cmd.go
```diff
@@ -1277,6 +1277,16 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
+	// Hosted clusters will be multiAZ by definition
+	if isHostedCP {
+		multiAZ = true
+		if cmd.Flags().Changed("multi-az") {
```
oriAdler, line 1283: Should we validate that the user didn't set the flag `multi-az` to **false**?

cmd/create/cluster/cmd.go
```diff
@@ -2362,12 +2383,20 @@ func handleByoOidcOptions(r *rosa.Runtime, cmd *cobra.Command, isSTS bool) (bool
...
 
-func minReplicaValidator(multiAZ bool) interactive.Validator {
+func minReplicaValidator(multiAZ bool, isHostedCP bool, privateSubnetsCount int) interactive.Validator {
 	return func(val interface{}) error {
 		minReplicas, err := strconv.Atoi(fmt.Sprintf("%v", val))
 		if err != nil {
 			return err
 		}
+		if isHostedCP {
+			if minReplicas%privateSubnetsCount != 0 {
```
oriAdler, line 2393: Do we need to check that `minReplica` is a positive number?

pkg/ocm/helpers.go
```diff
@@ -636,6 +638,49 @@ func ValidateSubnetsCount(multiAZ bool, privateLink bool, subnetsInputCount int)
...
+	}
+	vpcSubnets, vpcSubnetsErr := awsClient.GetVPCSubnets(subnetIDs[0])
```
oriAdler, line 648: We get all the VPC's subnets and then filtering them to find the subnets provided by the user.

I'm wondering if we can fetch from the AWS client only the subnets with the provided IDs in the first place.
`c.ec2Client.DescribeSubnets(&ec2.DescribeSubnetsInput{SubnetIds: []*string{<subnet1>,<subnet2>...}})`

Also, it's better not to expose `[]*ec2.Subnet` AWS SDK objects outside of the client.
Can we encapsulate this functionality in the client? So we pass a list of subnets to the AWS client, and then, get a private and public count. WDYT?

I hope it doesn't require too many changes under tight timelines...

cmd/create/cluster/cmd.go
lucasponce: It's deprecated in both cases, so, no matter what they set, internally an hypershift cluster is considered multiAZ for the control plane.
The number of private subnets will dynamically indicate it the dataplane is singleAZ or multiAZ.

the `cmd.Flags().Changed("multi-az")` should be `true` in the presence of the parameter with any value, right?

oriAdler: Yes, makes sense.

cmd/create/cluster/cmd.go
lucasponce: Added an extra guard for it; "privateSubnetsCount" is validated in a previous step, and that case should fail before entering this method, but it's a fair check to add it there too.

cmd/create/cluster/cmd.go
lucasponce: Yeah, I thought it was validated before, but not, added a condition there.
Thanks

pkg/ocm/helpers.go
lucasponce: It's a fair request; let me evaluate the effort of this.

oriAdler: In other cases in the code we do sometimes need all the subnets from a VPC, for example when provisioning a day-2 machine pool and choosing a subnet from the VPC.

I didn't mean to refactor `awsClient.GetVPCSubnets`, as it's needed.

But to create a new function for this specific case, to count private/public with the provided subnets.

oriAdler:
LGTM

lucasponce:
> I wonder if the user experience / error messages could be clearer if user were to provide separate --private-subnet-ids / --public-subnet-ids lists grey_question

I think that would impact also the existing experience for classic clusters.

The goal of this work was focused on deprecating the "--multi-az" flag for hosted clusters keeping the same UX as possible.

lucasponce:
This one has been held for some time to the reviewers: @oriAdler @cben. Do you see anything else pending for this work?

PR #1104, opened 2023-03-07
aws: Default to region from profile

vkareh:


jharrington22 (commented):
This will effect more than just the `whoami` command since a bunch of other cmds use it. Thinking about it changing existing behavior that is all but otherwise looks good.

PR #1105, opened 2023-03-09
[SDA-8451] Delete oidc-provider cmd should not check flags when is progmatically called

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8451
# Why
Fixes incorrect check when progmatically called


cmd/dlt/oidcprovider/cmd.go
```diff
@@ -71,12 +71,12 @@ func run(cmd *cobra.Command, argv []string) {
 	r := rosa.NewRuntime().WithAWS().WithOCM()
 	defer r.Cleanup()
 
-	skipInteractive := false
+	isProgmaticallyCalled := false
```
hunterkepley, line 74: Should this be `isProgrammaticallyCalled`?

PR #1109, opened 2023-03-09
Create Hypershift dedicated account roles

oriAdler:
1. The default flow creates two sets of roles: 3 for Hypershift and 4 for Hive.
2. Passing the `classic` flag will create only 4 roles for Hive.
3. Passing the `hosted-cp` flag will create only 3 roles for Hypershift.

Related: [[SDA-8665](https://issues.redhat.com/browse/SDA-8665)]

Default flow creates double sets:
image

Passing the `hosted-cp` flag creates one set:
image

### DIFF
--- a/cmd/create/accountroles/creators.go
+++ b/cmd/create/accountroles/creators.go
@@ -13,21 +13,31 @@ import (
 type creator interface {
 	createRoles(*rosa.Runtime, *accountRolesCreationInput) error
 	getRoleTags(string, *accountRolesCreationInput) map[string]string
-	buildCommands(*accountRolesCreationInput) (string, error)
+	printCommands(*rosa.Runtime, *accountRolesCreationInput) error
 }
 
-func initCreator(managedPolicies bool, hostedCP bool) creator {
-	// Hypershift policies are managed policies by default
+func initCreator(managedPolicies bool, classic bool, hostedCP bool, isClassicValueSet bool,
+	isHostedCPValueSet bool) (creator, bool) {
+	// Classic ROSA managed policies
+	if managedPolicies && !hostedCP {
+		return &managedPoliciesCreator{}, true
+	}
+
+	// If the user didn't select topologies (default flow creates both), or selected both topologies
+	if !isClassicValueSet && !isHostedCPValueSet || hostedCP && classic {
+		return &doubleRolesCreator{}, true
+	}
+
 	if hostedCP {
-		return &hcpManagedPoliciesCreator{}
+		return &hcpManagedPoliciesCreator{}, true
 	}
 
-	if managedPolicies {
-		return &managedPoliciesCreator{}
+	// Classic ROSA unmanaged policies
+	if classic {
+		return &unmanagedPoliciesCreator{}, true
 	}
 
-	// Default flow creates a set of roles with unmanaged policies
-	return &unmanagedPoliciesCreator{}
+	return nil, false
 }
 
 type accountRolesCreationInput struct {
@@ -57,6 +67,8 @@ func buildRolesCreationInput(prefix, permissionsBoundary, accountID, env string,
 type managedPoliciesCreator struct{}
 
 func (mp *managedPoliciesCreator) createRoles(r *rosa.Runtime, input *accountRolesCreationInput) error {
+	r.Reporter.Infof("Creating classic account roles using '%s'", r.Creator.ARN)
+
 	for file, role := range aws.AccountRoles {
 		accRoleName := aws.GetRoleName(input.prefix, role.Name)
 		assumeRolePolicy := getAssumeRolePolicy(file, input)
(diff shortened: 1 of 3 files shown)


pkg/aws/policies.go
```diff
@@ -74,8 +74,13 @@ const (
 	ControlPlaneAccountRole = "instance_controlplane"
 	WorkerAccountRole       = "instance_worker"
 	SupportAccountRole      = "support"
-	OCMRole                 = "OCM"
-	OCMUserRole             = "User"
+
+	HCPInstallerRole = "installer"
```
oriAdler, line 78: The keys are the same since both Hive roles and Hypershift roles are going to use the same trust policies.
`filename := fmt.Sprintf("sts_%s_trust_policy", file)`

gdbranco: Was going to ask about that, won't they be different at some point? Will there be a refactor later or would you rather wait for the correct policies?

cmd/create/accountroles/creators.go
```diff
@@ -139,35 +148,59 @@ func (up *unmanagedPoliciesCreator) createRoles(r *rosa.Runtime, input *accountR
...
+		accRoleName := aws.GetRoleName(input.prefix, role.Name)
+		if !confirm.Prompt(true, "Create the '%s' role?", accRoleName) {
+			continue
 		}
 
-		r.Reporter.Debugf("Attaching permission policy to role '%s'", filename)
-		err = r.AWSClient.AttachRolePolicy(accRoleName, policyARN)
+		assumeRolePolicy := getAssumeRolePolicy(file, input)
+		tagsList := hcp.getRoleTags(file, input)
+		filename := fmt.Sprintf("sts_hcp_%s_permission_policy", file)
```
oriAdler, line 201: The `_hcp` infix differentiates the classic rosa permission policies from the Hypershift permission policies.
SRE is updating the file names here https://github.com/openshift/managed-cluster-config/pull/1525 to match this convention.

pkg/aws/policies.go
gdbranco: Oh wait, this is for the trusted. This is true

cmd/create/accountroles/creators.go
gdbranco: This was what I was going to ask about, so you are set already but there's still work to do somewhere else. Did you download everything and renamed to test then?

oriAdler: Yes, I created the files locally to test the changes.
Changed the PR to draft until changes in managed cluster config are done, thanks.

oriAdler:
@andreadecorte @lucasponce PTAL

vkareh:
I'm not sure I like the `--only-foo` flag. Can we standardize on defaulting to whatever the default cluster type is? Effectively matching flags between roles and clusters.

For example, if running `rosa create cluster` (no flags) creates a classic STS cluster, then `rosa create account-roles` should create roles for classic STS clusters.

If running `rosa create cluster --hosted-cp` creates a HyperShift cluster, then `rosa create account-roles --hosted-cp` should create roles for HyperShift clusters. Etc.

The day we flip the default cluster type, then we also flip the default role type.

Any other model just causes confusion.

oriAdler:
> I'm not sure I like the `--only-foo` flag. Can we standardize on defaulting to whatever the default cluster type is? Effectively matching flags between roles and clusters.
> 
> For example, if running `rosa create cluster` (no flags) creates a classic STS cluster, then `rosa create account-roles` should create roles for classic STS clusters.
> 
> If running `rosa create cluster --hosted-cp` creates a HyperShift cluster, then `rosa create account-roles --hosted-cp` should create roles for HyperShift clusters. Etc.
> 
> The day we flip the default cluster type, then we also flip the default role type.
> 
> Any other model just causes confusion.

I agree with you 100%

If we can get BU confirmation on that, I can easily make the changes.

vkareh:
@arendej @wgordon17 do you agree on the above? That we should keep the flags between the `roles` and `clusters` commands consistent? I think it will help decrease user confusion and means that the state of their IAM resources matches their cluster creation pattern.

oriAdler:
@vkareh @zgalor @gdbranco though we can't merge this one yet, could you please re-review it?
Maybe I can ask QE to test this branch in the meantime, so once we have the managed policies it can be merged fast.

gdbranco:
lgtm

ciaranRoche:
@oriAdler what is the status of this, can we merge it?

davidleerh:
lgtm

oriAdler:
@davidleerh thanks for the profound review and checks, I highly appreciate it:)

The **create account roles** command, creates two sets only for non-interactive mode when all flags are provided.
The assumption is, that it will be used for the user's automation.
Otherwise, for interactive mode, we let the user choose the role type, to create one set.

PR #1112, opened 2023-03-09
[SDA-8369] Update versions to include hypershift_enabled and add validation

hunterkepley:
[SDA-8369](https://issues.redhat.com//browse/SDA-8369)

"Display only the versions that are hypershift supported in ocm and rosa cli"

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -1567,7 +1567,7 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 
 		// Validate subnets in the case the user has provided them using the `args.subnets`
-		if subnetsProvided && !isHostedCP {
+		if !isHostedCP {
 			err = ocm.ValidateSubnetsCount(multiAZ, privateLink, len(subnetIDs))
 			if err != nil {
 				r.Reporter.Errorf("%s", err)
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -42,7 +42,6 @@ import (
 	"github.com/openshift/rosa/pkg/fedramp"
 	"github.com/openshift/rosa/pkg/helper"
 	"github.com/openshift/rosa/pkg/helper/roles"
-	"github.com/openshift/rosa/pkg/helper/versions"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/interactive/confirm"
 	"github.com/openshift/rosa/pkg/ocm"
@@ -781,7 +780,7 @@ func run(cmd *cobra.Command, _ []string) {
 			os.Exit(1)
 		}
 	}
-	version, err = versions.ValidateVersion(version, versionList, channelGroup, isSTS, isHostedCP)
+	version, err = r.OCMClient.ValidateVersion(version, versionList, channelGroup, isSTS, isHostedCP)
 	if err != nil {
 		r.Reporter.Errorf("Expected a valid OpenShift version: %s", err)
 		os.Exit(1)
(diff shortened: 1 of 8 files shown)


hunterkepley:
Tested with a hypershift and STS cluster

hunterkepley:
Will un-hold after[https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/5401]( https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/5401) is merged

gdbranco:
Seems alright, why the hold?

oriAdler:
LGTM

pkg/ocm/versions.go
```diff
@@ -331,3 +309,53 @@ func CheckAndParseVersion(availableUpgrades []string, version string) (string, e
...
+		allVersions := strings.Join(versionList, " ")
+		err := fmt.Errorf("A valid version number must be specified\nValid versions: %s", allVersions)
+		return version, err
+	}
+
+	if isSTS && !HasSTSSupport(version, channelGroup) {
+		err := fmt.Errorf("Version '%s' is not supported for STS clusters", version)
+		return version, err
+	}
+	if !HasSTSSupportMinor(version) {
```
oriAdler, line 336: Should we check that only for **STS** clusters?

pkg/ocm/versions.go
```diff
@@ -331,3 +309,53 @@ func CheckAndParseVersion(availableUpgrades []string, version string) (string, e
...
+			Search(filter).
+			Page(1).
+			Size(1).
+			Send()
+		if err != nil {
+			return "", handleErr(response.Error(), err)
+		}
+		valid, err := HasHostedCPSupport(response.Items().Get(0))
+		if err != nil {
+			return "", fmt.Errorf("error while parsing OCP version '%s': %v", version, err)
```
oriAdler, line 353: NP
Maybe the error wrapping should be inside `HasHostedCPSupport`?

pkg/ocm/versions.go
oriAdler: Though I see we didn't check if it's STS previously with that `if` statement, so I'm not sure about this one.

oriAdler (commented):
Left two small comments, other than that LGTM.

cmd/create/cluster/cmd.go
```diff
@@ -1568,15 +1567,20 @@ func run(cmd *cobra.Command, _ []string) {
...
-		if subnetsProvided {
-			if isHostedCP {
-				// Hosted cluster should validate that
-				// - Public hosted clusters have at least one public subnet
-				// - Private hosted clusters have all subnets private
-				privateSubnetsCount, err = ocm.ValidateHostedClusterSubnets(awsClient, privateLink, subnetIDs)
-			} else {
-				err = ocm.ValidateSubnetsCount(multiAZ, privateLink, len(subnetIDs))
+		if subnetsProvided && !isHostedCP {
+			err = ocm.ValidateSubnetsCount(multiAZ, privateLink, len(subnetIDs))
```
gdbranco, line 1571: I know this was already like this, but I feel like it should be validated even if subnets are not provided. Maybe it is done like this to save a call

cmd/create/operatorroles/by_clusterkey.go
```diff
@@ -331,6 +340,9 @@ func buildCommands(r *rosa.Runtime, env string,
 		if managedPolicies {
 			iamTags[tags.ManagedPolicies] = helper.True
 		}
+		if hostedCPPolicies {
```
gdbranco, line 343: I think this is part of Ori's PR instead

cmd/create/cluster/cmd.go
oriAdler: Seems like this is from Lucas PR, we need to rebase this PR against the master branch.

davidleerh:
lgtm

PR #1114, opened 2023-03-14
Hypershift - create cluster

oriAdler:
Omit the control plane role when creating a Hypershift cluster.

**Note:** a follow-up PR will ensure the set of account roles is a Hypershift set.

Related: [SDA-8454](https://issues.redhat.com/browse/SDA-8454)

Only three account roles are required from the user:
image

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -865,8 +865,12 @@ func run(cmd *cobra.Command, _ []string) {
 			}
 			roleARN = roleARNs[0]
 		} else {
-			r.Reporter.Warnf("No account roles found. You will need to manually set them in the " +
-				"next steps or run 'rosa create account-roles' to create them first.")
+			createAccountRolesCommand := "rosa create account-roles"
+			if isHostedCP {
+				createAccountRolesCommand = createAccountRolesCommand + " --hosted-cp"
+			}
+			r.Reporter.Warnf(fmt.Sprintf("No account roles found. You will need to manually set them in the "+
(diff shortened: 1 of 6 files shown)


zgalor:
LGTM

vkareh:
This looks good, but at this point I wonder if we should start thinking 
of a refactor of the `cmd/create/cluster/` files and have a `classic.go` 
and `hypershift.go` with the code flow deltas, instead of adding a bunch 
of `if isHostedCP` conditionals everywhere.

Not saying that's the right path forward, but we should at least explore 
what it would look like to split it into different provisioning 
strategies.

-- 
Victor Kareh ***@***.***>
Principal Software Engineer, Cluster Management
IRC: vkareh | GitHub: vkareh

oriAdler:
> This looks good, but at this point I wonder if we should start thinking of a refactor of the `cmd/create/cluster/` files and have a `classic.go` and `hypershift.go` with the code flow deltas, instead of adding a bunch of `if isHostedCP` conditionals everywhere. Not saying that's the right path forward, but we should at least explore what it would look like to split it into different provisioning strategies.
> […](#)
> -- Victor Kareh ***@***.***> Principal Software Engineer, Cluster Management IRC: vkareh | GitHub: vkareh

I agree we should consider this path, as we assume the two flows are going to diverge.
With that being said, under the constraint of the tight timelines, I think we can postpone the implementation after we have the complete flow for Hypershift account roles: `upgrade cluster`, `create operator roles`, `upgrade roles`, etc.

oriAdler:
@zgalor @vkareh added more validations to check the managed policies are attached to the roles.
Can you re-PTAL?

PR #1115, opened 2023-03-14
Create Hypershift account roles with managed policies

oriAdler:
Run `rosa create account-roles --hosted-cp` to get a set of Hypershift roles.

**Note:** The default flow for Hypershift is going to be managed policies.

Related: [SDA-8494](https://issues.redhat.com/browse/SDA-8498)

Auto mode:
image

Manual mode:
image

### DIFF
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -41,6 +41,7 @@ var args struct {
 	channelGroup        string
 	managed             bool
 	forcePolicyCreation bool
+	hostedCP            bool
 }
 
 var Cmd = &cobra.Command{
@@ -118,6 +119,13 @@ func init() {
 		"Forces creation of policies skipping compatibility check",
 	)
 
+	flags.BoolVar(
+		&args.hostedCP,
+		"hosted-cp",
+		false,
+		"Enable the use of hosted control planes (HyperShift)",
+	)
+
 	aws.AddModeFlag(Cmd)
 
 	confirm.AddFlag(flags)
@@ -160,6 +168,12 @@ func run(cmd *cobra.Command, argv []string) {
 
 	if args.forcePolicyCreation && managedPolicies {
 		r.Reporter.Warnf("Forcing creation of policies only works for unmanaged policies")
+		os.Exit(1)
+	}
+
+	// Hosted cluster roles always use managed policies
+	if cmd.Flags().Changed("hosted-cp") && cmd.Flags().Changed("managed-policies") && !args.managed {
+		r.Reporter.Errorf("Setting `hosted-cp` as unmanaged policies is not supported")
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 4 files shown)


cmd/create/accountroles/creators.go
```diff
@@ -17,11 +17,16 @@ type creator interface {
 	buildCommands(*accountRolesCreationInput) (string, error)
 }
 
-func initCreator(managedPolicies bool) creator {
+func initCreator(managedPolicies bool, isHypershift bool) creator {
```
vkareh, line 20: The variable renaming is confusing: `--hosted-cp` stored in `args.hostedClusterEnabled` and read as `isHypershift bool`.

I suggest keeping it consistent:
`--hosted-cp` -> `args.hostedCP` -> `hostedCP bool`

cmd/create/accountroles/creators.go
```diff
@@ -217,6 +222,71 @@ func getAssumeRolePolicy(file string, input *accountRolesCreationInput) string {
 	})
 }
 
+type hcpManagedPoliciesCreator struct{}
+
+func (hcp *hcpManagedPoliciesCreator) createRoles(r *rosa.Runtime, input *accountRolesCreationInput) error {
+	for file, role := range aws.HCPAccountRoles {
+		accRoleName := aws.GetRoleName(input.prefix, role.Name)
+		if !confirm.Prompt(true, "Create the '%s' role?", accRoleName) {
```
vkareh, line 230: I wonder if we're ready to remove this. It's legacy code from when roles were still a new thing in ROSA.

Since the role creation is idempotent, we should probably just confirm `Create the roles? [Y/n]` once and be done with it.

vkareh: or to be honest, if you run `rosa create account-roles`, since it's not a destructive operation we probably shouldn't even confirm.

We don't say `rosa create cluster: are you sure?`. Like, yes, that's what I just typed, thank you. :laughing:

cmd/create/accountroles/cmd.go
```diff
@@ -163,6 +171,12 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
+	// Hosted cluster roles always use managed policies
+	if cmd.Flags().Changed("hosted-cp") && cmd.Flags().Changed("managed-policies") {
```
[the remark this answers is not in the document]
oriAdler: That makes sense, changed it, thanks.

cmd/create/accountroles/creators.go
oriAdler: I agree, changed.

cmd/create/accountroles/creators.go
oriAdler: Refactored.

gdbranco: We do the same for each on operator roles, might be interesting to refactor that when we get there as well

oriAdler: I agree, will refactor it when working on `create operator roles` for this epic.

cmd/create/accountroles/creators.go
vkareh: /nitpick
I think we still need to change this variable from `isHypershift` to `hostedCP`

oriAdler: Right, changed, thanks.

PR #1120, opened 2023-03-20
Create Hypershift operator roles

oriAdler:
If the account roles have Hypershift managed policies, attach the managed policies to the operator roles instead of creating policies in the user account.

Related: [SDA-8507](https://issues.redhat.com/browse/SDA-8507)

image

### DIFF
--- a/cmd/create/operatorroles/by_clusterkey.go
+++ b/cmd/create/operatorroles/by_clusterkey.go
@@ -10,7 +10,6 @@ import (
 	awscb "github.com/openshift/rosa/pkg/aws/commandbuilder"
 	"github.com/openshift/rosa/pkg/aws/tags"
 	"github.com/openshift/rosa/pkg/helper"
-	"github.com/openshift/rosa/pkg/interactive/confirm"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/output"
 	"github.com/openshift/rosa/pkg/rosa"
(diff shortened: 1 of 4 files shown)


oriAdler:
@gdbranco @zgalor @hunterkepley PTAL

cmd/create/operatorroles/by_clusterkey.go
```diff
@@ -331,6 +340,9 @@ func buildCommands(r *rosa.Runtime, env string,
 		if managedPolicies {
 			iamTags[tags.ManagedPolicies] = helper.True
 		}
+		if hostedCPPolicies {
+			iamTags[tags.HypershiftPolicies] = helper.True
```
gdbranco, line 344: Just noticed this now, this should be HcpPolicies or similar, as this tag is something the user can be exposed to.

cmd/create/operatorroles/by_clusterkey.go
```diff
@@ -94,13 +93,23 @@ func handleOperatorRoleCreationByClusterKey(r *rosa.Runtime, env string,
 	if err != nil {
 		r.Reporter.Errorf("%s", err)
 	}
+
+	var hostedCPPolicies bool
+	if cluster.Hypershift().Enabled() {
+		hostedCPPolicies, err = r.AWSClient.HasHostedCPPolicies(cluster.AWS().STS().RoleARN())
```
gdbranco, line 99: If we are taking it from this bool we can probably remove the --hosted-cp param from create operator roles

cmd/create/operatorroles/by_clusterkey.go
oriAdler: Are you referring to the tag's name in AWS?
I can change it, need to check with BU what is the preferred name.
Thanks

cmd/create/operatorroles/by_prefix.go
```diff
@@ -143,17 +142,25 @@ func handleOperatorRoleCreationByPrefix(r *rosa.Runtime, env string,
 		os.Exit(1)
 	}
 
+	var hostedCPPolicies bool
+	if args.hostedCp {
+		hostedCPPolicies, err = r.AWSClient.HasHostedCPPolicies(args.installerRoleArn)
+		if err != nil {
+			r.Reporter.Errorf("Failed to determine if cluster has hosted CP policies: %v", err)
```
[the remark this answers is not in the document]
oriAdler: Do you mean to check if the installer role is not empty?
We are checking it before calling the function.

cmd/create/operatorroles/by_clusterkey.go
gdbranco: Yes the tag name in AWS, just making sure whatever the customer's can see are all hcp instead of hs

cmd/create/operatorroles/by_clusterkey.go
oriAdler: If I'm not wrong we need this param for the flow when we create the operator roles before the cluster creation.

cmd/create/operatorroles/by_prefix.go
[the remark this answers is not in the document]
oriAdler: Fixed

cmd/create/operatorroles/by_clusterkey.go
gdbranco: But if the client is using hypershift account roles then we are sure to create hypershift operator roles, right?

I think we can just take whatever is from the installer role

oriAdler: We are currently using the param to fetch the operator roles for Hypershift https://github.com/openshift/rosa/blob/master/cmd/create/operatorroles/by_prefix.go#L118.

Since the Hypershift roles with managed policies are a hidden flow now, we still need this param, existing Hypershfit account roles don't have the Hypershift tag.

We can change it once we have the managed policies in place, and if we enforce the user to use Hypershift policies for Hypershift clusters.

PR #1121, opened 2023-03-21
Delete account-roles - handle Hypershift roles

oriAdler:
1. Default flow deletes classic ROSA roles.
2. If "--hosted-cp" is specified, delete the Hypershift roles (same behavior as `create account roles`).

**TODO:** we can delete both by default once we have the managed policies in place. 
Otherwise, deleting the Hypershift roles for users is meaningless.

Related: [SDA-8565](https://issues.redhat.com/browse/SDA-8565)

By default delete classic ROSA account roles:
image

When the **--hosted-cp** flag is provided, delete the Hypershift account roles:
image

### DIFF
--- a/cmd/dlt/accountroles/cmd.go
+++ b/cmd/dlt/accountroles/cmd.go
@@ -33,7 +33,8 @@ import (
 )
 
 var args struct {
-	prefix string
+	prefix   string
+	hostedCP bool
 }
 
 var Cmd = &cobra.Command{
@@ -56,6 +57,14 @@ func init() {
 		"",
 		"Prefix of the account roles to be deleted.",
 	)
+
+	flags.BoolVar(
+		&args.hostedCP,
+		"hosted-cp",
+		false,
+		"Delete hosted control planes roles (HyperShift)",
+	)
+	flags.MarkHidden("hosted-cp")
 
 	aws.AddModeFlag(Cmd)
 	confirm.AddFlag(flags)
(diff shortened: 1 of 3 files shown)


cmd/dlt/accountroles/cmd.go
```diff
@@ -190,13 +179,12 @@ func run(cmd *cobra.Command, _ []string) {
 				continue
 			}
 		}
-		r.Reporter.Infof("Successfully deleted the account roles")
+		r.Reporter.Infof(fmt.Sprintf("Successfully deleted the %saccount roles", roleTypeString))
```
gdbranco, line 182:
```suggestion
		r.Reporter.Infof(fmt.Sprintf("Successfully deleted the %s account roles", roleTypeString))
```

cmd/dlt/accountroles/cmd.go
```diff
@@ -177,8 +136,38 @@ func run(cmd *cobra.Command, _ []string) {
...
+	}
+}
+
+func deleteAccountRoles(r *rosa.Runtime, env string, prefix string, clusters []*cmv1.Cluster, mode string,
+	hostedCP bool) error {
+	var accountRolesMap map[string]aws.AccountRole
+	var roleTypeString string
+	if hostedCP {
+		accountRolesMap = aws.HCPAccountRoles
+		roleTypeString = "hosted CP "
```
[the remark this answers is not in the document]
oriAdler: Yes, the alternative is to have the space in the messages where `roleTypeString` is used.
If `roleTypeString == ""`, we'll have two spaces in the error message.

cmd/dlt/accountroles/cmd.go
oriAdler: It is for a couple of messages. 
I'm also not so happy with that, but the alternative was a duplicate code by checking `if hostedCP` before each error message and having two different error messages.

If we can find a better approach, I'll be happy to adapt it.

PR #1123, opened 2023-03-22
Fix hostedcp multiaz subnets validation on interactive mode

lucasponce:
cc @oriAdler 

For hosted clusters, the subnet validation is relaxed compared with classic clusters.
There was a bug in the interactive mode not updating the current logic.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -1555,7 +1555,7 @@ func run(cmd *cobra.Command, _ []string) {
 				Options:  options,
 				Default:  defaultOptions,
 				Validators: []interactive.Validator{
-					interactive.SubnetsCountValidator(multiAZ, privateLink),
+					interactive.SubnetsCountValidator(multiAZ, privateLink, isHostedCP),
 				},
 			})
 			if err != nil {
@@ -1568,15 +1568,20 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 
 		// Validate subnets in the case the user has provided them using the `args.subnets`
-		if subnetsProvided {
-			if isHostedCP {
-				// Hosted cluster should validate that
-				// - Public hosted clusters have at least one public subnet
-				// - Private hosted clusters have all subnets private
-				privateSubnetsCount, err = ocm.ValidateHostedClusterSubnets(awsClient, privateLink, subnetIDs)
-			} else {
-				err = ocm.ValidateSubnetsCount(multiAZ, privateLink, len(subnetIDs))
+		if subnetsProvided && !isHostedCP {
+			err = ocm.ValidateSubnetsCount(multiAZ, privateLink, len(subnetIDs))
+			if err != nil {
+				r.Reporter.Errorf("%s", err)
+				os.Exit(1)
 			}
+		}
+
+		// In both cases (interactive and subnetsProvided) hosted clusters need to validate the private subnets
+		if isHostedCP {
+			// Hosted cluster should validate that
+			// - Public hosted clusters have at least one public subnet
+			// - Private hosted clusters have all subnets private
+			privateSubnetsCount, err = ocm.ValidateHostedClusterSubnets(awsClient, privateLink, subnetIDs)
 			if err != nil {
 				r.Reporter.Errorf("%s", err)
 				os.Exit(1)
--- a/pkg/interactive/validation.go
+++ b/pkg/interactive/validation.go
@@ -139,9 +139,18 @@ func RegExpBoolean(restr string) Validator {
 
 // SubnetsCountValidator get a slice of `[]core.OptionAnswer` as an interface.
 // e.g. core.OptionAnswer { Value: subnet-04f67939f44a97dbe (us-west-2b), Index: 0 }
-func SubnetsCountValidator(multiAZ bool, privateLink bool) Validator {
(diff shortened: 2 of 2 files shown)


pkg/interactive/validation.go
```diff
@@ -139,9 +139,13 @@ func RegExpBoolean(restr string) Validator {
 
 // SubnetsCountValidator get a slice of `[]core.OptionAnswer` as an interface.
 // e.g. core.OptionAnswer { Value: subnet-04f67939f44a97dbe (us-west-2b), Index: 0 }
-func SubnetsCountValidator(multiAZ bool, privateLink bool) Validator {
+func SubnetsCountValidator(multiAZ bool, privateLink bool, hostedCP bool) Validator {
 	return func(input interface{}) error {
 		if answers, ok := input.([]core.OptionAnswer); ok {
+			if hostedCP {
```
oriAdler, line 145: What is the minimal number of subnets for a Hypershift cluster?

cmd/create/cluster/cmd.go
```diff
@@ -1569,14 +1569,19 @@ func run(cmd *cobra.Command, _ []string) {
...
 		// Validate subnets in the case the user has provided them using the `args.subnets`
 		if subnetsProvided {
-			if isHostedCP {
-				// Hosted cluster should validate that
-				// - Public hosted clusters have at least one public subnet
-				// - Private hosted clusters have all subnets private
-				privateSubnetsCount, err = ocm.ValidateHostedClusterSubnets(awsClient, privateLink, subnetIDs)
-			} else {
-				err = ocm.ValidateSubnetsCount(multiAZ, privateLink, len(subnetIDs))
+			err = ocm.ValidateSubnetsCount(multiAZ, privateLink, len(subnetIDs))
```
oriAdler, line 1572: Does this check applies also for Hypershift clusters?

pkg/interactive/validation.go
lucasponce: https://github.com/openshift/rosa/blob/master/pkg/ocm/helpers.go#L641

cmd/create/cluster/cmd.go
lucasponce: it's included in https://github.com/openshift/rosa/blob/master/pkg/ocm/helpers.go#L641

oriAdler: Just asking because previously we didn't run this check for Hypershift clusters:
image
Let me know if I missed anything.

lucasponce: In both interactive/non-interactive that validation is perform at this point:

https://github.com/openshift/rosa/pull/1123/files#diff-943c0ab57b82c7028652df55e1d49e4f4e4dc41fa094dc4a9b3f2c35887d44b5R1580

pkg/interactive/validation.go
lucasponce: The minimal number is 1 in private and 2 in public (singleaz/multiaz doesn't matter for hosted clusters).

We can add that validation there, but after out of this block we need to query AWS to check public/private subnets, and the private number is collected for further replicas validation.

So, +1 of adding the basic number in that block, but I didn't see fit to query AWS from it.

cmd/create/cluster/cmd.go
lucasponce: No. Line 1572 is when subnets are provided for classic clusters.

For interactive and subnets provider, the hosted cluster needs to evaluate the "ValidateHostedClusterSubnets" that's what's is out of that conditional block.

pkg/interactive/validation.go
oriAdler: I agree, just the minimal validation, to avoid this:
image
So we can give the user the opportunity to select more/fewer subnets instead of starting from the beginning again.

cmd/create/cluster/cmd.go
lucasponce: Ouch, now I see it :).
Good catch, yes, that's missing.

pkg/interactive/validation.go
lucasponce: Yeah, added.
Perhaps I was biased, as the same validation was done in the "ValidateHostedClusterSubnets".

cmd/create/cluster/cmd.go
lucasponce: Added.

cmd/create/cluster/cmd.go
```diff
@@ -1568,15 +1568,20 @@ func run(cmd *cobra.Command, _ []string) {
...
+		if subnetsProvided && !isHostedCP {
+			err = ocm.ValidateSubnetsCount(multiAZ, privateLink, len(subnetIDs))
+			if err != nil {
+				r.Reporter.Errorf("%s", err)
+				os.Exit(1)
 			}
+		}
+
+		// In both cases (interactive and subnetsProvided) hosted clusters need to validate the private subnets
+		if isHostedCP {
```
oriAdler, line 1580: Should we enter this block only if `subnetsProvided == true` or for all cases?

oriAdler (commented):

cmd/create/cluster/cmd.go
lucasponce: for all cases, interactive and subnetsProvided, in both cases, we need to:
- count the number of private/public
- return that number for further replicas validation

PR #1125, opened 2023-03-23
Improve taint validation

andreadecorte:
Fix panic
Accept empty values
Add key and values validation

Related: OCPBUGS-10432 [SDA-8272](https://issues.redhat.com//browse/SDA-8272)

### DIFF
--- a/cmd/create/machinepool/machinepool.go
+++ b/cmd/create/machinepool/machinepool.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/briandowns/spinner"
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/validation"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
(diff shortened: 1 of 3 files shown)


cmd/create/machinepool/machinepool_sute_test.go
tbrisker, line 1: NP: typo in file name

cmd/create/machinepool/machinepool_test.go
```diff
@@ -0,0 +1,43 @@
...
+			"Expected key=value:scheduleType format", 0),
+		Entry(
+			"Empty value taint",
+			"node-role.kubernetes.io/infra=val:NoSchedule,node-role.kubernetes.io/infra=:NoSchedule",
+			"Expected key=value:scheduleType format", 0),
+		Entry(
+			"Empty effect taint",
+			"node-role.kubernetes.io/infra=:",
+			"Expected key=value:scheduleType format", 0),
+	)
```
tbrisker, line 42: what about `node-role.kubernetes.io/infra=foobar:NoSchedule`? is that a valid taint?

andreadecorte: yes. This is what I am testing in https://github.com/openshift/rosa/pull/1125/files#diff-2c663107801b3bd651998b1d576dbf34aa8d2826e2078c57a865c22775183707R22 if I am not missing something

cmd/create/machinepool/machinepool_sute_test.go
andreadecorte: thanks

oriAdler:
LGTM

cmd/create/machinepool/machinepool_test.go
gdbranco: https://issues.redhat.com/browse/OPCT-13

Related btw

andreadecorte: interesting @gdbranco . This one is just to avoid the panic but indeed we need to allow empty values both here and in the CS. Should we do it separately or you want to fix the 2 issues together?
@tbrisker val is just a random string, I will add another one for remove any confusion

cmd/create/machinepool/machinepool.go
```diff
@@ -574,13 +575,33 @@ func parseTaints(taints string) ([]*cmv1.TaintBuilder, error) {
...
 		if !strings.Contains(taint, "=") || !strings.Contains(taint, ":") {
-			return nil, fmt.Errorf("Expected key=value:scheduleType format for taints")
+			return nil, fmt.Errorf("Expected key=value:scheduleType format for taints. Got '%s'", taint)
 		}
-		tokens := strings.FieldsFunc(taint, Split)
-		taintBuilders = append(taintBuilders, cmv1.NewTaint().Key(tokens[0]).Value(tokens[1]).Effect(tokens[2]))
+		// First split effect
+		splitEffect := strings.Split(taint, ":")
+		// Then split key and value
+		split2 := strings.Split(splitEffect[0], "=")
```
tzvatot, line 586: NP: I suggest to name it something more meaningful then `split2`

tzvatot (changes requested):

cmd/create/machinepool/machinepool.go
andreadecorte: thanks, fixed

PR #1127, opened 2023-03-24
[SDA-8610] Update oidc config commands to use /oidc_configs endpoints

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8610

# What
Create list oidc-config to list all oidc configurations within org taken from token
Update create oidc-config to register the oidc config with OCM at the end of the flow
Update delete oidc-config to supply only oidc-config-id param and delete it from registered oidc configs

# Why
Due to a redesign of this functionality it is now required to register the oidc configs with OCM


cmd/create/oidcconfig/cmd.go
```diff
@@ -150,30 +169,58 @@ func run(cmd *cobra.Command, argv []string) {
 		}
 	}
 
-	if !args.rawFiles {
-		r.Reporter.Infof("This command will create a S3 bucket populating it with documents " +
-			"to be compliant with OIDC protocol. It will also create a Secret in Secrets Manager containing the private key.")
+	if args.managed && mode != aws.ModeAuto {
```
davidleerh, line 172: Screenshot from 2023-03-24 14-27-22
Tried running the condition via CLI but not getting the warning. Am I approaching this incorrectly? Please kindly advise. @gdbranco

cmd/create/oidcconfig/cmd.go
```diff
@@ -101,6 +105,20 @@ func init() {
 		"Prefix for the OIDC configuration, secret and provider.",
 	)
 
+	flags.BoolVar(
+		&args.managed,
+		managedFlag,
+		false,
+		"Indicates whether it is a Red Hat managed or unmanaged (Customer's hosted) OIDC Configuration.",
```
hunterkepley, line 112: NP: `(Customer's hosted)` -- should this be `Customer hosted`? Feels a bit out of place having 's here

hunterkepley (approved):
looked fine, just one nitpick

cmd/create/oidcconfig/cmd.go
gdbranco: True

PR #1128, opened 2023-03-24
[SDA-8611] Use oidc-config-id in cluster flow

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8611

# What
Uses oidc-config-id when creating cluster

# Why
Facilitates CX to a single param when providing oidc configuration


hunterkepley (commented):
/lgtm

cmd/create/cluster/cmd.go
```diff
@@ -2347,82 +2333,34 @@ func validateOperatorRolesAvailabilityUnderUserAwsAccount(awsClient aws.Client,
...
+		if oidcConfigId == "" && interactive.Enabled() {
+			_oidcConfigId, err := interactive.GetString(
 				interactive.Input{
-					Question:   "OIDC Endpoint URL",
-					Help:       cmd.Flags().Lookup(OidcEndpointUrlFlag).Usage,
-					Required:   false,
-					Default:    oidcEndpointUrl,
-					Validators: []interactive.Validator{interactive.IsURL},
+					Question: "OIDC Config ID",
+					Help:     cmd.Flags().Lookup(oidcConfigId).Usage,
```
oriAdler, line 2343: @gdbranco if the `oidcConfigId` isn't provided by the user, we get panic here.
The `Lookup` function returns **nil**.
image

Is this flow still optional for Hypershift clusters?

PR #1132, opened 2023-03-26
[SDA-8621] Use oidcConfigIdFlag instead of var

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8621
# What
Correctly setup usage help

# Why
Avoid segfault


gdbranco:
Merging as is a minor fix to fix master branch quickly

oriAdler:
LGTM

PR #1140, opened 2023-03-28
[SDA-8657] Better UX when using oidc config id and related commands

gdbranco:
Related issues: 
* https://issues.redhat.com/browse/SDA-8657
* https://issues.redhat.com/browse/SDA-8653
* https://issues.redhat.com/browse/SDA-8667
* https://issues.redhat.com/browse/SDA-8669
* https://issues.redhat.com/browse/SDA-8671
* https://issues.redhat.com/browse/SDA-8672

# What
Adds UX improvements and bug fixes to some oidc config flows

# Why
Better CX

### DIFF
--- a/cmd/create/operatorroles/by_prefix.go
+++ b/cmd/create/operatorroles/by_prefix.go
@@ -63,21 +63,7 @@ func handleOperatorRolesPrefixOptions(r *rosa.Runtime, cmd *cobra.Command) {
 		os.Exit(1)
 	}
 	args.hostedCp = isHostedCP
-	installerRoleArn := args.installerRoleArn
-	installerRoleArn, err = interactive.GetString(interactive.Input{
-		Question: "Installer Role ARN",
-		Help:     cmd.Flags().Lookup(InstallerRoleArnFlag).Usage,
-		Default:  installerRoleArn,
-		Required: true,
-		Validators: []interactive.Validator{
-			aws.ARNValidator,
-		},
-	})
-	if err != nil {
-		r.Reporter.Errorf("Expected a valid ARN: %s", err)
-		os.Exit(1)
-	}
-	args.installerRoleArn = installerRoleArn
+	args.installerRoleArn = interactive.GetInstallerRoleArn(r, cmd, args.installerRoleArn, "")
 }
 
 func handleOperatorRoleCreationByPrefix(r *rosa.Runtime, env string,
--- a/cmd/dlt/oidcconfig/cmd.go
+++ b/cmd/dlt/oidcconfig/cmd.go
@@ -40,9 +40,9 @@ var Cmd = &cobra.Command{
 	Use:     "oidc-config",
 	Aliases: []string{"oidconfig, oidcconfig"},
 	Short:   "Delete OIDC Config",
-	Long:    "Cleans up OIDC config based on secret ARN.",
-	Example: `  # Delete OIDC config based on secret ARN that has been supplied
-	rosa delete oidc-config --oidc-private-key-secret-arn <oidc_private_key_secret_arn>`,
+	Long:    "Cleans up OIDC config based on registered OIDC Config ID.",
+	Example: `  # Delete OIDC config based on registered OIDC Config ID that has been supplied
+	rosa delete oidc-config --oidc-config-id <oidc_config_id>`,
 	Hidden: true,
 	Run:    run,
 }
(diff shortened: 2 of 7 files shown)


cmd/create/operatorroles/by_prefix.go
```diff
@@ -20,6 +20,10 @@ import (
 	"github.com/openshift/rosa/pkg/rosa"
 )
 
+const (
+	validMinMinorVersion = "4.0"
```
oriAdler, line 24: Is that the min minor version for creating operator roles by prefix?

cmd/dlt/oidcconfig/cmd.go
```diff
@@ -152,12 +140,26 @@ func buildOidcConfigInput(r *rosa.Runtime) OidcConfigInput {
...
+				"please run the command supplying region parameter.", parsedSecretArn.Region, args.region)
+			os.Exit(1)
+		}
+		secretResourceName, err := aws.GetResourceIdFromSecretArn(secretArn)
+		if err != nil {
+			r.Reporter.Errorf("There was a problem parsing secret ARN '%s' : %v", secretArn, err)
+			os.Exit(1)
+		}
+		bucketName = strings.TrimPrefix(secretResourceName, prefixForPrivateKeySecret)
+		index := strings.LastIndex(bucketName, "-")
```
oriAdler, line 157: Can we add an example with a bucket name? so it would be easier to understand the logic of getting the bucket name and the index?

pkg/interactive/helper.go
```diff
@@ -0,0 +1,100 @@
...
+
+func GetOidcConfigID(r *rosa.Runtime, cmd *cobra.Command) string {
+	oidcConfigs, err := r.OCMClient.ListOidcConfigs()
+	if err != nil {
+		r.Reporter.Warnf("There was a problem retrieving OIDC Configurations "+
+			"for your organization: %v", err)
+		return ""
+	}
+	if len(oidcConfigs) == 0 {
+		return ""
```
oriAdler, line 25: Previously we gave the user the option to enter the ID, so now they can only choose from the list retrieved from the backend?

pkg/interactive/helper.go
```diff
@@ -0,0 +1,100 @@
...
+		Required: true,
+	})
+	if err != nil {
+		r.Reporter.Errorf("Expected a valid OIDC Config ID: %s", err)
+		os.Exit(1)
+	}
+	return strings.TrimSpace(strings.Split(oidcConfigId, "|")[0])
+}
+
+func GetInstallerRoleArn(r *rosa.Runtime, cmd *cobra.Command,
```
oriAdler, line 45: The code is similar to lines `811-860` in `cmd/create/cluster/cmd.go`, can we please reuse the function in `create cluster` as well?

So it will be easier to maintain, and logic won't diverge in the future.

cmd/create/operatorroles/by_prefix.go
gdbranco: I just setup something that will always pass basically, because there's no min version. Maybe skip the validation within the function when it is empty instead?

cmd/dlt/oidcconfig/cmd.go
gdbranco: Oh sure!

pkg/interactive/helper.go
gdbranco: If the listing is empty they don't have anything they can enter. However, they can still pass the ID directly through the arg if needed

pkg/interactive/helper.go
gdbranco: I didn't do this one as I have plans to change the one in cluster creation to unify the set of roles by the prefix

cmd/create/operatorroles/by_prefix.go
gdbranco: Changed to skip validation if supplied version is empty, what do you think?

oriAdler: Ok

pkg/interactive/helper.go
oriAdler: Would it be the same for this flow?
When getting the installer role, select an installer role that is part of a complete set.
Otherwise, the user can select an installer role from an incomplete set.

We have to take into consideration also Hypershift roles, as the set is going to include only three roles.

oriAdler: Anyhow, I assume we can refactor this in a sequence PR.

PR #1142, opened 2023-03-29
Delete account-roles - delete both types for the default flow

oriAdler:
1. For `--hosted-cp` delete only hypershift account roles.
2. For `--classic` delete only classic account roles.

Related: [SDA-8666](https://issues.redhat.com/browse/SDA-8666)

Default flow deletes all account roles:
image

If role type is specified, delete only the adequate set:
image

### DIFF
--- a/cmd/dlt/accountroles/cmd.go
+++ b/cmd/dlt/accountroles/cmd.go
@@ -35,6 +35,7 @@ import (
 var args struct {
 	prefix   string
 	hostedCP bool
+	classic  bool
 }
 
 var Cmd = &cobra.Command{
@@ -65,6 +66,14 @@ func init() {
 		"Delete hosted control planes roles (HyperShift)",
 	)
 	flags.MarkHidden("hosted-cp")
+
+	flags.BoolVar(
+		&args.classic,
+		"classic",
+		false,
+		"Delete classic account roles",
+	)
+	flags.MarkHidden("classic")
 
 	aws.AddModeFlag(Cmd)
 	confirm.AddFlag(flags)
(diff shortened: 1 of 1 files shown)


cmd/dlt/accountroles/cmd.go
```diff
@@ -66,6 +67,14 @@ func init() {
 	)
 	flags.MarkHidden("hosted-cp")
 
+	flags.BoolVar(
+		&args.hostedCP,
```
gdbranco, line 71: classic

cmd/dlt/accountroles/cmd.go
```diff
@@ -137,11 +146,24 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
-	// TODO: delete both classic and hosted CP roles by default, once managed policies are in place
-	err = deleteAccountRoles(r, env, prefix, clusters, mode, args.hostedCP)
-	if err != nil {
-		r.Reporter.Errorf("%s", err)
-		os.Exit(1)
+	// Default flow deletes both classic and hosted-CP roles
+	if args.classic == false && args.hostedCP == false {
```
gdbranco, line 150: if !args.classic && !args.hostedCP

cmd/dlt/accountroles/cmd.go
oriAdler: Good catch!
Thanks

cmd/dlt/accountroles/cmd.go
oriAdler: Changed

PR #1144, opened 2023-03-30
[SDA-8685] Added hosted-cp to rosa list versions

hunterkepley:
[SDA-8685](https://issues.redhat.com//browse/SDA-8685) -- Add hosted-cp and only list HCP-enabled versions in rosa CLI

### DIFF
--- a/cmd/list/version/cmd.go
+++ b/cmd/list/version/cmd.go
@@ -32,6 +32,7 @@ import (
 
 var args struct {
 	channelGroup string
+	hostedCp     bool
 }
 
 var Cmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


cmd/list/version/cmd.go
```diff
@@ -67,7 +73,23 @@ func run(cmd *cobra.Command, _ []string) {
...
+		hcpVersions = append(hcpVersions, version)
+	}
+
+	// If hosted-cp arg is supplied, use the hcp enabled versions
+	if args.hostedCp {
+		versions = hcpVersions
+	}
 
 	// Remove disabled versions
 	for _, version := range versions {
```
gdbranco, line 95: What do you think of adding the if args.hostedCp && !version.HostedControlPlaneEnabled in the loop that sets up available versions instead of separating the slice?

hunterkepley: Smart, let me do that

PR #1153, opened 2023-04-04
[SDA-8752] Add type of cluster column in list clusters command

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8752
# What
Add type of cluster column in list clusters command

# Why
Easier for customer to check


hunterkepley (approved):
/lgtm

PR #1154, opened 2023-04-04
[SDA-8730] Force interactive enable if required params for hcp are not supplied

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8730
# What
Forces interactive mode if required params are not supplied

# Why
Some errors are show and the customer might be confused into next step


hunterkepley (approved):
LGTM

PR #1156, opened 2023-04-05
[SDA-8753] Add list operator-roles command

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8753
# What
Add list operator-roles command with version and prefix params

# Why
With the possibility of reusing operator-roles might be nice for the user to be able to check which are in place and which version they belong to


hunterkepley (approved):
/lgtm

PR #1161, opened 2023-04-11
chore: remove deprecated lib `io/ioutil`

ciaranRoche:
```
"io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.
```


renan-campos (approved):
/lgtm

PR #1165, opened 2023-04-11
SDA-8777 | feat: check if there are clusters using operator roles prefix

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8777
# What
Check if clusters are using operator roles prefix before removing roles

# Why
Ensures roles are not removed from working clusters


renan-campos (approved):
/lgtm

PR #1166, opened 2023-04-11
SDA-8786 | feat: remove public access block and set ownership to owner preferred

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8786
# Why
AWS introduced security changes where default behavior is to block public access
https://aws.amazon.com/blogs/aws/heads-up-amazon-s3-security-changes-are-coming-in-april-of-2023/


pkg/aws/client.go
```diff
@@ -984,6 +986,27 @@ func (c *awsClient) CreateS3Bucket(bucketName string, region string) error {
...
+	_, err = c.s3Client.PutBucketOwnershipControls(&s3.PutBucketOwnershipControlsInput{
+		Bucket: aws.String(bucketName),
+		OwnershipControls: &s3.OwnershipControls{
+			Rules: []*s3.OwnershipControlsRule{
+				{
+					ObjectOwnership: aws.String(OwnershipControlBucketOwnerPreferred),
+				},
+			},
+		},
+	})
```
jharrington22, line 1006: We can leave the default object ownership (which is recommended by AWS) and instead setup an object policy

See https://github.com/openshift/cloud-credential-operator/pull/526 for an example.

gdbranco: Yes, I got through that option as well. I'm just thinking to provide a temporary solution while we work things out to have it managed within repo (https://github.com/openshift/managed-cluster-config/tree/master/resources/sts)

gdbranco: @jharrington22 Let me know what you think so we can move this hardcoded solution temporarily and in the meantime work on the policy repo

pkg/aws/client.go
```diff
@@ -970,6 +986,28 @@ func (c *awsClient) CreateS3Bucket(bucketName string, region string) error {
...
+			IgnorePublicAcls:      aws.Bool(true),
+			BlockPublicPolicy:     aws.Bool(false),
+			RestrictPublicBuckets: aws.Bool(false),
+		},
+	})
+	if err != nil {
+		return err
+	}
+
+	_, err = c.s3Client.PutBucketPolicy(&s3.PutBucketPolicyInput{
```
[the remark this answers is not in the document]
gdbranco: ROSA runs within client side, so in general it may have all permissions as long as the company is not restricting access. If this happens that's why we have manual mode where they give the commands with someone who has permission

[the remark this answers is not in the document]
ciaranRoche: That is what I was thinking, do you know if we have it documented anywhere the permissions needed to actually fun the CLI?

PR #1167, opened 2023-04-12
Fix version validation error handling

andreadecorte:
As the `HasHostedCPSupport` method is used in the nodepool flow, so we need to handle errors in the caller


lucasponce:
lgtm

PR #1173, opened 2023-04-13
fix: small issue when customer sets up oidc provider with issuer url '/' at the end

gdbranco:
# What
When customer sets up oidc provider using '/' at the end of the URL the deletion process fails

# Why
Fixes a flow issue


ciaranRoche (approved):
/lgtm

PR #1174, opened 2023-04-14
SDA-6239 | fix: Add message to helper --delete param on init cmd

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-6239
# What
Adds message from the deprecated value this substituted

# Why
There was no helper message

### DIFF
--- a/cmd/initialize/cmd.go
+++ b/cmd/initialize/cmd.go
@@ -65,7 +65,7 @@ func init() {
 		&args.dlt,
 		"delete-stack",
 		false,
-		"Deletes stack template applied to your AWS account during the 'init' command.\n",
+		"Deletes stack template applied to your AWS account during the 'init' command.",
 	)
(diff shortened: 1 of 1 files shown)


cmd/initialize/cmd.go
```diff
@@ -65,15 +65,15 @@ func init() {
...
+		"Deletes stack template applied to your AWS account during the 'init' command.",
 	)
 	flags.MarkDeprecated("delete-stack", "use --delete instead")
 
 	flags.BoolVar(
 		&args.dlt,
 		"delete",
 		false,
-		"",
+		"Deletes stack template applied to your AWS account during the 'init' command.",
```
gdbranco, line 76: Couldn't really find more info on this command, would you say we can copy the deprecated usage?

@vkareh

vkareh: Ah sorry, I misunderstood your question.

Yes, you can copy the message. They do the same thing.

PR #1175, opened 2023-04-14
SDA-6562 | feat: skip subnet choice when AZ is supplied

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-6562
# What
Shows only subnets from supplied AZ

# Why
So as not to show user subnets he might not be interest in

### DIFF
--- a/cmd/create/machinepool/helper.go
+++ b/cmd/create/machinepool/helper.go
@@ -10,8 +10,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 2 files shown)


gdbranco:
@oriAdler PTAL

oriAdler:
For classic ROSA day 2 operation, if the user selected an AZ, it has to be from the cluster's AZ, and then the AZ is sufficient and the subnet is redundant.
The other option to select a subnet should be displayed when no AZ is provided.

PR #1176, opened 2023-04-14
SDA-8770 | fix: accept only official default prefix when selecting default prefix

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8770
# What
Selects only official default prefix when selecting as part of auto confirm mode

# Why
Non official default prefix shouldn't be selected without args being supplied

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -833,7 +833,11 @@ func run(cmd *cobra.Command, _ []string) {
 			defaultRoleARN := roleARNs[0]
 			// Prioritize roles with the default prefix
 			for _, rARN := range roleARNs {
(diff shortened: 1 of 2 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -833,7 +833,11 @@ func run(cmd *cobra.Command, _ []string) {
 			defaultRoleARN := roleARNs[0]
 			// Prioritize roles with the default prefix
 			for _, rARN := range roleARNs {
-				if strings.Contains(rARN, fmt.Sprintf("%s-%s-Role", aws.DefaultPrefix, role.Name)) {
+				roleName, err := aws.GetResourceIdFromARN(rARN)
```
gdbranco, line 836: Repeated behavior, this will be fixed in https://github.com/openshift/rosa/pull/927

PR #1177, opened 2023-04-17
SDA-8836 | feat: add output json/yaml to create oidc config

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8836
# What
Adds output option for auto mode in create OIDC config

# Why
Easier for automation flows

### DIFF
--- a/cmd/create/oidcconfig/cmd.go
+++ b/cmd/create/oidcconfig/cmd.go
@@ -50,6 +50,7 @@ import (
 	"github.com/openshift/rosa/pkg/helper"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/interactive/confirm"
+	"github.com/openshift/rosa/pkg/output"
 	"github.com/openshift/rosa/pkg/rosa"
 )
 
@@ -124,6 +125,7 @@ func init() {
 	confirm.AddFlag(flags)
 	interactive.AddFlag(flags)
 	arguments.AddRegionFlag(flags)
+	output.AddFlag(Cmd)
 }
 
 func checkInteractiveModeNeeded(cmd *cobra.Command) {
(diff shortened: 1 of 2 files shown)


cmd/create/oidcconfig/cmd.go
```diff
@@ -207,14 +214,14 @@ func run(cmd *cobra.Command, argv []string) {
 
 	if !args.managed {
 		if !args.rawFiles {
-			if r.Reporter.IsTerminal() {
```
oriAdler, line 210: I see in some places in the code we have `!output.HasFlag() && r.Reporter.IsTerminal()` and in other places we have `!output.HasFlag() || r.Reporter.IsTerminal()`.

So in case we need json/yaml output for automation we don't want to print the info messages, right?

cmd/create/oidcconfig/cmd.go
```diff
@@ -394,7 +401,7 @@ func (s *CreateUnmanagedOidcConfigAutoStrategy) execute(r *rosa.Runtime) {
 	privateKeySecretName := s.oidcConfig.PrivateKeySecretName
 	installerRoleArn := args.installerRoleArn
 	var spin *spinner.Spinner
-	if r.Reporter.IsTerminal() {
+	if output.HasFlag() && r.Reporter.IsTerminal() {
```
oriAdler, line 404: Should it be?
```suggestion
	if !output.HasFlag() && r.Reporter.IsTerminal() {
```

oriAdler (commented):
Left two small comments, other than that, LGTM.

cmd/create/oidcconfig/cmd.go
gdbranco: Thanks for catching!

cmd/create/oidcconfig/cmd.go
```diff
@@ -445,6 +452,14 @@ func (s *CreateUnmanagedOidcConfigAutoStrategy) execute(r *rosa.Runtime) {
 		r.Reporter.Warnf("Secret ARN: %s\tBucketUrl: %s", secretARN, bucketUrl)
 		os.Exit(1)
 	}
+	if output.HasFlag() {
+		err = output.Print(oidcConfig)
```
ciaranRoche, line 456: Curious about the spinner, should it be stopped before printing, or will printing interrupt it?

gdbranco: Spinner is not created if we have the output flag. However, I think we do need to revisit a lot of code that uses spinner to have some unified check error and stop spinner setup

oriAdler:
LGTM

cmd/create/oidcconfig/cmd.go
ciaranRoche: We can track it under this - https://issues.redhat.com/browse/SDA-8847

PR #1182, opened 2023-04-18
feat: Create cluster - filter roles by policy type

oriAdler:
For hosted CP cluster display only roles with hosted CP managed policies. 
For classic cluster display only roles with unmanaged policies.

Related: [SDA-8843](https://issues.redhat.com/browse/SDA-8843)

For hosted CP it chooses the only set with managed CP policies:
image

For classic it displays all roles excluding roles with CP managed policies:
image

### DIFF
--- a/cmd/create/oidcconfig/cmd.go
+++ b/cmd/create/oidcconfig/cmd.go
@@ -220,7 +220,8 @@ func run(cmd *cobra.Command, argv []string) {
 					"to be compliant with OIDC protocol. It will also create a Secret in Secrets Manager containing the private key")
 			}
 			if mode == aws.ModeAuto && (interactive.Enabled() || (confirm.Yes() && args.installerRoleArn == "")) {
-				args.installerRoleArn = interactive.GetInstallerRoleArn(r, cmd, args.installerRoleArn, minorVersionForGetSecret)
+				args.installerRoleArn = interactive.GetInstallerRoleArn(r, cmd, args.installerRoleArn,
+					minorVersionForGetSecret, false, false)
 			}
 			if interactive.Enabled() {
 				prefix, err := interactive.GetString(interactive.Input{
--- a/cmd/create/operatorroles/by_prefix.go
+++ b/cmd/create/operatorroles/by_prefix.go
@@ -55,7 +55,8 @@ func handleOperatorRolesPrefixOptions(r *rosa.Runtime, cmd *cobra.Command) {
 		os.Exit(1)
 	}
 	args.hostedCp = isHostedCP
-	args.installerRoleArn = interactive.GetInstallerRoleArn(r, cmd, args.installerRoleArn, "")
+	args.installerRoleArn = interactive.GetInstallerRoleArn(r, cmd, args.installerRoleArn, "",
+		true, args.hostedCp)
 }
 
 func handleOperatorRoleCreationByPrefix(r *rosa.Runtime, env string,
--- a/pkg/interactive/helper.go
+++ b/pkg/interactive/helper.go
@@ -43,14 +43,20 @@ func GetOidcConfigID(r *rosa.Runtime, cmd *cobra.Command) string {
 }
 
 func GetInstallerRoleArn(r *rosa.Runtime, cmd *cobra.Command,
-	defaultInstallerRoleArn string, minMinorVersion string) string {
+	defaultInstallerRoleArn string, minMinorVersion string, filterByPolicyType bool, requireHostedCPPolicies bool) string {
 	spin := spinner.New(spinner.CharSets[9], 100*time.Millisecond)
 	spin.Start()
 	awsClient := r.AWSClient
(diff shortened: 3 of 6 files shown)


pkg/interactive/helper.go
```diff
@@ -50,7 +50,7 @@ func GetInstallerRoleArn(r *rosa.Runtime, cmd *cobra.Command,
 	role := aws.AccountRoles[aws.InstallerAccountRole]
 	roleARN := defaultInstallerRoleArn
 	// Find all installer roles in the current account using AWS resource tags
-	roleARNs, err := awsClient.FindRoleARNs(aws.InstallerAccountRole, minMinorVersion)
```
oriAdler, line 53: @gdbranco I wasn't sure if it should be here only roles with hosted CP managed policies?

gdbranco: I think it would require to add this arg to GetInstallerRoleArn instead and we can check what is using it better

I think that for now all uses will be false, but worth checking to be sure
I know that for OIDC Config it shouldn't matter

oriAdler: Changed, thanks

cmd/create/oidcconfig/cmd.go
```diff
@@ -207,7 +207,8 @@ func run(cmd *cobra.Command, argv []string) {
 					"to be compliant with OIDC protocol. It will also create a Secret in Secrets Manager containing the private key")
 			}
 			if mode == aws.ModeAuto && (interactive.Enabled() || (confirm.Yes() && args.installerRoleArn == "")) {
-				args.installerRoleArn = interactive.GetInstallerRoleArn(r, cmd, args.installerRoleArn, minorVersionForGetSecret)
+				args.installerRoleArn = interactive.GetInstallerRoleArn(r, cmd, args.installerRoleArn,
+					minorVersionForGetSecret, false)
```
oriAdler, line 211: @gdbranco do you know what should be the value here? For `false` it will filter out non-HCP managed policies

cmd/create/operatorroles/by_prefix.go
```diff
@@ -54,7 +54,7 @@ func handleOperatorRolesPrefixOptions(r *rosa.Runtime, cmd *cobra.Command) {
 		os.Exit(1)
 	}
 	args.hostedCp = isHostedCP
-	args.installerRoleArn = interactive.GetInstallerRoleArn(r, cmd, args.installerRoleArn, "")
+	args.installerRoleArn = interactive.GetInstallerRoleArn(r, cmd, args.installerRoleArn, "", false)
```
oriAdler, line 57: @gdbranco I assume we should pass here the value `args.hostedCp`, can you confirm?

gdbranco: Yes, for this one this is correct

cmd/create/oidcconfig/cmd.go
gdbranco: hmm, we should be able to show either here, it only cares that the installer role has GetSecretValue permission

oriAdler: It was changed to ignore the policy type for creating the OIDC config.

gdbranco:
LGTM

renan-campos (approved):
lgtm

davidleerh:
hi @oriAdler 
I tried to create a classic cluster with the changes made; however, I was unable to see the list of installer roles being filtered (filter out HCP MP installer roles).

In the following example,
image
I created an HCP role with MP:
image
image
However, it was not being filtered out when listing the installer roles for classic cluster creation.

Vice versa,
I tried to create non mp, classic account roles; however, the installer role was being picked up when trying to create an hcp cluster via interactive mode:
image
image
image

Would you kindly advise / correct me if I am looking at this incorrectly? :slightly_smiling_face: 
Thank you!

oriAdler:
> hi @oriAdler I tried to create a classic cluster with the changes made; however, I was unable to see the list of installer roles being filtered (filter out HCP MP installer roles).
> 
> In the following example, image I created an HCP role with MP: image image However, it was not being filtered out when listing the installer roles for classic cluster creation.
> 
> Vice versa, I tried to create non mp, classic account roles; however, the installer role was being picked up when trying to create an hcp cluster via interactive mode: image image image
> 
> Would you kindly advise / correct me if I am looking at this incorrectly? slightly_smiling_face Thank you!

Hi David,
Thanks for the profound check.
I can't reproduce it. When I run the command I get filtered roles by policy type.
Let me schedule a time today or during ROSA office hours and we can go over it together.

davidleerh:
Hi @oriAdler 
My sincere apologies..
My manual tests were conducted using the wrong local setup / branch.
After testing again, I was able to see the filter is working properly based on the tags.

oriAdler:
Great, thanks for the reviews.
Let's merge this one once the CLB/NLB issues with managed policies are solved.
As this change is going to enforce users to use managed policies for HCP.

PR #1186, opened 2023-04-20
SDA-8897 | fix: shared operator roles with managed policies

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-8897
# What
When using shared operator roles with managed policies the validation does not need to consider the policies version

# Why
Managed policies are compatible with all cluster versions

### DIFF
--- a/pkg/ocm/helpers.go
+++ b/pkg/ocm/helpers.go
@@ -813,6 +813,14 @@ func ValidateOperatorRolesMatchOidcProvider(awsClient aws.Client,
 			return weberr.Errorf("Operator role '%s' does not have trusted relationship to '%s' issuer URL",
 				roleARN, parsedUrl.Host)
 		}
+		hasManagedPolicies, err := awsClient.HasManagedPolicies(roleARN)
+		if err != nil {
+			return err
+		}
+		if hasManagedPolicies {
+			// Managed policies should be compatible with all versions
+			continue
+		}
 		policiesDetails, err := awsClient.GetAttachedPolicy(roleObject.RoleName)
 		if err != nil {
 			return err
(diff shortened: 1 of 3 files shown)


pkg/ocm/helpers.go
```diff
@@ -813,6 +813,14 @@ func ValidateOperatorRolesMatchOidcProvider(awsClient aws.Client,
 			return weberr.Errorf("Operator role '%s' does not have trusted relationship to '%s' issuer URL",
 				roleARN, parsedUrl.Host)
 		}
+		hasManagedPolicies, err := awsClient.HasManagedPolicies(roleARN)
+		if err != nil {
+			return err
+		}
+		if hasManagedPolicies {
+			// Managed policies should be compatible with all versions
```
ciaranRoche, line 821: I am not a fan of the word `should`  is there any kind of validations we can do here to ensure they are compatible?

ciaranRoche: Never mind, I just read the ticket, we use a tag to check compatibility, and the tag does not exist on managed policies. 
That said I still feel we need some method to validate these policies

ciaranRoche: Given this solves the issue in the ticket, lets merge it.

gdbranco: To validate we would need to start going through all permission actions and have a list that matches each functionality

It is possible, just not something being done at the moment or expected to be done

ciaranRoche: Ill make note, and create a ticket for this, it be nice to validate at the granular level of permissions and not just on a tag

oriAdler: The `should` refers here to that the managed policies are owned by AWS and we are working under the assumption they are always up to date.
Unlike unmanaged policies, where the user has to upgrade.

oriAdler:
lgtm

PR #1187, opened 2023-04-20
SDA-8382|feat:Add tuning config support

andreadecorte:
* Add create/list/describe/update/delete tuning configs
* Add tuning config support for node pools

# UX
This PR adds a few new commands
```
rosa list tuning-configs -c <clusterid> [-o json]
rosa create tuning-config -c <clusterid> --name --spec-path <specFile>
rosa delete tuning-config -c <clusterid> <name>
rosa describe tuning-config -c <clusterid> <name> [-o json]
rosa edit tuning-config -c <clusterid> <name> --spec-path <specFile>
```
And an additional parameter for node pools
```
rosa create machinepool -c <clusterid> .. --tuning-configs="config1,config2"
rosa edit machinepool -c <clusterid> nodePoolName --tuning-configs="config1,config2"
```
And the relevant support for interactive mode for creation/edit. In node pool interactive mode, user is presented with the list of the tuning config existing in the cluster.

# References
* [general design doc](https://docs.google.com/document/d/1l1yYx52t3ckMzD0jqBrtTtBjnjJJ444nQmH-Wb0l3u0/edit#)
* JIRA: [SDA-8382](https://issues.redhat.com/browse/SDA-8382)


andreadecorte:
@wgordon17 PTAL if it looks ok to you in terms of UX

cmd/create/tuningconfigs/cmd.go
```diff
@@ -0,0 +1,133 @@
...
+	r.Reporter.Infof("To view all tuning configs, run 'rosa list tuning-configs -c %s'", clusterKey)
+}
+
+func buildTuningConfigFromInputFile(specPath string, name string, clusterKey string) (*cmv1.TuningConfig, error) {
+	specJson, err := input.UnmarshalInputFile(specPath)
+	if err != nil {
+		return nil, fmt.Errorf("Expected a valid TuneD spec file: %v", err)
+	}
+
+	tuningConfigBuilder := cmv1.NewTuningConfig().Name(name).Spec(specJson)
```
gdbranco, line 126: Does that also work with yaml? Should there be a warning for the user or validation that it is a json file?

gdbranco:
Overall looks good, is there any task to includes docs for this? I feel like it is necessary. Also should any of this be hidden for a while?

cmd/create/tuningconfigs/cmd.go
andreadecorte: Json only. Validation is done on backend side, you will get an error like.

```
rosa create tuningconfig -c ad2 --name tuned-2 --spec-path invalidspec.json
E: Expected a valid TuneD spec file: invalid character 'i' looking for beginning of value
```

wgordon17:
Cool, I think this looks good 👍 Any reason that there's no `rosa create machinepool --tuning-configs`? Or is that implied?

@gdbranco Yes, we'll _definitely_ need follow up docs

@andreadecorte This is HCP-*only*, right? This has no applicability to Classic

andreadecorte:
> Cool, I think this looks good 👍 Any reason that there's no `rosa create machinepool --tuning-configs`? Or is that implied?
> 
Just forgot to add in the description, it's there of course. Thanks for noticing
> @gdbranco Yes, we'll _definitely_ need follow up docs
Agree, I expect this to follow the normal ROSA doc flow? @wgordon17 do you have already something tracking this?
> 
> @andreadecorte This is HCP-_only_, right? This has no applicability to Classic
Correct. But good point, I will do the check directly here to avoid being rejected by the backend. Let me fix that

@gdbranco I don't think they should be hidden, it should be just a normal feature.

andreadecorte:
/hold
waiting for backend production load

wgordon17:
Newly minted docs card: https://issues.redhat.com/browse/OSDOCS-5816

PR #1194, opened 2023-04-21
fix: Add flag to bypass confirmation for hibernate/resume

renan-campos:
When calling `rosa hibernate|resume cluster` it always asks for a confirmation. However, the `--help` text does not show that the confirmation prompt can be bypassed with the `--yes`.

This PR adds the --yes flag to these operations.

### DIFF
--- a/cmd/hibernate/cluster/cmd.go
+++ b/cmd/hibernate/cluster/cmd.go
@@ -38,6 +38,7 @@ var Cmd = &cobra.Command{
 
 func init() {
 	ocm.AddClusterFlag(Cmd)
+	confirm.AddFlag(Cmd.Flags())
 }
 
 func run(cmd *cobra.Command, _ []string) {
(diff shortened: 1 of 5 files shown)


renan-campos:
I've broken the changes into two commits:

The first commits adds the --yes flag in a way that matches the pattern of the codebase.
The second commit refactors the code to not use global variables and init when creating the flags for hibernate/resume. The hope is to move away from global variables and init(). If this is an unwelcome change, this commit can be removed.

cmd/hibernate/cluster/cmd.go
```diff
@@ -55,7 +56,7 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
-	if !confirm.Confirm("hibernate cluster %s", clusterKey) {
+	if !confirm.Yes() && !confirm.Confirm("hibernate cluster %s", clusterKey) {
```
gdbranco, line 59: Please check to make sure, but I think you can use confirm.Prompt instead of confirm.Confirm when you want it to be used alongside confirm.Yes

gdbranco:
> I've broken the changes into two commits:
> 
> The first commits adds the --yes flag in a way that matches the pattern of the codebase. The second commit refactors the code to not use global variables and init when creating the flags for hibernate/resume. The hope is to move away from global variables and init(). If this is an unwelcome change, this commit can be removed.

I'd say this is good, would be worth to create a new task to change all commands into something similar and marking it as tech debt

PR #1198, opened 2023-04-24
SDA-8914 | fix: rosa upgrade hcp account roles - handle edge cases

oriAdler:
1. Prevent creating classic account roles with prefixes ending with `-HCP`.
2. Improve the error message for `rosa upgrade account-roles`.

Prevent `-HCP` for classic account roles:
image

Improve `upgrade account roles` error messages:
image
image

### DIFF
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -256,6 +256,10 @@ func run(cmd *cobra.Command, argv []string) {
 		r.Reporter.Errorf("Expected a valid role prefix matching %s", aws.RoleNameRE.String())
 		os.Exit(1)
 	}
+	if !args.hostedCP && strings.HasSuffix(prefix, "-HCP") {
+		r.Reporter.Errorf("The '-HCP' suffix is reserved for hosted CP managed policies")
+		os.Exit(1)
+	}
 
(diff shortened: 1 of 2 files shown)


oriAdler:
@gdbranco @ciaranRoche PTAL

cmd/create/accountroles/cmd.go
```diff
@@ -256,6 +256,10 @@ func run(cmd *cobra.Command, argv []string) {
 		r.Reporter.Errorf("Expected a valid role prefix matching %s", aws.RoleNameRE.String())
 		os.Exit(1)
 	}
+	if !args.hostedCP && strings.HasSuffix(prefix, "-HCP") {
+		r.Reporter.Errorf("'-HCP' suffix is not support for classic account roles")
```
gdbranco, line 260: I'd say something along the lines of "'-HCP' suffix is reserved/restricted to Hosted Control Planes managed policies" would be clearer for customer

oriAdler: Good point changed.

Why would we like to run this query?

gdbranco: If someone is using it already it might be confusing for the customer when we start doing this

PR #1199, opened 2023-04-25
Refactor oidcconfig and move a reuse-able code to a new helper package

bardielle:
Moving a reuse-able code from oidcconfig to an helper package


cmd/create/oidcconfig/cmd.go
```diff
@@ -265,7 +253,15 @@ func run(cmd *cobra.Command, argv []string) {
 		}
 	}
 
-	oidcConfigInput := buildOidcConfigInput(r)
+	oidcConfigInput := oidc_config.OidcConfigInput{}
+	if !args.managed {
```
gdbranco, line 257: Any reason to call this explicitly instead of letting the builder handle it?

pkg/helper/oidc_config/helper.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/openshift/rosa/pkg/helper"
+)
+
+const (
+	defaultLengthRandomLabel = 4
+	maxLengthUserPrefix      = 15
+
+	rawFilesFlag         = "raw-files"
```
gdbranco, line 25: I think some of these consts are duplicated and unused

cmd/create/oidcconfig/cmd.go
bardielle: In my idea the caller should be aware if the oidc config is empty or not (due to the managed attribute)

gdbranco: It's ok too I suppose

PR #1204, opened 2023-04-26
SDA-8716 | Fix: provide warning message when no OIDC config is found

willkutler:
ref: https://issues.redhat.com/browse/SDA-8716?filter=-1

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -2369,6 +2369,7 @@ func handleOidcConfigOptions(r *rosa.Runtime, cmd *cobra.Command, isSTS bool, is
 	}
 	if oidcConfigId == "" {
 		if !isHostedCP {
+			r.Reporter.Warnf("No OIDC Configuration found; will continue with the classic flow.")
 			return nil
 		}
 		if args.classicOidcConfig {


cmd/create/cluster/cmd.go
```diff
@@ -2368,10 +2368,8 @@ func handleOidcConfigOptions(r *rosa.Runtime, cmd *cobra.Command, isSTS bool, is
 		}
 	}
 	if oidcConfigId == "" {
-		if !isHostedCP {
-			return nil
```
gdbranco, line 2372: The only thing is that I feel like this message should show only for this one

If the cluster is a hosted CP cluster and the user specified args.classicOidcConfig it doesn't mean that an oidc configuration was not found but that the check was skipped, so it means the customer doesn't even want to check if there are any.

willkutler:
@gdbranco thanks for the review - does this look good now?

gdbranco:
> @gdbranco thanks for the review - does this look good now?

Would you mind adding a second check as a counter part to the args.classicOidcConfig?
Something similar to
```
if args.classicOidcConfig {
return nil
} else {
warn("No OIDC Configuration found;  Please create an OIDC Configuration through `rosa create oidc-configuration`")
os.Exit(1)
}
```
@pvasant mentioned would be better for UX

PR #1205, opened 2023-04-26
SDA-8245 | feat: ability to set billing account parameter for hypershift clusters

cristianoveiga:


pkg/ocm/billing.go
```diff
@@ -0,0 +1,58 @@
...
+
+	if len(billingAccounts) == 0 {
+		return billingAccounts, errors.New("no billing accounts found")
+	}
+
+	return billingAccounts, nil
+}
+
+func IsValidAWSAccount(account string) bool {
+	if len(account) == 0 || awsAccountRegexp.MatchString(account) {
```
oriAdler, line 54: Maybe we should call this function only if `account != ""`?
It might be confusing that we consider here an empty string as a valid AWS account.

cmd/create/cluster/cmd.go
```diff
@@ -682,6 +691,55 @@ func run(cmd *cobra.Command, _ []string) {
...
+				r.Reporter.Errorf("Expected a valid billing account: %s", err)
+				os.Exit(1)
+			}
+		} else {
+			billingAccount = ""
+		}
+	}
+
+	if !ocm.IsValidAWSAccount(billingAccount) {
+		r.Reporter.Errorf("Expected a valid AWS account")
```
oriAdler, line 739:
```suggestion
		r.Reporter.Errorf("Expected a valid AWS account billing account")
```

pkg/ocm/billing.go
cristianoveiga: My original idea was to keep the len(account) == 0 only until this attribute becomes mandatory (which will).

But I see how this can be confusing, as you pointed out. I will move this logic out of this function.

cmd/create/cluster/cmd.go
cristianoveiga: I've updated the message.

rawsyntax:
LGTM

PR #1207, opened 2023-04-27
filter out subnet in local zone for clusters lower than 4.12

marcolan018:
https://issues.redhat.com/browse/SDA-8111


cmd/create/machinepool/helper.go
```diff
@@ -65,6 +67,25 @@ func getSubnetFromUser(cmd *cobra.Command, r *rosa.Runtime, isSubnetSet bool,
...
+		if rawID, ok := version.GetRawID(); ok {
+			clusterVer, err := ver.NewVersion(rawID)
+			if err != nil {
+				return nil, err
+			}
+			if !clusterVer.GreaterThanOrEqual(lowestLocalZoneSupportVer) {
+				checkLocalZone = true
+			}
+		}
+	}
```
ciaranRoche, line 87: Would it be possible if we pull this out into its own function. It would help flatten out the nesting we have as we could return early if `!ok` or `err`

ciaranRoche: Could we add a unit test for this also please :pray:

ciaranRoche (commented):
Comment inline, pending approval of this being a bug or not.

cmd/create/machinepool/helper.go
marcolan018: @ciaranRoche 
new changes committed, thanks for the comments.

cmd/create/machinepool/helper.go
```diff
@@ -3,14 +3,17 @@ package machinepool
 import (
 	"os"
 
+	ver "github.com/hashicorp/go-version"
```
ciaranRoche, line 6: NP : Can you reorgnaise the imports. We try to group the imports when possible, put them into 3 seperate groups. Standard pkgs, external pkgs, and then internal pkgs, all with a space between them

Example -> https://github.com/openshift/rosa/blob/master/cmd/create/cluster/cmd.go#L20-L48

marcolan018: fixed

cmd/create/machinepool/helper.go
```diff
@@ -73,8 +83,41 @@ func getSubnetOptions(r *rosa.Runtime, cluster *cmv1.Cluster) ([]string, error)
...
+		return false, err
+	}
+
+	if version, ok := cluster.GetVersion(); ok {
+		if rawID, ok := version.GetRawID(); ok {
+			clusterVer, err := ver.NewVersion(rawID)
+			if err != nil {
+				return false, err
+			}
+			if !clusterVer.GreaterThanOrEqual(lowestLocalZoneSupportVer) {
```
oriAdler, line 115: NP
We can remove the `!` and change the order for simplicity.
```
if clusterVer.GreaterThanOrEqual(lowestLocalZoneSupportVer) {
    return false
} else {
    return true
} 
```

cmd/create/machinepool/helper.go
```diff
@@ -73,8 +83,41 @@ func getSubnetOptions(r *rosa.Runtime, cluster *cmv1.Cluster) ([]string, error)
...
+				return true, nil
+			} else {
+				return false, nil
+			}
+		}
+	}
+	return true, nil
```
oriAdler, line 122: Why do we return `true` here?

cmd/create/machinepool/helper_test.go
```diff
@@ -0,0 +1,49 @@
...
+			Entry("empty version",
+				"",
+				false,
+				nil,
+			),
+		)
+	})
+})
+
+func checkLocalZoneSupported(version string, expectedResult bool, expectedErr error) {
```
oriAdler, line 40: Where do we use this function?

PR #1209, opened 2023-04-27
SDA-8689 Adjust default replicas on hosted clusters

lucasponce:
On hosted clusters, when the user doesn't define the number of replicas, instead having a fixed number, it takes the number of private subnets as default.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -56,6 +56,9 @@ var kmsArnRE = regexp.MustCompile(
 const (
 	OidcConfigIdFlag      = "oidc-config-id"
 	ClassicOidcConfigFlag = "classic-oidc-config"
+
+	MinReplicasSingleAZ = 2
+	MinReplicaMultiAZ   = 3
 )
 
 var args struct {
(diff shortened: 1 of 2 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -1789,7 +1799,15 @@ func run(cmd *cobra.Command, _ []string) {
 	computeNodes := args.computeNodes
 	// Compute node requirements for multi-AZ clusters are higher
 	if multiAZ && !autoscaling && !isReplicasSet {
-		computeNodes = 3
+		if isHostedCP {
```
gdbranco, line 1802: the behavior for this is the same as in the other changes, right?
Could it be some function that computes amount of min replicas?

ciaranRoche: +1 to this, if we could break it to a function and add a unit test for this functionality :pray:

cmd/create/cluster/cmd.go
```diff
@@ -2795,3 +2796,35 @@ func buildCommand(spec ocm.Spec, operatorRolesPrefix string,
...
+	if !isMinReplicasSet {
+		if isHostedCP {
+			if privateSubnetsCount == 1 {
+				newMinReplicas = 2
+			} else {
+				newMinReplicas = privateSubnetsCount
+			}
+		} else {
+			if multiAZ {
+				newMinReplicas = 3
```
andreadecorte, line 2821: I think you forgot to use the const MinReplicaMultiAZ

cmd/create/cluster/cmd.go
```diff
@@ -2795,3 +2796,35 @@ func buildCommand(spec ocm.Spec, operatorRolesPrefix string,
...
+	isHostedCP bool,
+	multiAZ bool) (int, int) {
+
+	newMinReplicas := minReplicas
+	newMaxReplicas := maxReplicas
+
+	if !isMinReplicasSet {
+		if isHostedCP {
+			if privateSubnetsCount == 1 {
+				newMinReplicas = 2
```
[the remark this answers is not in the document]
lucasponce: +1

cmd/create/cluster/cmd.go
lucasponce: +1

PR #1213, opened 2023-05-01
SDA-8600 | feat: add interactive option for delete account roles

oriAdler:
For the interactive mode, prompt a message for deleting hosted CP account roles.

image

### DIFF
--- a/cmd/dlt/accountroles/cmd.go
+++ b/cmd/dlt/accountroles/cmd.go
@@ -100,6 +100,9 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	deleteClassic, deleteHostedCP := setDeleteRoles(cmd.Flags().Changed("classic"),
+		cmd.Flags().Changed("hosted-cp"))
+
 	clusters, err := r.OCMClient.GetAllClusters(r.Creator)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/dlt/accountroles/cmd.go
```diff
@@ -100,6 +100,14 @@ func run(cmd *cobra.Command, _ []string) {
...
 	}
 
+	var deleteClassic, deleteHostedCP bool
+	// Default flow deletes both classic and hosted-CP roles
+	if !args.classic && !args.hostedCP {
+		deleteClassic, deleteHostedCP = true, true
+	} else {
+		deleteClassic, deleteHostedCP = args.classic, args.hostedCP
+	}
+
```
ciaranRoche, line 110: If we are defaulting to true, can we not set the default to true on init?

oriAdler: We are defaulting to true only if the user didn't provide `--hosted-cp` and `--classic` flags.
Maybe I should change the condition here to check if the flags were provided instead.

oriAdler: Exported to a new function.
I returned the values instead of multiple `if` statements for each combination of the flags, and found it more concise. Let me know your thoughts.

ciaranRoche (commented):
/lgtm

PR #1217, opened 2023-05-02
SDA-8053 | Fix: make IDP password validation errors more readable

willkutler:
ref: https://issues.redhat.com/browse/SDA-8053?filter=-1

### DIFF
--- a/cmd/create/idp/htpasswd.go
+++ b/cmd/create/idp/htpasswd.go
@@ -187,19 +187,35 @@ func usernameValidator(val interface{}) error {
 
 func passwordValidator(val interface{}) error {
 	if password, ok := val.(string); ok {
-		notAsciiOnly, _ := regexp.MatchString(`[^\x20-\x7E]`, password)
+		re := regexp.MustCompile(`[^\x20-\x7E]`)
+		invalidChars := re.FindAllString(password, -1)
+		notAsciiOnly := len(invalidChars) > 0
 		containsSpace := strings.Contains(password, " ")
 		tooShort := len(password) < 14
+		pwdErrors := []string{}
+		if notAsciiOnly {
+			pwdErrors = append(pwdErrors, fmt.Sprintf("must not contain special characters [%s]",
+				strings.Join(invalidChars, ", ")))
+		}
+		if containsSpace {
+			pwdErrors = append(pwdErrors, "must not contain whitespace")
+		}
+		if tooShort {
+			pwdErrors = append(pwdErrors, fmt.Sprintf("must be at least 14 characters (got %d)", len(password)))
+		}
 		if notAsciiOnly || containsSpace || tooShort {
-			return fmt.Errorf(
-				"password must be at least 14 characters (ASCII-standard) without whitespaces")
+			if len(pwdErrors) > 1 {
+				pwdErrors[len(pwdErrors)-1] = "and " + pwdErrors[len(pwdErrors)-1]
+			}
+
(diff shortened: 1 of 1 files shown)


cmd/upgrade/cluster/cmd.go
```diff
@@ -212,8 +212,7 @@ func createUpgradePolicyHypershift(r *rosa.Runtime, cmd *cobra.Command, clusterK
 	cluster *cmv1.Cluster, version string, scheduleDate string, scheduleTime string) error {
 	upgradePolicyBuilder := cmv1.NewControlPlaneUpgradePolicy().ScheduleType("manual").
 		UpgradeType("ControlPlane").Version(version)
-	// TODO: remove forceNow when we support scheduling
-	nextRun := buildUpgradeSchedule(r, cmd, scheduleDate, scheduleTime, true)
+	nextRun := buildUpgradeSchedule(r, cmd, scheduleDate, scheduleTime)
```
gdbranco, line 215: Are these changes supposed to be included in this PR? Seems unrelated

cmd/create/idp/htpasswd.go
```diff
@@ -187,19 +187,39 @@ func usernameValidator(val interface{}) error {
...
+		}
+		if tooShort {
+			pwdErrors = append(pwdErrors, fmt.Sprintf("must be at least 14 characters (got %d)", len(password)))
+		}
+		if len(pwdErrors) == 1 {
+			return fmt.Errorf("Password " + pwdErrors[0])
+		} else if len(pwdErrors) > 1 {
+			msg := "Password "
+			for i := 0; i < len(pwdErrors)-1; i++ {
+				msg += pwdErrors[i] + ", "
```
gdbranco, line 211: Would it be possible to use strings.join instead and possibly drop the condition?

cmd/upgrade/cluster/cmd.go
willkutler: they aren't, just removed them

cmd/create/idp/htpasswd.go
willkutler: @gdbranco thanks for the review! I'd like to keep the condition since I think having the "and" before the last list item makes the message a bit more readable, but I switched the code to use `strings.Join` to make it cleaner.

cmd/create/idp/htpasswd.go
```diff
@@ -187,19 +187,35 @@ func usernameValidator(val interface{}) error {
...
+			pwdErrors = append(pwdErrors, fmt.Sprintf("must be at least 14 characters (got %d)", len(password)))
+		}
 		if notAsciiOnly || containsSpace || tooShort {
-			return fmt.Errorf(
-				"password must be at least 14 characters (ASCII-standard) without whitespaces")
+			if len(pwdErrors) > 1 {
+				pwdErrors[len(pwdErrors)-1] = "and " + pwdErrors[len(pwdErrors)-1]
+			}
+
+			return fmt.Errorf("Password " + strings.Join(pwdErrors, ", "))
```
gdbranco, line 211: Doesn't that outputs an extra ','?

gdbranco:
```suggestion
			output := strings.Join(pwdErrors, ", ")
	                i := strings.LastIndex(output, ",")
	                if i != -1 {
		                output = output[:i] + " and" + output[i+len(","):]
	                }
```

ciaranRoche:
Closing as solved by - https://github.com/openshift/rosa/pull/1475

PR #1226, opened 2023-05-04
Allow editing ingress privacy for private hypershift cluster

zgalor:
Related: [SDA-8109](https://issues.redhat.com//browse/SDA-8109)

### DIFF
--- a/cmd/edit/ingress/cmd.go
+++ b/cmd/edit/ingress/cmd.go
@@ -118,10 +118,15 @@ func run(cmd *cobra.Command, argv []string) {
 		interactive.Enable()
 	}
 
+	cluster := r.FetchCluster()
 	var labelMatch *string
 	if cmd.Flags().Changed(labelMatchFlag) {
+		if ocm.IsHyperShiftCluster(cluster) {
+			r.Reporter.Errorf("Updating route selectors is not supported for hosted cp clusters")
+			os.Exit(1)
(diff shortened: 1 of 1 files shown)


cmd/edit/ingress/cmd.go
```diff
@@ -118,10 +118,15 @@ func run(cmd *cobra.Command, argv []string) {
 		interactive.Enable()
 	}
 
+	cluster := r.FetchCluster()
 	var labelMatch *string
 	if cmd.Flags().Changed(labelMatchFlag) {
+		if ocm.IsHyperShiftCluster(cluster) {
+			r.Reporter.Errorf("Updating route selectors is not supported for hosted cp clusters")
```
gdbranco, line 125: Is this already done in clusters service as well?

zgalor: No. Just in rosa

zgalor: I was wrong. It is done indirectly:

```
	if patch.RouteSelectors != nil {
		if ingress.Default {
			err = errors.BadRequest.UserErrorf(
				"Can't add route selectors to default ingress '%s' in cluster '%s'",
				ingress.ID,
				cluster.ID)
			return
		}
		clauses["route_selectors"] = patch.RouteSelectors
	}
```

PR #1231, opened 2023-05-05
SDA-8963 Improve machine type not found message

lucasponce:
~~Minimal change to improve the wording of the machine type not found error.~~

I've prepared a patch for two similar issues (I see more complexity in splitting the patch, and I think it makes sense to put it together)

https://issues.redhat.com/browse/SDA-8963
https://issues.redhat.com/browse/SDA-9075

This PR will cover the following:
- A minimal refactoring to include the region and the availability zones in the MachineTypeList object.
- Improve the validation message when an instance type is not found in a given AZ.
- Fix a situation where the user may select a new subnet/AZ that is not part of day 1, and in the new AZ, there are different machine types.

Please, @andreadecorte @ciaranRoche @gregsheremeta, feel free to review. 

Thanks

### DIFF
--- a/pkg/ocm/machines.go
+++ b/pkg/ocm/machines.go
@@ -46,7 +46,7 @@ func (c *Client) GetMachineTypesInRegion(cloudProviderData *cmv1.CloudProviderDa
 		}
 
 		response.Items().Each(func(item *cmv1.MachineType) bool {
-			machineTypes = append(machineTypes, &MachineType{
+			machineTypes.Items = append(machineTypes.Items, &MachineType{
 				MachineType: item,
 			})
 			return true
@@ -78,11 +78,11 @@ func (c *Client) GetMachineTypes() (machineTypes MachineTypeList, err error) {
 			if errMsg == "" {
 				errMsg = err.Error()
 			}
-			return nil, errors.New(errMsg)
+			return MachineTypeList{}, errors.New(errMsg)
 		}
 
 		response.Items().Each(func(item *cmv1.MachineType) bool {
-			machineTypes = append(machineTypes, &MachineType{
+			machineTypes.Items = append(machineTypes.Items, &MachineType{
 				MachineType: item,
 			})
 			return true
@@ -137,6 +137,9 @@ func (c *Client) GetAvailableMachineTypesInRegion(region string, availabilityZon
 		return MachineTypeList{}, err
 	}
 
+	machineTypes.Region = region
+	machineTypes.AvailabilityZones = availabilityZones
+
 	quotaCosts, err := c.getQuotaCosts()
 	if err != nil {
 		return MachineTypeList{}, err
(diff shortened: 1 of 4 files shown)


pkg/ocm/machines.go
```diff
@@ -254,7 +254,8 @@ func (mtl *MachineTypeList) ValidateMachineType(machineType string, multiAZ bool
 
 	if v == nil {
 		allMachineTypes := strings.Join(mtl.IDs(), " ")
-		err := fmt.Errorf("A valid machine type number must be specified\nValid machine types: %s", allMachineTypes)
+		err := fmt.Errorf("Machine type '%s' not found in region machine type list: %s",
```
andreadecorte, line 257: What about adding the region there too for even better clarity? Like
```
Machine type '%s' not found in region '%s'. Available machine type list: %s"
```

lucasponce: Good idea, let me see if it's available in the call

gregsheremeta: "not found in region" is slightly misleading. The problem is that the machine type isn't found in the specific availability zone within the region. For example, even though us-east-1 (the region) has lots of m6i machines, us-east-1e (the AZ) might not have any.

lucasponce: Yes, I think we need to pass the AZ

lucasponce: Well, I think we need to refine the message.

But if the user selects a machine type that is available in one AZ in a region but not in another AZ, it should have that info.

At least for this PR.

Basically

gregsheremeta: I like the message that Andrea suggested, with a slight tweak:

`Machine type '%s' not found in availability zone '%s'. Available machine type list: %s"`

That list might be fairly long, though.

pkg/ocm/machines.go
```diff
@@ -254,7 +261,10 @@ func (mtl *MachineTypeList) ValidateMachineType(machineType string, multiAZ bool
 
 	if v == nil {
 		allMachineTypes := strings.Join(mtl.IDs(), " ")
-		err := fmt.Errorf("A valid machine type number must be specified\nValid machine types: %s", allMachineTypes)
+		allAvailabilityZones := strings.Join(mtl.AvailabilityZones, ",")
+		err := fmt.Errorf("Machine type '%s' not found in availability zones '%s' for region: '%s'\n"+
```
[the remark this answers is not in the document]
lucasponce: There can be more than one; I can change the wording, but 

```
allAvailabilityZones := strings.Join(mtl.AvailabilityZones, ",")
```

ciaranRoche (approved):
/lgtm 

Lets get this merged and through QE ASAP :smiley:

pkg/ocm/machines.go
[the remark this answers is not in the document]
ciaranRoche: keep the wording, since there can be multiples

PR #1237, opened 2023-05-08
SDA-9086 | fix: check error when retrieving operator-roles

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-9086

# What
Check aws errors before moving flow forward when deleting roles by prefix

# Why
Was not being checked


ciaranRoche (approved):
/lgtm

PR #1240, opened 2023-05-09
SDA-9087 | fix: validate installer role arn after selection

gdbranco:
Related issue: https://issues.redhat.com/browse/SDA-9087

# What
Validates version of installer role arn

# Why
When specifying the param directly version was not being validated

### DIFF
--- a/cmd/create/oidcconfig/cmd.go
+++ b/cmd/create/oidcconfig/cmd.go
@@ -246,6 +246,16 @@ func run(cmd *cobra.Command, argv []string) {
 					os.Exit(1)
 				}
 			}
+			isValid, err := r.AWSClient.ValidateAccountRoleVersionCompatibility(
+				roleName, aws.InstallerAccountRole, minorVersionForGetSecret)
+			if err != nil {
+				r.Reporter.Errorf("There was a problem listing role tags: %v", err)
+				os.Exit(1)
+			}
+			if !isValid {
+				r.Reporter.Errorf("Role '%s' is not of minimum version '%s'", args.installerRoleArn, minorVersionForGetSecret)
+				os.Exit(1)
+			}
 		}
 
 		args.userPrefix = strings.Trim(args.userPrefix, " \t")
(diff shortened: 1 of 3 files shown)


cmd/create/oidcconfig/cmd.go
```diff
@@ -246,6 +246,16 @@ func run(cmd *cobra.Command, argv []string) {
...
 				}
 			}
+			isValid, err := r.AWSClient.ValidateAccountRoleMinVersionTag(
+				roleName, aws.InstallerAccountRole, minorVersionForGetSecret)
+			if err != nil {
+				r.Reporter.Errorf("There was a problem listing role tags: %v", err)
+				os.Exit(1)
+			}
+			if !isValid {
+				r.Reporter.Errorf("Role '%s' is not of mininum version '%s'", args.installerRoleArn, minorVersionForGetSecret)
```
oriAdler, line 256:
```suggestion
				r.Reporter.Errorf("Role '%s' is not of minimum version '%s'", args.installerRoleArn, minorVersionForGetSecret)
```

pkg/aws/policies.go
```diff
@@ -479,55 +479,69 @@ func (c *awsClient) FindRoleARNs(roleType string, version string) ([]string, err
...
+	}
+	return roleARNs, nil
+}
 
-				if roleHasTag(listRoleTagsOutput.Tags, tags.ManagedPolicies, tags.True) {
-					// Managed policies will be up-to-date no need to check version tags
-					break
-				}
+// FIXME: refactor similar calls to use this instead
+func (c *awsClient) ValidateAccountRoleMinVersionTag(
```
oriAdler, line 495: This function also fetches the tags and checks the role type. I guess you extracted the code to reuse the function. But I wonder if we should change the function name or extract only the part under `case tags.OpenShiftVersion:` Which actually checks the minmum version, WDYT?

oriAdler: I see, `ValidateAccountRoleComptability` or `ValidateAccountRoleVersionComptability`?

PR #1241, opened 2023-05-09
OCM-263 | fix: behavior when creating operator-roles

gdbranco:
Related issue: https://issues.redhat.com/browse/OCM-263
# What
Adjusts behavior when creating operator-roles with `-f` param

# Why
No need to check s3 bucket as we can trust oidc-config spec, if something wrong happens will be caught creating the cluster

When using `-f` it should bypass validation that the operator already exist

### DIFF
--- a/cmd/create/operatorroles/by_clusterkey.go
+++ b/cmd/create/operatorroles/by_clusterkey.go
@@ -41,8 +41,10 @@ func handleOperatorRoleCreationByClusterKey(r *rosa.Runtime, env string,
 		if err != nil {
 			return err
 		}
-		r.Reporter.Infof("Cluster '%s' is using reusable OIDC Config and operator roles already exist.", clusterKey)
-		return nil
+		if !args.forcePolicyCreation {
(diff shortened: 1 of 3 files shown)


cmd/create/operatorroles/by_clusterkey.go
```diff
@@ -36,7 +36,7 @@ func handleOperatorRoleCreationByClusterKey(r *rosa.Runtime, env string,
 		}
 	}
 
-	if ocm.IsOidcConfigReusable(cluster) && len(missingRoles) == 0 {
```
oriAdler, line 39: I'm a little bit confused, the flag says `forcePolicyCreation` but we skip `ValidateOperatorRolesMatchOidcProvider`, and inside `ValidateOperatorRolesMatchOidcProvider` we check if the policy is compatible, but we don't create policies.

Can you please elaborate on that?

gdbranco: In the rush I ended up skipping more than wanted, adjusted so only the message and return are skipped

PR #1245, opened 2023-05-10
OCM-264 | fix: Properly checks the path when validating trusted relationship

gdbranco:
Related issue: https://issues.redhat.com/browse/OCM-264
# What
Properly checks the path when validating trusted relationship

# Why
Customer shouldn't be able to change the trusted relationship of existing roles


pkg/ocm/helpers.go
```diff
@@ -858,3 +859,20 @@ func ValidateOperatorRolesMatchOidcProvider(reporter *reporter.Object, awsClient
 	}
 	return nil
 }
+
+func validateIssuerUrlMatchesAssumePolicyDocument(
```
ciaranRoche, line 863: same as above can we unit test this function please

PR #1252, opened 2023-05-15
OCM-302 | fix: if no such entity move operator roles creation forward

gdbranco:
Related issue: https://issues.redhat.com/browse/OCM-302
# What
Allows to bypass issuer url does not match trusted relationship when first creating the operator roles

# Why
When specifying operator roles by prefix if no roles are found means the customer would like to create them in the first place

### DIFF
--- a/cmd/create/operatorroles/by_prefix.go
+++ b/cmd/create/operatorroles/by_prefix.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"strings"
 
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/iam"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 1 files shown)


cmd/create/operatorroles/by_prefix.go
```diff
@@ -140,8 +142,16 @@ func handleOperatorRoleCreationByPrefix(r *rosa.Runtime, env string,
 	err = ocm.ValidateOperatorRolesMatchOidcProvider(r.Reporter, r.AWSClient,
 		operatorRolesList, oidcConfig.IssuerUrl(), "4.0", path)
 	if err != nil {
-		r.Reporter.Errorf("%v", err)
-		os.Exit(1)
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
```
oriAdler, line 146: Not sure if we should do the error handling in the command file.
Maybe it's better to encapsulate the error handling in the client? and return an additional boolean of `exist`
We can check the error type inside `GetRoleByARN`

gdbranco: It is only this call of `ValidateOperatorRolesMatchOidcProvider` that would be ok to move forward though.

 I don't believe it is worth refactoring all calls of `GetRoleByARN` to have exist or check if they all allow going forward from NoSuchEntity. Also the same reason for not refactoring `ValidateOperatorRolesMatchOidcProvider` to have an extra exists return variable

oriAdler: After some thought, we could add a small function to the client, to check the error type, and call it from here.
Ideally, we should import **"github.com/aws/aws-sdk-go"** only in the aws client package, but I know there are other places in the code we import this package as well.

PR #1254, opened 2023-05-16
OCM-342 | feat: add a default value when selecting oidc configs

gdbranco:
Related issue: https://issues.redhat.com/browse/OCM-342
# What
Adds a default value and if no value is selectable a message indicating to create a new one

# Why
Clearer flow


gdbranco:
/hold

Until we have a new release holding this

PR #1255, opened 2023-05-16
OCM-285 | feat: unhide billing

gdbranco:
Related issue: https://issues.redhat.com/browse/OCM-285
# What
Unhide billing account param and add message informing tech preview when creating hosted cp cluster


cmd/create/cluster/cmd.go
```diff
@@ -715,6 +714,11 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	if isHostedCP && r.Reporter.IsTerminal() {
+		r.Reporter.Infof("A billing account is needed but subscription" +
+			" charges will not be billed, Infrastructures costs still apply.")
```
ciaranRoche, line 719: Should we add something here, that billing is not charged while in tech preview.

gdbranco: Fixed, but this should not be merged for now

PR #1262, opened 2023-05-17
OCM-768 | show default version first

gdbranco:
Related issue: https://issues.redhat.com/browse/OCM-768


ciaranRoche:
@gdbranco just update the commit message and we can merge this in if its still wanted, else we can close it out.

PR #1265, opened 2023-05-17
[SDA-8997] Day 1 audit log forwarding cli support

davidleerh:
Do Not Merge.
A. PR raised for review purposes only.
B. Pending internal review on flag naming conventions.

Purpose:
To add ROSA CLI command that will allow users to enable and add the audit log forwarding (AWS Cloudwatch) role with their new cluster, via flag "--audit-log-role-arn". In addition, implemented user assiting prompts via interactive mode.
image


davidleerh:
Keeping record of previous commit in case a revert is required. Will squash commits once approved.

cmd/create/cluster/cmd.go
```diff
@@ -2209,6 +2219,47 @@ func run(cmd *cobra.Command, _ []string) {
...
+	if auditLogRoleARN != "" && !isHostedCP {
+		r.Reporter.Errorf("Audit log forwarding to AWS CloudWatch is only supported for Hosted Control Plane clusters")
+		os.Exit(1)
+	}
+
+	if interactive.Enabled() && isHostedCP {
+		requestAuditLogForwarding, err := interactive.GetBool(interactive.Input{
+			Question: "Enable audit log forwarding to AWS CloudWatch (optional)",
+			Default:  false,
+			Required: true,
```
lucasponce, line 2234: Is this required (in the question is marked as optional).

cmd/edit/cluster/cmd.go
```diff
@@ -473,6 +484,93 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
+	/*******  AuditLogForwarding *******/
```
lucasponce, line 487: Unify comments style :)

davidleerh: I see a couple examples of the following comment block style being used (within the same file)
image
image

Is this not appropriate for this repo? Please advise :)

cmd/create/cluster/cmd.go
davidleerh: The answer (yes or no) is required for the question prompt.
You are correct that the '(optional)' does not align with the user requirement that they must input a yes or no answer.
Will go ahead and remove the '(optional)' string.

cmd/edit/cluster/cmd.go
```diff
@@ -473,6 +484,93 @@ func run(cmd *cobra.Command, _ []string) {
...
+	if cmd.Flags().Changed("audit-log-arn") {
+		auditLogRoleValue = args.auditLogRoleARN
+		auditLogRole = &auditLogRoleValue
+	}
+	// user did not supply value via flag in cli
+	if interactive.Enabled() && !auditLogRoleExists {
+		requestAuditLogForwarding, err := interactive.GetBool(interactive.Input{
+			Question: auditLogEnableOrUpdatePrompt(cluster.AWS().AuditLog().RoleArn()),
+			Default:  false,
+			Required: true,
```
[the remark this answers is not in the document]
davidleerh: The answer to the audit log arn prompt 'auditLogEnableOrUpdatePrompt' informs that the user needs to input a yes or no answer to the inquiry:
image

If the user says 'YES' to either enable or update an audit log role arn, they are required to input the arn of the role in the next prompt:

image

When using the interactive mode as above, the user is required to input a properly formatted aws role IAM arn and will be validated via the following validator:
```
	aws.ARNValidator,
```
an empty value and/or empty string here will not pass the validation.


It is a different story with the flag option "--audit-log-arn".
Users can pass either a properly formatted arn 'or' an empty string.
An empty string indicates that the user wants to disable the audit log forwarding feature, and passes the request to disable it to CS.

image

cmd/edit/cluster/cmd.go
lucasponce: I tend to prefer 

```
//  AuditLogForwarding
```

Instead of those old classic styles

```
// /*******  AuditLogForwarding *******/
```

Anyway, it's a nitpick comment, not a blocker.

lucasponce:
LGTM

cmd/edit/cluster/cmd.go
```diff
@@ -473,6 +484,93 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
+	/*******  AuditLogForwarding *******/
+	var auditLogRole *string
+	var auditLogRoleValue string
+	auditLogRoleExists := cluster.AWS().AuditLog().RoleArn() != ""
+
+	// user supplied value via flag in cli
+	if cmd.Flags().Changed("audit-log-arn") {
```
gdbranco, line 493: Not checking if param is available for classic

davidleerh:
@gdbranco 
Resolved missing items:
1. rosa create cluster --audit-log-arn (validate arn using iam role regex)
2. Replaced generic arn validator with iam role regex when validating audit log role arn
3. rosa edit cluster --audit-log-arn (returns error msg when user uses flag with classic cluster)
4. rosa edit cluster (audit log forwarding interactive prompts only appear for hypershift clusters)

cmd/edit/cluster/cmd.go
```diff
@@ -473,6 +484,99 @@ func run(cmd *cobra.Command, _ []string) {
...
+			if !aws.AuditLogArnRE.MatchString(auditLogRoleValue) {
+				r.Reporter.Errorf("Expected a valid value for audit-log-arn matching %s", aws.AuditLogArnRE.String())
+				os.Exit(1)
+			}
+			r.Reporter.Warnf("You are editing audit log forwarding with the provided arn '%s'.", auditLogRoleValue)
+			if !confirm.Confirm("edit audit log forwarding for cluster '%s'", clusterKey) {
+				os.Exit(0)
+			}
+			// validation for user supplied value via flag
+		} else if auditLogRoleValue == "" && auditLogRoleExists {
```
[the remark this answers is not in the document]
gdbranco: I don't think we do anything like that for commands today. So because of that I'd say it is overkill. But if this is something that has been worked out with BU then we move it forward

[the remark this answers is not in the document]
davidleerh: I don't believe there was any disagreement from BU side, when I added the screen shots of the reactive prompts (within the Jira Card associated with this PR)

Hi @ciaranRoche @oriAdler 
would like to get consensus / vote on what would be the best approach here to better the user experience as well as meeting the standards for the repo. Please kindly advise.  :slightly_smiling_face:

cmd/create/cluster/cmd.go
```diff
@@ -2209,6 +2219,52 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	// Audit Log Forwarding
+	auditLogRoleARN := args.AuditLogRoleARN
+
+	if auditLogRoleARN != "" && !isHostedCP {
```
oriAdler, line 2225: NP
I wonder if we can check that earlier, to give faster feedback.

cmd/edit/cluster/cmd.go
```diff
@@ -473,6 +484,99 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
+	/*******  AuditLogForwarding *******/
+	var auditLogRole *string
+	var auditLogRoleValue string
+	auditLogRoleExists := cluster.AWS().AuditLog().RoleArn() != ""
+
+	// user supplied value via flag in cli
+	if cmd.Flags().Changed("audit-log-arn") {
```
oriAdler, line 493: Can we please extract this block to a new function? 
Something like `SetAuditLogForwarding`.

cmd/edit/cluster/cmd.go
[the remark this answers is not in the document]
oriAdler: If we can somehow break it into smaller functions, or rearrange the conditions, maybe a `switch` statement? I don't know.
That could be great.
I have no doubt the user experience here is good, but I agree with Guilherme there are a lot of `else if` conditions here.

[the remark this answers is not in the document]
ciaranRoche: Yeah i agree with both Ori and Guilherme here. If we can flatten this out, potentially abstract to smaller functions and have unit tests on it to make sure we do not break it in the future.

cmd/create/cluster/cmd.go
davidleerh: Not sure where the best area / placement should be here :slightly_frowning_face: 
I personally thought it was best to group this check alongside other code statements related to audit log forwarding.

cmd/edit/cluster/cmd.go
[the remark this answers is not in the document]
davidleerh: Thank you all for the review:
I went ahead and refactored the code to reduce multiple if else if statements.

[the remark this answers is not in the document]
davidleerh: image
First statement covers case when user explicitly defines the flag / arg value for audit log role arn
The if statement following, covers case when user did not define a value and must go through the interactive mode

cmd/edit/cluster/cmd.go
```diff
@@ -551,3 +572,103 @@ func parseRFC3339(s string) (time.Time, error) {
...
+	}
+	return fmt.Sprintf("Update existing audit log forwarding role '%s'", currentAuditLogRole)
+}
+
+func setAuditLogForwarding(r *rosa.Runtime, cmd *cobra.Command, cluster *cmv1.Cluster, auditLogArn string) (
+	argValuePtr *string) {
+	if cmd.Flags().Changed("audit-log-arn") {
+		if !aws.IsHostedCP(cluster) {
+			r.Reporter.Errorf("Audit log forwarding to AWS CloudWatch is only supported for Hosted Control Plane clusters")
+			os.Exit(1)
```
oriAdler, line 592: NP
I think in general it's better to return an error and exit from the main function.

oriAdler (commented):

cmd/edit/cluster/cmd.go
```diff
@@ -552,3 +577,102 @@ func parseRFC3339(s string) (time.Time, error) {
...
+
+	if !auditLogRoleExists(cluster) {
+		requestAuditLogForwarding, err := interactive.GetBool(interactive.Input{
+			Question: auditLogEnableOrUpdatePrompt(cluster.AWS().AuditLog().RoleArn()),
+			Default:  false,
+			Required: true,
+		})
+		if err != nil {
+			r.Reporter.Errorf("Expected a valid value: %s", err)
+			os.Exit(1)
```
oriAdler, line 639: NP
We can return an error here as well instead of exiting the program.

davidleerh: Thank you for the find:)

PR #1269, opened 2023-05-18
Revert "OCM-140 | chore : reverting IMDSv2"

ciaranRoche:
Reverts openshift/rosa#1257


renan-campos:
We revert the revert here? :)

PR #1272, opened 2023-05-23
SDA-7710 | fix: validate against tags used in configuration

robdolares:
Related Issue: https://issues.redhat.com/browse/SDA-7710

# What
When running `rosa create cluster`, tag validation should throw error for following tag keys:
-  `kubernetes.io`
- `k8s.io`
- `openshift.io`

# Why
These values are reserved for configuration tags used in internal processes.

### DIFF
--- a/pkg/aws/helpers.go
+++ b/pkg/aws/helpers.go
@@ -38,11 +38,15 @@ var UserTagValueRE = regexp.MustCompile(`^[\pL\pZ\pN_.:/=+\-@]{0,256}$`)
 // first pattern is to validate IPv4 address
 // second,is for IPv4 CIDR range validation
 // third pattern is to validate domains
-// and the fifth petterrn is to be able to remove the existing no-proxy value by typing empty string ("").
+// and the fifth pattern is to be able to remove the existing no-proxy value by typing empty string ("").
 // nolint
 var UserNoProxyRE = regexp.MustCompile(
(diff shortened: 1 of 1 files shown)


oriAdler:
LGTM, can we please add unit testing for `UserTagValidator`?

pkg/aws/helpers.go
```diff
@@ -249,6 +253,11 @@ func UserTagValidator(input interface{}) error {
 				return fmt.Errorf("invalid tag format, Tags are comma separated, for example: 'foo:bar,bar:baz'")
 			}
 			tag := strings.Split(t, ":")
+
+			if UserTagReservedKeysRE.MatchString(strings.ToLower(tag[0])) {
```
ciaranRoche, line 257: I think this would be better if we moved it to the backend. 

If you have a look at - https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/cmd/clusters-service/servecmd/apiserver/validation_helpers.go#L1941

We can also extend this test to capture this new logic there - https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/cmd/clusters-service/servecmd/apiserver/validation_helpers_test.go#L3136

ciaranRoche: WDYT?

robdolares: @ciaranRoche 
Would moving to backend be in addition to keeping in rosa-cli?  So that there is immediate error response in interactive mode?

ciaranRoche: As long as the tags can be parsed into a key value pair, i dont think we need to add additional validations as a immediate error response, this can be handled in the backend and returned to the user. 

If the need for immediate response comes up, we will need to create a third party package to share this logic between CS and ROSA. To avoid maintaining the same logic in both the client and the service.

gdbranco, line 257: Would it be better to use insensitive check on the regex instead by adding `(?i)` to the start of the regex?

robdolares:
@gdbranco @ciaranRoche 
Moved to CS - https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/5857

robdolares:
@gdbranco @ciaranRoche 
Moved to CS - https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/5857

PR #1273, opened 2023-05-23
[SDA-9172] Update worker IAM role suffix for HCP Only (*-ROSA-Worker)

davidleerh:


davidleerh:
A. If HCP Account roles with old worker IAM role name is used, it will prompt users to delete and recreate new account roles for use.
image

B. Manual test results using updated worker IAM role with installer policy defined: 

Please note, there is a validation that checks if the correct policy is tied to the installer role here
https://github.com/openshift/rosa/blob/master/pkg/helper/roles/helpers.go#L98-L99
In order to test, this validation needed to be skipped temporarily and installer role created via "rosa create account-roles --hosted-cp" needed to be attached manually to a new / manually created permission policy matching the contents defined in JSON:
https://github.com/openshift/managed-cluster-config/blob/master/resources/sts/4.12/hypershift/sts_hcp_installer_permission_policy.json#L36C4-L38

image

cmd/create/accountroles/creators.go
```diff
@@ -303,3 +303,10 @@ func buildAttachRolePolicyCommand(accRoleName string, policyARN string) string {
 		AddParam(awscb.PolicyArn, policyARN).
 		Build()
 }
+
+func getAccountRoleName(prefix string, roleName string) string {
```
oriAdler, line 307: Not sure if this will be sufficient, as we are iterating over this map `aws.HCPAccountRoles` in other commands as well: `upgrade account roles`, `create cluster`, and `delete account roles`.

I was thinking of another approach to changing the value here https://github.com/openshift/rosa/blob/master/pkg/aws/policies.go#L103.

Also started this thread https://redhat-internal.slack.com/archives/C03SZLX3A10/p1684916329073869 to check if we can keep the `-Role` suffix to simplify flow, and be consistent with the other flows.

cmd/create/cluster/cmd.go
```diff
@@ -964,15 +966,28 @@ func run(cmd *cobra.Command, _ []string) {
...
 					resourceId, err := aws.GetResourceIdFromARN(rARN)
 					if err != nil {
 						r.Reporter.Errorf("Failed to get resource ID from arn. %s", err)
 						os.Exit(1)
 					}
 					lowerCaseResourceIdToCheck := strings.ToLower(resourceId)
-					if lowerCaseResourceIdToCheck == expectedResourceIDForAccRole {
+
+					// Searches for outdated worker role and prompts users to delete and recreate
+					if lowerCaseResourceIdToCheck == getOutdatedWorkerRole(isHostedCP, rolePrefix) {
```
oriAdler, line 978: I'm not sure if we need this, creating roles with the `-HCP` infix is currently only available for QE and DEV.
Users creating roles during the public preview are using classic account roles without the `-HCP` prefix.

oriAdler (commented):
I hope we can simplify the changes, I started a thread with SRE to check if we can have the `-Role` suffix for the worker role as well.

cmd/create/cluster/cmd.go
davidleerh: hi @oriAdler 
For clarification:
Is it expected that users, after the public preview, will create hcp account roles with the `-HCP` infix in the account role names? (same as QE and DEV)

oriAdler: Yes, for GA we are expecting it will be available for all users.
The user won't set the `-HCP` explicitly, but by choosing `create account-roles --hosted-cp` they will get a set with `-HCP` infix.

davidleerh:
Closing PR and re-created with: 
https://github.com/openshift/rosa/pull/1281

PR #1280, opened 2023-05-30
OCM-2103 | feat: add 'in use' column list oidc providers

gdbranco:
Related issue: https://issues.redhat.com/browse/OCM-2103

### DIFF
--- a/pkg/aws/helpers.go
+++ b/pkg/aws/helpers.go
@@ -781,6 +781,26 @@ func GetResourceIdFromARN(stringARN string) (string, error) {
 	return parsedARN.Resource[index+1:], nil
 }
 
+func GetResourceIdFromOidcProviderARN(stringARN string) (string, error) {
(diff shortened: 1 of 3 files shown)


pkg/aws/helpers.go
```diff
@@ -781,6 +781,26 @@ func GetResourceIdFromARN(stringARN string) (string, error) {
 	return parsedARN.Resource[index+1:], nil
 }
 
+func GetResourceIdFromOidcProviderARN(stringARN string) (string, error) {
```
oriAdler, line 784: Can we re-use https://github.com/openshift/rosa/blob/master/pkg/aws/helpers.go#L764?

gdbranco: Not really, that one takes the last index and the new one takes the first.

I'm going to refactor this based on the service of the parsed arn

PR #1282, opened 2023-05-31
OCM-1361 | fix: Add flag to bypass confirmation on edit cluster

cameronmwall:
JIRA: https://issues.redhat.com/browse/OCM-1361

This PR adds functionality to the edit cluster command to accept a "-y" flag to bypass confirmation on certain changes. Editing proxy values remains unchanged as the confirmation question was never asked on these commands

### DIFF
--- a/cmd/edit/cluster/cmd.go
+++ b/cmd/edit/cluster/cmd.go
@@ -234,7 +234,7 @@ func run(cmd *cobra.Command, _ []string) {
 		private = &privateValue
 	} else if privateValue {
 		r.Reporter.Warnf("You are choosing to make your cluster API private. %s", privateWarning)
-		if !confirm.Yes() && !confirm.Confirm("set cluster '%s' as private", clusterKey) {
+		if !confirm.Confirm("set cluster '%s' as private", clusterKey) {
 			os.Exit(0)
 		}
 	}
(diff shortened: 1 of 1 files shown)


cmd/edit/cluster/cmd.go
```diff
@@ -233,7 +234,7 @@ func run(cmd *cobra.Command, _ []string) {
 		private = &privateValue
 	} else if privateValue {
 		r.Reporter.Warnf("You are choosing to make your cluster API private. %s", privateWarning)
-		if !confirm.Confirm("set cluster '%s' as private", clusterKey) {
+		if !confirm.Yes() && !confirm.Confirm("set cluster '%s' as private", clusterKey) {
```
gdbranco, line 237: I think we only needed to add the flag due to this

https://github.com/openshift/rosa/blob/510a8c7e2c25b85a2ec5e30be3c0deeaa5d44ac9/pkg/interactive/confirm/confirm.go#L50

Would you have some time to test removing the confirm.yes?

cameronmwall: After more exploration, I think you are right. Because the "confirm.confirm" returns yes if the flag has been passed, we do not seem to need the "confirm.yes" check. I have tested after removing this check and the confirmation question is still not being asked

PR #1283, opened 2023-05-31
OCM-842|Feat: Accept multiple users for htpasswd IDP

ckandag:
-- Currently, ROSACLI for htpasswd idp, accepts only one username and password to add to the IDP 
-- the changes in this PR allow user to input multiple users at a time through the -users flag 
-- the old flags of username/password will continue to be supported for backward comptability 
-- and so as not to break existing automation

Refers to https://issues.redhat.com/browse/OCM-842

### DIFF
--- a/cmd/create/idp/cmd.go
+++ b/cmd/create/idp/cmd.go
@@ -78,6 +78,7 @@ var args struct {
 	// HTPasswd
 	htpasswdUsername string
 	htpasswdPassword string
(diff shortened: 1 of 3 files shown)


gdbranco:
LGTM, would like a second opinion @oriAdler would you have some time too?

renan-campos (approved):
\lgtm

PR #1284, opened 2023-05-31
OCM-1410 | fix: correct error message when 'rosa link user-role' with account-id under diff org

robdolares:
Related Issue: https://issues.redhat.com/browse/OCM-1410


thomasmckay:
lgtm

PR #1287, opened 2023-06-01
fix/filter-subnets-by-cidr

cameronmwall:
JIRA: https://issues.redhat.com/browse/OCM-1227

Two changes have been made. A filter has been added so that only the subnets which fall within the range of the machine CIDR will be shown as options. Additionally, the prompt for machine CIDR has been moved above that of the subnets so that the comparison can be made.

The prompts for the other two CIDR's have also been moved so that they follow sequentially

If special attention could be paid to this reordering, that would be helpful.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -31,6 +31,7 @@ import (
 	mpHelpers "github.com/openshift/rosa/pkg/helper/machinepools"
 	"github.com/spf13/cobra"
 
+	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/openshift/rosa/cmd/create/oidcprovider"
 	"github.com/openshift/rosa/cmd/create/operatorroles"
 	clusterdescribe "github.com/openshift/rosa/cmd/describe/cluster"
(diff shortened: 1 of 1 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -1552,16 +1610,39 @@ func run(cmd *cobra.Command, _ []string) {
...
+			}
+
+			if machineNetwork.Contains(subnetIP) {
+				subnets = append(subnets, subnet)
+			} else {
+				exclude = true
+			}
+		}
 
+		if exclude {
```
gdbranco, line 1643: This could also be len(subnets) != len(initialSubnets), so less branch prediction is required :thinking:

cmd/create/cluster/cmd.go
```diff
@@ -1552,16 +1610,41 @@ func run(cmd *cobra.Command, _ []string) {
...
+			os.Exit(1)
+		}
+		for _, subnet := range initialSubnets {
+			subnetIP, _, err := net.ParseCIDR(*subnet.CidrBlock)
+			if err != nil {
+				r.Reporter.Errorf("Unable to parse subnet CIDR")
+				os.Exit(1)
+			}
+
+			if machineNetwork.Contains(subnetIP) && !serviceNetwork.Contains(subnetIP) {
```
gdbranco, line 1640: For the service we check both sides

Intersect(n1, n2 *net.IPNet) bool {
	return n2.Contains(n1.IP) || n1.Contains(n2.IP)
}

gdbranco:
LGTM
Could you squash the commits?

cameronmwall:
> LGTM Could you squash the commits?

done

PR #1295, opened 2023-06-08
OCM-152 | feat: 'rosa verify network' command

thomasmckay:
[OCM-152](https://issues.redhat.com//browse/OCM-152)

https://docs.google.com/document/d/1pLN83t5kh7Pv0iBtJFFQnIQvdpagG4WZIyel9YRFE3s/edit#

Note: The `go mod vendor` commit will be removed. Dependent on:
https://github.com/openshift-online/ocm-api-model/pull/757
https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/5884

### DIFF
--- a/cmd/verify/network/cmd.go
+++ b/cmd/verify/network/cmd.go
@@ -1,0 +1,234 @@
+/*
+Copyright (c) 2023 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package network
+
+import (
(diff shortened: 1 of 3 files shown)


thomasmckay:
@willkutler Removing `--status-only` would imply removing the `--watch` too which means it would always be "interactive". I haven't used rosa enough to know users' patterns. Would it be sufficient to just always block and wait when the command is run?

willkutler:
@thomasmckay Hmm, thinking it about it more, it's probably better to not require users to be blocked and wait for a minute-plus long command to run. I think it should be good as-is

cmd/verify/network/cmd.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+
+		region, err := aws.GetRegion(arguments.GetRegion())
+		if err != nil {
+			r.Reporter.Errorf("Error getting region: %v", err)
+			os.Exit(1)
+		}
+		args.region = region
+	} else if cmd.Flags().Changed(clusterFlag) {
+		if cmd.Flags().Changed(subnetIDsFlag) {
```
gdbranco, line 124: Couldn't this check and the one above be merged into a single condition which checks if both have changed?

gdbranco: I don't mind that much, but personally find it cleaner

renan-campos (approved):
/lgtm

PR #1296, opened 2023-06-08
[OCM-942] ROSA CLI support for configurable ingress load balancer type

srampal:
ROSA CLI support for adding/ modifying ingress load balancer type


srampal:
cc @gdbranco PTAL also

cmd/create/ingress/cmd.go
```diff
@@ -139,6 +158,41 @@ func run(cmd *cobra.Command, _ []string) {
 	if len(routeSelectors) > 0 {
 		ingressBuilder = ingressBuilder.RouteSelectors(routeSelectors)
 	}
+
+	var lbType *string
+	if cmd.Flags().Changed("lb-type") {
+		lbType = &args.lbType
```
gdbranco, line 164: Nitpick: Don't need to care if it is changed, just set it to args.lbType and if it is interactive it should go through the next part anyways.

cmd/edit/ingress/cmd.go
```diff
@@ -155,6 +175,28 @@ func run(cmd *cobra.Command, argv []string) {
 		private = &privArg
 	}
 
+	var lbType *string
+	if cmd.Flags().Changed(lbTypeFlag) {
```
gdbranco, line 179: Same NP

cmd/edit/ingress/cmd.go
```diff
@@ -155,6 +175,28 @@ func run(cmd *cobra.Command, argv []string) {
...
+		lbType = &args.lbType
+	} else {
+		if interactive.Enabled() {
```
gdbranco, line 182: Can we put this in https://github.com/openshift/rosa/blob/9a4f60c1dd9fbae4e6f1a41649e2c2ed1142e857/pkg/interactive/helper.go#L103-L104?

We are trying to have the interactive helper for new interactive getters which are reused for more commands

cmd/edit/ingress/cmd.go
srampal: Using the Changed() flag is the pattern used for the other two parameters in the same function (see above these lines) so I retained the same pattern. Should I change it ? and just for this parameter ?

gdbranco: Oh no, if that's the case let's keep it that way

srampal: Also, another issue I found is that if we don't check the flag change, then even if the user provides the desired lbType as a flag and has the interactive flag also set, then it ignores the command line flag which is not good. So I think it may be best to retain the Changed() flag for both create and edit functions and that seems to work better in testing.

cmd/edit/ingress/cmd.go
srampal: I understand but again, we should double check before making such a change. I am using the same pattern used for other parameters of Ingress. If I just move the load balancer type to another file, but other parameters of Ingress stay in the current files, it may not be clean.

PR #1299, opened 2023-06-11
OCM-2259 | feat: Expose `--hosted-cp` flag

oriAdler:
AWS published all the managed policies, therefore, exposing the flag.

Related: [OCM-2259](https://issues.redhat.com/browse/OCM-2259)


oriAdler:
> I noticed in some of these commands we have also hidden `classic` should we be exposing them along side of `hosted-cp` or is there a card tracking this?

Good catch! I missed that, let me open a PR to take care of that.

ciaranRoche (commented):
/lgtm

PR #1300, opened 2023-06-11
OCM-2258 | fix: Adjust error message for HCP account roles

oriAdler:
Mention the `--hosted-cp` flag.

Related: [OCM-2258](https://issues.redhat.com/browse/OCM-2258)


cmd/upgrade/accountroles/cmd.go
```diff
@@ -135,7 +135,12 @@ func run(cmd *cobra.Command, argv []string) error {
 
 	roleARN, err := awsClient.GetAccountRoleARN(prefix, role.Name)
 	if err != nil {
-		r.Reporter.Errorf("Failed to get account roles ARN: %v", err)
+		if args.hostedCP {
+			r.Reporter.Errorf("Failed to get hosted CP account roles ARN: %v", err)
+		} else {
+			r.Reporter.Errorf("Failed to get classic account roles ARN: %v. "+
+				"To upgrade hosted CP account roles use the '--hosted-cp' flag", err)
```
ciaranRoche, line 142: is this correct, for classic account roles?

oriAdler: The assumption was that the user want to upgrade HCP roles, but forgets the flag, and then we guide them.
Do you find it confusing? Maybe I can improve the wording here (it is possible they aren't aware of HCP at all)

gdbranco: I wonder if we should alert both sides then

PR #1301, opened 2023-06-12
OCM-23 | feat: Removing default pool from rosa cli

tsorya:
[OCM-23](https://issues.redhat.com//browse/OCM-23): Removing default pool from rosa cli
Worker pool is added to ocm and it will match cluster nodes values, we should not show default pool anymore


renan-campos (approved):
/lgtm good to merge once the backend changes have been made.

ciaranRoche (commented):
/lgtm

PR #1307, opened 2023-06-14
OCM-2180 | Fix: hide available upgrades for HCP in list versions

andreadecorte:
The target is to transition to cluster based updates recommandations,
which means no more available upgrades for a single version, only for a specific cluster.
So the rosa list version --hosted-cp should not show anymore the available upgrades,
as in any case the current list is not correct (also showing versions which are disabled for HCP).
The user is informed that it should use rosa list upgrades -c clusterID instead
No change for rosa classic versions.

CC @wgordon17 

Related: [OCM-2180](https://issues.redhat.com//browse/OCM-2180)

### DIFF
--- a/cmd/list/version/cmd.go
+++ b/cmd/list/version/cmd.go
@@ -25,6 +25,7 @@ import (
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 
+	"github.com/openshift/rosa/cmd/list/upgrade"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/output"
 	"github.com/openshift/rosa/pkg/rosa"
@@ -64,6 +65,7 @@ func init() {
 func run(cmd *cobra.Command, _ []string) {
 	r := rosa.NewRuntime().WithOCM()
 	defer r.Cleanup()
+	isHostedCp := args.hostedCp
 
 	// Try to find the cluster:
 	r.Reporter.Debugf("Fetching versions")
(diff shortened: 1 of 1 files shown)


wgordon17:
@andreadecorte Just confirming, so this PR will _remove_ the version list if a customer runs _specifically_ `rosa list versions --hosted-cp`, correct?

Do we have any other way for a customer to see what upgrades _are_ available to them? Like a `rosa list versions -c <cluster-name>`?

andreadecorte:
> @andreadecorte Just confirming, so this PR will _remove_ the version list if a customer runs _specifically_ `rosa list versions --hosted-cp`, correct?

Correct.
> 
> Do we have any other way for a customer to see what upgrades _are_ available to them? Like a `rosa list versions -c <cluster-name>`?

We have a cluster aware command which is `rosa list upgrades -c <cluster>`.
The idea is that user uses that one (coupled with --control-plane/--machine-pool flags, not implemented yet) to see the list of available upgrades for the specific cluster.

wgordon17:
That's excellent, thanks @andreadecorte!

Could we, for the specific use case of `rosa list versions --hosted-cp`, provide some kind of message or pointer to the `rosa list upgrades` option? Otherwise, I worry customers may get lost

andreadecorte:
Added an info message:
```
rosa list version --channel-group=stable --hosted-cp
I: To list available upgrades, please use 'rosa list upgrades'
VERSION    DEFAULT
4.13.2     no
```

andreadecorte:
@wgordon17 any other comments after the changes in https://github.com/openshift/rosa/pull/1307#issuecomment-1597212843 or can we proceed with this change?

wgordon17:
+1 lgtm

PR #1308, opened 2023-06-14
OCM-2347: Accept path setting thru iam role check - regex

davidleerh:
Description of problem:
The AuditLogArnRE (regex) should allow the iam role arn with path setting and pass the validation.

BEFORE FIX
Example that passes:
arn:aws:iam::000000000000:role/asdf-audit-r
Example that fails:
arn:aws:iam::000000000000:role/aa/asd/asdf-audit-r

AFTER FIX
Example that passes:
arn:aws:iam::000000000000:role/asdf-audit-r
arn:aws:iam::000000000000:role/aa/asd/asdf-audit-r

### DIFF
--- a/pkg/aws/helpers.go
+++ b/pkg/aws/helpers.go
@@ -31,7 +31,7 @@ import (
 var RoleNameRE = regexp.MustCompile(`^[\w+=,.@-]+$`)
 
 var AuditLogArnRE = regexp.MustCompile(
-	`^arn:aws:iam::\d{12}:role/[a-zA-Z0-9_-]+$`,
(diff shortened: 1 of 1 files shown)


pkg/aws/helpers.go
```diff
@@ -31,7 +31,7 @@ import (
 var RoleNameRE = regexp.MustCompile(`^[\w+=,.@-]+$`)
 
 var AuditLogArnRE = regexp.MustCompile(
-	`^arn:aws:iam::\d{12}:role/[a-zA-Z0-9_-]+$`,
+	`^arn:aws:iam::\d{12}:role(?:\/[\w+=,.@-]+)*(?:\/[a-zA-Z0-9_-]+)$`,
```
ciaranRoche, line 34: Can we document what this is exactly checking. Ideally a link to a spec or schema would be great. Just for the next engineer.

PR #1315, opened 2023-06-16
OCM 2112 | feat: add -c flag for listing operator-roles

robdolares:
Related Issue: https://issues.redhat.com/browse/OCM-2112

### DIFF
--- a/cmd/list/operatorroles/cmd.go
+++ b/cmd/list/operatorroles/cmd.go
@@ -35,6 +35,7 @@ import (
 var args struct {
 	version string
 	prefix  string
+	cluster string
 }
 
 var Cmd = &cobra.Command{
@@ -68,6 +69,7 @@ func init() {
 		"List only operator-roles that are associated with the given prefix."+
 			" The prefix must match up to openshift|kube-system",
 	)
+	ocm.AddOptionalClusterFlag(Cmd)
 	output.AddFlag(Cmd)
 }
 
@@ -97,7 +99,13 @@ func run(cmd *cobra.Command, _ []string) {
 		spin.Start()
 	}
 
-	operatorsMap, err := r.AWSClient.ListOperatorRoles(args.version)
+	clusterId := ""
+	if cmd.Flags().Changed("cluster") {
+		clusterKey := r.GetClusterKey()
+		clusterId = clusterKey
+	}
+
+	operatorsMap, err := r.AWSClient.ListOperatorRoles(args.version, clusterId)
 
 	if spin != nil {
 		spin.Stop()
@@ -135,6 +143,13 @@ func run(cmd *cobra.Command, _ []string) {
 
 	// Create the writer that will be used to print the tabulated results:
 	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
+	if clusterId != "" {
+		for key, value := range operatorsMap {
+			if value[0].ClusterID == clusterId {
+				args.prefix = key
+			}
+		}
+	}
 	if args.prefix == "" {
 		fmt.Fprintf(writer, "ROLE PREFIX\tAMOUNT IN BUNDLE\n")
 		keys := []string{}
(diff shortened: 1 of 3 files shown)


pkg/aws/policies.go
```diff
@@ -853,14 +854,23 @@ func (c *awsClient) ListOperatorRoles(version string) (map[string][]Role, error)
...
+						operatorRole.ClusterID = tagValue 
 				}
-			}
 			if isTagged && !skip {
 				operatorRole.RoleName = aws.StringValue(role.RoleName)
 				operatorRole.RoleARN = aws.StringValue(role.Arn)
 				operatorMap[foundPrefix] = append(operatorMap[foundPrefix], operatorRole)
 			}
 		}
+		}
```
gdbranco, line 873: By doing this you moved `if isTagged && !skip {` this condition to within the loop, that's why it's got duplicates in the map now

pkg/aws/policies.go
```diff
@@ -853,14 +854,23 @@ func (c *awsClient) ListOperatorRoles(version string) (map[string][]Role, error)
 					}
 					isTagged = true
 					operatorRole.Version = tagValue
+				
+				case tags.ClusterID:
```
gdbranco, line 858: This actually needs to be on line 826, after the ManagedPolicies tag check, that's because the cluster ID tag is in the role and not the policy

cmd/list/operatorroles/cmd.go
```diff
@@ -173,17 +189,18 @@ func run(cmd *cobra.Command, _ []string) {
...
 			if operatorRole.ManagedPolicy {
 				awsManaged = "Yes"
 			}
 			fmt.Fprintf(
 				writer,
-				"%s\t%s\t%s\t%s\n",
+				"%s\t%s\t%s\t%s\t%s\n",
 				operatorRole.RoleName,
 				operatorRole.RoleARN,
+				clusterId,
```
gdbranco, line 203: This should be operatorRole.ClusterID, right?

cmd/list/operatorroles/cmd.go
```diff
@@ -135,6 +144,13 @@ func run(cmd *cobra.Command, _ []string) {
 
 	// Create the writer that will be used to print the tabulated results:
 	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
+	if clusterId != "" {
+		for key := range operatorsMap {
```
gdbranco, line 148:
```suggestion
		for key, value := range operatorsMap {
		       if value[0].ClusterID == clusterId {
				args.prefix = key
			}
		}
```

cmd/list/operatorroles/cmd.go
robdolares: An empty field was returned when using operatorRole.ClusterID - The above returned the clusterId so maybe I am missing something elsewhere?

gdbranco: It is due to this https://github.com/openshift/rosa/pull/1315#discussion_r1236859168

If you put the clusterId variable the user supplied all roles will have the same ID, which wouldn't make a lot of sense

robdolares:
@gdbranco - PTAL

gdbranco:
@renan-campos @oriAdler I believe my opinion would be biased as I've worked a bit with @robdolares on this. Could you also review?

pkg/aws/policies.go
```diff
@@ -813,14 +814,27 @@ func (c *awsClient) ListOperatorRoles(version string) (map[string][]Role, error)
...
 			return operatorMap, err
 		}
+		skip := false
 		for _, tag := range listRoleTagsOutput.Tags {
 			switch aws.StringValue(tag.Key) {
 			case tags.ManagedPolicies:
 				if aws.StringValue(tag.Value) == tags.True {
 					operatorRole.ManagedPolicy = true
 				}
+			case tags.ClusterID:
```
oriAdler, line 824: In case an operator role is missing the cluster tag, should we skip it?
Otherwise, it will be displayed among the other operator roles.

gdbranco: Most operator roles will not have the tag anymore due to reusability from oidc configs. But they still should show up if you are listing all

oriAdler: In case the user provides the `-c` flag, do we want to display only the operators that are associated with the cluster?

gdbranco: Yes, correct

oriAdler: I think in that case if we set `skip := false`, and an operator role won't have the cluster tag, we won't get into this case `tags.ClusterID` and will add the operator to the list. So it will be displayed to the user, but not associated with the cluster.

gdbranco: This adds it to the map, but then https://github.com/openshift/rosa/pull/1315/files/262d693de4c85497097136a3c342884841b53826#diff-03b84a3ca830edbf2664b62deb2a277a56f599670638d5dc179467f926c92ca1R149 this retrieves only the prefix that matches the correct cluster ID

gdbranco: But you are correct, if we ever want to use that function outside of list/operator-roles it wouldn't work just by itself

PR #1319, opened 2023-06-20
OCM-2257 | fix: Ensure only one of users or username flags is provided

ckandag:
REF: https://issues.redhat.com/browse/OCM-2257

Remove  flags  username/password from help/usage msg as they are only for backwards compatibility.
Validate users and username flags are mutually exclusive


ciaranRoche (approved):
/lgtm

PR #1320, opened 2023-06-20
OCM-2433 | feat: adds admin flag for fedramp

tonytheleg:
REF: [OCM-2433](https://issues.redhat.com/browse/OCM-2433) and [OSD-16451](https://issues.redhat.com/browse/OSD-16451)

* adds new --admin flag that overrides aliases for FedRAMP for the SREP Cognito endpoints (hidden flag)
* adds new login/url aliases for SREP endpoints for FedRAMP
* updates all related logic to check for this flag and set the correct aliases accordingly

### DIFF
--- a/cmd/login/cmd.go
+++ b/cmd/login/cmd.go
@@ -157,7 +157,11 @@ func run(cmd *cobra.Command, argv []string) {
 		if env == sdk.DefaultURL {
 			env = "production"
 		}
-		uiTokenPage = fedramp.LoginURLs[env]
+		if fedramp.HasAdminFlag(cmd) {
+			uiTokenPage = fedramp.AdminLoginURLs[env]
+		} else {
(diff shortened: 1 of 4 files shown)


tonytheleg:
@vkareh new PR to fix and comply with contributors requirements. Pending full validation in FedRAMP.

vkareh:
@tonytheleg - I've pushed changes to this that allow it to run other commands that require knowledge of the environment it connects to. Please let me know if this works or not.

tonytheleg:
@vkareh I've tested the CLI for both the current domain and the new admin domain and both work for all command, including ocm and backplane commands. We should be good to go! Thanks for knocking the rest of it out!

markturansky:
is this mergeworthy?  I see Tony tested it.

PR #1321, opened 2023-06-21
OCM-1150 | fix: Validation for 'value' in tags when creating a non-sts cluster

hunterkepley:
[OCM-1150](https://issues.redhat.com/browse/OCM-1150)
Tags previously had no validation for a value, leaving them susceptible to being empty upon cluster creation. This leads to an error when creating the cluster, and the whole interactive process must be redone to try again.

This validation checks if the substring after ":" is empty. If it is, throw an error and ask for tags again.

PS if someone could let me know if this is a proper way of handling, or if I should log an error then keep going/ignore the tags? I feel like it's best to re-try, but this might affect scripts to create clusters if they have invalid tags in them.

### DIFF
--- a/pkg/aws/helpers.go
+++ b/pkg/aws/helpers.go
@@ -258,6 +258,9 @@ func UserTagValidator(input interface{}) error {
 			if len(tag) != 2 {
 				return fmt.Errorf("invalid tag format. Expected tag format: 'key:value'")
 			}
+			if tag[1] == "" {
+				return fmt.Errorf("invalid tag format, Tags must have values. Expected tag format: 'key:value'")
+			}
 			if !UserTagKeyRE.MatchString(tag[0]) {
 				return fmt.Errorf("expected a valid user tag key '%s' matching %s", tag[0], UserTagKeyRE.String())
 			}


hunterkepley:
Upon rethinking this, i dont think this will cause issues with user scripts to create clusters. The original bug caused an error later in the process, so no one should be able to create clusters with empty tags even before this change.

thomasmckay:
lgtm

pkg/aws/helpers.go
```diff
@@ -258,6 +258,9 @@ func UserTagValidator(input interface{}) error {
 			if len(tag) != 2 {
 				return fmt.Errorf("invalid tag format. Expected tag format: 'key:value'")
 			}
+			if tag[1] == "" {
+				return fmt.Errorf("invalid tag format, Tags must have values. Expected tag format: 'key:value'")
+			}
```
renan-campos, line 263: This way totally works, I just wonder if it would instead be better to just update the UserTagValueRE expression. It is `^[\pL\pZ\pN_.:/=+\-@]{0,256}$`, I believe `^[\pL\pZ\pN_.:/=+\-@]{1,256}$` would cover what we need, which is consisten with the way that the UserTagKeyRE is validated.

renan-campos (commented):
Can you add some unit test coverage for this helper?

ciaranRoche:
> Can you add some unit test coverage for this helper?

@renan-campos I have a follow up MR for this that has test coverage. Lets just merge this as is and catch the coverage in my MR.

PR #1326, opened 2023-06-22
OCM-2427 | fix: Added command in INFO message to register oidc config after creating

hunterkepley:
[OCM-2427](https://issues.redhat.com/browse/OCM-2427) requires extra information in a info message to  the user. The reporter requested suggesting the user run the following command: `rosa register oidc-config`

### DIFF
--- a/cmd/create/oidcconfig/cmd.go
+++ b/cmd/create/oidcconfig/cmd.go
@@ -515,8 +515,9 @@ func (s *CreateUnmanagedOidcConfigManualStrategy) execute(r *rosa.Runtime) {
 	fmt.Println(awscb.JoinCommands(commands))
 	if r.Reporter.IsTerminal() {
 		r.Reporter.Infof("Please run commands above to generate OIDC compliant configuration in your AWS account. " +
-			"After running the commands please refer to the documentation to register your unmanaged OIDC Configuration " +
-			"with OCM.")
+			"To register this OIDC Configuration, please run the following command:\n" +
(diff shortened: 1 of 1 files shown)


cmd/create/oidcconfig/cmd.go
```diff
@@ -516,7 +516,8 @@ func (s *CreateUnmanagedOidcConfigManualStrategy) execute(r *rosa.Runtime) {
 	if r.Reporter.IsTerminal() {
 		r.Reporter.Infof("Please run commands above to generate OIDC compliant configuration in your AWS account. " +
 			"After running the commands please refer to the documentation to register your unmanaged OIDC Configuration " +
-			"with OCM.")
```
hunterkepley, line 519: Should I remove "After running the commands please refer to the documentation" ? @gdbranco

hunterkepley: Made a happy medium change, let me know what you think

gdbranco: I think it is okay to do that, but we actually should adjust so that these two matches
https://github.com/openshift/rosa/blob/08377c478b454db2c0037cbee25b3863d72ba491/cmd/create/oidcconfig/cmd.go#L393


Which means it is probably better to have a const unifying them

hunterkepley: Though, isn't L517 about just running commands? It seems to run at the end, so do we even need "try again through" for L517? :thinking:

PR #1329, opened 2023-06-23
OCM-2424 | fix: Made info message only print when on auto mode

hunterkepley:
[OCM-2424](https://issues.redhat.com/browse/OCM-2424) Is about an info message that wasn't supposed to pop up when using the create oidc-config command in Manual mode, but it did. Now the info message should only pop up when the mode is Auto

### DIFF
--- a/cmd/create/oidcconfig/cmd.go
+++ b/cmd/create/oidcconfig/cmd.go
@@ -230,7 +230,7 @@ func run(cmd *cobra.Command, argv []string) {
 			}
 			roleName, _ := aws.GetResourceIdFromARN(args.installerRoleArn)
 			if roleName != "" {
-				if !output.HasFlag() && r.Reporter.IsTerminal() {
+				if !output.HasFlag() && r.Reporter.IsTerminal() && mode == aws.ModeAuto {
 					r.Reporter.Infof("Using %s for the installer role", args.installerRoleArn)
 				}
 				err := aws.ARNValidator(args.installerRoleArn)


cmd/create/oidcconfig/cmd.go
```diff
@@ -230,7 +230,7 @@ func run(cmd *cobra.Command, argv []string) {
 			}
 			roleName, _ := aws.GetResourceIdFromARN(args.installerRoleArn)
 			if roleName != "" {
-				if !output.HasFlag() && r.Reporter.IsTerminal() {
+				if !output.HasFlag() && r.Reporter.IsTerminal() && mode == aws.ModeAuto {
```
oriAdler, line 233: The ticket says the installer role won't be used for the manual mode.
I wonder if all the other steps afterward: `ARNValidator`, `CheckRoleExists`, etc are applicable to the manual flow?

hunterkepley: @oriAdler Im wondering the same, do you know someone I can talk to?

oriAdler: @gdbranco in the auto mode flow, should we skip lines `233-260`?

gdbranco: We already skip it, as roleName would be empty in this case

What happens here is that the customer supplied the role arn even though it wasn't needed. We should probably fail first and say the role is not required for manual mode at the start.

But for this specific issue I  think it is enough to just not show the message

PR #1334, opened 2023-06-26
OCM-1295 | fix: add existence  check for ocm-roles

cameronmwall:
JIRA: https://issues.redhat.com/browse/OCM-1295

Previously, we were allowed to link ocmroles that did not exist. This fix adds a check to make sure that the role exists on AWS and throws an error and quits the wizard if the role does not exist.

### DIFF
--- a/cmd/link/ocmrole/cmd.go
+++ b/cmd/link/ocmrole/cmd.go
@@ -62,7 +62,7 @@ func init() {
 }
 
 func run(cmd *cobra.Command, argv []string) (err error) {
-	r := rosa.NewRuntime().WithOCM()
+	r := rosa.NewRuntime().WithAWS().WithOCM()
 	defer r.Cleanup()
 
 	if len(argv) > 0 {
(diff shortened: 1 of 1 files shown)


cmd/link/ocmrole/cmd.go
```diff
@@ -114,6 +114,17 @@ func run(cmd *cobra.Command, argv []string) (err error) {
 			os.Exit(1)
 		}
 	}
+
+	roleExists, _, err := r.AWSClient.CheckRoleExists(args.roleArn)
+	if err != nil {
+		r.Reporter.Errorf("There was a problem checking if role '%s' exists: %v", args.roleArn, err)
```
gdbranco, line 120:
```suggestion
		r.Reporter.Errorf("There was a problem checking if role '%s' exists: %v", roleArn, err)
```

cmd/link/ocmrole/cmd.go
```diff
@@ -114,6 +114,17 @@ func run(cmd *cobra.Command, argv []string) (err error) {
...
+		os.Exit(1)
+	}
+	if !roleExists {
+		r.Reporter.Errorf("Role '%s' does not exist", args.roleArn)
```
gdbranco, line 124:
```suggestion
		r.Reporter.Errorf("Role '%s' does not exist", roleArn)
```

PR #1335, opened 2023-06-26
OCM 844 | feat:  Support loading users from htpasswd file

ckandag:
https://issues.redhat.com/browse/OCM-844

Add flag to provide a htpasswdfile , create IDP from the users and hashed passwords loaded  from the htpasswd file

### DIFF
--- a/cmd/create/idp/htpasswd.go
+++ b/cmd/create/idp/htpasswd.go
@@ -17,6 +17,7 @@ limitations under the License.
 package idp
 
 import (
+	"bufio"
 	"fmt"
 	"os"
 	"regexp"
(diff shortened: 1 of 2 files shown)


cmd/create/idp/htpasswd.go
```diff
@@ -106,21 +107,33 @@ func createHTPasswdIDP(cmd *cobra.Command,
...
+	}
+	usersfile := args.htpasswdFile
+	if usersfile != "" {
+		argsCounter++
+	}
 
-	if username != "" && len(users) > 0 {
-		r.Reporter.Errorf("Only one of  'users' or 'username/password' may be specified. " +
-			"Choose the option 'users' to add one or more users to the IDP.")
+	if argsCounter > 1 {
```
gdbranco, line 128: I feel like we should move this to a function, like hasMultipleUserArgs or something. Just so it is clearer

cmd/create/idp/htpasswd.go
```diff
@@ -134,6 +147,15 @@ func buildUserList(cmd *cobra.Command, r *rosa.Runtime) *cmv1.HTPasswdUserListBu
...
 		}
+	} else if usersfile != "" {
+		err := parseHtpasswordFile(&userList, usersfile)
+		if err != nil {
+			r.Reporter.Errorf(
+				"Failed to load Htpasswd file '%s': %v", usersfile, err)
+			os.Exit(1)
+		}
+		//password in htpasswd are already and do not need to be hashed again in CS
+		hashed = true
```
gdbranco, line 158: Is it always expected that a file input will be pre hashed? Is there any way we can check this somehow :thinking:

ckandag: @gdbranco : if you  look at the htpsswd utility used to build htpasswd file, by default they are hashed ( you can ofcourse  force change the type/status of encryption )

This concept is the same  in OCP as well.  For Htpasswd IDP, OCP just lets you upload any well-formed htpasswd file and I think the idea here  is to be able to achieve the same thing with  ROSA.

While doing the design review on this,  I did bring up with @vkareh the fact that we may not be able to validate the passwords like we do with the plain-text ones.
But everybody seemed ok with it, we want to give the users the ability to upload htpasswd file and let them choose the encryption type and status.

cmd/create/idp/htpasswd.go
ckandag: @gdbranco : ok, I will refractor it.

PR #1338, opened 2023-06-28
[OCM-2348] Add audit log details in rosa describe cluster

davidleerh:
Related issue: https://issues.redhat.com/browse/OCM-2348

What
Adds a row in "rosa describe cluster -c $CLUSTER" output that indicates whether audit log forwarding is enabled as well as the audit log ARN (if forwarding is enabled). 

Why
Better UX

### DIFF
--- a/cmd/describe/cluster/cmd.go
+++ b/cmd/describe/cluster/cmd.go
@@ -353,6 +353,15 @@ func run(cmd *cobra.Command, argv []string) {
 		}
 	}
 
+	if isHypershift {
+		str = fmt.Sprintf("%s"+
+			"Audit Log Forwarding:       %s\n", str, getAuditLogForwardingStatus(cluster))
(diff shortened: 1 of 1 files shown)


cmd/describe/cluster/cmd.go
```diff
@@ -353,6 +353,13 @@ func run(cmd *cobra.Command, argv []string) {
 		}
 	}
 
+	str = fmt.Sprintf("%s"+
+		"Audit Log Forwarding:       %s\n", str, getAuditLogForwardingStatus(cluster))
```
gdbranco, line 357: This is only for HCP right? Should we check the cluster type prior to showing these field? It's just that current classic customers maybe inquire how to do it and there isn't a way at the moment

davidleerh: that is a great idea!!!
Will make it appear only for hcp.

PR #1339, opened 2023-06-29
OCM-954 | Feat: Hypershift node pool upgrades

andreadecorte:
Adds support for HyperShift node pool upgrades:
- upgrade machinepool
- delete upgrade
- describe upgrade
- list upgrade 

Marking also as deprecated the version change in editing machinepools.
Related: [OCM-954](https://issues.redhat.com//browse/OCM-954)

Upgrade UX design document: https://docs.google.com/document/d/15S3TXJXu6_iUJFDFT-mpcUp2uC4cWVKwZbt_bwgwqZ4/edit#

### DIFF
--- a/cmd/create/machinepool/nodepool.go
+++ b/cmd/create/machinepool/nodepool.go
@@ -76,7 +76,7 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
 		}
 
 		// Filter the available list of versions for a hosted machine pool
-		filteredVersionList := versions.GetFilteredVersionList(versionList, minVersion, clusterVersion)
+		filteredVersionList := versions.GetFilteredVersionListForCreation(versionList, minVersion, clusterVersion)
 		if err != nil {
 			r.Reporter.Errorf("%s", err)
 			os.Exit(1)
--- a/cmd/describe/upgrade/cmd.go
+++ b/cmd/describe/upgrade/cmd.go
@@ -19,16 +19,18 @@ package upgrade
 import (
 	"fmt"
 	"os"
+	"strings"
 
-	"github.com/spf13/cobra"
-
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/openshift/rosa/pkg/interactive/confirm"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/rosa"
+	"github.com/spf13/cobra"
 )
 
 var Cmd = &cobra.Command{
 	Use:     "upgrade",
-	Aliases: []string{"appliance", "upgrade"},
+	Aliases: []string{"appliance", "upgrades"},
 	Short:   "Show details of an upgrade",
 	Long:    "Show details of an upgrade",
 	Example: `  # Describe an upgrade-policy"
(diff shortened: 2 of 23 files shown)


cmd/describe/upgrade/cmd.go
```diff
@@ -37,74 +39,128 @@ var Cmd = &cobra.Command{
...
 	_, upgradeState, err := r.OCMClient.GetScheduledUpgrade(clusterID)
 	if err != nil {
-		r.Reporter.Errorf("Failed to get scheduled upgrades for cluster '%s': %v", clusterID, err)
-		os.Exit(1)
+		return fmt.Errorf("Failed to get scheduled upgrades for cluster '%s': %v", clusterID, err)
 	}
 	if len(upgrades) < 1 {
-		r.Reporter.Warnf("No scheduled upgrades for cluster id '%s'", clusterID)
-		os.Exit(1)
+		r.Reporter.Infof("No scheduled upgrades for cluster id '%s'", clusterID)
```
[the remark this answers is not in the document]
andreadecorte: yes. We have both in the code and I think Infof here makes more sense, as this is not an error, but just an information.

cmd/edit/machinepool/cmd.go
```diff
@@ -129,6 +129,8 @@ func init() {
 			"Tuning config must already exist. "+
 			"This list will overwrite any modifications made to node tuning configs on an ongoing basis.",
 	)
+
+	flags.MarkDeprecated("version", "for upgrades, please use 'rosa upgrade machinepool' instead")
```
lucasponce, line 133: +1

cmd/upgrade/machinepool/cmd.go
```diff
@@ -0,0 +1,248 @@
...
+			Options:  filteredVersionList,
+			Default:  version,
+		})
+		if err != nil {
+			return "", fmt.Errorf("Expected a valid machine pool version from interactive prompt: %s", err)
+		}
+	}
+	// This is called in HyperShift, but we don't want to exclude version which are HCP disabled for node pools
+	// so we pass the relative parameter as false
+	version, err = r.OCMClient.ValidateVersion(version, filteredVersionList, channelGroup, true, false)
```
lucasponce, line 216: Not a blocker or anything to change, but I wonder if we want to do something here.

Just thinking that the UI will need to use something similar.

lucasponce:
I put a couple of minor comments (nothing blocking).

LGTM, it's a big change and I'd like it if QA can start to test it all together.

cmd/upgrade/machinepool/cmd.go
andreadecorte: yes, this is a consequence of https://issues.redhat.com/browse/OCM-2182, which means that the requirement is to check HCP_enabled only for control plane. The only thing I could see to improve this is to explictly mark the parameter as HCP_control_plane_enabled or something, but it's a bit late to change the API, so I don't think we can do that.

cmd/create/machinepool/nodepool.go
```diff
@@ -76,7 +76,7 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
 		}
 
 		// Filter the available list of versions for a hosted machine pool
-		filteredVersionList := versions.GetFilteredVersionList(versionList, minVersion, clusterVersion)
+		filteredVersionList := versions.GetFilteredVersionList(versionList, minVersion, clusterVersion, false)
```
gdbranco, line 79: Not a fan of adding bools to functions, would rather setup a strategy (https://refactoring.guru/design-patterns/strategy) or have the condition here separating two functions

gdbranco:
Such a big PR, overall seems good to me. Glad you added all the tests!

ciaranRoche (commented):
@andreadecorte can you have a look at - https://github.com/openshift/rosa/pull/1419/files 

Would be awesome if you could update the tests, to follow similar to build the responses, just to cut down :crossed_fingers:

cmd/create/machinepool/nodepool.go
andreadecorte: updated to have 2 different functions, not really worth a strategy. It should be clearer in terms of usage

andreadecorte:
> @andreadecorte can you have a look at - https://github.com/openshift/rosa/pull/1419/files
> 
> Would be awesome if you could update the tests, to follow similar to build the responses, just to cut down 🤞

updated with that, thanks for the heads up

PR #1337, opened 2023-06-28
fixes keycloak url aliases

markturansky:

PR #1341, opened 2023-06-29
OCM-347 | fix: Align FedRAMP non-admin token url and client IDs for integration

tylercreller:
Adapted from https://github.com/openshift/rosa/pull/1337

### DIFF
--- a/pkg/fedramp/config.go
+++ b/pkg/fedramp/config.go
@@ -68,12 +68,13 @@ var AdminURLAliases = map[string]string{
 }
 
 const cognitoURL = "auth-fips.us-gov-west-1.amazoncognito.com/oauth2/token"
+const keycloakURL = "realms/redhat-external/protocol/openid-connect/token"
 
 // TokenURLs allows the value of the `--env` option to map to the various AWS Cognito token URLs.
 var TokenURLs = map[string]string{
 	"production":  fmt.Sprintf("https://ocm-ra-production-domain.%s", cognitoURL),
 	"staging":     fmt.Sprintf("https://ocm-ra-stage-domain.%s", cognitoURL),
-	"integration": fmt.Sprintf("https://rh-ocm-appsre-integration.%s", cognitoURL),
+	"integration": fmt.Sprintf("https://sso.int.openshiftusgov.com/%s", keycloakURL),
 }
 
 // AdminTokenURLs allows the value of the `--env` option to map to the various Admin AWS Cognito token URLs.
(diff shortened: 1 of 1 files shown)


pkg/fedramp/config.go
```diff
@@ -83,11 +88,11 @@ var AdminTokenURLs = map[string]string{
...
 }
 
-// ClientIDs allows the value of the `--env` option to map to the various AWS Cognito user pool clients.
+// ClientIDs allows the value of the `--env` option to map to the Keycloak clients.
 var ClientIDs = map[string]string{
-	"production":  "72ekjh5laouap6qcfis521jlgi",
-	"staging":     "1lb687dlpsmsfuj53r3je06vpp",
-	"integration": "20fbrpgl28f8oehp6709mk3nnr",
+	"production":  "console-dot",
+	"staging":     "console-dot",
```
vkareh, line 94: Since the stg/prod token URLs are not changing _yet_, I would keep these commented out/removed until we migrated all envs.

pkg/fedramp/config.go
```diff
@@ -83,11 +88,12 @@ var AdminTokenURLs = map[string]string{
 	"integration": fmt.Sprintf("https://rh-ocm-appsre-integration.%s", cognitoURL),
 }
 
-// ClientIDs allows the value of the `--env` option to map to the various AWS Cognito user pool clients.
+// ClientIDs allows the value of the `--env` option to map to the Keycloak clients.
 var ClientIDs = map[string]string{
-	"production":  "72ekjh5laouap6qcfis521jlgi",
-	"staging":     "1lb687dlpsmsfuj53r3je06vpp",
```
vkareh, line 89: These still need to be in place, otherwise stage/prod will break. Maybe just a single line change to update integration to `console-dot`.

pkg/fedramp/config.go
```diff
@@ -68,12 +68,17 @@ var AdminURLAliases = map[string]string{
...
-	"production":  fmt.Sprintf("https://ocm-ra-production-domain.%s", cognitoURL),
-	"staging":     fmt.Sprintf("https://ocm-ra-stage-domain.%s", cognitoURL),
-	"integration": fmt.Sprintf("https://rh-ocm-appsre-integration.%s", cognitoURL),
+	"integration": fmt.Sprintf("https://sso.int.openshiftusgov.com/%s", keycloakURL),
+
+	//TODO:  switch these aliases as the changes propagate to prod, then remove this TODO
+	//"production":  fmt.Sprintf("https://sso.openshiftusgov.com/%s", keycloakURL),
+	"production": fmt.Sprintf("https://ocm-ra-production-domain.%s", cognitoURL),
+	//"staging":     fmt.Sprintf("https://sso.stage.openshiftusgov.com/%s", keycloakURL),
+	"staging": fmt.Sprintf("https://ocm-ra-stage-domain.%s", cognitoURL),
```
vkareh, line 81: Can we just clean this up? Make it a single line change updating the `integration` token URL. Once we have the final stage/prod URLs in working order, we will be able to update them just fine.

vkareh:
@tylercreller Prow is failing because there's not JIRA in the commit message.

pkg/fedramp/config.go
tylercreller: Updated to single line

pkg/fedramp/config.go
tylercreller: Cleaned up from the original MR

tylercreller:
@vkareh looking at the jira format now

vkareh:
@tylercreller Prow still fails. Try the following format for the title:
```
OCM-347 | fix: Align FedRAMP non-admin token url and client IDs for integration
```

PR #1342, opened 2023-06-30
OCM-2506 | feat : Allow creation of multiple htpasswd idps

ckandag:
Ref: https://issues.redhat.com/browse/OCM-2506

As part of this work,

Any number of htpasswd IDPs can now be created just as any other IDP type like LDAP, OPENID etc.
admin user is created in a separate htpasswd IDP instance  named "cluster-admin".

Changes have been  made to the following flows to implement this design change.
admin - create, describe, delete flow to allow for this.
htpasswd idp - create, delete.

### DIFF
--- a/cmd/create/admin/cmd.go
+++ b/cmd/create/admin/cmd.go
@@ -24,12 +24,14 @@ import (
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 
-	"github.com/openshift/rosa/cmd/create/idp"
 	"github.com/openshift/rosa/pkg/object"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/output"
 	"github.com/openshift/rosa/pkg/rosa"
 )
+
+const ClusterAdminUsername = "cluster-admin"
+const ClusterAdminIDPname = "cluster-admin"
 
 var Cmd = &cobra.Command{
 	Use:   "admin",
@@ -69,10 +71,9 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
-	// Try to find an existing htpasswd identity provider and
-	// check if cluster-admin user already exists
-	existingHTPasswdIDP, existingUserList := idp.FindExistingHTPasswdIDP(cluster, r)
-	if idp.HasClusterAdmin(existingUserList) {
+	// Try to find an existing cluster admin IDP
+	existingClusterAdminIDP, _ := FindExistingClusterAdminIDP(cluster, r)
+	if existingClusterAdminIDP != nil {
 		r.Reporter.Errorf("Cluster '%s' already has an admin", clusterKey)
 		os.Exit(1)
 	}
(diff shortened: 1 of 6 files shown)


cmd/create/admin/cmd.go
```diff
@@ -24,13 +24,15 @@ import (
...
 
-	"github.com/openshift/rosa/cmd/create/idp"
 	"github.com/openshift/rosa/pkg/object"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/output"
 	"github.com/openshift/rosa/pkg/rosa"
 )
 
+const ClusterAdminUsername = "cluster-admin"
+const ClusterAdminIDPname = "cluster-admin"
```
ciaranRoche, line 34: Do we need the two consts for the same value?

cmd/create/admin/cmd.go
```diff
@@ -222,3 +205,53 @@ func generateRandomPassword(length int) (string, error) {
...
+func FindExistingClusterAdminIDP(cluster *cmv1.Cluster, r *rosa.Runtime) (
+	htpasswdIDP *cmv1.IdentityProvider, userList *cmv1.HTPasswdUserList) {
+
+	// admin user will now be created in a htpasswd IDP named 'cluster-admin'
+	// It should suffice to look for this idp but for back-compatibility ,
+	// i.e for cases where the cluster was created  with previous versions of rosacli
+	// we are going to search  all htpasswd idps
+
+	r.Reporter.Debugf("Loading cluster's identity providers")
+	r.Reporter.Infof("Loading cluster's identity providers")
```
ciaranRoche, line 218: Do we need the two different logs?

cmd/create/admin/cmd.go
```diff
@@ -222,3 +205,53 @@ func generateRandomPassword(length int) (string, error) {
...
+	idps, err := r.OCMClient.GetIdentityProviders(cluster.ID())
+	if err != nil {
+		r.Reporter.Errorf("Failed to get identity providers for cluster '%s': %v", r.ClusterKey, err)
```
ciaranRoche, line 221: Just curious, why are we using `r.ClusterKey` in the error, but are using `cluster.ID` when fetching idps?

cmd/create/admin/cmd.go
```diff
@@ -222,3 +205,53 @@ func generateRandomPassword(length int) (string, error) {
...
+		if ocm.IdentityProviderType(item) == ocm.HTPasswdIDPType {
+
+			itemUserList, err := r.OCMClient.GetHTPasswdUserList(cluster.ID(), item.ID())
+			r.Reporter.Infof("user list %s: %v", item.Name(), itemUserList)
+			if err != nil {
+				r.Reporter.Errorf("Failed to get user list of the HTPasswd IDP of '%s: %s': %v", item.Name(), r.ClusterKey, err)
+				os.Exit(1)
+			}
+			if HasClusterAdmin(itemUserList) {
+				r.Reporter.Infof("Has admintrue")
```
ciaranRoche, line 235: Is this a log we want to expose to the customer?

cmd/create/admin/cmd.go
```diff
@@ -222,3 +205,53 @@ func generateRandomPassword(length int) (string, error) {
...
+
+				htpasswdIDP = item
+				userList = itemUserList
+				return
+			} else {
+				r.Reporter.Infof("Has adminfalse")
```
ciaranRoche, line 241: Same as above, do we need this log?

cmd/create/admin/cmd.go
```diff
@@ -222,3 +205,53 @@ func generateRandomPassword(length int) (string, error) {
...
+func HasClusterAdmin(userList *cmv1.HTPasswdUserList) bool {
+	hasAdmin := false
+	userList.Each(func(user *cmv1.HTPasswdUser) bool {
+		if user.Username() == ClusterAdminUsername {
+			hasAdmin = true
+		}
+		return true
+	})
+	return hasAdmin
+}
```
ciaranRoche, line 257: I dont think we need to itterate the full list if we find a user with the admin user name, we should just return, can either update the current function : 
```
if user.Username() == ClusterAdminUsername {
			hasAdmin = true
			return false
		}
```
`Each` function will keep itterating till it exhaustes the list or until the func returns false. 

Alternatively we can just use `range` and return true from that :  
```suggestion
func HasClusterAdmin(userList *cmv1.HTPasswdUserList) bool {
	for _, user := range userList.Slice() {
		if user.Username() == ClusterAdminUsername {
			return true
		}
	}
	return false
}
```

What ever you think yourself

cmd/create/idp/htpasswd.go
```diff
@@ -24,11 +24,9 @@ import (
 	"strings"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
-	"github.com/spf13/cobra"
-
 	"github.com/openshift/rosa/pkg/interactive"
-	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/rosa"
+	"github.com/spf13/cobra"
```
ciaranRoche, line 29: NP : the ordering of the imports, we try keep imports ordered in groups, (std, external, internal) all with a whitespace between them 

```
	"github.com/spf13/cobra"

```

cmd/create/admin/cmd.go
ckandag: will fix this

cmd/create/admin/cmd.go
ckandag: that was not intentional , remnants from the old code I guess. Will fix this

cmd/create/admin/cmd.go
ckandag: @ciaranRoche ,  I do think we need two separate constants to distinguish between user-name and idp-name for clarity/readability of the code.  

It just so happened that in this case , we are choosing the same  name for  user as well as the idp holding the user.

cmd/create/admin/cmd.go
ckandag: nope, will fix that

cmd/create/admin/cmd.go
gdbranco: I think it is because the user might have supplied the name or ID, but to get identity providers we need to supply the ID, but to show the error back to the user we use whatever he supplied.

cmd/create/admin/cmd.go
ckandag: This too is a remnant of the old code , just happened to  move it from one file to another.

I will fix that.

cmd/create/admin/cmd.go
ckandag: nope, will fix that

cmd/create/admin/cmd.go
ckandag: yes, agree. It seems to make sense to use the clusterKey vs clusterID in the error msg.

As clusterID is a long ID string vs clusterKey is the more readable cluster-name

cmd/create/admin/cmd.go
ciaranRoche: cool cool cool

ckandag:
> overall looks good, can you share some verification steps, before we merge it please 🙏

@ciaranRoche : As you only one IDP of type htpasswd was allowed earlier, with this PR you should be able to create multiple different IDPs and also admin will be created in its own separate IDP.

please try the below ROSA Cli commands

1. Create admin user
`rosa create admin -c <CLUSTERNAME>`

2. List IDPs to verify the idp "cluster-admin" is listed in the results
`rosa list idps -c <CLUSTERNAME>`

3. Create more htpasswd idps

`rosa create idp --type htpasswd --name testhtpasswd1 -c cktest6 --users='tester1:Aq1sw2de3fr4gt5!'`

`rosa create idp --type htpasswd --name testhtpasswd2 -c cktest6 --users='tester2:Aq1sw2de3fr4gt5!'`

4. List IDPs to verify all three IDPs are listed i.e  "cluster-admin", "testhtpasswd1", "testhtpasswd2"

5. For additional verification, you can wait a while and go to the cluster console,  you should see all idps listed on the console,  choose a idp and login with  a user who is  in that idp.
i.e if you chose IDP - testhtpasswd1 then user/password is tester1:Aq1sw2de3fr4gt5!

PR #1358, opened 2023-07-12
OCM-2694 | fix: update valid cluster states for verify network

thomasmckay:


cmd/verify/network/cmd.go
```diff
@@ -129,8 +129,9 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 		cluster := r.FetchCluster()
 
-		if cluster.Status() != nil && cluster.Status().State() != cmv1.ClusterStateReady {
-			r.Reporter.Errorf("Cluster state is not ready: %v", cluster.Status().State())
+		if cluster.Status() != nil &&
+			(cluster.Status().State() != cmv1.ClusterStateReady || cluster.Status().State() != cmv1.ClusterStateError) {
```
gdbranco, line 133:
```suggestion
			helper.Contains([]cmv1.ClusterState{cmv1.ClusterStateReady, cmv1.ClusterStateError}, cluster.Status().State()) {
```

https://github.com/openshift/rosa/blob/edcbeaf6aad55007793523a20fe48014465389cf/pkg/helper/helpers.go#L81
```
func Contains[T comparable](slice []T, element T) bool {
	for _, sliceElement := range slice {
		if reflect.DeepEqual(sliceElement, element) {
			return true
		}
	}

	return false
}
```

PR #1362, opened 2023-07-12
OCM-2695 | fix: determine role-arn for verify network

thomasmckay:
* role arn is pulled from cluster, if not specified
* region is pulled from cluster, if not specified
* subnets are pulled from cluster, if not specified

### DIFF
--- a/cmd/verify/network/cmd.go
+++ b/cmd/verify/network/cmd.go
@@ -28,7 +28,6 @@ import (
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/openshift/rosa/pkg/arguments"
 	"github.com/openshift/rosa/pkg/aws"
-	"github.com/openshift/rosa/pkg/helper"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/reporter"
 	"github.com/openshift/rosa/pkg/rosa"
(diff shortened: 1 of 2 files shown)


cmd/verify/network/cmd.go
```diff
@@ -106,43 +106,42 @@ func init() {
...
+		} else {
+			r.Reporter.Errorf("At least one subnet IDs is required")
 			os.Exit(1)
 		}
+	}
 
-		args.subnetIDs = cluster.AWS().SubnetIDs()
-		if len(args.subnetIDs) == 0 {
-			r.Reporter.Errorf("No subnets on cluster")
+	if cmd.Flags().Changed("region") {
```
ciaranRoche, line 134: Getting the region value, can we flatten out this whole block. Bring it out to its own func, and return early. Getting rid of the `else` and `else if`.

cmd/verify/network/cmd.go
```diff
@@ -153,8 +152,19 @@ func run(cmd *cobra.Command, _ []string) {
...
 		}
 	} else if !cmd.Flags().Changed(statusOnlyFlag) {
-		r.Reporter.Errorf("%s is required", roleArnFlag)
-		os.Exit(1)
+		if cmd.Flags().Changed(clusterFlag) {
+			if cluster == nil {
+				cluster = r.FetchCluster()
+			}
+			if args.roleArn = cluster.AWS().STS().RoleARN(); args.roleArn == "" {
+				r.Reporter.Errorf("Unable to get role ARN from non-STS cluster '%s'.",
```
ciaranRoche, line 160: I dont think this error is clear, should just return something like `NV is only supported for STS clusters`. 

Pretty much if roleARN is empty is the toggle indicating if a cluster is STS or not.

cmd/verify/network/cmd.go
```diff
@@ -153,8 +152,19 @@ func run(cmd *cobra.Command, _ []string) {
...
+					cluster.Name())
```
ciaranRoche, line 161: FWIW its best to return `cluster.ID()` then `cluster.Name()` this is common pattern in OCM. 

Just makes our life easier for debugging issues when people report them to search by ID than Name

cmd/verify/network/cmd.go
```diff
@@ -106,43 +106,42 @@ func init() {
...
-			r.Reporter.Errorf("Error getting region: %v", err)
-			os.Exit(1)
-		}
-		args.region = region
-	} else if cmd.Flags().Changed(clusterFlag) {
-		if cmd.Flags().Changed(subnetIDsFlag) {
-			r.Reporter.Errorf("One and only one of %s or %s is required",
-				clusterFlag, subnetIDsFlag)
+		if cluster.Status() != nil && cluster.Status().State() != cmv1.ClusterStateReady {
+			r.Reporter.Errorf("Cluster state is not ready: %v", cluster.Status().State())
```
ciaranRoche, line 116: Curious as to why the cluster has to be ready to run verify the network?

gdbranco: There's nothing that truly impedes it, but when running from a cluster if it is not ready it will either run from the inflights instead of on demand or have already ran from the inflights.

cmd/verify/network/cmd.go
thomasmckay: this doesn't seem complex enough to pull out into separate func, it's pretty compact and legible imo

cmd/verify/network/cmd.go
thomasmckay: updated, follow-on PRs will handle non-sts as well. thanks!

cmd/verify/network/cmd.go
thomasmckay: i changed to `cluster.ID()` but should both name and id be output? do users typically specify id or name as a cli argument?

cmd/verify/network/cmd.go
thomasmckay: i removed all checks for cluster state

cmd/verify/network/cmd.go
ciaranRoche: I agree this current block is not complex, but keeping indentation to a minimum is a best practice for golang. 

ROSA is full of nested if's and else statements. We need to start cutting back on them and follow the best practices. 

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow
https://go.dev/doc/effective_go

thomasmckay: updated, thanks!

ciaranRoche (approved):
/lgtm

PR #1366, opened 2023-07-19
OCM-1220 | feat: supports use local credentials cluster properties

gdbranco:


pkg/aws/client.go
```diff
@@ -271,6 +275,12 @@ func (b *ClientBuilder) AccessKeys(value *AccessKey) *ClientBuilder {
 	return b
 }
 
+func (b *ClientBuilder) UseLocalCredentials(value bool) *ClientBuilder {
+	// fmt.Printf("Using new access key %s\n", value.AccessKeyID)
```
vkareh, line 279: Remove this output

cmd/initialize/cmd.go
```diff
@@ -83,6 +85,14 @@ func init() {
 		"Indicates if cloud permission checks are disabled when attempting installation of the cluster.",
 	)
 
+	flags.BoolVar(
+		&args.useLocalCredentials,
+		"use-local-credentials",
+		false,
+		"Use local AWS credentials instead of the 'osdCcsAdmin' user. This is not supported.",
+	)
+	flags.MarkHidden("use-local-credentials")
```
vkareh, line 94: Why add this to the `init` command? Part of the purpose of this feature is that we can skip the `init` command entirely and rely on the user's own credentials (assuming they have the right policy access).

gdbranco: That was just because there was a builder there and I wasn't sure if it would be needed to have this option.
Keeping it as hidden would allow if there is some strange edge case, but most of the time would retain original functionality where this would be false.

ciaranRoche:
Going to merge is as the flags are hidden, changes have been tested and look good. If we want to remove from init, we can revert the change.

PR #1367, opened 2023-07-19
'verify network' unit tests

thomasmckay:


cmd/verify/network/cmd_test.go
```diff
@@ -0,0 +1,488 @@
...
+		cmd.Flags().Set(clusterFlag, "tomckay-vpc")
+		cmd.Flags().Set(roleArnFlag, "arn:aws:iam::765374464689:role/tomckay-Installer-Role")
+		stdout, stderr, err := captureStdout(r, cmd)
+		Expect(err).To(BeNil())
+		Expect(stderr).To(Equal(""))
+		Expect(stdout).To(Equal(successOutputPendingComplete))
+	})
+})
+
+func captureStdout(runtime *rosa.Runtime, cmd *cobra.Command) (string, string, error) {
```
andreadecorte, line 463: nice, so we can also compare output. I suggest to move this outside the specific package, so it can be reused, as it is not specific to this use case

thomasmckay: agreed, perhaps `pkg/helpers/tests/helpers.go` (i'm open to alternative)

andreadecorte: `pkg/test/helpers.go` ? close to what we have on the backend

gdbranco (approved):

PR #1368, opened 2023-07-19
OCM-2741 | fix: Do not Allow IDP name cluster-admin

ckandag:
Ref: https://issues.redhat.com/browse/OCM-2741

### DIFF
--- a/cmd/create/idp/cmd.go
+++ b/cmd/create/idp/cmd.go
@@ -384,17 +384,18 @@ func run(cmd *cobra.Command, _ []string) {
 	if !cmd.Flags().Changed("name") {
 		idps := getIdps(r, cluster)
 		idpName = GenerateIdpName(idpType, idps)
-	} else {
-		isValidIdpName := idRE.MatchString(idpName)
-		if !isValidIdpName {
-			r.Reporter.Errorf("Invalid identifier '%s' for 'name'", idpName)
-			os.Exit(1)
-		}
 	}
+
 	if interactive.Enabled() {
 		idpName = getIDPName(cmd, idpName, r)
 	}
 	idpName = strings.Trim(idpName, " \t")
+
+	err = ValidateIdpName(idpName)
+	if err != nil {
+		r.Reporter.Errorf("%v", err)
+		os.Exit(1)
+	}
 
 	var idpBuilder cmv1.IdentityProviderBuilder
 	switch idpType {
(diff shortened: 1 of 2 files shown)


cmd/create/idp/cmd.go
```diff
@@ -434,6 +438,18 @@ func getIDPName(cmd *cobra.Command, idpName string, r *rosa.Runtime) string {
...
 }
 
+func validateIDPName(idpName interface{}) error {
+
+	if !idRE.MatchString(idpName.(string)) {
+		return fmt.Errorf("Invalid identifier '%s' for 'name'", idpName)
+	}
+
+	if strings.EqualFold(idpName.(string), "cluster-admin") {
+		return fmt.Errorf("The name \"cluster-admin\" is reserved for admin user IDP.")
```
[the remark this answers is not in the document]
ckandag: ok

cmd/create/idp/cmd.go
```diff
@@ -434,6 +438,18 @@ func getIDPName(cmd *cobra.Command, idpName string, r *rosa.Runtime) string {
 	return strings.Trim(idpName, " \t")
 }
 
+func validateIDPName(idpName interface{}) error {
+
+	if !idRE.MatchString(idpName.(string)) {
```
ciaranRoche, line 443: If a non string is passed into this function it will cause a panic. 

Can we ensure `idpName` is the expected type.

ciaranRoche (commented):
Just one comment inline, can we also add a test case for the `validateIDPName` function.

cmd/create/idp/cmd.go
```diff
@@ -434,6 +438,18 @@ func getIDPName(cmd *cobra.Command, idpName string, r *rosa.Runtime) string {
 	return strings.Trim(idpName, " \t")
 }
 
+func validateIDPName(idpName interface{}) error {
```
ciaranRoche, line 441: @oriAdler do we adopt the naming convention that CS has, which would make it `validateIdpName` or do we stick with `validateIDPName`

gdbranco: IMO should be `Idp`

ckandag: @ciaranRoche @gdbranco @oriAdler  -  So within the ROSACLI code, there are already several instances of using "IDP" , e.g doCreateIDP, getIDPName 
Also, I believe golang convention for acronyms is to use caps like in URL or ID etc

But if the consensus is to use Idp, I am happy to change it.

ciaranRoche: Lets not block this MR over this, we should discuss it on the next office hours, and get consensus on what we adopt for ROSA

cmd/create/idp/cmd.go
ckandag: ok

ckandag:
@ciaranRoche : Addressed the review comments, Can you pls take a look again.

cmd/create/idp/cmd.go
```diff
@@ -434,6 +438,24 @@ func getIDPName(cmd *cobra.Command, idpName string, r *rosa.Runtime) string {
 	return strings.Trim(idpName, " \t")
 }
 
+func ValidateIdpName(idpName interface{}) error {
```
ciaranRoche, line 441: Do we have similar validations in the API?

ckandag: i dont think so

ckandag:
@ciaranRoche , PTAL Again !

cmd/create/idp/cmd.go
ciaranRoche: awesome, thanks for the context. Okay makes sense to just have it here for this use case.

ciaranRoche (approved):
/lgtm

PR #1371, opened 2023-07-20
updating create-cluster command with autoscaler configurations

Gregory-Pereira:
Addresses: [OCM-210](https://issues.redhat.com/browse/OCM-210)

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -69,6 +69,26 @@ const (
 	defaultIngressWildcardPolicyFlag           = "default-ingress-wildcard-policy"
 	defaultIngressNamespaceOwnershipPolicyFlag = "default-ingress-namespace-ownership-policy"
 
+	autoscalerBalanceSimilarNodeGroupsFlag      = "autoscaler-balance-similar-node-groups"
+	autoscalerSkipNodesWithLocalStorageFlag     = "autoscaler-skip-nodes-with-local-storage"
+	autoscalerLogVerbosityFlag                  = "autoscaler-log-verbosity"
+	autoscalerMaxPodGracePeriodFlag             = "autoscaler-max-pod-grace-period"
+	autoscalerPodPriorityThresholdFlag          = "autoscaler-pod-priority-threshold"
+	autoscalerIgnoreDaemonsetsUtilizationFlag   = "autoscaler-ignore-daemonsets-utilization"
+	autoscalerMaxNodeProvisionTimeFlag          = "autoscaler-max-node-provision-time"
+	autoscalerBalancingIgnoredLabelsFlag        = "autoscaler-balancing-ignored-labels"
+	autoscalerMaxNodesTotalFlag                 = "autoscaler-max-nodes-total"
+	autoscalerMinCoresFlag                      = "autoscaler-min-cores"
+	autoscalerMaxCoresFlag                      = "autoscaler-max-cores"
+	autoscalerMinMemoryFlag                     = "autoscaler-min-memory"
+	autoscalerMaxMemoryFlag                     = "autoscaler-max-memory"
+	autoscalerScaleDownEnabledFlag              = "autoscaler-scale-down-enabled"
+	autoscalerScaleDownUnneededTimeFlag         = "autoscaler-scale-down-unneeded-time"
+	autoscalerScaleDownUtilizationThresholdFlag = "autoscaler-scale-down-utilization-threshold"
+	autoscalerScaleDownDelayAfterAddFlag        = "autoscaler-scale-down-delay-after-add"
+	autoscalerScaleDownDelayAfterDeleteFlag     = "autoscaler-scale-down-delay-after-delete"
+	autoscalerScaleDownDelayAfterFailureFlag    = "autoscaler-scale-down-delay-after-failure"
+
 	MinReplicasSingleAZ = 2
 	MinReplicaMultiAZ   = 3
 )
@@ -116,6 +136,18 @@ var args struct {
 	minReplicas              int
 	maxReplicas              int
 	defaultMachinePoolLabels string
+
+	// Autoscaler Configurations
+	autoscalerBalanceSimilarNodeGroups    bool
+	autoscalerSkipNodesWithLocalStorage   bool
+	autoscalerLogVerbosity                int
+	autoscalerMaxPodGracePeriod           int
+	autoscalerPodPriorityThreshold        int
+	autoscalerIgnoreDaemonsetsUtilization bool
+	autoscalerMaxNodeProvisionTime        string
+	autoscalerBalancingIgnoredLabels      []string
+	autoscalerResourceLimits              ResourceLimits
+	autoscalerScaleDown                   ScaleDownConfig
 
 	// Networking options
 	networkType string
@@ -189,6 +221,26 @@ var args struct {
 	privateHostedZoneID string
 	sharedVPCRoleARN    string
 	baseDomain          string
+}
+
+type ResourceLimits struct {
+	MaxNodesTotal int
+	Cores         ResourceRange
+	Memory        ResourceRange
+}
+
+type ResourceRange struct {
+	Min int
+	Max int
+}
+
+type ScaleDownConfig struct {
+	Enabled              bool
+	UnneededTime         string
+	UtilizationThreshold float64
+	DelayAfterAdd        string
+	DelayAfterDelete     string
+	DelayAfterFailure    string
 }
 
 var Cmd = &cobra.Command{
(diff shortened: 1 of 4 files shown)


romfreiman:
@ciaranRoche PTAL

ciaranRoche (changes requested):

go.mod
```diff
@@ -71,6 +71,7 @@ require (
 	github.com/prometheus/procfs v0.7.3 // indirect
 	github.com/russross/blackfriday/v2 v2.0.1 // indirect
 	github.com/shurcooL/sanitized_anchor_name v1.0.0 // indirect
+	gitlab.cee.redhat.com/service/uhc-clusters-service v0.0.11
```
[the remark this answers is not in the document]
osherdp: yeah, with clusters-service being located D/S it shouldn't be possible to import it U/S

rather, client implementation should be made via ``github.com/openshift-online/ocm-sdk-go``, as everything should be available starting from release https://github.com/openshift-online/ocm-sdk-go/releases/tag/v0.1.354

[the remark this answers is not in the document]
osherdp: @Gregory-Pereira did you manage to use ``github.com/openshift-online/ocm-sdk-go``?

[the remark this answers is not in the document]
screeley44: @Gregory-Pereira - any status on this?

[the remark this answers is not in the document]
Gregory-Pereira: Yeah sorry I was treating this PR as a Work in Progress so was focused on fixing implementing the feature, but these points should now be adressed.

osherdp:
@gdbranco @renan-campos

cmd/create/cluster/cmd.go
```diff
@@ -490,6 +522,152 @@ func init() {
...
+	flags.StringVar(
+		&args.scaleDown.UnneededTime,
+		"scale-down-unneeded-time",
+		"",
+		"How long a node should be unneeded before it is eligible for scale down.",
+	)
+
+	flags.Float64Var(
+		&args.scaleDown.UtilizationThreshold,
+		"scale-down-utilization-threshold",
```
gdbranco, line 642: Please add these as constants instead to make sure the consts are used when setting up the manual command for creating the cluster with the same config, and everywhere else where it references these values

cmd/create/cluster/cmd.go
```diff
@@ -2020,6 +2198,27 @@ func run(cmd *cobra.Command, _ []string) {
 		isHostedCP,
 		multiAZ)
 
+	var isClusterAutoscalerSet bool
```
gdbranco, line 2201: Should this be considered true by default if there are changes to any of the autoscaler params from the args?

cmd/create/cluster/cmd.go
```diff
@@ -2067,6 +2266,382 @@ func run(cmd *cobra.Command, _ []string) {
 			r.Reporter.Errorf("%s", err)
 			os.Exit(1)
 		}
+
+		// Autoscaler configurations
+		isClusterAutoscalerSet, err = interactive.GetBool(interactive.Input{
```
gdbranco, line 2271: Any way to consider this as set without going through interactive?

cmd/create/cluster/cmd.go
Gregory-Pereira: Not sure If I am understanding this correctly. These are CLI flags passed by the user and therefore variable. Could you expand on how I would set them as constants?

cmd/create/cluster/cmd.go
Gregory-Pereira: I changed this to pull out all the variables that check if the flags got changed and use that to determine if autoscaler is set. Let me know if this implementation does not work.

cmd/create/cluster/cmd.go
Gregory-Pereira: Should be resolved same as above, let me know if this does not work.

Gregory-Pereira:
@osherdp I planned to keep create and delete together in a singular PR (effectively creating update/patch). Would this be implemented on openshift-online/ocm-sdk-go/clustersmgmt/v1/cluster_autoscaler/cluster_type.go / would (this)[https://github.com/openshift-online/ocm-sdk-go/pull/821/files] be a potential solution in tandem with the autoscaler package I have created?

cmd/edit/cluster/cmd.go
```diff
@@ -48,10 +48,42 @@ var args struct {
 	noProxySlice              []string
 	additionalTrustBundleFile string
 
+	// Autoscaler options
```
osherdp, line 51: editing the autoscaler cannot be done from the cluster object, so it should only be possible from the dedicated autoscaler resource (meaning only by ``dlt/autoscaler/``

cmd/create/cluster/cmd.go
osherdp: Done!

cmd/create/cluster/cmd.go
osherdp: I'm changing the implementation a bit, because currently it doesn't let the user to set the variables interactively unless one of them were passed at the CLI

ciaranRoche:
@osherdp @Gregory-Pereira the ticket linked says day 2, these changes are just on cluster create?

osherdp:
> @osherdp @Gregory-Pereira the ticket linked says day 2, these changes are just on cluster create?

yeah it's a mistake. I'll update the jira ticket

cmd/create/cluster/cmd.go
```diff
@@ -2114,6 +2365,364 @@ func run(cmd *cobra.Command, _ []string) {
...
+			}
+		}
+
+		if isClusterAutoscalerSet {
+			if interactive.Enabled() && !isAutoscalerBalanceSimilarNodeGroupsSet {
+				autoscalerBalanceSimilarNodeGroups, err = interactive.GetBool(interactive.Input{
+					Question: "Balance similar node groups",
+					Help:     cmd.Flags().Lookup(autoscalerBalanceSimilarNodeGroupsFlag).Usage,
+					Default:  false,
+					Required: false,
```
gdbranco, line 2388: All of them are set of not required. If isClusterAutoscalerSet is true should they be required? Today the customer has the option to set `-y` and the non required fields would assume default or empty values. Has that been checked to cover if needed?

osherdp: From what I'm seeing, ``-y`` is only supposed to be used for automatically answering "yes" to confirm operations. Those questions are mainly asking for values rather than confirmation, so I think it shouldn't auto-set them to the defaults in this case

gdbranco: `--mode auto -y` usually skips interactive mode as a whole :thinking: . Maybe there are some edge cases around it that I'm not familiar with that still goes into interactive, but I have the expectation that is not the case

cc @ciaranRoche @oriAdler Would you have some extra context here?

osherdp: hmm, also ``--mode auto`` mentions handling of AWS resources iiuc: https://github.com/openshift/rosa/blob/49ef038958eef794bea1d6e3f9065f9367b6e318/pkg/aws/flag.go#L41

would you say it's a blocker for the PR, or rather an enhancement that can be made in a follow-up?
Either way I'll look into that, for sure it can also help me with some of the manual testing that I'm doing :sweat_smile:

oriAdler: As Guilherme said, the current behavior is that if the customer is running:
```
oadler@fedora:rosa (master)$ ./rosa create cluster --cluster-name oadler-test --mode auto -y
W: --mode is only valid for STS clusters
W: --mode is only valid for STS clusters
I: Creating cluster 'oadler-test'
I: To view a list of clusters and their status, run 'rosa list clusters'
I: Cluster 'oadler-test' has been created.
I: Once the cluster is installed you will need to add an Identity Provider before you can login into the cluster. See 'rosa create idp --help' for more information.

Name:                       oadler-test
ID:                         25lsfc4po2h6grn0kd3hgcguiur9ljik
External ID:                
OpenShift Version:          
DNS:                        Not ready
API URL:                    
Console URL:                
Region:                     us-east-1
Nodes:
 - Control plane:           3
 - Infra:                   2
 - Compute:                 4
Network:
 - Type:                    OVNKubernetes
 - Service CIDR:            172.30.0.0/16
 - Machine CIDR:            10.0.0.0/16
 - Pod CIDR:                10.128.0.0/14
 - Host Prefix:             /23
Ec2 Metadata Http Tokens:   optional
State:                      pending (Preparing account)
Created:                    Aug 17 2023 12:42:27 UTC
Details Page:               https://qaprodauth.console.redhat.com/openshift/details/s/2U729cFTwWK0TWexjDCLso2jvLa

I: To determine when your cluster is Ready, run 'rosa describe cluster -c oadler-test'.
I: To watch your cluster installation logs, run 'rosa logs install -c oadler-test --watch'.
```


It skips the interactive mode and sets the default values.
I would prefer to be consistent with the current ROSA CLI behavior unless there is a good reason to change it.

osherdp: ack. either way I can verify that running the following command:
```
./rosa create cluster --cluster-name odepaz-asg --autoscaler-scale-down-enabled --region us-west-1 --availability-zones us-west-1a --mode auto -y
```

does result with the defaults being automatically set for the cluster:
```
$ ocm get /api/clusters_mgmt/v1/clusters/<id>/autoscaler
{
  "kind": "ClusterAutoscaler",
  "href": "/api/clusters_mgmt/v1/clusters/<id>/autoscaler",
  "balance_similar_node_groups": false,
  "skip_nodes_with_local_storage": true,
  "log_verbosity": 1,
  "max_pod_grace_period": 0,
  "pod_priority_threshold": 0,
  "ignore_daemonsets_utilization": false,
  "resource_limits": {
    "max_nodes_total": 1000,
    "cores": {
      "min": 0,
      "max": 100
    },
    "memory": {
      "min": 0,
      "max": 4096
    }
  },
  "scale_down": {
    "enabled": true,
    "utilization_threshold": "0.500000"
  }
}
```

gdbranco:
SGTM, can we squash before merging?

osherdp:
> SGTM, can we squash before merging?

sure. done!

PR #1387, opened 2023-07-27
OCM-2910 | fix: Adding default value for http tokens

tsorya:
We should set http tokens default value in order for user to see which version of ec2 metadata service he is using

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -805,6 +805,12 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	// setting default for ec2MetadataHttpTokens in case cluster is not hosted
+	// or ec2MetadataHttpTokens was not set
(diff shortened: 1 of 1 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -803,6 +803,9 @@ func run(cmd *cobra.Command, _ []string) {
 	if isHostedCP && args.ec2MetadataHttpTokens != "" {
 		r.Reporter.Errorf("ec2-metadata-http-tokens can't be set with hosted-cp")
 		os.Exit(1)
+	} else if !isHostedCP && args.ec2MetadataHttpTokens == "" {
```
oriAdler, line 806: NP
We can simplify the condition:
```suggestion
	if !isHostedCP && args.ec2MetadataHttpTokens == "" {
```
Do we need the `else` here? if the previous condition didn't apply we know that:
`isHostedCp == false` or `args.ec2MetadataHttpTokens == ""`

PR #1399, opened 2023-08-01
OCM-2276 | fix: more clarity into why it might have not found the prefix

gdbranco:
Related issue: https://issues.redhat.com/browse/OCM-2276


thomasmckay:
lgtm

PR #1417, opened 2023-08-09
OCM-3143 | Create operator-roles - ensure shared VPC policy

oriAdler:
Multiple operator roles point to the same policy.
When creating a new set of operator roles with the `--shared-vpc-role-arn` flag provided, 
we need to ensure the ingress operator role is updated with the new policy.

### DIFF
--- a/cmd/create/operatorroles/by_prefix.go
+++ b/cmd/create/operatorroles/by_prefix.go
@@ -312,7 +312,7 @@ func createRolesByPrefix(r *rosa.Runtime, prefix string, permissionsBoundary str
 				tags.OperatorName:      operator.Name(),
 			}
 
-			if args.forcePolicyCreation {
+			if args.forcePolicyCreation || (isSharedVpc && credrequest == aws.IngressOperatorCloudCredentialsRoleType) {
 				_, err := r.AWSClient.ForceEnsurePolicy(policyArn, policyDetails,
 					defaultPolicyVersion, operatorPolicyTags, path)
 				if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/create/operatorroles/by_prefix.go
```diff
@@ -312,7 +312,7 @@ func createRolesByPrefix(r *rosa.Runtime, prefix string, permissionsBoundary str
 				tags.OperatorName:      operator.Name(),
 			}
 
-			if args.forcePolicyCreation {
+			if args.forcePolicyCreation || (isSharedVpc && credrequest == aws.IngressOperatorCloudCredentialsRoleType) {
```
gdbranco, line 315: Didn't quite follow why this is needed

oriAdler: Operator role policy is created once per account role, with the account role prefix, afterward multiple operator roles are using the same policy.

If you have created a cluster with a set of account roles, and then trying to create operator roles for the shared VPC flow, you'll be missing this permission https://github.com/openshift/managed-cluster-config/blob/master/resources/sts/4.13/shared_vpc_openshift_ingress_operator_cloud_credentials_policy.json#L31
Therefore, we have to create a policy with the `sts:AssumeRole` permission.

gdbranco: I gotcha, but wouldn't that be the point of having the force option? By simply supplying the sharedvpc options I feel like we should go towards the basic flow and if they supply both we would move towards recreating.

oriAdler: Yes, on the one hand, I prefer to keep it simple as you suggest, the user can supply `-f` and we are done:)

On the other hand, for the shared VPC we should **create a policy version only for the ingress operator**, also wanted to keep it transparent to the user, as the flow is already complex, to do it for them. They might forget to set the `-f` option...

gdbranco: I suppose they wouldn't need to do it often

PR #1421, opened 2023-08-14
OCM-1338 | fix: use utility function to retrieve acc role suffix

gdbranco:
Related issue: https://issues.redhat.com/browse/OCM-1338


willkutler:
LGTM

PR #1422, opened 2023-08-14
OCM-2865 | feat: support shared vpc in rosa

marcolan018:
https://issues.redhat.com/browse/OCM-2865

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -184,6 +184,11 @@ var args struct {
 
 	// Storage
 	machinePoolRootDiskSize string
+
+	// Shared VPC
+	privateHostedZoneID string
+	sharedVPCRoleARN    string
+	baseDomain          string
 }
 
 var Cmd = &cobra.Command{
@@ -693,6 +698,27 @@ The password must
 			strings.Join(ingress.ValidNamespaceOwnershipPolicies, ","), ingress.DefaultNamespaceOwnershipPolicy),
 	)
 
+	flags.StringVar(
+		&args.privateHostedZoneID,
+		"private-hosted-zone-id",
+		"",
+		`ID of shared private hosted zone`,
+	)
+
+	flags.StringVar(
+		&args.sharedVPCRoleARN,
+		"shared-vpc-role-arn",
+		"",
+		`Role ARN for permissions in the shared private hosted zone`,
+	)
+
+	flags.StringVar(
+		&args.baseDomain,
+		"base-domain",
+		"",
+		`Base domain name for the shared private hosted zone`,
+	)
+
 	aws.AddModeFlag(Cmd)
 	interactive.AddFlag(flags)
 	output.AddFlag(Cmd)
@@ -706,6 +732,23 @@ func networkTypeCompletion(cmd *cobra.Command, args []string, toComplete string)
 func run(cmd *cobra.Command, _ []string) {
 	r := rosa.NewRuntime().WithAWS().WithOCM()
 	defer r.Cleanup()
+
+	// validate flags for shared vpc
+	isSharedVPC := false
+	privateHostedZoneID := strings.Trim(args.privateHostedZoneID, " \t")
+	sharedVPCRoleARN := strings.Trim(args.sharedVPCRoleARN, " \t")
+	baseDomain := strings.Trim(args.baseDomain, " \t")
+	if privateHostedZoneID != "" ||
+		sharedVPCRoleARN != "" {
+		if privateHostedZoneID == "" ||
+			sharedVPCRoleARN == "" ||
+			baseDomain == "" {
+			r.Reporter.Errorf("To install a cluster into a shared VPC, " +
+				"private-hosted-zone-id, shared-vpc-role-arn and base-domain should be provided together")
+			os.Exit(1)
+		}
+		isSharedVPC = true
+	}
 
 	supportedRegions, err := r.OCMClient.GetDatabaseRegionList()
 	if err != nil {
@@ -1874,6 +1917,11 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 	r.Reporter.Debugf("Found the following availability zones for the subnets provided: %v", availabilityZones)
 
+	if len(subnetIDs) == 0 && isSharedVPC {
+		r.Reporter.Errorf("Installing a cluster into a shared VPC is only supported for BYO VPC clusters")
+		os.Exit(1)
+	}
+
 	// Select availability zones for a non-BYOVPC cluster
 	var selectAvailabilityZones bool
 	if !useExistingVPC && !subnetsProvided {
@@ -2645,6 +2693,11 @@ func run(cmd *cobra.Command, _ []string) {
 		clusterConfig.ClusterAdminUser = clusterAdminUser
 		clusterConfig.ClusterAdminPassword = clusterAdminPassword
 	}
+	if isSharedVPC {
+		clusterConfig.PrivateHostedZoneID = privateHostedZoneID
+		clusterConfig.SharedVPCRoleArn = sharedVPCRoleARN
+		clusterConfig.BaseDomain = baseDomain
+	}
 
 	props := args.properties
 	if args.fakeCluster {
@@ -3067,6 +3120,11 @@ func buildCommand(spec ocm.Spec, operatorRolesPrefix string,
 	}
 	if len(spec.SubnetIds) > 0 {
 		command += fmt.Sprintf(" --subnet-ids %s", strings.Join(spec.SubnetIds, ","))
+	}
+	if spec.PrivateHostedZoneID != "" {
+		command += fmt.Sprintf(" --private-hosted-zone-id %s", spec.PrivateHostedZoneID)
+		command += fmt.Sprintf(" --shared-vpc-role-arn %s", spec.SharedVPCRoleArn)
+		command += fmt.Sprintf(" --base-domain %s", spec.BaseDomain)
 	}
 	if spec.FIPS {
 		command += " --fips"
--- a/pkg/ocm/clusters.go
+++ b/pkg/ocm/clusters.go
@@ -143,6 +143,11 @@ type Spec struct {
 
 	// Machine pool's storage
 	MachinePoolRootDisk *Volume
+
+	// Shared VPC
+	PrivateHostedZoneID string
+	SharedVPCRoleArn    string
+	BaseDomain          string
 }
 
 // Volume represents a volume property for a disk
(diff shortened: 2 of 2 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -1917,6 +1943,113 @@ func run(cmd *cobra.Command, _ []string) {
...
 	}
 
+	isSharedVPC := false
+	privateHostedZoneID := strings.Trim(args.privateHostedZoneID, " \t")
+	privateHostedZoneRoleARN := strings.Trim(args.privateHostedZoneRoleARN, " \t")
+	baseDomain := strings.Trim(args.baseDomain, " \t")
+	if privateHostedZoneID != "" ||
+		privateHostedZoneRoleARN != "" ||
+		baseDomain != "" {
+		if !isSTS {
```
oriAdler, line 1953: We're adding this check to the backend here https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/6333, I think it's ok to remove it from the CLI.
If the feature will be supported for non-STS, we can easily change the backend, unlike the CLI, where we need a release.

cmd/create/cluster/cmd.go
```diff
@@ -1917,6 +1943,113 @@ func run(cmd *cobra.Command, _ []string) {
...
+			r.Reporter.Errorf("Installing a cluster into a shared VPC is only supported for STS clusters")
+			os.Exit(1)
+		}
+		if len(subnetIDs) == 0 {
+			r.Reporter.Errorf("Installing a cluster into a shared VPC is only supported for BYO VPC clusters")
+			os.Exit(1)
+		}
+		if isHostedCP {
```
oriAdler, line 1961: Same here, we have a check in the backend, I assume at a certain point it will be supported for HCP. I think it's ok to remove the validation, as the error will be returned from the backend.

cmd/create/cluster/cmd.go
```diff
@@ -1917,6 +1943,113 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
+	isSharedVPC := false
+	privateHostedZoneID := strings.Trim(args.privateHostedZoneID, " \t")
+	privateHostedZoneRoleARN := strings.Trim(args.privateHostedZoneRoleARN, " \t")
+	baseDomain := strings.Trim(args.baseDomain, " \t")
+	if privateHostedZoneID != "" ||
+		privateHostedZoneRoleARN != "" ||
+		baseDomain != "" {
```
oriAdler, line 1952: Currently, the way we identify a shared VPC in the backend is:
```
func (c *Cluster) IsSharedVpc() bool {
	return c.AWS != nil && (c.AWS.PrivateHostedZoneID != nil || c.AWS.PrivateHostedZoneRoleARN != nil)
}
```
It is true that `baseDomain` is required as well, but there is another effort to enable the user to provide `baseDomain` outside of the scope of the shared VPC. So I think it's ok to check only `privateHostedZoneID` or `privateHostedZoneRoleARN`.

cmd/create/cluster/cmd.go
```diff
@@ -1917,6 +1943,113 @@ func run(cmd *cobra.Command, _ []string) {
...
+			r.Reporter.Errorf("Installing a cluster into a shared VPC is only supported for BYO VPC clusters")
+			os.Exit(1)
+		}
+		if isHostedCP {
+			r.Reporter.Errorf("Installing a cluster into a shared VPC is only supported for classic ROSA clusters")
+			os.Exit(1)
+		}
+		isSharedVPC = true
+	}
+	if (privateHostedZoneID == "" ||
```
oriAdler, line 1967: Per Bala's comment here https://docs.google.com/document/d/1f97Xwube-KYscLXQN3SnOMTH3IeK5JAfseJk_vEnlT0/edit?disco=AAAA1ZP5dRU we can identify if the subnets belong to a shared VPC and prompt the interactive questions accordingly.
I think the assumption is that 95% of customers won't use the shared VPC at first, and it can confuse them, therefore, we want to expose it only if they actively shared subnets and chose them.

If possible we should defer the interactive mode handling to a separate PR, and concentrate in this release on enabling the flow via flags for the MVP. Also, we should finalize with BU how is it going to look like for interactive.

cmd/create/cluster/cmd.go
```diff
@@ -1917,6 +1943,113 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
+	isSharedVPC := false
+	privateHostedZoneID := strings.Trim(args.privateHostedZoneID, " \t")
```
oriAdler, line 1947: Maybe we can add a validation that if `privateHostedZoneID` or `privateHostedZoneRoleARN` are provided all three new flags should be provided together.
We check this combination in the backend as well, but since it's simple logic, I think it's ok to validate in the CLI.
https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/cmd/clusters-service/servecmd/apiserver/validation_helpers.go#L2180
@ciaranRoche WDYT?

pkg/ocm/clusters.go
```diff
@@ -929,6 +934,19 @@ func (c *Client) createClusterSpec(config Spec, awsClient aws.Client) (*cmv1.Clu
 		awsBuilder = awsBuilder.EtcdEncryption(cmv1.NewAwsEtcdEncryption().KMSKeyARN(config.EtcdEncryptionKMSArn))
 	}
 
+	// shared vpc
+	if config.PrivateHostedZoneID != "" {
+		awsBuilder = awsBuilder.PrivateHostedZoneID(config.PrivateHostedZoneID)
+		clusterProperties["install_config"] = "featureSet: TechPreviewNoUpgrade"
```
oriAdler, line 940: We can remove this property, it applied only for the 4.14 nightly version.

cmd/create/cluster/cmd.go
ciaranRoche: Lets keep the validations to a min in the client, until everything is worked out. For MVP I think its okay to have this in the backend for now. Better to report late then wrong, and give us some freedom to change if needed and not tied to a release. 

@oriAdler lets set a call for tomorrow, and we capture these things in an improvements epic

cmd/create/cluster/cmd.go
ciaranRoche: i agree with that, we should find out first how people are consuming this to better judge how we want to prompt them with out causing confusion. Even the wording shared VPC is confusing. 

Lets capture this in an improvements epic

cmd/create/cluster/cmd.go
marcolan018: remove the check

cmd/create/cluster/cmd.go
marcolan018: removed the check

cmd/create/cluster/cmd.go
marcolan018: updated logic here

pkg/ocm/clusters.go
marcolan018: removed

cmd/create/cluster/cmd.go
```diff
@@ -693,6 +698,27 @@ The password must
...
+		"private-hosted-zone-role-arn",
+		"",
+		`Role ARN for permissions in private hosted zone`,
+	)
+
+	flags.StringVar(
+		&args.baseDomain,
+		"base-domain",
+		"",
+		`Base domain name for the private hosted zone`,
```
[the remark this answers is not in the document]
marcolan018: thanks for the comments, revised the description

cmd/create/cluster/cmd.go
marcolan018: removed the changes for interactive mode

cmd/create/cluster/cmd.go
```diff
@@ -1917,6 +1943,26 @@ func run(cmd *cobra.Command, _ []string) {
...
+	isSharedVPC := false
+	privateHostedZoneID := strings.Trim(args.privateHostedZoneID, " \t")
+	privateHostedZoneRoleARN := strings.Trim(args.privateHostedZoneRoleARN, " \t")
+	baseDomain := strings.Trim(args.baseDomain, " \t")
+	if privateHostedZoneID != "" ||
+		privateHostedZoneRoleARN != "" {
+		if privateHostedZoneID == "" ||
+			privateHostedZoneRoleARN == "" ||
+			baseDomain == "" {
+			r.Reporter.Errorf("To install a cluster into a shared VPC," +
```
oriAdler, line 1955:
```suggestion
			r.Reporter.Errorf("To install a cluster into a shared VPC, " +
```

cmd/create/cluster/cmd.go
```diff
@@ -1917,6 +1943,26 @@ func run(cmd *cobra.Command, _ []string) {
...
+				"private-hosted-zone-id and private-hosted-zone-role-arn should be provided together")
```
oriAdler, line 1956:
```suggestion
				"private-hosted-zone-id, private-hosted-zone-role-arn , and base-domain should be provided together")
```

cmd/create/cluster/cmd.go
oriAdler: I would suggest having the flags validation at the beginning of the command before the user goes through all the stages.

cmd/create/cluster/cmd.go
ciaranRoche, line 1956: The error does not match the check, we are also checking for the `Domain`

marcolan018: @oriAdler 
Some confused by this comment.
you mentioned above

It is true that baseDomain is required as well, but there is another effort to enable the user to provide baseDomain outside of the scope of the shared VPC. So I think it's ok to check only privateHostedZoneID or privateHostedZoneRoleARN.

so I removed the requirement for base-domain.  Any misunderstanding here?

cmd/create/cluster/cmd.go
marcolan018: make sense. And we can remove it directly once we have the interactive mode support later. Will make this change

cmd/create/cluster/cmd.go
marcolan018: fixed

cmd/create/cluster/cmd.go
marcolan018: made the change

cmd/create/cluster/cmd.go
```diff
@@ -693,6 +698,27 @@ The password must
...
+	flags.StringVar(
+		&args.privateHostedZoneID,
+		"private-hosted-zone-id",
+		"",
+		`ID of shared private hosted zone`,
+	)
+
+	flags.StringVar(
+		&args.privateHostedZoneRoleARN,
+		"private-hosted-zone-role-arn",
```
oriAdler, line 710: I wonder if we should stick with the naming convention we have in `create operator-roles` which is `--shared-vpc-role-arn` because it's the same value in both commands.

oriAdler (commented):
Left a small comment, overall LGTM.
Just we need BU's input about the help messages and we can merge.

PR #1424, opened 2023-08-16
WIP OCM-210 | feat: Allow setting cluster-autoscaler on cluster creation

osherdp:
Dupe of https://github.com/openshift/rosa/pull/1371 as I don't have permissions to push to that other branch.

This should allow setting cluster-autoscaler when creating a ROSA cluster. It does not address the following:
* Allowing day2 operations, while it's fully supported in the server-side
* Setting GPU limitations. This struct is a bit more complex so I would refrain from implementing it at the moment.


osherdp:
sorry, we'll continue in https://github.com/openshift/rosa/pull/1371
/close

PR #1426, opened 2023-08-16
OCM-2892 | fix: Adding default value of http tokens metadata for describe command

tsorya:
We should always show value for http tokens and currently backend (cause of backward compatibility issues) will still return value as empty string


ciaranRoche (commented):
/lgtm

Merging as work around until we solve issue blocking this from being set in the backend.

PR #1428, opened 2023-08-16
OCM-2611 / fix: Added check for duplcate deletion of same cluster

den-rgb:
[OCM-2611](https://issues.redhat.com/browse/OCM-2611)

# DONE

Added log statement to inform the client that the cluster is already being uninstalled if a duplicate removal of the same cluster is executed.
Deletion code is skipped if cluster is in Uninstall stage.

# REVIEW

- Build rosa cli binary locally `make rosa`
- Create a rosa cluster 
Replace:
- - Name
- - Prefix

``` 
rosa create cluster --cluster-name <name> --sts --role-arn arn:aws:iam::765374464689:role/ManagedOpenShift-Installer-Role --support-role-arn arn:aws:iam::765374464689:role/ManagedOpenShift-Support-Role --controlplane-iam-role arn:aws:iam::765374464689:role/ManagedOpenShift-ControlPlane-Role --worker-iam-role arn:aws:iam::765374464689:role/ManagedOpenShift-Worker-Role --operator-roles-prefix <prefix> --region eu-west-1 --version 4.12.28 --replicas 2 --compute-machine-type m5.xlarge --machine-cidr 10.0.0.0/16 --service-cidr 172.30.0.0/16 --pod-cidr 10.128.0.0/14 --host-prefix 23
```

- Delete the rosa cluster `rosa delete cluster --cluster <name>`
- Delete the rosa cluster again 

The logs should return an info statement saying the cluster is already being uninstalled.

image

### DIFF
--- a/cmd/dlt/cluster/cmd.go
+++ b/cmd/dlt/cluster/cmd.go
@@ -72,13 +72,13 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(0)
 	}
 
-	r.Reporter.Debugf("Deleting cluster '%s'", clusterKey)
-	cluster, err := r.OCMClient.DeleteCluster(clusterKey, r.Creator)
+	cluster := r.FetchCluster()
+
+	err := handleClusterDelete(r, cluster, clusterKey)
 	if err != nil {
 		r.Reporter.Errorf("%s", err)
 		os.Exit(1)
 	}
-	r.Reporter.Infof("Cluster '%s' will start uninstalling now", clusterKey)
 
 	if cluster.AWS().STS().RoleARN() != "" {
 		interactive.Enable()
@@ -108,6 +108,27 @@ func run(cmd *cobra.Command, _ []string) {
 			clusterKey,
 		)
 	}
+}
+
+func handleClusterDelete(r *rosa.Runtime, cluster *cmv1.Cluster, clusterKey string) error {
+	clusterState, err := r.OCMClient.GetClusterState(cluster.ID())
+	if err != nil {
+		return err
+	}
+
+	if clusterState == cmv1.ClusterStateUninstalling {
+		r.Reporter.Infof("cluster '%s' is already uninstalling", clusterKey)
+		return nil
+	}
+
+	r.Reporter.Debugf("Deleting cluster '%s'", clusterKey)
+	_, err = r.OCMClient.DeleteCluster(clusterKey, r.Creator)
+	if err != nil {
+		return err
+	}
+
+	r.Reporter.Infof("Cluster '%s' will start uninstalling now", clusterKey)
+	return nil
 }
 
 func buildCommands(cluster *cmv1.Cluster) string {


cmd/dlt/cluster/cmd.go
```diff
@@ -110,6 +110,27 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 }
 
+func checkDeletion(r *rosa.Runtime, cluster *cmv1.Cluster, clusterKey string) string {
```
ciaranRoche, line 113: this is doing more than checking the deletion, this is actually handling the deletion. I would suggest renaming this fuction to be more descriptive of what it is doing, even just `handleClusterDelete` or something like that

cmd/dlt/cluster/cmd.go
```diff
@@ -110,6 +110,27 @@ func run(cmd *cobra.Command, _ []string) {
...
+    clusterState, err := r.OCMClient.GetClusterState(cluster.ID())
+    if err != nil {
+        r.Reporter.Errorf("%s", err)
```
ciaranRoche, line 116: Lets return errors from this function, and leave the caller handle the error.

cmd/dlt/cluster/cmd.go
```diff
@@ -110,6 +110,27 @@ func run(cmd *cobra.Command, _ []string) {
...
 
+func handleClusterDelete(r *rosa.Runtime, cluster *cmv1.Cluster, clusterKey string) error {
+	clusterState, err := r.OCMClient.GetClusterState(cluster.ID())
+	if err != nil {
+		r.Reporter.Errorf("%s", err)
+		os.Exit(1)
+	}
+
+	if clusterState == cmv1.ClusterStateUninstalling {
+		r.Reporter.Infof("cluster '%s' is already uninstalling", clusterKey)
```
ciaranRoche, line 121: need to return nil after this

cmd/dlt/cluster/cmd.go
```diff
@@ -110,6 +110,27 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 }
 
+func handleClusterDelete(r *rosa.Runtime, cluster *cmv1.Cluster, clusterKey string) error {
+	clusterState, err := r.OCMClient.GetClusterState(cluster.ID())
+	if err != nil {
+		r.Reporter.Errorf("%s", err)
```
ciaranRoche, line 116: return the error, instead of handling it here

cmd/dlt/cluster/cmd.go
```diff
@@ -72,13 +72,13 @@ func run(cmd *cobra.Command, _ []string) {
...
-	r.Reporter.Debugf("Deleting cluster '%s'", clusterKey)
-	cluster, err := r.OCMClient.DeleteCluster(clusterKey, r.Creator)
+	cluster, err := r.OCMClient.GetCluster(clusterKey, r.Creator)
 	if err != nil {
 		r.Reporter.Errorf("%s", err)
 		os.Exit(1)
 	}
-	r.Reporter.Infof("Cluster '%s' will start uninstalling now", clusterKey)
+
+	handleClusterDelete(r, cluster, clusterKey)
```
ciaranRoche, line 81: need to handle the error returned from this function

cmd/dlt/cluster/cmd.go
```diff
@@ -72,13 +72,17 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(0)
 	}
 
-	r.Reporter.Debugf("Deleting cluster '%s'", clusterKey)
-	cluster, err := r.OCMClient.DeleteCluster(clusterKey, r.Creator)
+	cluster, err := r.OCMClient.GetCluster(clusterKey, r.Creator)
```
gdbranco, line 75: I think you are able to rely on fetchCluster instead 
https://github.com/openshift/rosa/blob/master/cmd/describe/cluster/cmd.go#L69


Not all commands are using it yet but I think we should be moving towards it

gdbranco:
lgtm

PR #1429, opened 2023-08-16
OCM-3284 | fix: allow to choose load balancer type for sts clusters

gdbranco:
Related issue: https://issues.redhat.com/browse/OCM-3284

### DIFF
--- a/cmd/edit/ingress/cmd.go
+++ b/cmd/edit/ingress/cmd.go
@@ -284,30 +284,22 @@ func run(cmd *cobra.Command, argv []string) {
 			r.Reporter.Errorf("Updating Load Balancer Type is not supported for Hosted Control Plane clusters")
 			os.Exit(1)
 		}
-		if ocm.IsSts(cluster) {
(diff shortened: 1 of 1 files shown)


cmd/edit/ingress/cmd.go
```diff
@@ -284,30 +284,22 @@ func run(cmd *cobra.Command, argv []string) {
...
 			os.Exit(1)
 		}
-		if ocm.IsSts(cluster) {
-			r.Reporter.Errorf("Updating Load Balancer Type is not supported for STS clusters")
-			os.Exit(1)
-		}
 		lbType = &args.lbType
-	} else {
-		if interactive.Enabled() {
-			if !ocm.IsSts(cluster) {
```
oriAdler, line 294: With this check removed, do we need to check `!ocm.IsHyperShiftCluster(cluster)` before prompting the message?

gdbranco: Oh true

gdbranco: Fixed, thanks!

PR #1432, opened 2023-08-17
OCM-2906 | fix: ROSA CLI for IMDSv2 - Help displays same prompt instead of anything helpful

tsorya:
[OCM-2906](https://issues.redhat.com//browse/OCM-2906) | fix: ROSA CLI for IMDSv2 - Help displays same prompt instead of anything helpful

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -443,7 +443,8 @@ func init() {
 		&args.ec2MetadataHttpTokens,
 		"ec2-metadata-http-tokens",
 		"",
-		"Configure the use of IMDSv2 for ec2 instances, 'optional' or 'required'.",
+		"Should cluster nodes use both v1 and v2 endpoints or just v2 endpoint "+
+			"of EC2 Instance Metadata Service (IMDS)",
 	)
 
 	flags.StringSliceVar(


cmd/create/cluster/cmd.go
```diff
@@ -1015,8 +1015,8 @@ func run(cmd *cobra.Command, _ []string) {
 	httpTokens := args.ec2MetadataHttpTokens
 	if interactive.Enabled() && !isHostedCP {
 		httpTokens, err = interactive.GetString(interactive.Input{
-			Question: fmt.Sprintf("Configure the use of IMDSv2 for ec2 instances %s/%s",
-				v1.Ec2MetadataHttpTokensOptional, v1.Ec2MetadataHttpTokensRequired),
+			Question: "Should cluster nodes use both v1 and v2 endpoints or just v2 endpoint " +
+				"of EC2 Instance Metadata Service (IMDS)",
```
ciaranRoche, line 1019: I dont think this is what is asked in the JIRA. 

I think we need to update the help text and not the question.

ciaranRoche:
```
╰─$ rosa create cluster
I: Enabling interactive mode
? Cluster name: test
? Deploy cluster with Hosted Control Plane (optional): No
? Create cluster admin user: No
? Deploy cluster using AWS STS: Yes
? OpenShift version: 4.12.28
? Should cluster nodes use both v1 and v2 endpoints or just v2 endpoint of EC2 Instance Metadata Service (IMDS)
? Configure the use of IMDSv2 for ec2 instances optional/required: (optional) 
```

ciaranRoche (approved):
/lgtm

PR #1438, opened 2023-08-18
OCM-210 | fix: adding GPULimits to autoscaler configuration

Gregory-Pereira:
/cc @osherdp 
/cc @screeley44

This PR introduces support for GPULimits back into scope for [OCM-210](https://issues.redhat.com/browse/OCM-210).

PR #1439, opened 2023-08-18
OCM-210 | fix: adding GPULimits to autoscaler configuration

Gregory-Pereira:
Stable version of #1438. Making this one the production PR and converting other one to draft until I confirm that deadline is end of the 21st, which would give me enough time to finish the full implementation of the feature and get it reviewed and merged Monday for `1.2.26`.

/cc @osherdp
/cc @screeley44

This PR introduces support for GPULimits back into scope for [OCM-210](https://issues.redhat.com/browse/OCM-210).

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -88,6 +88,7 @@ const (
 	autoscalerScaleDownDelayAfterAddFlag        = "autoscaler-scale-down-delay-after-add"
 	autoscalerScaleDownDelayAfterDeleteFlag     = "autoscaler-scale-down-delay-after-delete"
 	autoscalerScaleDownDelayAfterFailureFlag    = "autoscaler-scale-down-delay-after-failure"
+	autoscalerGPULimitsFlag                     = "autoscaler-gpu-limits"
 
 	MinReplicasSingleAZ = 2
 	MinReplicaMultiAZ   = 3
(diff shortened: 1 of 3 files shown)


ciaranRoche:
@Gregory-Pereira @osherdp how does setting GPU limits work, if a user has not selected a GPU node for the cluster? 

I am just wondering if this is something we should be validating before asking for user input.

osherdp:
> @Gregory-Pereira @osherdp how does setting GPU limits work, if a user has not selected a GPU node for the cluster?
> 
> I am just wondering if this is something we should be validating before asking for user input.

Those nodes might be added in the future of cluster-creation (i.e. on day2), so there's no good way to validate it at the time we set this parameter.

Actually it applies to other parameters as well. max-nodes-total is somewhat related to the max-replicas fields of machine-pools (it's a cluster-wide restriction whereas max-replicas are at the scope of a machine-pool). All in all it's cluster-wide configuration, so it's ok if not every configuration is yet applicable on cluster-creation (and can only be applicable in the future)

ciaranRoche:
@osherdp can we close this if its been replaced by - https://github.com/openshift/rosa/pull/1473

osherdp:
> @osherdp can we close this if its been replaced by - https://github.com/openshift/rosa/pull/1473

Sure!
/close

PR #1441, opened 2023-08-21
OCM-3336 | fix: Describe cluster compute node shows wrong value as we should not look on cluster.nodes.compute anymore

tsorya:
[OCM-3336](https://issues.redhat.com//browse/OCM-3336): Describe cluster compute node shows wrong value as we should not look on cluster.nodes.compute anymore


ciaranRoche (approved):
/lgtm

PR #1442, opened 2023-08-22
OCM-3356 | fix: change the wording for autoscaler params

osherdp:
It almost seems like the 'help' option ('?') doesn't show more information for some parameters.

Changing it into a bit more comprehensive messages.


osherdp:
/retest
seems like an infra problem. checking if it persists

PR #1443, opened 2023-08-22
OCM-3353 | fix: utilization-threshold missing from interactive mode

osherdp:
Because of a redundant check, this parmeters is not configurable from the interactive mode for cases where the user didn't set it via CLI (``--autoscaler-scale-down-utilization-threshold``).

I also fixed the validation because it wasn't actually been tested.

The changes for ``delay-after-*`` parameters are mainly refactoring. If a parameter has not been set it's also necessarily empty so checking ``string == ""`` is pretty much redundant.


ciaranRoche (approved):
/lgtm

PR #1446, opened 2023-08-22
OCM-3357 | fix: add a min/max validation

osherdp:
Making sure that when the user gives some value for a max parameter it's greater than or equal to a previous min value that was given.


ciaranRoche (approved):
/lgtm

PR #1452, opened 2023-08-23
OCM-3392 | fix: don't show RH managed VPC options when selecting subnets

gdbranco:
Related issue: https://issues.redhat.com/browse/OCM-3392

```
? Install into an existing VPC (optional): Yes
W: The following subnets were excluded because they belong to a VPC that is managed by Red Hat: [subnet-xxx, subnet-xxx, subnet-xx, subnet-xxx, subnet-xxx, subnet-xxx, subnet-xxx, subnet-xxx, subnet-xxx, subnet-xxx, subnet-xxx, subnet-xxx]
W: The following subnets have been excluded because they do not fit into chosen CIDR ranges: [subnet-xxx, subnet-xxx]
```

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -42,6 +42,7 @@ import (
 	installLogs "github.com/openshift/rosa/cmd/logs/install"
 	"github.com/openshift/rosa/pkg/arguments"
 	"github.com/openshift/rosa/pkg/aws"
+	"github.com/openshift/rosa/pkg/aws/tags"
 	"github.com/openshift/rosa/pkg/clusterautoscaler"
 	"github.com/openshift/rosa/pkg/fedramp"
 	"github.com/openshift/rosa/pkg/helper"
@@ -55,6 +56,7 @@ import (
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/output"
 	"github.com/openshift/rosa/pkg/properties"
+	"github.com/openshift/rosa/pkg/reporter"
 	"github.com/openshift/rosa/pkg/rosa"
 )
 
@@ -1519,13 +1521,13 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	// Custom tags for AWS resources
-	tags := args.tags
+	_tags := args.tags
 	tagsList := map[string]string{}
 	if interactive.Enabled() {
 		tagsInput, err := interactive.GetString(interactive.Input{
 			Question: "Tags",
 			Help:     cmd.Flags().Lookup("tags").Usage,
-			Default:  strings.Join(tags, ","),
+			Default:  strings.Join(_tags, ","),
 			Validators: []interactive.Validator{
 				aws.UserTagValidator,
 				aws.UserTagDuplicateValidator,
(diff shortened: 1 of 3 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -1985,11 +1986,32 @@ func run(cmd *cobra.Command, _ []string) {
...
+			hasRHManaged := false
+			for _, tag := range subnet.Tags {
+				switch *tag.Key {
+				case tags.RedHatManaged:
+					boolValue, err := strconv.ParseBool(*tag.Value)
+					if err != nil {
+						r.Reporter.Debugf("An error occurred whilst parsing bool value of '%s' tag, value '%s'", *tag.Key, *tag.Value)
+						boolValue = false
+					}
+					hasRHManaged = boolValue
```
hunterkepley, line 2005: NP: Do we need `boolValue`? Couldn't we just assign `hasRHManaged` directly here?

hunterkepley:
other than the nitpick, LGTM -- ping me again after you respond to my comment

cmd/create/cluster/cmd.go
gdbranco: In the backend there is a similar check, I expect that if red-hat-managed is available it would be true. But who knows if some user decided to tag or change to false for some reason

cmd/create/cluster/cmd.go
```diff
@@ -3375,3 +3377,32 @@ func getExpectedResourceIDForAccRole(hostedCPPolicies bool, roleARN string, role
...
+		return initialValidSubnets, err
+	}
+	for _, subnet := range allSubnets {
+		hasRHManaged := false
+		for _, tag := range subnet.Tags {
+			switch *tag.Key {
+			case tags.RedHatManaged:
+				boolValue, err := strconv.ParseBool(*tag.Value)
+				if err != nil {
+					reporter.Debugf("An error occurred whilst parsing bool value of '%s' tag, value '%s'", *tag.Key, *tag.Value)
```
ciaranRoche, line 3394: Can we wrap these derefs in `aws.StringValue` just for a bit of added safety. I know the risk is small, but just to be safe.

cmd/create/cluster/cmd.go
```diff
@@ -3330,3 +3332,34 @@ func getExpectedResourceIDForAccRole(hostedCPPolicies bool, roleARN string, role
...
+	initialValidSubnets := []*ec2.Subnet{}
+	allSubnets, err := awsClient.GetSubnetIDs()
+	if err != nil {
+		return initialValidSubnets, err
+	}
+	for _, subnet := range allSubnets {
+		hasRHManaged := false
+		for _, tag := range subnet.Tags {
+			stringKey := awssdk.StringValue(tag.Key)
+			switch stringKey {
```
oriAdler, line 3346: Do we need a `switch` here if we have only one case?

cmd/create/cluster/cmd.go
```diff
@@ -3330,3 +3332,34 @@ func getExpectedResourceIDForAccRole(hostedCPPolicies bool, roleARN string, role
...
+				stringValue := awssdk.StringValue(tag.Value)
+				boolValue, err := strconv.ParseBool(stringValue)
+				if err != nil {
+					reporter.Debugf("An error occurred whilst parsing bool value of '%s' tag, value '%s'", stringKey, stringValue)
+					boolValue = false
+				}
+				hasRHManaged = boolValue
+			}
+			if hasRHManaged {
+				break
```
oriAdler, line 3357: Maybe we can create a slice with all the subnets that were excluded? 
then we can inform the user:
`I: "The following subnets were excluded because they belong to a VPC that is managed by Red Hat: <subnet_ids>"`
So the user knows which subnets were excluded and why.

cmd/create/cluster/cmd.go
```diff
@@ -1519,13 +1521,13 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	// Custom tags for AWS resources
-	tags := args.tags
+	_tags := args.tags
```
oriAdler, line 1524: Why do we need the underscore here?

gdbranco: Renamed because tags is the name of a package :laughing:

cmd/create/cluster/cmd.go
gdbranco: I see, that can be useful. But should we even show it? RH vpcs aren't suppose to be selectable anyway :thinking:

oriAdler: It is similar to this thread https://redhat-internal.slack.com/archives/C054PJX8LE8/p1693487192971549
https://issues.redhat.com/browse/OCM-3532.

If the user is expecting to find all subnets displayed, and we are omitting some of them, that can be confusing for the user, they might think it's just a bug on the CLI not displaying the subnets they are looking for.

oriAdler: If they provide the subnets using the flags, they are going to get an error: "the subnet doesn't exist", without further explanation.

PR #1456, opened 2023-08-24
OCM-3381 | fix: adjust so that it doesn't show new ingress attributes when not supported

gdbranco:
Related issue: https://issues.redhat.com/browse/OCM-3381

### DIFF
--- a/cmd/edit/ingress/cmd.go
+++ b/cmd/edit/ingress/cmd.go
@@ -208,6 +208,7 @@ func run(cmd *cobra.Command, argv []string) {
 	}
 
 	clusterKey := r.GetClusterKey()
+	cluster := r.FetchCluster()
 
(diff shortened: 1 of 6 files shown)


cmd/edit/ingress/cmd.go
```diff
@@ -174,6 +179,15 @@ func init() {
 		"Components route TLS secret reference for oauth, console, download.",
 	)
 
+	flags.BoolVar(
+		&args.bypassLegacyIngressSupportVersionValidation,
+		bypassLegacyIngressSupportVersionValidationFlag,
+		false,
```
gdbranco, line 185: Allows customers that opened the support case to move their 4.13 clusters to be under the managed ingress improvements support.

gdbranco:
/hold

depends on backend change

ciaranRoche (commented):
/lgtm

PR #1457, opened 2023-08-24
OCM-3397 Support machine pools with zero replicas on day two

lucasponce:
As part of https://issues.redhat.com/browse/OCM-2522 backend work.

This PR updates the validation of machine pool replicas for hosted clusters on day 2.

Non autoscaling machine pools support zero replicas, meanwhile, machine pools with autoscaling must validate at least one replica.

Related to https://issues.redhat.com/browse/OCM-3397

### DIFF
--- a/pkg/helper/machinepools/helpers.go
+++ b/pkg/helper/machinepools/helpers.go
@@ -14,14 +14,20 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation"
 )
 
-func MinNodePoolReplicaValidator() interactive.Validator {
+func MinNodePoolReplicaValidator(autoscaling bool) interactive.Validator {
 	return func(val interface{}) error {
 		minReplicas, err := strconv.Atoi(fmt.Sprintf("%v", val))
 		if err != nil {
 			return err
 		}
-		if minReplicas < 1 {
-			return fmt.Errorf("min-replicas must be greater than zero")
+		if autoscaling {
+			if minReplicas < 1 {
+				return fmt.Errorf("min-replicas must be greater than zero")
+			}
+		} else {
+			if minReplicas < 0 {
+				return fmt.Errorf("replicas must be a non-negative integer")
+			}
 		}
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


pkg/helper/machinepools/helpers.go
```diff
@@ -14,14 +14,20 @@ import (
...
 	return func(val interface{}) error {
 		minReplicas, err := strconv.Atoi(fmt.Sprintf("%v", val))
 		if err != nil {
 			return err
 		}
-		if minReplicas < 1 {
-			return fmt.Errorf("min-replicas must be greater than zero")
+		if autoscaling {
+			if minReplicas < 1 {
+				return fmt.Errorf("min-replicas must be greater than zero")
```
andreadecorte, line 25: do we have such kind of messages? I would have expected something expanded 
```suggestion
				return fmt.Errorf("The minimum number of replicas must be greater than zero")
```

lucasponce: Note that I've reused similar wording for same validation in the classic machine pools.

I.e.
image

Where the message focuses on the field first "replicas" for non autoscaling vs "min-replicas" for autoscaling.

andreadecorte: ok, let's be consistent then

ciaranRoche:
@lucasponce is this only supported in hcp or can classic clusters support 0 replicas?

lucasponce:
> @lucasponce is this only supported in hcp or can classic clusters support 0 replicas?

0 replicas is supported (untouched) in ROSA Classic.

Hosted Control planes don't support 0 replicas when autoscaling is enabled, then the original decision was to align the min replicas to 1 in both autoscaling enabled/disabled.

But after discussed, we wanted to allow 0 replicas on autoscaling disabled; when autoscaling is enabled, the min replicas is 1.

This is only for hosted control planes, ROSA classic is untouched.

ciaranRoche (approved):
/lgtm

PR #1458, opened 2023-08-24
OCM-3418 | fix: handle verify network 'running' state

thomasmckay:
The `--watch` loop needs to continue on both "pending" and "running"


willkutler (approved):

willkutler:
needs a lint fix but LGTM otherwise!

PR #1459, opened 2023-08-25
OCM-2447 | feat: add automatic upgrades for node pools

andreadecorte:
Features and UX should be aligned with control plane upgrades Related: [OCM-2447](https://issues.redhat.com//browse/OCM-2447)

### DIFF
--- a/cmd/upgrade/cluster/cmd.go
+++ b/cmd/upgrade/cluster/cmd.go
@@ -110,6 +110,8 @@ func init() {
 			"minor release, e.g. 4.12.20 -> 4.13.2. By default only z-stream updates will be scheduled. "+
 			"This is currently supported only for Hosted Control Planes. ",
 	)
+	// Hidden for now as not supported yet
+	flags.MarkHidden("allow-minor-version-updates")
 
 	flags.StringVar(
 		&args.nodeDrainGracePeriod,
@@ -214,9 +216,10 @@ func runWithRuntime(r *rosa.Runtime, cmd *cobra.Command) error {
 		}
 	}
 
-	// Check mandatory parameters and enable interactive mode if needed
-	if currentUpgradeScheduling.ScheduleDate == "" && currentUpgradeScheduling.ScheduleTime == "" &&
-		currentUpgradeScheduling.Schedule == "" {
+	// Enable interactive mode if needed
+	// We need to specify either both date and time or nothing
+	if (currentUpgradeScheduling.ScheduleDate != "" && currentUpgradeScheduling.ScheduleTime == "") ||
+		currentUpgradeScheduling.ScheduleDate == "" && currentUpgradeScheduling.ScheduleTime != "" {
 		interactive.Enable()
 	}
 
@@ -265,7 +268,7 @@ func runWithRuntime(r *rosa.Runtime, cmd *cobra.Command) error {
 					Required: false,
 				})
 				if err != nil {
-					return fmt.Errorf("Expected an choice on the versions to target: %s", err)
+					return fmt.Errorf("Expected a choice on the versions to target: %s", err)
 				}
 			}
 		}
--- a/cmd/upgrade/machinepool/cmd.go
+++ b/cmd/upgrade/machinepool/cmd.go
@@ -27,13 +27,16 @@ import (
 	"github.com/openshift/rosa/pkg/interactive/confirm"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/rosa"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
 var args struct {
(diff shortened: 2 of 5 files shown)


cmd/upgrade/cluster/cmd.go
```diff
@@ -215,7 +217,7 @@ func runWithRuntime(r *rosa.Runtime, cmd *cobra.Command) error {
 	}
 
 	// Check mandatory parameters and enable interactive mode if needed
-	if currentUpgradeScheduling.ScheduleDate == "" && currentUpgradeScheduling.ScheduleTime == "" &&
+	if (currentUpgradeScheduling.ScheduleDate == "" && currentUpgradeScheduling.ScheduleTime == "") &&
```
gdbranco, line 220: Considering it's 'ands' why have the ()?

gdbranco: Should this be an 'or' to match https://github.com/openshift/rosa/pull/1459/files#diff-3f508bee7428efffa6218f9a113aa7c939e1fc6dd8e30554b969aa1ccd7cc47aR142?

andreadecorte: it should be an or indeed! good catch thanks

cmd/upgrade/cluster/cmd.go
```diff
@@ -215,7 +217,7 @@ func runWithRuntime(r *rosa.Runtime, cmd *cobra.Command) error {
 	}
 
 	// Check mandatory parameters and enable interactive mode if needed
-	if currentUpgradeScheduling.ScheduleDate == "" && currentUpgradeScheduling.ScheduleTime == "" &&
+	if (currentUpgradeScheduling.ScheduleDate == "" || currentUpgradeScheduling.ScheduleTime == "") &&
```
lucasponce, line 220: Is this intentional? So, user needs to provide both "ScheduleDate" and "ScheduleTime" to not jump to the interactive mode?

cmd/upgrade/machinepool/cmd.go
```diff
@@ -102,69 +124,194 @@ func runWithRuntime(r *rosa.Runtime, cmd *cobra.Command, argv []string) error {
...
 	machinePoolID := argv[0]
-	scheduleDate := args.scheduleDate
-	scheduleTime := args.scheduleTime
+	var err error
+	currentUpgradeScheduling := ocm.UpgradeScheduling{
+		Schedule:                 args.schedule,
+		ScheduleDate:             args.scheduleDate,
+		ScheduleTime:             args.scheduleTime,
+		AllowMinorVersionUpdates: args.allowMinorVersionUpdates,
+		AutomaticUpgrades:        args.schedule != "",
```
lucasponce, line 133: Should the "AutomaticUpgrade" flag have also the or args.schedule != "" || (args.scheduleDate != "" && args.scheduleTime != "") here ?

andreadecorte: automatic upgrade requires a schedule in cron format, that is passed with `schedule` paramater. 
manual schedule requires a `scheduleDate` and `scheduleTime` because it is a single schedule.
So here we say that enabling automatic upgrades is equivalent to specify `schedule`. So it is correct.

cmd/upgrade/cluster/cmd.go
```diff
@@ -110,6 +110,8 @@ func init() {
 			"minor release, e.g. 4.12.20 -> 4.13.2. By default only z-stream updates will be scheduled. "+
 			"This is currently supported only for Hosted Control Planes. ",
 	)
+	// Hidden for now as not supported yet
+	flags.MarkHidden("allow-minor-version-updates")
```
gdbranco, line 114: Does it require this check here as well? https://github.com/openshift/rosa/pull/1459/files#diff-3f508bee7428efffa6218f9a113aa7c939e1fc6dd8e30554b969aa1ccd7cc47aR149

cmd/upgrade/cluster/cmd.go
andreadecorte: updated to cover this case

cmd/upgrade/cluster/cmd.go
andreadecorte: actually after Lucas comment, I fixed again the check as the case of no schedule specified is valid (and should schedule an update immediately)

cmd/upgrade/cluster/cmd.go
andreadecorte: if you mean to explictly block the parameter usage, I am debated.
The parameter is right now rejected by the backend, but I hesitate to block it into the CLI too, as we will reenable later. So for now I just made it hidden when specified directly. WDYT? Should we just block or completely remove for now and readd later?

ciaranRoche (approved):

PR #1463, opened 2023-08-28
OCM-3444 | fix: edit nodepool was broken if no upgrade available

andreadecorte:
In interactive mode, we should not exit if no upgrade is available but just skip the version part.

Related: [OCM-3444](https://issues.redhat.com//browse/OCM-3444)

### DIFF
--- a/cmd/edit/machinepool/nodepool.go
+++ b/cmd/edit/machinepool/nodepool.go
@@ -98,11 +98,16 @@ func editNodePool(cmd *cobra.Command, nodePoolID string, clusterKey string, clus
 			os.Exit(1)
 		}
 		if version == "" {
-			r.Reporter.Infof("No upgrade available for the machine pool '%s' on cluster '%s'", nodePoolID,
-				clusterKey)
-			os.Exit(0)
+			if !interactive.Enabled() {
+				// Not interactive mode, stop if user specifies a version but none is available
+				r.Reporter.Infof("No upgrade available for the machine pool '%s' on cluster '%s'. Exiting",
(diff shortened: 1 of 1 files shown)


cmd/edit/machinepool/nodepool.go
```diff
@@ -98,11 +98,15 @@ func editNodePool(cmd *cobra.Command, nodePoolID string, clusterKey string, clus
...
-				clusterKey)
-			os.Exit(0)
+			if !interactive.Enabled() {
+				// Not interactive mode, stop if user specifies a version but none is available
+				r.Reporter.Infof("No upgrade available for the machine pool '%s' on cluster '%s'. Exiting",
+					nodePoolID, clusterKey)
+				os.Exit(0)
+			}
+		} else {
+			npBuilder.Version(cmv1.NewVersion().ID(version))
```
ciaranRoche, line 108: What happens if you pass an empty string for the version id to the node pools API?

andreadecorte:
```
time=2023-08-28T14:50:12+02:00 level=debug msg={
  "kind": "Error",
  "id": "400",
  "href": "/api/clusters_mgmt/v1/errors/400",
  "code": "CLUSTERS-MGMT-400",
  "reason": "The 'version.id' attribute is mandatory when 'version' is used",
  "operation_id": "6fefb350-6139-46cc-942f-2435ac64bfe0"
}
```
I'll add a comment

andreadecorte: added

PR #1466, opened 2023-08-29
OCM-3247 | feat: add root disk size validation

marcolan018:
https://issues.redhat.com/browse/OCM-3247


ciaranRoche (changes requested):
This is similar to the validation we got in the backend, could we move it to https://github.com/openshift-online/ocm-common and share it between the CLI and UI

PR #1467, opened 2023-08-30
OCM-3473 | fix: create operator-roles for shared VPC

oriAdler:
Operator roles' policies are associated with the installer role prefix. In case the operator role policy already has `sts:AssumeRole` action with shared VPC role ARN, and the user provides a different shared VPC role ARN, error out, and guide the user to use a different account role ARN.

**Note:** Currently the command overrides the shared VPC role ARN value with the value provided by the user.

Examples:
1. User create operator roles outside of the shared VPC scope:
image
image
2. User overrides the policy created in [1], while creating operator roles with shared VPC role ARN:
(A new policy version is created with a `sts:AssumeRole` permission)
image
image
3. The user tries to create operator roles with the same installer role from [2] with a different **shared-vpc-role-arn** and is prompted to use a different installer role:
image
4. The user is using a different role ARN, and creates successfully the operator roles:
image
image

### DIFF
--- a/cmd/create/operatorroles/common_utils.go
+++ b/cmd/create/operatorroles/common_utils.go
@@ -2,9 +2,14 @@ package operatorroles
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/openshift/rosa/pkg/aws"
+	"github.com/openshift/rosa/pkg/rosa"
+	errors "github.com/zgalor/weberr"
 )
+
+const assumePolicyAction = "AssumeRole"
 
 func computePolicyARN(accountID string, prefix string, namespace string, name string, path string) string {
 	if prefix == "" {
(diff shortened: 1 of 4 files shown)


cmd/create/operatorroles/common_utils.go
```diff
@@ -19,3 +24,30 @@ func computePolicyARN(accountID string, prefix string, namespace string, name st
...
+	}
+
+	policyDocument, err := r.AWSClient.GetDefaultPolicyDocument(policyArn)
+	if err != nil {
+		return err
+	}
+
+	// The policy associated with the installer role. In the case it contains a different shared VPC role ARN,
+	// don't override it.
+	if strings.Contains(policyDocument, assumePolicyAction) && !strings.Contains(policyDocument, sharedVpcRoleArn) {
```
davidleerh, line 43: NP:
Would it be possible to retrieve the existing shared vpc arn listed in the op role assume policy?
To allow a more transparent error message like:

```
"Policy with ARN 'A' contains 'sts:AssumeRole' action with an unexpected shared VPC role ARN. [Expected: 'B', Provided: 'C']
To create operator roles with shared vpc role ARN 'C', please provide a different value for '--installer-role-arn'."
```

davidleerh:
aside from small NP, looking good to me~

cmd/create/operatorroles/common_utils.go
oriAdler: I agree that the error message you're suggesting is more helpful.
We have a place in the code where we parse the policy document JSON, and then it would be possible, let me play with the code and see what it looks like.
I took this approach, of `strings.Contains`, because it was sort of a last-minute change, and tried to keep the PR as simple as possible.

pkg/aws/policies.go
```diff
@@ -1134,6 +1135,42 @@ func (c *awsClient) deletePolicies(policies []string) (*iam.DeletePolicyOutput,
...
+	policyVersionsOutput, err := c.iamClient.ListPolicyVersions(&iam.ListPolicyVersionsInput{
+		PolicyArn: aws.String(policyArn),
+	})
+	if err != nil {
+		return "", err
+	}
+
+	for _, version := range policyVersionsOutput.Versions {
+		if aws.BoolValue(version.IsDefaultVersion) {
+			return *version.VersionId, nil
```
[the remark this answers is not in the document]
oriAdler: Good point, fixed.

cmd/create/operatorroles/common_utils.go
ciaranRoche: lets merge and get this through QE asap, we can follow up with error message improvements.

oriAdler: Created this ticket https://issues.redhat.com/browse/OCM-3522 to follow up on the next iteration.
Thanks for the feedback.

PR #1468, opened 2023-08-30
OCM-3375 | feat: Add 'Workload Monitoring' to describe/cluster

hunterkepley:
Workload monitoring will either say `Enabled` if it is enabled, `Disabled` if it is disabled, or `Unknown` if the check fails for whatever reason

### DIFF
--- a/cmd/describe/cluster/cmd.go
+++ b/cmd/describe/cluster/cmd.go
@@ -179,13 +179,13 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
-	workloadMonitoring, ok := cluster.GetDisableUserWorkloadMonitoring()
-	isWorkloadMonitoringEnabled := "Disabled"
+	workloadMonitoringDisabled, ok := cluster.GetDisableUserWorkloadMonitoring()
+	isWorkloadMonitoringEnabled := "Enabled"
 	if !ok {
 		isWorkloadMonitoringEnabled = "Unknown"
 	}
-	if workloadMonitoring {
-		isWorkloadMonitoringEnabled = "Enabled"
+	if workloadMonitoringDisabled {
+		isWorkloadMonitoringEnabled = "Disabled"
 	}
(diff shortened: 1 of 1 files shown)


cmd/describe/cluster/cmd.go
```diff
@@ -179,6 +179,15 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
+	workloadMonitoringDisabled, ok := cluster.GetDisableUserWorkloadMonitoring()
+	isWorkloadMonitoringEnabled := "Enabled"
+	if !ok {
+		isWorkloadMonitoringEnabled = "Unknown"
```
marcolan018, line 185: is "Unkonwn" a valid value? 
Another option is to return the default value("Enabled") if cannot get the value.

ciaranRoche: I agree with @marcolan018 `unknown` is not a valid value and I would rather not show anything than to show an invalid value. 
Same as we can not just go with Enabled if the value is not found. 

Lets just prefix it to `str` if we have the value, something like 
```
	workloadMonitoringDisabled, ok := cluster.GetDisableUserWorkloadMonitoring()
	if ok {
		isWorkloadMonitoringEnabled := "Enabled"
		if workloadMonitoringDisabled {
			isWorkloadMonitoringEnabled = "Disabled"
		}
		str = fmt.Sprintf("Workload Monitoring:        %s\n", isWorkloadMonitoringEnabled)	
	}
```

hunterkepley: changed

ciaranRoche: Can you update this again, so we are not passing in an empty string to `str` similar to the example in the previous comment. 

We can prefix the workload monitoring to `str`

hunterkepley: updated

gdbranco (approved):

PR #1469, opened 2023-08-30
OCM-1478 | fix: Moved getMode to top of cluster creation so it checks mode firstly

hunterkepley:
In case of invalid mode, the command will not be run

```
./rosa create cluster --sts --mode atuo
E: Invalid mode. Allowed values are [auto manual]
```


ciaranRoche (commented):
/lgtm

PR #1470, opened 2023-08-31
OCM-3313 | refactor: extract autoscaler code to new modules

osherdp:
In order to make code less repetitive when we add day2 support of cluster-autoscaler (and also to reduce the amount of code in ``cmd/create/cluster/cmd.go``), this refactors the code to be placed in new modules that are also more easily imported by new commands.

In the step 2 of this we want to add a ``rosa create autoscaler`` command that will use the new location of the code

### DIFF
--- a/pkg/ocm/validators.go
+++ b/pkg/ocm/validators.go
@@ -1,0 +1,68 @@
+package ocm
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+)
+
+func IntValidator(val interface{}) error {
+	if val == "" { // if a value is not passed it should not throw an error (optional value)
+		return nil
+	}
+	_, err := strconv.Atoi(fmt.Sprintf("%v", val))
+	return err
+}
+
+func NonNegativeIntValidator(val interface{}) error {
+	if val == "" { // if a value is not passed it should not throw an error (optional value)
+		return nil
+	}
+	number, err := strconv.Atoi(fmt.Sprintf("%v", val))
+	if err != nil {
+		return fmt.Errorf("Failed parsing '%v' to an integer number.", val)
+	}
+
+	if number < 0 {
+		return fmt.Errorf("Number must be greater or equal to zero.")
+	}
+
+	return nil
+}
+
+func DurationStringValidator(val interface{}) error {
+	if val == "" {
+		return nil
+	}
+	input, ok := val.(string)
+
+	if !ok {
+		return fmt.Errorf("Can only validate strings, got %v", val)
+	}
+
+	re := regexp.MustCompile("^([0-9]+(.[0-9]+)?(ns|us|µs|ms|s|m|h))+$")
+	regexPass := re.MatchString(input)
+	if !regexPass {
+		return fmt.Errorf("Expecting an integer plus unit of time (without spaces). " +
+			"Options for time units include: ns, us, µs, ms, s, m, h. Examples: 2000000ns, 180s, 2m, etc.")
+	}
+	return nil
+
+}
+
+func PercentageValidator(val interface{}) error {
+	if val == "" {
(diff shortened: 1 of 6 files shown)


osherdp:
/hold
waiting for reviews

vkareh:
@osherdp Do you have a ticket and/or DDR for this change? I have some reservations about introducing a `create` command for something that's not a resource in our API. Would `rosa edit cluster` be a better command to tackle this?

gdbranco:
Mostly moving code around, SGTM

pkg/ocm/validators.go
```diff
@@ -0,0 +1,72 @@
+package ocm
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+)
+
+func IntValidator(val interface{}) error {
```
ciaranRoche, line 9: Can we add unit tests for all these validators :pray:

vkareh:
@osherdp but considering that this is also a day-1 option, it seems more fitting for `rosa create|edit cluster`. Sounds like a mistake to have put it in a separate API endpoint and should be part of `Cluster.AutoScaler{}`?

This code change means users can no longer set it during cluster creation, so it seems we're changing/removing functionality

osherdp:
> but considering that this is also a day-1 option, it seems more fitting for rosa create|edit cluster. Sounds like a mistake to have put it in a separate API endpoint and should be part of Cluster.AutoScaler{}?

Maybe we can incorporate that into ``rosa edit cluster`` rather than ``rosa create|edit autoscaler``, but either way it's out of scope for this specific PR. Whatever the command we choose we prefer to have the code re-usable across the different commands (instead of repeating a lot of code)

The reason I think ``rosa create|edit autoscaler`` is more clear is that it's more consistent with the existing behavior of default machine-pool | oidc-providers. I don't think we can justify those being separate resources and autoscaler being a cluster property

> This code change means users can no longer set it during cluster creation, so it seems we're changing/removing functionality

after this code change users are still able to set it during cluster creation. it's only a preparation for handling it in a day2 scenario

vkareh:
okay, I see your point, thanks for clarifying. So basically customers can set it during `rosa create cluster` as well as `rosa create autoscaler` if I understand correctly? If that's the case, then I'm fine with it. Just want to make sure it doesn't cause confusion to end-users. Since day-1 `create cluster` resources are perceived as part of the cluster spec, users will likely expect them to also be editable as part of the cluster spec (which implies `edit cluster`). Does that make sense?

osherdp:
> 

@vkareh yeah, I get what you mean. Actually originally I also wanted to make it all part of the cluster object, but we've got into a situation where the lack of support for ``PUT`` requests in OCM made it impossible to partially override autoscaler configuration and still preserve the ``PATCH`` contract. Because requirements were to allow that for day1 and day2, we didn't have much choice and had to have it this way

In this case we partly rely on the docs making this kind of behavior clearer to users. Either way configuring cluster-autoscaler is something I consider a bit more advance, so I assume people will first turn to the docs and won't expect it to be intuitive on first sight. Hope it makes some sense :smile:

vkareh:
@osherdp yes! Our API unfortunately doesn't have a way to unset attributes, apart from issuing a `DELETE` request. Sigh...

Okay then, this all makes more sense to me. Thanks for all this context! :+1:

osherdp:
@ciaranRoche @gdbranco please let me know if anything else is needed
Tests are added to the validators and this should only refactor the code (we still have the next step in order to introduce the new commands)

osherdp:
/hold
sorry, doing one more manual test

ciaranRoche (approved):
/lgtm

osherdp:
/unhold

thanks everyone!

PR #1472, opened 2023-08-31
OCM-3370 | fix: re-attach policies after detaching from operator roles

davidleerh:
Found a small bug that occurs when a user elects to detach existing policies from an existing "operator" role and reattach updated the operator policy.

(When running commands "rosa upgrade roles" or "rosa upgrade cluster" only)

Conditions to emulate issue:

A. Operator roles only. Account roles working properly.

B. Attach custom policy onto existing operator role.

C. Upgrade operator role via "rosa upgrade roles" or "rosa upgrade cluster" only.

"rosa upgrade operator-roles" or "rosa upgrade account-roles"command is works fine.

D. When prompted if you'd like to "Would you like to dettach current policies and setup a new one", select YES.

 

Steps to emulate issue:

A. Create classic account roles via "rosa create account-roles --version 4.10"

B. Create classic cluster with classic account roles via "rosa create cluster --mode=auto". Please make sure version selected is from previous major version (e.g. 4.10.3)

C. In AWS Console or AWS CLI, attach a custom policy to one of the auto created operator role.

D. Upgrade roles via "rosa upgrade roles" or "rosa upgrade cluster".

E. When prompted if you'd like to "Would you like to dettach current policies and setup a new one", select YES. (This will appear when upgrading the operator role containing more than one policy [custom + operator policy])

**Expected Outcome:**

ROSA detaches all policies (custom policies and respective operator policy) from an operator role, and reattaches it with a single policy with the updated version tag.

**Bugged Outcome:**

ROSA detaches all policies (custom policies and respective operator policy) from an operator role, updates the policy version but never reattaches it back to the role.

Based on refinement, it is agreed to keep the prompt, and ensure we attach the correct policy.

**SUMMARY
Fixed via adding method to reattach policy to role for operator roles.**

### DIFF
--- a/cmd/upgrade/roles/cmd.go
+++ b/cmd/upgrade/roles/cmd.go
@@ -493,7 +493,7 @@ func handleAccountRolePolicyARN(
 	}
 
 	promptString := fmt.Sprintf("More than one policy attached to account role '%s'.\n"+
-		"\tWould you like to dettach current policies and setup a new one ?", roleName)
+		"\tWould you like to detach current policies and setup a new one ?", roleName)
 	if !confirm.Prompt(true, promptString) {
 
 		attachedPoliciesArns := make([]string, len(attachedPoliciesDetail))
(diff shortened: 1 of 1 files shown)


cmd/upgrade/roles/cmd.go
```diff
@@ -842,6 +843,18 @@ func upgradeOperatorRolePoliciesFromCluster(
 		if err != nil {
 			return err
 		}
+
+		if operatorRoleName != "" {
+			err = awsClient.AttachRolePolicy(operatorRoleName, policyARN)
```
gdbranco, line 848: If it is already attached does this error out?

davidleerh: I don't believe it errors out. It (awsClient.AttachRolePolicy) is currently being done with account roles in "rosa upgrade roles".

PR #1477, opened 2023-09-05
OCM-3535 | fix: set max-nodes-total default value to 180

osherdp:
1000 is way over the number of nodes we're currently supporting in OpenShift (it's rather 180 for most versions).

Changing the value to make it more sensible.


ciaranRoche (approved):
/lgtm

PR #1478, opened 2023-09-05
OCM-3574 | fix: cores range validation

osherdp:
When creating autoscaler interactively, the cores range is being compared wrongly. For example, when given the value 99 it will try to compare it with the default value of 100:

```
$ rosa create cluster 
...
? Maximum number of cores to deploy in cluster: [? for help] (100)
<giving it 99>
X Sorry, your reply was invalid: max value must be greater or equal than
min value 100.
```

FYI @sunzhaohua2 
/cc @gdbranco @ciaranRoche


ciaranRoche (approved):
/lgtm

PR #1479, opened 2023-09-06
OCM-3580 | fix: Validation issues of configuring autoscaler interactively

osherdp:
* pod-priority-threshold was missing integer validation interactively.
* log verbosity wasn't validated to be a non-negative number interactively. same goes for maximum-pod-grace-period
* duration strings validation wasn't properly catching issues like in "1 00m". changed the code to just try doing the parsing using golang's stdlib


ciaranRoche (approved):
/lgtm

PR #1484, opened 2023-09-06
add a function to escape special characters when printing out Cluster…

supernovae:
Special characters were getting interpolated by the shell when the password was echoed backed to the user.  Created the escapeSpcialChars function and used that to print the password so it would escape out any special strings still allowed for passwords.


robpblake:
@supernovae Do you want to progress this one? It will need a rebase and also a corresponding issue in OCM JIRA project?

cheers,
Rob

PR #1486, opened 2023-09-07
OCM-3609 | feat: Disable prompt & flag for HCP while using a govcloud account

hunterkepley:
This MR accomplishes 2 things with `create accountroles`:

1. Return an error when AWS caller is a govcloud account
2. Disable prompt for asking if user if they'd like to create HCP roles when AWS caller is a govcloud account

### DIFF
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -188,6 +188,7 @@ func run(cmd *cobra.Command, argv []string) {
 	callerIdentityOutput, err := r.AWSClient.GetCallerIdentity()
 	if err != nil {
 		r.Reporter.Warnf("Unable to get AWS Caller Identity.")
+		os.Exit(1)
 	}
 
 	callerIsGovcloud := false
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -204,6 +204,7 @@ func run(cmd *cobra.Command, argv []string) {
 
 	if cmd.Flags().Changed("hosted-cp") && callerIsGovcloud {
 		r.Reporter.Errorf("Setting `hosted-cp` is not supported for Govcloud AWS accounts")
+		os.Exit(1)
 	}
 
 	// Validate AWS credentials for current user
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -183,18 +183,12 @@ func run(cmd *cobra.Command, argv []string) {
 			"any supported ROSA version can be installed with managed policies")
 	}
 
-	callerIsGovcloud := false
-
-	if r.Creator.IsGovcloud {
-		callerIsGovcloud = true
-	}
-
 	// Hosted cluster roles always use managed policies
 	if cmd.Flags().Changed("hosted-cp") && cmd.Flags().Changed("managed-policies") && !args.managed {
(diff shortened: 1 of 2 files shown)


cmd/create/accountroles/cmd.go
```diff
@@ -182,12 +182,30 @@ func run(cmd *cobra.Command, argv []string) {
 			"any supported ROSA version can be installed with managed policies")
 	}
 
+	r.WithAWS()
```
ciaranRoche, line 185: Can we move this to where we instantiate the runtime at the top. To avoid this getting missed later on, and someone potentially overriding it.

cmd/create/accountroles/cmd.go
```diff
@@ -182,12 +182,30 @@ func run(cmd *cobra.Command, argv []string) {
...
+	// Check whether or not AWS account is govcloud
+	callerIdentityOutput, err := r.AWSClient.GetCallerIdentity()
+	if err != nil {
+		r.Reporter.Warnf("Unable to get AWS Caller Identity.")
+		os.Exit(1)
+	}
+
+	callerIsGovcloud := false
+
+	if (*callerIdentityOutput.Arn)[:14] == "arn:aws-us-gov" {
```
ciaranRoche, line 196: What is the ARN is nil, this will crash, best to use `aws.StringValue(callerIdentityOutput.Arn)` to safely deref the string. 

aws have an arn Parse function, we can then pass the string value too, and from there we can check the `Partition` value.

ciaranRoche (commented):
I wonder would it be better to extend the creator struct to have a govCloud bool. 

In the `getCreator` func we are already call get caller identity, so we can extend this to check the Partition from the parsed ARN. 

WDYT?

cmd/create/accountroles/cmd.go
hunterkepley: not sure i understand the Parse/Partition comment, but i will implement the first

hunterkepley: now understood

cmd/create/accountroles/cmd.go
```diff
@@ -140,6 +140,7 @@ func init() {
 
 func run(cmd *cobra.Command, argv []string) {
 	r := rosa.NewRuntime()
+	r.WithAWS()
```
gdbranco, line 143: NP: Can be inline with last line

cmd/create/accountroles/cmd.go
```diff
@@ -182,12 +183,22 @@ func run(cmd *cobra.Command, argv []string) {
 			"any supported ROSA version can be installed with managed policies")
 	}
 
+	callerIsGovcloud := false
+
+	if r.Creator.IsGovcloud {
```
gdbranco, line 188: NP: wouldn't it be possible to just assigned directly in last line?

pkg/aws/client.go
```diff
@@ -606,15 +607,21 @@ func (c *awsClient) GetCreator() (*Creator, error) {
...
 		}
 
-		// resolveSTSRole esures a parsed valid ARN before
+		// resolveSTSRole ensures a parsed valid ARN before
 		// returning it so we don't need to parse it again
 		creatorARN = *stsRole
 	}
 
+	isGovcloud := false
+	if creatorParsedARN.Partition == "aws-us-gov" {
```
gdbranco, line 616: Could it be a const?

gdbranco (approved):

PR #1487, opened 2023-09-10
OCM-3366 | fix: CLI options not validated client-side

osherdp:
Applying the same validations that are done interactively when the fields are provided via CLI options.


ciaranRoche:
@osherdp do we want to include this in the release?

osherdp:
> @osherdp do we want to include this in the release?

@ciaranRoche this one is not critical

osherdp:
/hold
requires some more testing

PR #1488, opened 2023-09-11
OCM-3640 | fix: autoscaler created when not enabled

osherdp:
When not enabling autoscaling feature, the cluster-autoscaler is still being created due to us not unsetting the autoscaler configuration for that case.

Also, we were missing including the prefix for the arguments, which made us miss detecting that autoscaler options were used. This should also be fixed now.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -2092,7 +2092,10 @@ func run(cmd *cobra.Command, _ []string) {
 		isHostedCP,
 		multiAZ)
 
-	if autoscaling {
+	var clusterAutoscaler *clusterautoscaler.AutoscalerArgs
+	if !autoscaling {
+		clusterAutoscaler = nil
+	} else {
 		// if the user set compute-nodes and enabled autoscaling
 		if isReplicasSet {
 			r.Reporter.Errorf("Compute-nodes can't be set when autoscaling is enabled")
@@ -2140,7 +2143,7 @@ func run(cmd *cobra.Command, _ []string) {
 			os.Exit(1)
 		}
 
-		autoscalerArgs, err = clusterautoscaler.GetAutoscalerOptions(
+		clusterAutoscaler, err = clusterautoscaler.GetAutoscalerOptions(
 			cmd.Flags(), clusterAutoscalerFlagsPrefix, true, autoscalerArgs)
 		if err != nil {
 			r.Reporter.Errorf("%s", err)
@@ -2749,8 +2752,8 @@ func run(cmd *cobra.Command, _ []string) {
 		clusterConfig.SharedVPCRoleArn = sharedVPCRoleARN
 		clusterConfig.BaseDomain = baseDomain
 	}
-	if autoscalerArgs != nil {
-		autoscalerConfig, err := clusterautoscaler.CreateAutoscalerConfig(autoscalerArgs)
+	if clusterAutoscaler != nil {
+		autoscalerConfig, err := clusterautoscaler.CreateAutoscalerConfig(clusterAutoscaler)
 		if err != nil {
 			r.Reporter.Errorf("Failed creating autoscaler configuration: %s", err)
 			os.Exit(1)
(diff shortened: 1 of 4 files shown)


lucasponce:
Thanks @osherdp, let me test it with my working branch in the backend

osherdp:
/hold
waiting for the tests

cmd/create/cluster/cmd.go
```diff
@@ -2092,7 +2092,10 @@ func run(cmd *cobra.Command, _ []string) {
 		isHostedCP,
 		multiAZ)
 
-	if autoscaling {
+	var clusterAutoscaler *clusterautoscaler.AutoscalerArgs
+	if !autoscaling {
```
lucasponce, line 2096: Does this mean the global autoscaler will be enabled only when the machine pool/node pool autoscaling is enabled?

(I'm thinking in side effects for the hosted cluster area)

osherdp: HCP is out of scope at the moment for the cluster-autoscaler

lucasponce:
I'm confused with this block:

image

Isn't still testing the parameters without the prefixes?

cmd/create/cluster/cmd.go
osherdp: cool, didn't know you're working on that

refer to the DDR at https://issues.redhat.com/browse/OCM-70, it has most of the relevant information for the design of the feature
basically speaking, on the backend we had to make those two constructs (machine-pools, cluster-autoscaler) independent from each other

lucasponce: It's not a big change, it can reuse the whole design and apply it to the provisioner for hosted clusters.

It's experimental, but there is a draft with day 1 enabled under this:
https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/6442

The only relevant change is that not all fields apply to the hosted clusters and there should be a validation for those cases.

But it should be transparent for this current OCM-70 work.

osherdp:
> I'm confused with this block:
> 
> image
> 
> Isn't still testing the parameters without the prefixes?

you're right. thanks!

ciaranRoche:
@osherdp @lucasponce anything stopping this from being merged now, we want it to land before we cut our first RC

lucasponce:
I'm not sure if this is going to give problems for hive clusters:

image

It seems that ROSA will always populate all auto-scaler fields; is that intentional?

What about populating only if the value is provided by the argument in ROSA?

osherdp:
> I'm not sure if this is going to give problems for hive clusters:
> 
> image
> 
> It seems that ROSA will always populate all auto-scaler fields; is that intentional?
> 
> What about populating only if the value is provided by the argument in ROSA?

@lucasponce we're getting out-of-scope here, and as @ciaranRoche mentions above we need to do the cut-off today
Does this change fixes the described bug? My testing concludes that it is, so I think it should first go in and then we can optimize other stuff

osherdp:
/unhold
seems to fix the problem, if the situation arises I'm always available for more changes in the behavior

lucasponce:
Ok, I won't block the PR.

The main bug, which is https://issues.redhat.com/browse/OCM-3640 seems addressed.

Thanks

lucasponce:
> > Ok, I won't block the PR.
> > The main bug, which is https://issues.redhat.com/browse/OCM-3640 seems addressed.
> > Thanks
> 
> @lucasponce thank you! again, as needed I'm open to suggestions. Thanks for noticing this bug, it could be pretty painful to see this landing in production

For the record, the rosa logic is:
- If no autoscaler parameters are set, the autoscaler is not set.
- If one autoscaler parameter is set, the whole structure is set. 

I think that ROSA should populate ONLY the parameter(s) that is/are provided and let the backend how to deal with the defaults.

Right now, the backend has no info about what is set by default, and what by the user I'm not sure if that is going to create side effects.

But I don't have enough information about that.

I'll re-arrange OCM-3231considering this behavior, but my preference would be that the defaults parameters that customers don't specify, those should be handled in the backend.

gdbranco:
/lgtm

Moving forward given above considerations

PR #1492, opened 2023-09-11
OCM-3616 | fix: resource limits range defaults

osherdp:
Trying to be more similar to the kubernetes default values, but instead we allow 180 nodes in OCM clusters rather than 5000 in kubernetes.

See: https://github.com/kubernetes/autoscaler/blob/632512c9dfdf18caebd55a9652d455c6c68ad496/cluster-autoscaler/config/const.go#L20-L23


ciaranRoche:
@osherdp do we want to include this in the release?

osherdp:
@ciaranRoche yes that would be great!

ciaranRoche (approved):
/lgtm

PR #1495, opened 2023-09-11
OCM-2559 | [fix] newer version available point to console.redhat.com

thomasmckay:


cmd/verify/rosa/cmd.go
```diff
@@ -42,6 +42,7 @@ var Cmd = &cobra.Command{
 const (
 	DownloadLatestMirrorFolder = "https://mirror.openshift.com/pub/openshift-v4/clients/rosa/latest/"
 	baseReleasesFolder         = "https://mirror.openshift.com/pub/openshift-v4/clients/rosa/"
+	consoleLatestFolder        = "https://mirror.openshift.com/pub/openshift-v4/clients/rosa/latest/"
```
gdbranco, line 45: This is the same URL btw

gdbranco:
commit format is wrong

Should be
`OCM-2559 | fix: newer version available point to console.redhat.com`

PR #1497, opened 2023-09-12
OCM-3616 | fix: autoscaler default values

osherdp:
Setting those two fields to be consistent with OpenShift's defaults.


sunzhaohua2:
/lgtm
Thanks!

PR #1500, opened 2023-09-12
OCM-3663 | AWS billing support

chenz4027:
https://issues.redhat.com/browse/OCM-3663

image
image


cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,84 @@ func run(cmd *cobra.Command, _ []string) {
...
+			billingAccount, err = interactive.GetOption(interactive.Input{
+				Question: "Billing Account",
+				Help:     cmd.Flags().Lookup("billing-account").Usage,
+				Default:  billingAccount,
+				Required: true,
+				Options:  cloudAccounts,
+			})
+
+			pattern := regexp.MustCompile(`[-]?\d[\d,]*[\.]?[\d{2}]*`)
+			billingAccount = pattern.FindString(billingAccount)
```
ciaranRoche, line 967: Can we abstract this to a small function with some unit tests. So its easier to maintain and we understand exactly what that regexp is expecting?

chenz4027: Yeah sure, will add some unit texts regarding that. Thanks for the review @ciaranRoche :)

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,83 @@ func run(cmd *cobra.Command, _ []string) {
...
+				os.Exit(1)
+			}
+			for _, contract := range contracts {
+				dimensions := contract.Dimensions()
+				r.Reporter.Infof(
+					"+---------------------+----------------+ \n"+
+						"   | Start Date          |'%s'  | \n"+
+						"   | End Date            |'%s'  | \n"+
+						"   +---------------------+----------------+ \n",
+					contract.StartDate().Format(format), contract.EndDate().Format(format))
```
cristianoveiga, line 1003: we could probably move this to its own function `printContracts`

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,83 @@ func run(cmd *cobra.Command, _ []string) {
...
+				for _, dimension := range dimensions {
+					numberOfVPCs := ocm.GetDimensionValue(dimension, "four_vcpu_hour")
+					numberOfClusters := ocm.GetDimensionValue(dimension, "control_plane")
+					r.Reporter.Infof(
+						"+---------------------+----------------+ \n"+
+							"   | Number of vCPUs:    |'%s'             | \n"+
+							"   | Number of clusters: |'%s'             | \n"+
+							"   +---------------------+----------------+ \n",
+						strconv.Itoa(numberOfVPCs), strconv.Itoa(numberOfClusters))
```
cristianoveiga, line 1012: do we want to show these dimensions even if they are zero?

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,83 @@ func run(cmd *cobra.Command, _ []string) {
...
+			if billingAccount != awsCreator.AccountID {
+				r.Reporter.Infof("The selected AWS billing account is a different account than your AWS infrastructure account." +
+					"The AWS billing account will be charged for subscription usage. " +
+					"The AWS infrastructure account will be used for managing the cluster. \n" +
+					"   Contract enabled for this billing account. \n")
+			} else {
+				r.Reporter.Infof("Using '%s' as billing account. \n"+
+					"   Contract enabled for this billing account. \n",
+					billingAccount)
+			}
```
cristianoveiga, line 989: I believe we want to show the text `Contract enabled for this billing account` only if the account has pre purchased amounts of vCPUs or Clusters.

cmd/create/cluster/cmd.go
cristianoveiga: The output will also look a little different from the mockup - we probably want to show start date, end date, number of cpus and cluster in the same "block"

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,83 @@ func run(cmd *cobra.Command, _ []string) {
...
+			if err != nil {
+				r.Reporter.Errorf("Expected a valid billing account: '%s'", err)
+				os.Exit(1)
+			}
+		}
+
+		if billingAccount != "" && !ocm.IsValidAWSAccount(billingAccount) {
+			r.Reporter.Errorf("Expected a valid billing account")
+			os.Exit(1)
+		} else {
```
cristianoveiga, line 977: unnecessary `else`

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,83 @@ func run(cmd *cobra.Command, _ []string) {
...
+			format := "Jan 02, 2006"
```
cristianoveiga, line 978: nit: I would move this variable closer to its usage, if possible.

pkg/ocm/billing.go
```diff
@@ -38,18 +40,73 @@ func (c *Client) GetBillingAccounts() ([]string, error) {
...
+		return billingAccounts, errors.New("No billing account associated. " +
+			"Go to https://docs.openshift.com/rosa/rosa_architecture/rosa-understanding.html" +
+			"to learn how to associate your billing account." +
+			" You must associate a billing account to continue.")
 	}
 
 	return billingAccounts, nil
 }
 
+func (c *Client) GetBillingAccountContracts(cloudAccoutID string) ([]*v1.Contract, error) {
```
cristianoveiga, line 68: I don't think  we need a new function for this. We can probably refactor/modify `GetBillingAccounts`.

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,83 @@ func run(cmd *cobra.Command, _ []string) {
...
+		cloudAccounts, err := r.OCMClient.GetBillingAccounts()
+		if err != nil {
+			r.Reporter.Errorf("%s", err)
+			os.Exit(1)
+		}
+
+		if len(cloudAccounts) > 0 {
+			billingAccount, err = interactive.GetOption(interactive.Input{
+				Question: "Billing Account",
+				Help:     cmd.Flags().Lookup("billing-account").Usage,
```
cristianoveiga, line 960: the billing-account flag is currently hidden - we probably want to make it visible now.

cristianoveiga (changes requested):

cmd/create/cluster/cmd.go
chenz4027: Looks like UI would show 0 as default. It might be better if we keep the consistency.

image
image

pkg/ocm/billing.go
chenz4027: k will work on that

cmd/create/cluster/cmd.go
chenz4027: Enabled

cmd/create/cluster/cmd.go
chenz4027: Will show it only conditions are met.

pkg/ocm/billing.go
chenz4027: Removed GetBillingAccounts.

cmd/create/cluster/cmd.go
chenz4027: Moved.

cmd/create/cluster/cmd.go
chenz4027: Removed.

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,95 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	if billingAccount == "" && !interactive.Enabled() {
+		billingAccount = aws.GetJumpAccount("local")
```
mikeshng, line 945: Minor comment. In `cmd/create/ocmrole/cmd.go` the env is from below and passed in to the GetJumpAccount
```
	env, err := ocm.GetEnv()
```
@cristianoveiga is it necessary to do this or just leave it hardcode as local?

cristianoveiga: nice catch @mikeshng . We need to use the ` awsCreator.AccountID` instead of the hard coded one.

See https://github.com/openshift/rosa/pull/1227/files#diff-943c0ab57b82c7028652df55e1d49e4f4e4dc41fa094dc4a9b3f2c35887d44b5R712-R715

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,97 @@ func run(cmd *cobra.Command, _ []string) {
...
+			billingAccounts = append(billingAccounts, account.CloudAccountID()+contractString)
+		}
+	}
+
+	if billingAccount == "" && !interactive.Enabled() {
+		if helper.HasPrefix(billingAccounts, awsCreator.AccountID) {
+			billingAccount = awsCreator.AccountID
+			r.Reporter.Infof("Using '%s' as billing account", billingAccount)
+			r.Reporter.Infof("To use a different billing account, add --billing-account xxxxxxxxxx to previous command")
+		}
```
mikeshng, line 970: Question: Should we just error out here?
The billing account is empty and we can't find a default value and the user is not using interactive so we can't supply another billing account value.

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,97 @@ func run(cmd *cobra.Command, _ []string) {
...
+						contractString = " [Contract enabled]"
+					}
+				}
+			}
+			billingAccounts = append(billingAccounts, account.CloudAccountID()+contractString)
+		}
+	}
+
+	if billingAccount == "" && !interactive.Enabled() {
+		if helper.HasPrefix(billingAccounts, awsCreator.AccountID) {
```
mikeshng, line 966: I like Cristiano's comment in the other PR and should be included.
```
// if a billing account is not provided we will try to use the infrastructure account as default
```

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,97 @@ func run(cmd *cobra.Command, _ []string) {
...
+						ocm.GetDimensionValue(dimension, "control_plane") > 0 {
+						contractString = " [Contract enabled]"
+					}
+				}
+			}
+			billingAccounts = append(billingAccounts, account.CloudAccountID()+contractString)
+		}
+	}
+
+	if billingAccount == "" && !interactive.Enabled() {
```
mikeshng, line 965: I feel like this whole block should be moved above the data massaging of cloudAccounts above. 
This block doesn't require the above logic besides a tiny bit of optimization.
Even if we want the optimization I think we can just separate out the aws filter into another helper function or loop.

pkg/helper/helpers.go
```diff
@@ -92,6 +92,15 @@ func Contains[T comparable](slice []T, element T) bool {
 	return false
 }
 
+func HasPrefix(slice []string, element string) bool {
```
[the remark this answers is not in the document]
chenz4027: Kk will change the name.

cmd/create/cluster/cmd.go
chenz4027: @mikeshng I think we are making it mandatory now

pkg/helper/helpers.go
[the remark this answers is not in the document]
chenz4027: Changed the name.

cmd/create/cluster/cmd.go
chenz4027: Done

cmd/create/cluster/cmd.go
chenz4027: @mikeshng I moved it around a bit, would you mind taking another look and see if it's where you expected?

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,97 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	var billingAccounts []string
+	cloudAccounts, err := r.OCMClient.GetBillingAccounts()
```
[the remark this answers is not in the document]
chenz4027: Added.

cmd/create/cluster/cmd.go
chenz4027: Refactored.

pkg/helper/helpers.go
```diff
@@ -92,6 +92,15 @@ func Contains[T comparable](slice []T, element T) bool {
 	return false
 }
 
+func ContainsPrefix(slice []string, element string) bool {
```
mikeshng, line 95:
```suggestion
func ContainsPrefix(slice []string, prefix string) bool {
```

mikeshng: minor nit

cmd/create/cluster/cmd.go
mikeshng: Looks good.

cmd/create/cluster/cmd.go
mikeshng: Can you double check? If it's mandatory then I think you should do an else and err out here.

mikeshng: Confirm with Lucas that we should allow empty => calls the backend and the backend will decide to error or not.

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,85 @@ func run(cmd *cobra.Command, _ []string) {
...
+				Question: "Billing Account",
+				Help:     cmd.Flags().Lookup("billing-account").Usage,
+				Default:  billingAccount,
+				Required: true,
+				Options:  billingAccounts,
+			})
+
+			billingAccount = ocm.RegexpReturnOnlyInteger(billingAccount)
+
+			if err != nil {
```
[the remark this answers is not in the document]
chenz4027: Moved.

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,85 @@ func run(cmd *cobra.Command, _ []string) {
...
+		var contracts []*accountsv1.Contract
+		var contractEnableString string
+		for _, account := range cloudAccounts {
+			if account.CloudAccountID() == billingAccount {
+				contracts = account.Contracts()
+				for _, contract := range contracts {
+					for _, dimension := range contract.Dimensions() {
+						if ocm.GetDimensionValue(dimension, "four_vcpu_hour") > 0 ||
+							ocm.GetDimensionValue(dimension, "control_plane") > 0 {
+							contractEnableString = "   Contract enabled for this billing account. \n"
```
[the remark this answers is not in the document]
chenz4027: Added a break for now.

pkg/helper/helpers.go
chenz4027: Changed

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,85 @@ func run(cmd *cobra.Command, _ []string) {
...
+		}
+	}
+
+	if interactive.Enabled() {
+		if err != nil {
+			r.Reporter.Errorf("%s", err)
+			os.Exit(1)
+		}
+
+		if len(cloudAccounts) > 0 {
```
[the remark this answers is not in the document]
chenz4027: Yep, done.

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,85 @@ func run(cmd *cobra.Command, _ []string) {
...
+		// if a billing account is not provided we will try to use the infrastructure account as default
+		if helper.ContainsPrefix(billingAccounts, awsCreator.AccountID) {
+			billingAccount = awsCreator.AccountID
+			r.Reporter.Infof("Using '%s' as billing account", billingAccount)
+			r.Reporter.Infof("To use a different billing account, add --billing-account xxxxxxxxxx to previous command")
+		}
+	}
+
+	if interactive.Enabled() {
+		if err != nil {
```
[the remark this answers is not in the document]
chenz4027: Removed.

mikeshng:
@chenz4027 I think we should unhold this PR. WDYT?

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,80 @@ func run(cmd *cobra.Command, _ []string) {
...
+		// Get contract info
+		var contracts []*accountsv1.Contract
+		var contractEnableString string
+		for _, account := range cloudAccounts {
+			if account.CloudAccountID() == billingAccount {
+				contracts = account.Contracts()
+				for _, contract := range contracts {
+					for _, dimension := range contract.Dimensions() {
+						if ocm.GetDimensionValue(dimension, "four_vcpu_hour") > 0 ||
+							ocm.GetDimensionValue(dimension, "control_plane") > 0 {
```
cristianoveiga, line 993: This logic is now used in 3 places - we probably could create a function for it

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,80 @@ func run(cmd *cobra.Command, _ []string) {
...
+							contractEnableString = "   Contract enabled for this billing account. \n"
```
cristianoveiga, line 994: maybe a typo here?  `contractEnabled`

chenz4027: Hmm got it from the design. I'm thinking that as it's in a string, probably the two words are split.
image

cmd/create/cluster/cmd.go
```diff
@@ -941,6 +941,80 @@ func run(cmd *cobra.Command, _ []string) {
...
+		}
+
+		if billingAccount != awsCreator.AccountID {
+			r.Reporter.Infof("The selected AWS billing account is a different account than your AWS infrastructure account." +
+				"The AWS billing account will be charged for subscription usage. " +
+				"The AWS infrastructure account will be used for managing the cluster. \n" +
+				contractEnableString)
+		} else {
+			r.Reporter.Infof("Using '%s' as billing account. \n"+
+				contractEnableString,
```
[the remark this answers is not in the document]
chenz4027: Looks like the string would exist in both cases though
image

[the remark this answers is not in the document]
chenz4027: Refactored

cmd/create/cluster/cmd.go
chenz4027: Changed the var.

cmd/create/cluster/cmd.go
chenz4027: Added the function.

cmd/create/cluster/cmd.go
chenz4027: Hide it for invalid contracts.

cmd/create/cluster/cmd.go
```diff
@@ -991,6 +991,79 @@ func run(cmd *cobra.Command, _ []string) {
...
+				"The AWS billing account will be charged for subscription usage. " +
+				"The AWS infrastructure account will be used for managing the cluster. \n")
+		} else {
+			r.Reporter.Infof("Using '%s' as billing account.",
+				billingAccount)
+		}
+
+		if isContractEnabled {
+			PrintContracts(contracts, r)
+		}
```
cristianoveiga, line 1063: This will print contracts for all accounts - you probably want to print the contracts for the selected account

cmd/create/cluster/cmd.go
```diff
@@ -3017,6 +3090,34 @@ func run(cmd *cobra.Command, _ []string) {
...
+func PrintContracts(contracts []*accountsv1.Contract, r *rosa.Runtime) {
+	for _, contract := range contracts {
+		format := "Jan 02, 2006"
+		dimensions := contract.Dimensions()
+		var numberOfString string
+
+		for _, dimension := range dimensions {
+			numberOfVPCs := ocm.GetDimensionValue(dimension, "four_vcpu_hour")
+			numberOfClusters := ocm.GetDimensionValue(dimension, "control_plane")
+			numberOfString = numberOfString + fmt.Sprintf("   | Number of vCPUs:    |'%s'             | \n"+
```
cristianoveiga, line 3102: nit: `numberOfString` is not mnemonic - it might be worth it to find another name.

cmd/create/cluster/cmd.go
```diff
@@ -991,6 +991,79 @@ func run(cmd *cobra.Command, _ []string) {
...
+					for _, dimension := range contract.Dimensions() {
+						if ocm.HasValidContracts(dimension) {
+							isContractEnabled = true
+							break
+						}
+					}
+				}
+
+			}
+		}
```
cristianoveiga, line 1050: This can probably be moved to its own function

cmd/create/cluster/cmd.go
chenz4027: It's getting the contracts from the selected billing account: https://github.com/openshift/rosa/pull/1500/files#diff-943c0ab57b82c7028652df55e1d49e4f4e4dc41fa094dc4a9b3f2c35887d44b5R1039

cmd/create/cluster/cmd.go
chenz4027: How about prePurchaseInfo?

cmd/create/cluster/cmd.go
chenz4027: k will pull that out

chenz4027: How about now? Will do a rebase and squash once all issues are addressed.

cmd/create/cluster/cmd.go
chenz4027: Updated with prePurchaseInfo, but we can change the name

cmd/create/cluster/cmd.go
```diff
@@ -3017,6 +3075,54 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 }
 
+func GenerateContracts(cloudAccounts []*accountsv1.CloudAccount, billingAccount string) ([]*accountsv1.Contract, bool) {
```
cristianoveiga, line 3078: maybe something like `GetBillingAccountContracts` would be more representative, what do you think?

chenz4027: Agreed, changing...

chenz4027: Done

cmd/create/cluster/cmd.go
```diff
@@ -991,6 +991,64 @@ func run(cmd *cobra.Command, _ []string) {
...
+
+	billingAccounts := ocm.GenerateBillingAccountsList(cloudAccounts)
+
+	if billingAccount == "" && !interactive.Enabled() {
+		// if a billing account is not provided we will try to use the infrastructure account as default
+		if helper.ContainsPrefix(billingAccounts, awsCreator.AccountID) {
+			billingAccount = awsCreator.AccountID
+			r.Reporter.Infof("Using '%s' as billing account", billingAccount)
+			r.Reporter.Infof("To use a different billing account, add --billing-account xxxxxxxxxx to previous command")
+		}
```
cristianoveiga, line 1008: @chenz4027 I think we have a gap here - what happens if the infrastructure account is not on the list of accounts?

Looks like it will try to create the cluster with an empty billing account.

chenz4027: Added error message.

cmd/create/cluster/cmd.go
```diff
@@ -3017,6 +3077,55 @@ func run(cmd *cobra.Command, _ []string) {
...
+func PrintContracts(contracts []*accountsv1.Contract, r *rosa.Runtime) {
+	for _, contract := range contracts {
+		format := "Jan 02, 2006"
+		dimensions := contract.Dimensions()
+		var prePurchaseInfo string
+
+		for _, dimension := range dimensions {
+			numberOfVPCs := ocm.GetDimensionValue(dimension, "four_vcpu_hour")
+			numberOfClusters := ocm.GetDimensionValue(dimension, "control_plane")
+			prePurchaseInfo = prePurchaseInfo + fmt.Sprintf("   | Number of vCPUs:    |'%s'             | \n"+
```
cristianoveiga, line 3110: prePurchaseInfo will be empty at this point - probably safe to remove it from the right side of the assignment.

cmd/create/cluster/cmd.go
```diff
@@ -3017,6 +3077,55 @@ func run(cmd *cobra.Command, _ []string) {
...
+
+func PrintContracts(contracts []*accountsv1.Contract, r *rosa.Runtime) {
+	for _, contract := range contracts {
+		format := "Jan 02, 2006"
+		dimensions := contract.Dimensions()
+		var prePurchaseInfo string
+
+		for _, dimension := range dimensions {
+			numberOfVPCs := ocm.GetDimensionValue(dimension, "four_vcpu_hour")
+			numberOfClusters := ocm.GetDimensionValue(dimension, "control_plane")
```
cristianoveiga, line 3109: @chenz4027 I believe this logic needs to be adjusted. My understanding is that it's not implementing the behaviour you would expect, i.e: the dimensions are not correctly assigned to the vars. PTAL

pkg/ocm/billing.go
```diff
@@ -34,22 +36,61 @@ func (c *Client) GetBillingAccounts() ([]string, error) {
...
+
+func HasValidContracts(dimension *v1.ContractDimension) bool {
+	if GetDimensionValue(dimension, "four_vcpu_hour") > 0 ||
+		GetDimensionValue(dimension, "control_plane") > 0 {
+		return true
+	}
+
+	return false
+
+}
```
cristianoveiga, line 92: A ` HasValidContracts` function that accepts a cloudAccount parameter and iterates over its dimensions would make more sense to me.

pkg/ocm/billing.go
```diff
@@ -34,22 +36,61 @@ func (c *Client) GetBillingAccounts() ([]string, error) {
...
+		}
+	}
+	return billingAccounts
+
+}
+
+func GetDimensionValue(dimension *v1.ContractDimension, resource string) int {
+	var value = 0
+	if dimension.Name() == resource {
+		value, _ = strconv.Atoi(dimension.Value())
```
cristianoveiga, line 79: I'm not sure we can assume that dimensions are always going to be convertible  to integers (except for four_vcpu_hour and control_plane) - probably better to deal with it separately.

cmd/create/cluster/cmd.go
```diff
@@ -991,6 +991,66 @@ func run(cmd *cobra.Command, _ []string) {
...
+			os.Exit(1)
+		}
+
+		// Get contract info
+		contracts, isContractEnabled := GetBillingAccountContracts(cloudAccounts, billingAccount)
+
+		if billingAccount != awsCreator.AccountID {
+			r.Reporter.Infof("The selected AWS billing account is a different account than your AWS infrastructure account." +
+				"The AWS billing account will be charged for subscription usage. " +
+				"The AWS infrastructure account will be used for managing the cluster. \n")
```
cristianoveiga, line 1042: The `\n` seems to be causing an extra/undesired line break

cmd/create/cluster/cmd.go
```diff
@@ -991,6 +991,66 @@ func run(cmd *cobra.Command, _ []string) {
...
+				Required: true,
+				Options:  billingAccounts,
+			})
+
+			if err != nil {
+				r.Reporter.Errorf("Expected a valid billing account: '%s'", err)
+				os.Exit(1)
+			}
+
+			billingAccount = ocm.RegexpReturnOnlyInteger(billingAccount)
```
cristianoveiga, line 1028: The `ParseOption` function can be reused here. See `pkg/aws/helpers.go`

cmd/create/cluster/cmd.go
```diff
@@ -3017,6 +3077,55 @@ func run(cmd *cobra.Command, _ []string) {
...
+			numberOfClusters := ocm.GetDimensionValue(dimension, "control_plane")
+			prePurchaseInfo = prePurchaseInfo + fmt.Sprintf("   | Number of vCPUs:    |'%s'             | \n"+
+				"   | Number of clusters: |'%s'             | \n",
+				strconv.Itoa(numberOfVPCs), strconv.Itoa(numberOfClusters))
+
+		}
+
+		r.Reporter.Infof(
+			"\n"+
+				"   Contract enabled for this billing account. \n"+
```
cristianoveiga, line 3118: Looking at the design doc, it seems "Using [account number ] as billing account" is missing.

It should go before the "Contract enabled for this billing account" text, PTAL.

cmd/create/cluster/cmd.go
chenz4027: @cristianoveiga The prePurchaseInfo will not be empty if there are multiple dimensions though. I recall that we would need to accommodate the scenario where there could be multiple dimensions and multiple contracts.

cmd/create/cluster/cmd.go
chenz4027: @cristianoveiga I'm a bit lost on this one, would you mind elaborating a bit more on this?

pkg/ocm/billing.go
chenz4027: So far it's only used for four_vcpu_hour and control_plane, but I got you, we can make it more generic. Just to confirm, are you suggesting that we add two cases for those two types. For the rest, we don't do the number conversion, but just return the value?

cmd/create/cluster/cmd.go
mikeshng: @chenz4027 I think what @cristianoveiga is saying is you might overwrite your previously set value with a 0 with the current logic.

So let say dimensions is [ 4 vpcs , 2 clusters ]
The first time it goes through the loop 
numberOfVPCs get set to 4
numberOfClusters get set to 0
then second time it goes through the loop
numberOfVPCs is now set to 0 (!?)
numberOfClusters is now set to 2.

I think what you need to do to fix it is use a if or case statements to only set numberOfVPCs when the dimension name matches four_vcpu_hour and only set numberOfClusters when dimension matches control_plane

pkg/ocm/billing.go
mikeshng: @chenz4027 how about GetNumsOfVPCsAndClusters(takes in a dimension slice) return (int, int) first being numberOfVPCs and second being numberOfClusters.
In the function iterate over the dimensions slice and set the numbers according and return them at the end. return 0s as fall back.
I think this might help address @cristianoveiga other concern about HasValidContracts as well.

cmd/create/cluster/cmd.go
chenz4027: It will always show up if the billing account is not the AWS infrastructure account.
image

cmd/create/cluster/cmd.go
chenz4027: Thanks! Replaced with the ParseOption

cmd/create/cluster/cmd.go
chenz4027: Removed \n.

pkg/ocm/billing.go
chenz4027: Thanks @mikeshng @cristianoveiga , updated the function as https://github.com/openshift/rosa/pull/1500/files#diff-54e259e8c0b289af4e61624a5e697f0285de9d9e0e14375c7d4fa82c772038eaR69

pkg/ocm/billing.go
chenz4027: Refactored as: https://github.com/openshift/rosa/pull/1500/files#diff-54e259e8c0b289af4e61624a5e697f0285de9d9e0e14375c7d4fa82c772038eaR82

cmd/create/cluster/cmd.go
chenz4027: New implementation: https://github.com/openshift/rosa/pull/1500/files#diff-54e259e8c0b289af4e61624a5e697f0285de9d9e0e14375c7d4fa82c772038eaR86

cristianoveiga: IMO, this is still a bit off. The main point is that you are calling GetDimensionValue inside the dimensions loop. This way will either get a valid value for the four_vcpu_hour or control_plane.

chenz4027: Updated

chenz4027:
Closing as it was messed up. New PR: https://github.com/openshift/rosa/pull/1540

PR #1502, opened 2023-09-14
OCM-3267 | feat: Add interactive mode for shared VPC

oriAdler:
1. If the user didn't provide all the flags for the shared VPC flow, default to interactive mode.
2. Display AWS owner ID alongside the subnet selection.

Subnet selection:
```
? Install into an existing VPC (optional): Yes
W: Some subnets have been excluded because they do not fit into chosen CIDR ranges
? Subnet IDs (optional):  [Use arrows to move, space to select, <right> to all, <left> to none, type to filter, ? for more help]
> [ ]  subnet-0457887d7b27a56d3 ('disktest-bthz8-public-us-east-1a','vpc-04df95f8a614f7851','us-east-1a', Owner ID: '501358428071')
  [ ]  subnet-04a42219d1aa4297b ('disktest-bthz8-private-us-east-1a','vpc-04df95f8a614f7851','us-east-1a', Owner ID: '501358428071')
  [ ]  subnet-0be38de5605f1c4ad ('','vpc-05c422525aadd0731','us-east-1a', Owner ID: '962993289388')
  [ ]  subnet-0ed8ae29f147bf7ae ('','vpc-05c422525aadd0731','us-east-1d', Owner ID: '962993289388')
  [ ]  subnet-0ed6f282dc9e22e9a ('','vpc-05c422525aadd0731','us-east-1a', Owner ID: '962993289388')
  [ ]  subnet-0676575e02b630a9c ('','vpc-05c422525aadd0731','us-east-1c', Owner ID: '962993289388')
  [ ]  subnet-00c6d8c20e8ffb2c1 ('','vpc-05c422525aadd0731','us-east-1c', Owner ID: '962993289388')
```

Informing the user about shared VPC:
`I: Subnet with ID 'subnet-0be38de5605f1c4ad' is shared by AWS account '962993289388', Defaulting to installing a cluster into a shared VPC.`

Shared VPC input fields:
```
? Private hosted zone ID: Z02058382IP472TYY7PGG
? Shared VPC role ARN: arn:aws:iam::962993289388:role/oadler-shared-vpc
? Base Domain:  [Use arrows to move, type to filter, ? for more help]
> 31v6.s1.devshift.org
  34nb.s1.devshift.org
  sg9f.s1.devshift.org
  u2gj.s1.devshift.org
  xio1.s1.devshift.org
  xj4p.s1.devshift.org
  zi0e.s1.devshift.org
```

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -764,23 +764,6 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
-	// validate flags for shared vpc
-	isSharedVPC := false
-	privateHostedZoneID := strings.Trim(args.privateHostedZoneID, " \t")
-	sharedVPCRoleARN := strings.Trim(args.sharedVPCRoleARN, " \t")
-	baseDomain := strings.Trim(args.baseDomain, " \t")
-	if privateHostedZoneID != "" ||
-		sharedVPCRoleARN != "" {
-		if privateHostedZoneID == "" ||
-			sharedVPCRoleARN == "" ||
-			baseDomain == "" {
-			r.Reporter.Errorf("To install a cluster into a shared VPC, " +
-				"private-hosted-zone-id, shared-vpc-role-arn and base-domain should be provided together")
-			os.Exit(1)
-		}
-		isSharedVPC = true
-	}
-
 	supportedRegions, err := r.OCMClient.GetDatabaseRegionList()
 	if err != nil {
 		r.Reporter.Errorf("Unable to retrieve supported regions: %v", err)
@@ -1807,6 +1790,7 @@ func run(cmd *cobra.Command, _ []string) {
 
 	var availabilityZones []string
 	var subnets []*ec2.Subnet
+	mapSubnetIDToSubnet := make(map[string]aws.Subnet)
 	if useExistingVPC || subnetsProvided {
 		initialSubnets, err := awsClient.GetSubnetIDs()
 		if err != nil {
@@ -1851,7 +1835,6 @@ func run(cmd *cobra.Command, _ []string) {
 			useExistingVPC = false
 			subnetsProvided = false
 		}
-		mapSubnetToAZ := make(map[string]string)
 		mapAZCreated := make(map[string]bool)
 		options := make([]string, len(subnets))
 		defaultOptions := make([]string, len(subnetIDs))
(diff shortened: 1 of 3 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -1879,6 +1863,7 @@ func run(cmd *cobra.Command, _ []string) {
 			subnetID := awssdk.StringValue(subnet.SubnetId)
 			availabilityZone := awssdk.StringValue(subnet.AvailabilityZone)
 			mapSubnetToAZ[subnetID] = availabilityZone
+			mapSubnetToOwnerID[subnetID] = awssdk.StringValue(subnet.OwnerId)
```
gdbranco, line 1866: Can we create a map of subnet ID to some middle structure so we can keep a single map to all information required? Basically unify the AZ and OwnerID

cmd/create/cluster/shared_vpc.go
```diff
@@ -0,0 +1,96 @@
...
+func getBaseDomain(r *rosa.Runtime, cmd *cobra.Command, baseDomain string) (string, error) {
+	dnsDomains, err := getAvailableBaseDomains(r)
+	if err != nil {
+		return "", err
+	}
+
+	res, err := interactive.GetOption(interactive.Input{
+		Question: "Base Domain",
+		Help:     cmd.Flags().Lookup("base-domain").Usage,
+		Default:  baseDomain,
```
marcolan018, line 67: should we validate the provided baseDomain by users exist or not, like the current logic for subnet ids?

cmd/create/cluster/shared_vpc.go
```diff
@@ -0,0 +1,96 @@
...
+
+	dnsDomains, err := r.OCMClient.ListDNSDomains(
+		fmt.Sprintf("user_defined='true' and cluster.id='' and organization.id='%s'", organizationID))
+	if err != nil {
+		return nil, errors.Errorf("Failed to list DNS domains: %s", err)
+	}
+
+	var dnsDomainsIDs []string
+	for _, dnsDomain := range dnsDomains {
+		dnsDomainsIDs = append(dnsDomainsIDs, dnsDomain.ID())
```
marcolan018, line 92: do we need to filter out the dnsdomain which already has a cluster linked here?

cmd/create/cluster/shared_vpc.go
oriAdler: If the user is in an interactive mode, they can choose only from the list of DNS domains retrieved from the backend.
So if they provide a domain that doesn't exist, they won't be able to continue. 
Do you mean checking for the option of providing a base domain using a flag?

cmd/create/cluster/shared_vpc.go
oriAdler: We do filter out using this query `"user_defined='true' and cluster.id='' and organization.id='%s'"`
Do you think it's better to display all domains? and get the error from the backend?

marcolan018: that makes sense. I ignored the cluster.id='' before

cmd/create/cluster/shared_vpc.go
marcolan018: I mean the similar code in subnet ids here: https://github.com/openshift/rosa/blob/ca5ffb4daa2b3748975f8eea1c2d6a9a14855a6a/cmd/create/cluster/cmd.go#L1859-L1873

oriAdler: Since we have validation in the backend, I think it's ok.
Just thinking about duplicate login across the CLI and the backend.
Also, the error message returned from the backend will include more details in case the base domain exists but is associated with another cluster.

cmd/create/cluster/cmd.go
oriAdler: Changed, thanks.

PR #1505, opened 2023-09-14
OCM-3366 | fix: CLI options dependency

osherdp:
Marking some CLI options as dependant on each other (required together). This should make it more clear that the user should enable autoscaling in general when trying to apply cluster-autoscaler options.
In addition, min and max options are marked required together to make it clearer what value is used (rather than implicitly assuming the default values).

Some examples for how it's showing:
```
$ rosa edit autoscaler -c my-cluster --min-memory 0
...
Failed to execute root command: if any flags in the group [min-memory max-memory] are set they must all be set; missing [max-memory]
```

```
$ rosa create cluster -n my-cluster --autoscaler-scale-down-enabled
...
Failed to execute root command: if any flags in the group [enable-autoscaling autoscaler-scale-down-enabled] are set they must all be set; missing [enable-autoscaling]
```


osherdp:
/hold
making sure this is the behavior we'd like to have

PR #1507, opened 2023-09-14
OCM-3714 | fix: don't ask for disk size for hosted cp

marcolan018:
https://issues.redhat.com/browse/OCM-3714

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -2262,7 +2262,8 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	var machinePoolRootDisk *ocm.Volume
-	if args.machinePoolRootDiskSize != "" || interactive.Enabled() {
+	if !isHostedCP &&
+		(args.machinePoolRootDiskSize != "" || interactive.Enabled()) {
 		var machinePoolRootDiskSizeStr string
 		if args.machinePoolRootDiskSize == "" {
 			// We don't need to parse the default since it's returned from the OCM API and AWS


cmd/create/cluster/cmd.go
```diff
@@ -2262,7 +2262,8 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	var machinePoolRootDisk *ocm.Volume
-	if args.machinePoolRootDiskSize != "" || interactive.Enabled() {
+	if !isHostedCP &&
```
oriAdler, line 2265: Is that going to skip setting a default value for `machinePoolRootDisk` for hosted CP clusters?

oriAdler: Just before the end of the block, we have a check, that sets the default value if needed:
```
if machinePoolRootDiskSize != defaultMachinePoolRootDiskSize {
			machinePoolRootDisk = &ocm.Volume{
				Size: machinePoolRootDiskSize,
			}
		}
```

oriAdler: Or do we just have to pass an empty value here to the backend for hosted CP?

marcolan018: I think any settings for disk size are valid only for classic rosa cluster? If that's true, the default machinePoolRootDisk is not necessary for hosted cp.

oriAdler: Can you please check the flow with the changes for hosted CP?
Previously, we set a default value for hosted CP as well and now we skip it.
Just to verify the backend knows how to handle that.

marcolan018: @oriAdler 
in https://github.com/openshift/rosa/blob/af33827c75cfcd84a6b8161a2b76b5ef43cc4306/cmd/create/cluster/cmd.go#L2338-L2341
a non-empty machinePoolRootDisk will be passed into cluster config only for non-default value.
so with this change, the flow for hosted cp should be as same as before.

btw, in CS side, if there is any value disksize passed in for hosted cp, it will return bad request
https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/cmd/clusters-service/service/cluster_service.go#L1280-1281

oriAdler: Okay, thanks for the clarification.

PR #1508, opened 2023-09-15
OCM-3737 | fix: align version computation create machinepool for HCP

andreadecorte:
HCP control plane version needs to be hcp enabled versions. For HCP node pools instead, node pool version can be any version which is rosa enabled. This is already checked at upgrade time, we now align also the behaviour at creation time.

Related: [OCM-3737](https://issues.redhat.com//browse/OCM-3737)


andreadecorte:
@lucasponce this is the minimal change to align behaviour with the one on backend side. I keep separated the discussion on improving version handling

ciaranRoche (approved):
/lgtm

PR #1511, opened 2023-09-16
OCM-3694 | feat: support day 1 additional compute sg ids

gdbranco:
Related issue: https://issues.redhat.com/browse/OCM-3694


cmd/create/cluster/cmd.go
```diff
@@ -735,6 +739,13 @@ The password must
 			"associated with intended shared VPC, e.g., '1vo8.p1.openshiftapps.com'.",
 	)
 
+	flags.StringSliceVar(
+		&args.additionalComputeSecurityGroupIds,
+		additionalComputeSecurityGroupIdsFlag,
+		nil,
```
hunterkepley, line 745: Would it be better to avoid nil and use an empty slice?

pkg/ocm/versions.go
```diff
@@ -28,16 +28,18 @@ import (
...
+	DefaultChannelGroup                            = "stable"
+	NightlyChannelGroup                            = "nightly"
+	LowestSTSSupport                               = "4.7.11"
+	LowestHttpTokensRequiredSupport                = "4.11.0"
+	LowestSTSMinor                                 = "4.7"
+	MinVersionForManagedIngressV2                  = "4.14-0"
+	MinVersionForAdditionalComputeSecurityGroupIds = "4.14-0"
+	VersionPrefix                                  = "openshift-v"
+
+	LowestHostedCPSupport = "4.12.0-0.a" //TODO: Remove the 0.a once stable 4.12 builds are available
```
hunterkepley, line 42: NP: `LowestHostedCpSupport`

hunterkepley (commented):
overall LGTM, only a couple of comments

cmd/create/cluster/cmd.go
gdbranco: Nil helps in identifying if the value changed in some cases where we can't use cmd.Flags for some reason, that seems to be a standard looking at other `StringSliceVar`. We may want to revisit that later to adjust so all aren't null

PR #1515, opened 2023-09-22
OCM-2083 | fix: prompts with default value are not 'optional'

thomasmckay:
https://issues.redhat.com/browse/OCM-2083

Prompts during interactive mode with a default value should not be called "optional". The only prompt type that was violating this is the "boolean" type (y/n).


willkutler (approved):

willkutler:
LGTM

PR #1519, opened 2023-09-26
OCM-1625 | feat: Add the ability to list clusters using an Account Role

robpblake:
This PR adds the flag `--account-role-arn` to the `rosa list cluster` command to supporting listing clusters that are using the specified role.

When a user specifies the role of an ARN via the flag, we:

1. Lookup the Role to see if it exists
2. If the Role exists, check the `type` of the role
3. Perform a query to the `/api/clusters_mgmt/v1/clusters` endpoint, setting the query string to look for the role based upon the role type

### DIFF
--- a/cmd/list/cluster/cmd.go
+++ b/cmd/list/cluster/cmd.go
@@ -18,12 +18,13 @@ package cluster
 
 import (
 	"fmt"
-	"github.com/openshift/rosa/pkg/aws"
 	"os"
 	"text/tabwriter"
 
+	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 
+	"github.com/openshift/rosa/pkg/aws"
 	"github.com/openshift/rosa/pkg/output"
 	"github.com/openshift/rosa/pkg/rosa"
 )
@@ -39,15 +40,31 @@ var Cmd = &cobra.Command{
 	Run:  run,
 }
 
-var listAll bool
+const clusterCount = 1000
+
+var args struct {
+	listAll        bool
+	accountRoleArn string
+}
 
 func init() {
 	flags := Cmd.Flags()
 	flags.SortFlags = false
 
 	output.AddFlag(Cmd)
-	flags.BoolVarP(&listAll, "all", "a", false, "List all clusters across different AWS "+
+	flags.BoolVarP(&args.listAll, "all", "a", false, "List all clusters across different AWS "+
 		"accounts under the same Red Hat organization")
+	flags.StringVar(&args.accountRoleArn, "account-role-arn", "", "List all clusters "+
+		"using the account role identified by the ARN")
+}
+
+func listClustersUsingAccountRole(creator *aws.Creator, runtime *rosa.Runtime) ([]*v1.Cluster, error) {
+	role, err := runtime.AWSClient.GetAccountRoleByArn(args.accountRoleArn)
+	if err != nil {
+		return []*v1.Cluster{}, err
+	}
+
+	return runtime.OCMClient.GetClustersUsingAccountRole(creator, role, clusterCount)
 }
 
 func run(_ *cobra.Command, _ []string) {
@@ -56,12 +73,21 @@ func run(_ *cobra.Command, _ []string) {
 
 	// Retrieve the list of clusters:
 	var creator *aws.Creator
-	if listAll {
+	if args.listAll {
 		creator = nil
 	} else {
 		creator = r.Creator
 	}
-	clusters, err := r.OCMClient.GetClusters(creator, 1000)
+
+	var clusters []*v1.Cluster
+	var err error
+
+	if args.accountRoleArn != "" {
+		clusters, err = listClustersUsingAccountRole(creator, r)
+	} else {
+		clusters, err = r.OCMClient.GetClusters(creator, clusterCount)
+	}
+
 	if err != nil {
 		r.Reporter.Errorf("Failed to get clusters: %v", err)
 		os.Exit(1)
--- a/pkg/aws/client_test.go
+++ b/pkg/aws/client_test.go
@@ -1,6 +1,9 @@
 package aws_test
 
 import (
+	"fmt"
+
+	awsSdk "github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
@@ -12,6 +15,7 @@ import (
 
 	"github.com/openshift/rosa/pkg/aws"
 	"github.com/openshift/rosa/pkg/aws/mocks"
+	rosaTags "github.com/openshift/rosa/pkg/aws/tags"
 )
 
 var _ = Describe("Client", func() {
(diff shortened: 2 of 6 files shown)


cmd/list/cluster/cmd.go
```diff
@@ -18,6 +18,7 @@ package cluster
 
 import (
 	"fmt"
+	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
```
ciaranRoche, line 21: NP: to sort the ordering on these imports. These imports are already wrong in this file, but we try to keep them in groups, from std, external and then internal pkgs. :pray:

pkg/aws/client_test.go
```diff
@@ -1,13 +1,16 @@
...
+	"fmt"
+	aws2 "github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	tags2 "github.com/openshift/rosa/pkg/aws/tags"
```
gdbranco, line 13: NP: personally I'd rather name the import tags and the variable something else

pkg/aws/client_test.go
```diff
@@ -1,13 +1,16 @@
 package aws_test
 
 import (
+	"fmt"
+	aws2 "github.com/aws/aws-sdk-go/aws"
```
ciaranRoche, line 5: NP: same as above, just add some whitespace between the `fmt` import and the `aws2` import

gdbranco:
lgtm

gdbranco (approved):

ciaranRoche (commented):
small nit picks, but other than that looks good to me 

Validated in prod
```
╭─croche@fedora ~/Work/ocm/rosa ‹ocm-1625-list-clusters-using-accountrole●› 
╰─$ rosa list clusters --account-role-arn arn:aws:iam::765374464689:role/ManagedOpenShift-Installer-Role
ID                                NAME           STATE  TOPOLOGY
1u42ijpktv18lobegtb6min5m5gscgb7  testovn1       ready  Classic (STS)
20agiuq3dmuu7oer2hiedhpqq0p4st5q  isaac-cluster  error  Classic (STS)
╭─croche@fedora ~/Work/ocm/rosa ‹ocm-1625-list-clusters-using-accountrole●› 
╰─$ rosa list clusters --account-role-arn arn:aws:iam::765374464689:role/nargaman-account-tf-Installer-Role
ID                                NAME         STATE  TOPOLOGY
26gqvkbu974p331kkmnc7m9huigdj3ra  nargaman-tf  ready  Classic (STS)
╭─croche@fedora ~/Work/ocm/rosa ‹ocm-1625-list-clusters-using-accountrole●› 
╰─$ rosa list clusters --account-role-arn arn:aws:iam::765374464689:role/nargaman-account-tf-Worker-Role

ID                                NAME         STATE  TOPOLOGY
26gqvkbu974p331kkmnc7m9huigdj3ra  nargaman-tf  ready  Classic (STS)
╭─croche@fedora ~/Work/ocm/rosa ‹ocm-1625-list-clusters-using-accountrole●› 
╰─$ rosa list clusters --account-role-arn arn:aws:iam::765374464689:role/not-a-role                     

E: Failed to get clusters: NoSuchEntity: The role with name not-a-role cannot be found.
	status code: 404, request id: 62d026a1-39a2-4be7-8949-f5852e5ed87c
```

pkg/aws/client_test.go
ciaranRoche: +1 this, same for `aws2`

cmd/list/cluster/cmd.go
```diff
@@ -39,15 +40,34 @@ var Cmd = &cobra.Command{
...
+}
+
+func listClustersUsingAccountRole(creator *aws.Creator, runtime *rosa.Runtime) ([]*v1.Cluster, error) {
+	role, err := runtime.AWSClient.GetAccountRoleByArn(accountRoleArn)
+	if err != nil {
+		return []*v1.Cluster{}, err
+	}
+
+	if role == nil {
+		return nil, fmt.Errorf("no Account role with ARN '%s' exists", accountRoleArn)
```
oriAdler, line 67: In case the role doesn't exist on AWS, do we get this error from this API call `iamClient.GetRole`?
Regardless of that, I wonder if we should do this check and the error handling inside `GetAccountRoleByArn`?

cmd/list/cluster/cmd.go
```diff
@@ -39,15 +40,34 @@ var Cmd = &cobra.Command{
 	Run:  run,
 }
 
+const clusterCount = 1000
+
 var listAll bool
 
+var accountRoleArn string
```
oriAdler, line 47: Can we please wrap this variable in a struct named `args`?
Just the convention we have in many other commands.

pkg/aws/policies.go
```diff
@@ -743,16 +752,34 @@ func (c *awsClient) listPolicies(role *iam.Role) ([]Policy, error) {
...
-	roles, err := c.ListRoles()
+func (c *awsClient) GetAccountRoleByArn(arn string) (*Role, error) {
+	role, err := c.GetRoleByARN(arn)
 	if err != nil {
-		return accountRoles, err
+		return nil, err
 	}
+
+	roles := []*iam.Role{role}
+	accountRoles, err := c.mapToAccountRoles("", roles)
```
oriAdler, line 762: Should we refactor only the part that handles a single role to `mapToAccountRole`? That way we don't need `roles := []*iam.Role{role}`, and can reuse the function inside a loop in `ListAccountRoles`. WDYT?

cmd/list/cluster/cmd.go
gdbranco: We should adopt the same for backend to auto format this. Will look into that later

pkg/aws/policies.go
robpblake: Yep, good call. I'll do that 👍

cmd/list/cluster/cmd.go
robpblake: Done.

cmd/list/cluster/cmd.go
robpblake: Yeah you're right. The error bubbles up already from the `iamClient.GetRole` call if the role doesn't exist, so this check is redundant

pkg/aws/client_test.go
robpblake: @gdbranco @ciaranRoche I'm not assigning a variable using these packages, only relying on a function from them. There is a clash between the ROSA `aws` package and the AWS SDK `aws` package. I'll rename to the following:

- `aws2` --> `awsSdk`
- `tags2` --> `rosaTags`

Let me know if that works for you 👍

cmd/list/cluster/cmd.go
robpblake: @gdbranco I created us a JIRA for this a while back: https://issues.redhat.com/browse/OCM-3618

robpblake:
@gdbranco @ciaranRoche @oriAdler All your comments I believe are addressed. Could you please take another look.

cheers!

Rob

cmd/list/cluster/cmd.go
```diff
@@ -39,15 +40,35 @@ var Cmd = &cobra.Command{
...
+		"using the account role identified by the ARN")
+}
+
+func listClustersUsingAccountRole(creator *aws.Creator, runtime *rosa.Runtime) ([]*v1.Cluster, error) {
+	role, err := runtime.AWSClient.GetAccountRoleByArn(args.accountRoleArn)
+	if err != nil {
+		return []*v1.Cluster{}, err
+	}
+
+	if role == nil {
```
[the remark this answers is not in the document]
robpblake: @oriAdler I removed this check in favour of `GetAccountRoleByArn` returning an error if the role does not exist.

If the role doesn't have tags, then it is still returned from the `mapToAccountRole` function, but the `Policy` fields are not filled. I've updated it so that the `RoleName` and `RoleARN` fields are always completed, which is a slight change to existing behaviour,  but from what I can see, that shouldn't be a problem.

PR #1520, opened 2023-09-26
OCM-3875 | feat: remove printing operator roles, and oidc provider on…

ciaranRoche:


ciaranRoche:
@gdbranco what was the motivation to print the commands in the past? Was wondering why we did not just omit them if the user passed in an oidc?

gdbranco:
> @gdbranco what was the motivation to print the commands in the past? Was wondering why we did not just omit them if the user passed in an oidc?

The user might not have pre created the resources even though the configuration exists. If we go to this route we need to at least check within AWS then

PR #1522, opened 2023-09-29
OCM-3831 | fix: 'rosa describe admin' cannot find 'cluster-admin' htpasswd idp and user (updated)

davidleerh:
Covers fixes for both cards:
[OCM-3831](https://issues.redhat.com//browse/OCM-3831), [OCM-3293](https://issues.redhat.com//browse/OCM-3293)

**Initial Issue:**
The `rosa describe admin -c <cluster_id>` command is unable to find the idp that has the admin user, even though the admin user was created previously.

Returned response:
"There is no admin on cluster '%s'. To create it run the following command:\n"+
" rosa create admin -c %s"
 
Expected response:
"There is an admin on cluster '%s'. To login, run the following command:\n"+
" oc login %s --username %s"

**Reason**
For commands:
`rosa create admin` - always created admin username "cluster-admin" under the htpasswd IDP named "Cluster-Admin"
`rosa create cluster `--cluster-admin-user` - required users to generate admin username not equal to "cluster-admin" under the htpasswd IDP named "Cluster-Admin"

To cover both issues [OCM-3831](https://issues.redhat.com//browse/OCM-3831), [OCM-3293](https://issues.redhat.com//browse/OCM-3293); 
It was decided that the cluster admin user created via ROSA will always have the username "cluster-admin" under the htpasswd IDP named "Cluster-Admin".
Users will no longer be able to generate a customer username for the admin user created via ROSA.

A1. Indicate `--cluster-admin-user` flag is deprecated.
image

A2. Indicate `--create-admin-user` and `--cluster-admin-password` flags can only be used with classic ROSA.
image

A3. Using `--create-admin-user` and `--cluster-admin-password` flags and electing `hosted cp` via interactive prompt will return an error.
image

B. Users using flag `--create-admin-user` without `--cluster-admin-password` flag will auto generate the password
[Please note, --create-admin-user does not take in an input as it is now a boolean flag]
image

C. Users using flag `--create-admin-user` AND `--create-admin-password` flag will allow user specified password
image

D. Users can use `--cluster-admin-password` flag without `--create-admin-user` flag
[Using only `--cluster-admin-password` flag lets ROSA know that the user wants to create a cluster admin user]
image

E. Users can auto generate password via interactive prompts by answering `no` to  `Create custom password for cluster admin` question
image

F. Users can input custom password via interactive prompts by answering `yes` to  `Create custom password for cluster admin` question
image

G. Build command output generated at the end of create cluster prompts correctly includes new admin flags:
image

With the fixes, 
`rosa delete admin` and `rosa describe admin` will always work now if the cluster admin user was created via ROSA; since both commands search for the username, 'cluster-admin' within the htpasswd idp to confirm whether a cluster admin user exists.

**EDIT**
Kept --cluster-admin-password instead of --create-admin-password, following request.

### DIFF
--- a/cmd/create/admin/cmd.go
+++ b/cmd/create/admin/cmd.go
@@ -32,6 +32,8 @@ import (
 
 const ClusterAdminUsername = "cluster-admin"
 const ClusterAdminIDPname = "cluster-admin"
+const GeneratingRandomPasswordString = "Generating random password"
+const MaxPasswordLength = 23
 
 var Cmd = &cobra.Command{
 	Use:   "admin",
@@ -83,8 +85,8 @@ func run(cmd *cobra.Command, _ []string) {
 	var err error
 	passwordArg := args.passwordArg
 	if len(passwordArg) == 0 {
-		r.Reporter.Debugf("Generating random password")
-		password, err = generateRandomPassword(23)
+		r.Reporter.Debugf(GeneratingRandomPasswordString)
+		password, err = GenerateRandomPassword()
 		if err != nil {
 			r.Reporter.Errorf("Failed to generate a random password")
 			os.Exit(1)
@@ -171,7 +173,7 @@ func run(cmd *cobra.Command, _ []string) {
 	r.Reporter.Infof("It may take several minutes for this access to become active.")
 }
 
-func generateRandomPassword(length int) (string, error) {
+func GenerateRandomPassword() (string, error) {
 	const (
 		lowerLetters = "abcdefghijkmnopqrstuvwxyz"
 		upperLetters = "ABCDEFGHIJKLMNPQRSTUVWXYZ"
@@ -179,7 +181,7 @@ func generateRandomPassword(length int) (string, error) {
 		all          = lowerLetters + upperLetters + digits
 	)
 	var password string
-	for i := 0; i < length; i++ {
+	for i := 0; i < MaxPasswordLength; i++ {
 		n, err := rand.Int(rand.Reader, big.NewInt(int64(len(all))))
 		if err != nil {
 			return "", err
@@ -188,7 +190,7 @@ func generateRandomPassword(length int) (string, error) {
 		if password == "" {
 			password = newchar
 		}
-		if i < length-1 {
+		if i < MaxPasswordLength-1 {
 			n, err = rand.Int(rand.Reader, big.NewInt(int64(len(password)+1)))
 			if err != nil {
 				return "", err
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -35,6 +35,7 @@ import (
 	passwordValidator "github.com/openshift-online/ocm-common/pkg/idp/validations"
 	diskValidator "github.com/openshift-online/ocm-common/pkg/machinepool/validations"
 	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/openshift/rosa/cmd/create/admin"
 	"github.com/openshift/rosa/cmd/create/idp"
 	"github.com/openshift/rosa/cmd/create/oidcprovider"
 	"github.com/openshift/rosa/cmd/create/operatorroles"
@@ -176,8 +177,10 @@ var args struct {
 	billingAccount       string
 
 	// Cluster Admin
-	clusterAdminUser     string
+	createAdminUser      bool
 	clusterAdminPassword string
+	// Deprecated Cluster Admin
+	clusterAdminUser string
 
 	// Audit Log Forwarding
 	AuditLogRoleARN string
@@ -657,23 +660,29 @@ func init() {
 	)
 	flags.MarkHidden("billing-account")
 
+	flags.BoolVar(
+		&args.createAdminUser,
+		"create-admin-user",
+		false,
+		`Create cluster admin`,
+	)
+
 	flags.StringVar(
 		&args.clusterAdminUser,
 		"cluster-admin-user",
 		"",
-		`Username of Cluster Admin.
-Username must not contain /, :, or %%`,
+		`Deprecated cluster admin flag. Please use --create-admin-user.`,
 	)
-
 	flags.StringVar(
 		&args.clusterAdminPassword,
 		"cluster-admin-password",
 		"",
-		`Password of Cluster Admin.
-The password must
-- Be at least 14 characters (ASCII-standard) without whitespaces
-- Include uppercase letters, lowercase letters, and numbers or symbols (ASCII-standard characters only)`,
+		`The password must
+		- Be at least 14 characters (ASCII-standard) without whitespaces
+		- Include uppercase letters, lowercase letters, and numbers or symbols (ASCII-standard characters only)`,
 	)
+	// cluster admin flags deprecated to be removed
+	flags.MarkHidden("cluster-admin-user")
 
 	flags.StringVar(
 		&args.AuditLogRoleARN,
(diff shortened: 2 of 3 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -871,27 +899,39 @@ func run(cmd *cobra.Command, _ []string) {
 		args.ec2MetadataHttpTokens = string(v1.Ec2MetadataHttpTokensOptional)
 	}
 
+	// Errors when users elects for cluster admin via flags and elects for hosted control plane via interactive prompt"
+	if isHostedCP && (createAdminUser || createAdminPassword != "") {
```
marcolan018, line 903: better to check this as early as possible

cmd/create/cluster/cmd.go
```diff
@@ -3135,8 +3192,8 @@ func buildCommand(spec ocm.Spec, operatorRolesPrefix string,
 		}
 	}
 	if spec.ClusterAdminUser != "" {
-		command += fmt.Sprintf(" --cluster-admin-user %s", spec.ClusterAdminUser)
-		command += fmt.Sprintf(" --cluster-admin-password %s", spec.ClusterAdminPassword)
+		command += " --create-admin-user"
+		command += fmt.Sprintf(" --create-admin-password %s", spec.ClusterAdminPassword)
```
marcolan018, line 3196: should we hide the password here?

cmd/create/cluster/cmd.go
```diff
@@ -3135,8 +3192,8 @@ func buildCommand(spec ocm.Spec, operatorRolesPrefix string,
 		}
 	}
 	if spec.ClusterAdminUser != "" {
-		command += fmt.Sprintf(" --cluster-admin-user %s", spec.ClusterAdminUser)
-		command += fmt.Sprintf(" --cluster-admin-password %s", spec.ClusterAdminPassword)
+		command += " --create-admin-user"
```
marcolan018, line 3195: can "--create-admin-password" be used w/o "--create-admin-user", based on above code, seems it's allowed?

cmd/create/cluster/cmd.go
davidleerh: I have another check that occurs earlier that checks for the combination of 3 flags 
https://github.com/davidleerh/rosa/blob/OCM-3831-ADMIN/cmd/create/cluster/cmd.go#L783
`--create-admin-user --create-admin-password --hosted-cp`
image


Condition on line 903 is a separate check that occurs if the user supplies flags related to cluster admin, but 
hosted-cp is selected during interactive prompt.

image

cmd/create/cluster/cmd.go
davidleerh: I believe we may want to have the password shown since it might interfere with any automation if the user decides to copy and paste the command with a redacted password, as indicated here: https://redhat-internal.slack.com/archives/C04C3G5MDKN/p1695329663534979

cmd/create/cluster/cmd.go
davidleerh: Yes, it was intentional.
During office hours, I believe the consensus was that we wanted to let the users have three command combinations:
A. --create-admin-user 
> creates the admin user with random password

B. --create-admin-user --create-admin-password
> creates the admin user with user specified password

C. --create-admin-password
> creates the admin user with user specified password

Please correct me if I misunderstood or if I didn't remember correctly. :)

marcolan018: right, that's what we want.
But here, if user has selected any option including option C, there will be flag "--create-admin-password". I am not sure it's ok, even I know this will not have impact on the behavior

davidleerh: Ah I see what you mean :) Thank you for pointing it out.
Will discuss this during our next call to agree which build command output meets best practice.

cmd/create/cluster/cmd.go
```diff
@@ -871,27 +899,39 @@ func run(cmd *cobra.Command, _ []string) {
...
-			if err != nil {
-				r.Reporter.Errorf("%s", err)
-				os.Exit(1)
-			}
+		if createAdminUser {
 			isClusterAdmin = true
+			// user supplies create-admin-user flag without create-admin-password will generate random password
+			if createAdminPassword == "" {
+				r.Reporter.Debugf("Generating random password")
+				createAdminPassword, err = admin.GenerateRandomPassword(23)
```
oriAdler, line 916: NP
Can we please use a const here?

cmd/create/cluster/cmd.go
```diff
@@ -871,27 +899,39 @@ func run(cmd *cobra.Command, _ []string) {
 		args.ec2MetadataHttpTokens = string(v1.Ec2MetadataHttpTokensOptional)
 	}
 
+	// Errors when users elects for cluster admin via flags and elects for hosted control plane via interactive prompt"
+	if isHostedCP && (createAdminUser || createAdminPassword != "") {
+		r.Reporter.Errorf("Setting Cluster Admin is only supported in classic ROSA clusters")
+		os.Exit(1)
+	}
+
+	// isClusterAdmin is aflag indicating if user wishes to create cluster admin
```
oriAdler, line 908:
```suggestion
	// isClusterAdmin is a flag indicating if user wishes to create cluster admin
```

cmd/create/cluster/cmd.go
```diff
@@ -653,23 +658,38 @@ func init() {
...
+	flags.BoolVar(
+		&args.createAdminUser,
+		"create-admin-user",
+		false,
+		`Create cluster admin`,
+	)
+
+	flags.StringVar(
+		&args.createAdminPassword,
+		"create-admin-password",
```
oriAdler, line 670: I'm not sure about the flag naming convention.
With `create-admin-password` one can expect this flag to be a boolean.
I wonder if we should go for something like `cluster-admin-password`, WDYT?

ciaranRoche:
adding a hold on this until after we cut the release today. Want to make sure QE have plenty of time to test these changes before adding to a release

cmd/create/cluster/cmd.go
davidleerh: I will go ahead and add it to the create/admin file as the GenerateRandomPassword method also originates from there.

cmd/create/cluster/cmd.go
davidleerh: Edited via sep. commit. Thank you for the find!

cmd/create/cluster/cmd.go
davidleerh: Hi @oriAdler
The changes requested here definitely seems more intuitive. I will go ahead and make this change.
In addition, since it will now be `create-admin-user` and `cluster-admin-password`, should we enforce flag `cluster-admin-password` to always be used with `create-admin-user`? or should we allow users to be able to 
use  `cluster-admin-password` without `create-admin-user`?

Related to confusion here: https://github.com/openshift/rosa/pull/1522#discussion_r1341568029

cmd/create/cluster/cmd.go
davidleerh: Moving discussion to: https://github.com/openshift/rosa/pull/1522#discussion_r1342137338

cmd/create/cluster/cmd.go
oriAdler: IMO only if the user assigns a password.
So if the user selected an auto-generated password option, we don't print `cluster-admin-password`, and the next time the user runs the command the password will be auto-generated.
With that being said, I think both options apply.

davidleerh: the output of the build command after the commit will show as 
`rosa create cluster --create-admin-user` if the user decides to create admin user without custom password.
`rosa create cluster --create-admin-user` if the user decides to create admin user with custom password.

cmd/create/admin/cmd.go
```diff
@@ -83,8 +84,8 @@ func run(cmd *cobra.Command, _ []string) {
 	var err error
 	passwordArg := args.passwordArg
 	if len(passwordArg) == 0 {
-		r.Reporter.Debugf("Generating random password")
-		password, err = generateRandomPassword(23)
+		r.Reporter.Debugf(GeneratingRandomPasswordString)
+		password, err = GenerateRandomPassword(23)
```
oriAdler, line 88: NP
Can we use a const here for the password length?

cmd/create/cluster/cmd.go
```diff
@@ -657,23 +660,27 @@ func init() {
...
-
 	flags.StringVar(
 		&args.clusterAdminPassword,
 		"cluster-admin-password",
 		"",
-		`Password of Cluster Admin.
-The password must
-- Be at least 14 characters (ASCII-standard) without whitespaces
-- Include uppercase letters, lowercase letters, and numbers or symbols (ASCII-standard characters only)`,
+		`Password of Cluster Admin.`,
```
oriAdler, line 680: Should we provide details about the password requirements in the help message? 
Like we have done previously.

oriAdler (commented):
Left two small comments, other than that, LGTM.

cmd/create/admin/cmd.go
davidleerh: I will go ahead and remove the param requirement of the length of the password, this method is solely used for admin user password generation. I will store the length of the password as a const where the method is defined.
/cmd/create/admin/cmd.go

cmd/create/cluster/cmd.go
davidleerh: Thanks for the find!

davidleerh:
Thanks :1st_place_medal: :1st_place_medal: :1st_place_medal:  @oriAdler :1st_place_medal: :1st_place_medal: :1st_place_medal: for the detailed review.

ciaranRoche (approved):
/lgtm

PR #1524, opened 2023-10-02
OCM-3833 | fix: only show interactive SGs if there are any

gdbranco:
Related issue: [OCM-3833](https://issues.redhat.com//browse/OCM-3833)


ciaranRoche (commented):
/lgtm

PR #1531, opened 2023-10-09
OCM-3526 | feat: create operator roles for shared VPC with cluster key

oriAdler:
Adding support to create operator roles after cluster creation. 
The shared VPC role ARN is available as part of the cluster spec.

### DIFF
--- a/cmd/create/operatorroles/cmd.go
+++ b/cmd/create/operatorroles/cmd.go
@@ -180,11 +180,6 @@ func run(cmd *cobra.Command, argv []string) error {
 		cluster = r.FetchCluster()
 	}
 
-	if args.prefix == "" && args.sharedVpcRoleArn != "" {
-		r.Reporter.Errorf("Shared VPC role ARN is only supported when setting the 'prefix' flag")
-		os.Exit(1)
-	}
-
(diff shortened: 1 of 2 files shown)


gdbranco:
Does this means we'll be waiting for the shared vpc role in waiting state now?

oriAdler:
> Does this means we'll be waiting for the shared vpc role in waiting state now?

Yes, I'm going to work on a separate PR in the backend, which will validate that the ingress operator can assume the shared VPC role.

cmd/create/operatorroles/cmd.go
```diff
@@ -180,11 +180,6 @@ func run(cmd *cobra.Command, argv []string) error {
 		cluster = r.FetchCluster()
 	}
 
-	if args.prefix == "" && args.sharedVpcRoleArn != "" {
```
marcolan018, line 183: should we continue to support those parameters, but only prompt that those parameters already deprecreated?

oriAdler: We still use both parameters for creating operator roles prior to the cluster creation.
The main flow for shared VPC is creating the operator roles using these two parameters.
In addition, it would be possible to create the operator roles after cluster creation.

gdbranco (approved):

PR #1536, opened 2023-10-09
OCM-4123 | fix: Remove unnecessary warning messages about Red Hat owned subnets when user explicitly specifies subnets to use

robpblake:
This PR updates the `getInitialValidSubnets` function in the `rosa create cluster` command to consider the list of subnets passed as a command line argument by the user via the `--subnet-ids` switch.

The `getInitialValidSubnets` function treats the subnets specified via `--subnet-ids` as the only ones it check, rather than pulling and checking all available subnets (which is the cause of the erroneous warning). The function still checks if any of the subnets specified by the user via `--subnet-ids` are Red Hat managed and will print the warning if they are.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -1815,7 +1815,7 @@ func run(cmd *cobra.Command, _ []string) {
 	var subnets []*ec2.Subnet
 	mapSubnetIDToSubnet := make(map[string]aws.Subnet)
 	if useExistingVPC || subnetsProvided {
-		initialSubnets, err := getInitialValidSubnets(awsClient, r.Reporter)
+		initialSubnets, err := getInitialValidSubnets(awsClient, args.subnetIDs, r.Reporter)
 		if err != nil {
 			r.Reporter.Errorf("Failed to get the list of subnets: %s", err)
 			os.Exit(1)
@@ -3444,14 +3444,16 @@ func getExpectedResourceIDForAccRole(hostedCPPolicies bool, roleARN string, role
 	return strings.ToLower(fmt.Sprintf("%s-%s-Role", rolePrefix, accountRoles[roleType].Name)), rolePrefix, nil
 }
 
-func getInitialValidSubnets(awsClient aws.Client, reporter *reporter.Object) ([]*ec2.Subnet, error) {
+func getInitialValidSubnets(aws aws.Client, ids []string, r *reporter.Object) ([]*ec2.Subnet, error) {
 	initialValidSubnets := []*ec2.Subnet{}
 	excludedSubnets := []string{}
-	allSubnets, err := awsClient.GetSubnetIDs()
+
+	validSubnets, err := aws.ListSubnets(ids...)
+
 	if err != nil {
 		return initialValidSubnets, err
 	}
-	for _, subnet := range allSubnets {
+	for _, subnet := range validSubnets {
 		hasRHManaged := tags.Ec2ResourceHasTag(subnet.Tags, tags.RedHatManaged, strconv.FormatBool(true))
 		if !hasRHManaged {
 			initialValidSubnets = append(initialValidSubnets, subnet)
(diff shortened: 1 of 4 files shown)


pkg/aws/client.go
```diff
@@ -431,10 +432,24 @@ func (c *awsClient) GetRegion() string {
 	return aws.StringValue(c.awsSession.Config.Region)
 }
 
-func (c *awsClient) GetSubnetIDs() ([]*ec2.Subnet, error) {
+func (c *awsClient) ListSubnets() ([]*ec2.Subnet, error) {
 	return c.getSubnetIDs(&ec2.DescribeSubnetsInput{})
 }
 
+func (c *awsClient) GetSubnetsById(subnetIds []string) ([]*ec2.Subnet, error) {
```
gdbranco, line 439: I feel like these two are essentially the same and we could probably have a single function that receives one or more subnets with the spread operator. What do you think?

pkg/aws/client.go
```diff
@@ -431,10 +432,24 @@ func (c *awsClient) GetRegion() string {
...
+
+	// Convert the user provided slice of strings to pointer
+	ids := make([]*string, len(subnetIds))
+	for _, id := range subnetIds {
+		id2 := id
+		ids = append(ids, &id2)
+	}
```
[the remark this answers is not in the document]
robpblake: Thanks for the tip @ziccardi . I'm new to Go, so all help is appreciated :-)

pkg/aws/client.go
robpblake: Yeah that's a good idea @gdbranco . I'll make the changes.

pkg/aws/client.go
```diff
@@ -431,10 +431,21 @@ func (c *awsClient) GetRegion() string {
 	return aws.StringValue(c.awsSession.Config.Region)
 }
 
-func (c *awsClient) GetSubnetIDs() ([]*ec2.Subnet, error) {
-	return c.getSubnetIDs(&ec2.DescribeSubnetsInput{})
-}
+func (c *awsClient) ListSubnets(subnetIds ...string) ([]*ec2.Subnet, error) {
+
+	if subnetIds == nil {
```
gdbranco, line 436: This could be
```suggestion
	if len(subnetIds) == 0 {
```

cmd/create/cluster/cmd.go
```diff
@@ -3444,23 +3444,32 @@ func getExpectedResourceIDForAccRole(hostedCPPolicies bool, roleARN string, role
...
-func getInitialValidSubnets(awsClient aws.Client, reporter *reporter.Object) ([]*ec2.Subnet, error) {
+func getInitialValidSubnets(aws aws.Client, ids []string, r *reporter.Object) ([]*ec2.Subnet, error) {
 	initialValidSubnets := []*ec2.Subnet{}
 	excludedSubnets := []string{}
-	allSubnets, err := awsClient.GetSubnetIDs()
+
+	var validSubnets []*ec2.Subnet
+	var err error
+
+	if len(ids) == 0 {
```
gdbranco, line 3454: If this is checked internally there's no need to have this separation when calling

pkg/aws/client.go
```diff
@@ -431,10 +431,21 @@ func (c *awsClient) GetRegion() string {
...
 }
 
-func (c *awsClient) GetSubnetIDs() ([]*ec2.Subnet, error) {
-	return c.getSubnetIDs(&ec2.DescribeSubnetsInput{})
-}
+func (c *awsClient) ListSubnets(subnetIds ...string) ([]*ec2.Subnet, error) {
+
+	if subnetIds == nil || len(subnetIds) == 0 {
+		return c.getSubnetIDs(&ec2.DescribeSubnetsInput{})
+	}
```
ziccardi, line 438: No need to check for `nil`
```suggestion
	if len(subnetIds) == 0 {
		return c.getSubnetIDs(&ec2.DescribeSubnetsInput{})
	}
```

pkg/aws/client_test.go
```diff
@@ -261,4 +263,59 @@ var _ = Describe("Client", func() {
...
+		subnetTwoId := "test-subnet-2"
+		subnet := ec2.Subnet{
+			SubnetId: helpers.NewString(subnetOneId),
+		}
+
+		subnet2 := ec2.Subnet{
+			SubnetId: helpers.NewString(subnetTwoId),
+		}
+
+		subnets := make([]*ec2.Subnet, 0)
```
ziccardi, line 279:
```suggestion
		var subnets []*ec2.Subnet
```

PR #1537, opened 2023-10-10
OCM-3366 | fix: force --enable-autoscaling flag when autoscaling flags are used

davidleerh:
Please see two separate calls
image

In Day 1 cluster create, it now properly forces the flag `--enable-autoscaling` if any of the autoscaling flags are used.


ciaranRoche (approved):
/lgtm

PR #1542, opened 2023-10-13
OCM-3911 | feat: use HCP specific endpoints for versions and upgrades

andreadecorte:
We need to handle HCP versions in the backend side and they may also depend on the customer.

To obtain this, the current change:
1) introduces the use of an HCP specific version endpoint 
2) switch all the available upgrades calls to what is available in the cluster or the node pool API response, as this is the source of truth. This will ensure that the list is the correct one and reduce the number of calls in the client needed to get the version properties.

Functionally there should be no change.

Improved also the testing to use always SDK objects instead of raw JSON which will help test clarity and mantainence.

Related: [OCM-3911](https://issues.redhat.com//browse/OCM-3911)

### DIFF
--- a/cmd/upgrade/machinepool/cmd.go
+++ b/cmd/upgrade/machinepool/cmd.go
@@ -21,7 +21,6 @@ import (
 	"os"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
-	"github.com/openshift/rosa/pkg/helper/versions"
 	"github.com/openshift/rosa/pkg/input"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/interactive/confirm"
@@ -334,11 +333,9 @@ func checkNodePoolExistingScheduledUpgrade(r *rosa.Runtime, cluster *cmv1.Cluste
 
 func ComputeNodePoolVersion(r *rosa.Runtime, cmd *cobra.Command, cluster *cmv1.Cluster,
 	nodePool *cmv1.NodePool, version string) (string, error) {
+	var err error
 	channelGroup := cluster.Version().ChannelGroup()
-	filteredVersionList, err := GetAvailableVersion(r, cluster, nodePool)
-	if err != nil {
-		return "", err
-	}
+	filteredVersionList := ocm.GetNodePoolAvailableUpgrades(nodePool)
 	// No updates available
 	if len(filteredVersionList) == 0 {
 		return "", nil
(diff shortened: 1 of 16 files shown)


cmd/upgrade/machinepool/cmd.go
```diff
@@ -369,20 +367,10 @@ func ComputeNodePoolVersion(r *rosa.Runtime, cmd *cobra.Command, cluster *cmv1.C
...
+func formatNodePool(nodePool *cmv1.NodePool) string {
+	var nodePoolJson bytes.Buffer
 
-	// Filter the available list of versions for a hosted machine pool
-	filteredVersionList := versions.GetFilteredVersionListForUpdate(versionList, nodePoolVersion, clusterVersion)
-	if err != nil {
-		return nil, err
-	}
-	return filteredVersionList, nil
+	cmv1.MarshalNodePool(nodePool, &nodePoolJson)
```
lucasponce, line 373: Mmm this is weird, using the SDK, why do you need to perform a marshall at this point?

(perhaps I missed context in the PR)

andreadecorte: good catch, this is an helper function for the tests, but I have now a generic helper so I remove it in favour of the generic method. Nothing for the normal flow in any case

lucasponce:
LGTM

pkg/helper/versions/helpers.go
```diff
@@ -26,7 +33,7 @@ func GetVersionList(r *rosa.Runtime, channelGroup string, isSTS bool, isHostedCP
 		if isSTS && !ocm.HasSTSSupport(v.RawID(), v.ChannelGroup()) {
 			continue
 		}
-		if isHostedCP {
+		if filterHostedCP {
```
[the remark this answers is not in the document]
andreadecorte: because we have 2 main use case:
- control plane versions
- node pool versions 
Both needs to an HCP minimal version filtering.
The CP versions are filtered based on `hosted_control_plane` field in each version, while nodes can have any ROSA enabled version compatible (so we don't filter if `hosted_control_plane` is enabled or not).

Hence having 2 flags: isHostedCP to apply minimal filter
filterHostedCP to filter differently based on control plan or node pool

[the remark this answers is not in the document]
oriAdler: Thanks for the clarification.

pkg/ocm/versions.go
```diff
@@ -77,6 +82,7 @@ func (c *Client) GetVersions(channelGroup string, defaultFirst bool) (versions [
 			Order(order).
 			Page(page).
 			Size(size).
+			Parameter("product", product).
```
oriAdler, line 85: Should we send the `product` parameter also when it's an empty string?

oriAdler (commented):
Left two small comments, other than that LGTM.

pkg/ocm/versions.go
andreadecorte: good point, sending now only if it's not empty, it is cleaner

pkg/ocm/versions.go
```diff
@@ -204,6 +210,20 @@ func (c *Client) GetAvailableUpgrades(versionID string) ([]string, error) {
 	return availableUpgrades, nil
 }
 
+func (c *Client) GetAvailableUpgradesByCluster(cluster *cmv1.Cluster) []string {
```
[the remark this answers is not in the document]
andreadecorte: removed

oriAdler:
LGTM

PR #1557, opened 2023-10-23
OCM-4449 | feat: add information about inflight check failures on cluster describe

gdbranco:
Related issue: OCM-4449


cmd/describe/cluster/cmd.go
```diff
@@ -413,12 +414,103 @@ func run(cmd *cobra.Command, argv []string) {
...
+	}
+
 	str = fmt.Sprintf("%s\n", str)
 
 	// Print short cluster description:
 	fmt.Print(str)
 }
 
+var mapInflightErrorTypeToTitle = map[string]string{
+	"egress_url_errors": "Egress URL access",
```
davidleerh, line 455: Egress URL access -> Egress URL errors?

cmd/describe/cluster/cmd.go
```diff
@@ -413,12 +414,103 @@ func run(cmd *cobra.Command, argv []string) {
...
+					r.Logger.Errorf("Unexpected error parsing inflight details '%s: %v", inflight.ID(), err)
+					continue
+				}
+				summary += details
+				summaries = append(summaries, summary)
+			}
+		}
+		if len(summaries) > 0 {
+			str += fmt.Sprintf("Failed Inflight Checks:\n%s\n", strings.Join(summaries, "\n"))
+			str += fmt.Sprintf("\tPlease run `rosa verify network -c %s` after adjusting"+
```
davidleerh, line 442: After tomorrow's discussion, 
should we output this for inflight.Type() == "egress" only?

cmd/describe/cluster/cmd.go
gdbranco: So we call it egress errors within, I wanted something like connectivity but I find it too long. Access felt fairly ok

cmd/describe/cluster/cmd.go
gdbranco: So I think we should be able to resolve all eventually. Like if a user adjusts the subnet tags and has the warning they need to be able to clear up the inflight too

gdbranco: Showing only the egress errors now

cmd/describe/cluster/cmd.go
gdbranco: Added 'access issues'

PR #1564, opened 2023-10-27
OCM-3963

den-rgb:
## Describe your changes

This pr is to update the `aws sdk go` version from `v1` to `v2`, the majority of these changes are refactoring the code for it to work with the updated library.

### JIRA link
[JIRA](https://issues.redhat.com/browse/OCM-3963)

### What's new?

- Refactoring of functions to work with the updated library
- Updated test cases
- Added interfaces and associated functions from the v2 [repo](https://github.com/aws/aws-sdk-go-v2/tree/main/service)

### How to test?

- All test cases should pass within the repo
- All basic commands should work as prior
- QE testing pipelines pass

## PR Author Check List

* [x]  All commit messages adhere to the project [standard](https://github.com/openshift/rosa/blob/master/CONTRIBUTING.md) 

* [ ]  All commits are squashed into a single commit `git rebase -i HEAD~<number of commits>`

* [x]  All code changes have unit test coverage

* [ ]  All changes have been tested locally and the steps documented in the `How to Test?` section


pkg/aws/helpers.go
```diff
@@ -506,15 +506,19 @@ func GetOIDCProviderARN(accountID string, providerURL string) string {
...
+	}
 	region, err := GetRegion(arguments.GetRegion())
 	if err != nil || region == "" {
-		return endpoints.AwsPartitionID
+		return "aws"
 	}
-	partition, ok := endpoints.PartitionForRegion(endpoints.DefaultPartitions(), region)
-	if !ok || partition.ID() == "" {
-		return endpoints.AwsPartitionID
+	e, err := cfg.EndpointResolver.ResolveEndpoint(region, "sts")
```
den-rgb, line 517: need to find a way to implement this function using `aws.EndPointResolver`

cmd/create/cluster/cmd.go
```diff
@@ -1983,27 +1983,28 @@ func run(cmd *cobra.Command, _ []string) {
 			for _, subnetArg := range subnetIDs {
 				verifiedSubnet := false
 				for _, subnet := range subnets {
-					if awssdk.StringValue(subnet.SubnetId) == subnetArg {
+					if awssdk.ToString(subnet.SubnetId) == subnetArg {
 						verifiedSubnet = true
 					}
 				}
 				if !verifiedSubnet {
+					reg, _ := r.AWSClient.GetRegion()
```
gdbranco, line 1991: Is the second return param an error now? If so we might need to check it

pkg/aws/client.go
```diff
@@ -423,32 +403,45 @@ func (b *ClientBuilder) Build() (Client, error) {
...
+	if err != nil {
+		return aws.Credentials{}, err
+	}
+	return creds, nil
 }
 
-func (c *awsClient) GetRegion() string {
-	return aws.StringValue(c.awsSession.Config.Region)
+func (c *awsClient) GetRegion() (string, error) {
+	cfg, err := config.LoadDefaultConfig(context.TODO())
```
oriAdler, line 420: Previously we stored the `awsSession` in the `awsClient` struct, and accessed it to fetch the region.
Since the `session` isn't available with V2, should we store the `cfg` in the `awsClient`, so we can access here `cfg.Region` instead of loading the config, that way we also don't need to handle an error here.

cmd/create/cluster/cmd.go
```diff
@@ -2026,27 +2026,32 @@ func run(cmd *cobra.Command, _ []string) {
 			for _, subnetArg := range subnetIDs {
 				verifiedSubnet := false
 				for _, subnet := range subnets {
-					if awssdk.StringValue(subnet.SubnetId) == subnetArg {
+					if awssdk.ToString(subnet.SubnetId) == subnetArg {
 						verifiedSubnet = true
 					}
 				}
 				if !verifiedSubnet {
+					reg, err := r.AWSClient.GetRegion()
```
oriAdler, line 2034: Changing the `GetRegion` function (left a comment in the client) will save us error handling here.

cmd/create/cluster/cmd.go
```diff
@@ -3360,8 +3365,9 @@ func validateAvailabilityZones(multiAZ bool, availabilityZones []string, awsClie
 	}
 	for _, az := range availabilityZones {
 		if !helper.Contains(regionAvailabilityZones, az) {
+			reg, _ := awsClient.GetRegion()
```
oriAdler, line 3368: We should check the error returned from the function, by changing the `GetRegion` to get the region from the `cfg` stored in the `awsClient` struct, the function will return only a string.

cmd/create/ocmrole/cmd.go
```diff
@@ -310,6 +310,7 @@ func run(cmd *cobra.Command, argv []string) {
 		}
 		var commands string
 		commands, err = buildCommands(
+			r,
```
oriAdler, line 313: We can omit `r.Creator.AccountID` in line `318` and access both the partition and the account ID from within the function.

cmd/create/ocmrole/cmd.go
```diff
@@ -382,7 +383,7 @@ func buildCommands(prefix string, roleName string, rolePath string, permissionsB
 			return "", err
 		}
 	} else {
-		policyARN = aws.GetPolicyARN(accountID, roleName, rolePath)
+		policyARN = aws.GetPolicyARN(r.WithAWS().AWSClient, accountID, roleName, rolePath)
```
oriAdler, line 386: We can pass here `r.Creator.partition` instead of the client.

cmd/create/ocmrole/cmd.go
```diff
@@ -417,7 +418,7 @@ func buildCommands(prefix string, roleName string, rolePath string, permissionsB
 				return "", err
 			}
 		} else {
-			policyARN = aws.GetAdminPolicyARN(accountID, roleName, rolePath)
+			policyARN = aws.GetAdminPolicyARN(r.WithAWS().AWSClient, accountID, roleName, rolePath)
```
oriAdler, line 421: We can pass here `r.Creator.partition` instead.

cmd/create/ocmrole/cmd.go
```diff
@@ -451,15 +452,15 @@ func createRoles(r *rosa.Runtime, prefix string, roleName string, rolePath strin
 			return "", err
 		}
 	} else {
-		policyARN = aws.GetPolicyARN(accountID, roleName, rolePath)
+		policyARN = aws.GetPolicyARN(r.WithAWS().AWSClient, accountID, roleName, rolePath)
```
oriAdler, line 455:
```suggestion
		policyARN = aws.GetPolicyARN(r.Creator.partition, accountID, roleName, rolePath)
```

cmd/create/ocmrole/cmd.go
```diff
@@ -451,15 +452,15 @@ func createRoles(r *rosa.Runtime, prefix string, roleName string, rolePath strin
...
 	}
 	if !confirm.Prompt(true, "Create the '%s' role?", roleName) {
 		os.Exit(0)
 	}
 	filename := fmt.Sprintf("sts_%s_trust_policy", aws.OCMRolePolicyFile)
 	policyDetail := aws.GetPolicyDetails(policies, filename)
-	policy := aws.InterpolatePolicyDocument(policyDetail, map[string]string{
-		"partition":           aws.GetPartition(),
+	policy := aws.InterpolatePolicyDocument(r.WithAWS().AWSClient, policyDetail, map[string]string{
```
oriAdler, line 462:
```suggestion
	policy := aws.InterpolatePolicyDocument(r.Creator.partition, policyDetail, map[string]string{
```

cmd/create/ocmrole/cmd.go
```diff
@@ -516,7 +517,7 @@ func createRoles(r *rosa.Runtime, prefix string, roleName string, rolePath strin
 				return "", err
 			}
 		} else {
-			policyARN = aws.GetAdminPolicyARN(accountID, roleName, "")
+			policyARN = aws.GetAdminPolicyARN(r.WithAWS().AWSClient, accountID, roleName, "")
```
oriAdler, line 520:
```suggestion
			policyARN = aws.GetAdminPolicyARN(r.Creator.partition, accountID, roleName, "")
```

cmd/create/ocmrole/cmd.go
```diff
@@ -534,8 +535,8 @@ func generateOcmRolePolicyFiles(r *rosa.Runtime, env string, orgID string, isAdm
 	filename := fmt.Sprintf("sts_%s_trust_policy", aws.OCMRolePolicyFile)
 
 	policyDetail := aws.GetPolicyDetails(policies, filename)
-	policy := aws.InterpolatePolicyDocument(policyDetail, map[string]string{
-		"partition":           aws.GetPartition(),
+	policy := aws.InterpolatePolicyDocument(r.WithAWS().AWSClient, policyDetail, map[string]string{
```
oriAdler, line 538:
```suggestion
	policy := aws.InterpolatePolicyDocument(r.Creator.partition, policyDetail, map[string]string{
```

cmd/create/operatorroles/by_clusterkey.go
```diff
@@ -303,7 +304,7 @@ func buildCommands(r *rosa.Runtime, env string,
 				return "", err
 			}
 		} else {
-			policyARN = computePolicyARN(r.Creator.AccountID, prefix, operator.Namespace(), operator.Name(), path)
+			policyARN = computePolicyARN(r, r.Creator.AccountID, prefix, operator.Namespace(), operator.Name(), path)
```
oriAdler, line 307: Since we send `r` to the function already, we can access `r.Creator.AccountID` from within the function.
```suggestion
			policyARN = computePolicyARN(r, prefix, operator.Namespace(), operator.Name(), path)
```

cmd/create/accountroles/cmd.go
```diff
@@ -406,7 +406,7 @@ func run(cmd *cobra.Command, argv []string) {
 			ocm.Version:  policyVersion,
 		})
 	case aws.ModeManual:
-		err = aws.GeneratePolicyFiles(r.Reporter, env, true, false, policies, nil, managedPolicies, "")
+		err = aws.GeneratePolicyFiles(r.WithAWS().AWSClient, r.Reporter, env, true, false, policies, nil, managedPolicies, "")
```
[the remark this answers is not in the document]
den-rgb: I think it'd be better to just pass in the partition rather than the whole runtime as that's the only thing the runtime is being used for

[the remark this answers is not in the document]
oriAdler: We already passing the `reporter` from the `Runtime`.
So we can either pass `r.Reporter` and `r.Creator.partition`, or just the `Runtime` and reference the fields inside the function.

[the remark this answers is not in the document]
den-rgb: I cant pass in the runtime because I cant reference it in the `pkg/aws` helper file

pkg/aws/client.go
```diff
@@ -587,28 +581,36 @@ func (c *awsClient) getSubnetRouteTable(subnetID *string, routeTables []*ec2.Rou
...
-	res, err := c.ec2Client.DescribeSubnets(describeSubnetsInput)
+func (c *awsClient) getSubnetIDs(describeSubnetsInput *ec2.DescribeSubnetsInput) ([]*ec2types.Subnet, error) {
+	res, err := c.ec2Client.DescribeSubnets(context.Background(), describeSubnetsInput)
 	if err != nil {
 		return nil, err
 	}
-	return res.Subnets, nil
+
+	var subnetPointers []*ec2types.Subnet
+	for _, subnet := range res.Subnets {
```
oriAdler, line 591: I wonder if this convertion is needed, from a quick glance, when we use the subnets later we either dereference the pointer or use `awssdk.ToString`.
As AWS simplified their SDK to return a slice of subnets, instead of a slice of pointers to subnets maybe it's worth adjusting our code if we eventually dereference the pointers.

pkg/aws/client.go
```diff
@@ -873,24 +876,22 @@ func (c *awsClient) DeleteAccessKeys(username string) error {
...
+		})
 	if err != nil {
-		if aerr, ok := err.(awserr.Error); ok {
-			switch aerr.Code() {
-			case iam.ErrCodeNoSuchEntityException:
-				return false, "", nil
-			default:
-				return false, "", err
-			}
+		var noSuchEntity *iamtypes.NoSuchEntityException
```
oriAdler, line 884: Should we use here the new function `IsNoSuchEntityException`?

pkg/aws/client.go
```diff
@@ -967,9 +970,12 @@ func (c *awsClient) DetachRolePolicies(roleName string) error {
 		if err != nil {
 			return err
 		}
-		isTruncated = *resp.IsTruncated
+		isTruncated = resp.IsTruncated
 		marker = resp.Marker
-		attachedPolicies = append(attachedPolicies, resp.AttachedPolicies...)
+		for _, attachedPolicy := range resp.AttachedPolicies {
```
oriAdler, line 975: If we create a slice of `iamtypes.AttachedPolicy` instead of a slice of pointers, we can remove the conversion here.

cmd/create/operatorroles/by_prefix.go
```diff
@@ -145,17 +143,9 @@ func handleOperatorRoleCreationByPrefix(r *rosa.Runtime, env string,
...
-			switch aerr.Code() {
-			case iam.ErrCodeNoSuchEntityException:
-				// If ErrCodeNoSuchEntityException we want to create so we may continue
-				break
-			default:
-				r.Reporter.Errorf("%v", err)
-				os.Exit(1)
-			}
-		}
+	if err != nil && !aws.IsNoSuchEntityException(err) {
```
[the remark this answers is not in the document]
den-rgb: Thank you

pkg/aws/client.go
den-rgb: Good point I'll refactor the related functions along with that.

pkg/aws/client.go
den-rgb: Good spot thank you

pkg/aws/client.go
```diff
@@ -343,71 +373,26 @@ func (b *ClientBuilder) Build() (Client, error) {
...
-	// TODO: No need to do this twice, we're essentially doing the
-	// same thing in getClientDetails()
-	// We should implement getClientDetails() here or a new validation func
-	_, err = sess.Config.Credentials.Get()
-	if err != nil {
-		b.logger.Debugf("Failed to find credentials: %v", err)
-		return nil, fmt.Errorf("Failed to find credentials. Check your AWS configuration and try again")
-	}
-
-	// Check that the region is set:
```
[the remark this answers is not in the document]
den-rgb: I can' seem to find this comment

pkg/aws/client.go
```diff
@@ -286,29 +283,72 @@ func (b *ClientBuilder) UseLocalCredentials(value bool) *ClientBuilder {
...
+func (b *ClientBuilder) BuildSessionWithOptions(logLevel aws.ClientLogMode) (*aws.Config, error) {
+	cfg, err := config.LoadDefaultConfig(context.TODO(),
+		config.WithSharedConfigProfile(profile.Profile()),
+		config.WithRegion(*b.region),
+		config.WithHTTPClient(&http.Client{
+			Transport: http.DefaultTransport,
+		}),
+		config.WithClientLogMode(logLevel),
+		config.WithAPIOptions([]func(stack *middleware.Stack) error{
+			smithyhttp.AddHeaderValue("rosa.ROSAVersionUserAgentHandler",
```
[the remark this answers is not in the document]
den-rgb: would it be worth extracting the whole value of `config.WithAPIOptions` as a var since it's all used twice

cmd/create/accountroles/cmd.go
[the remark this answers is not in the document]
oriAdler: Right, the cyclic import, thanks for the clarification.

pkg/aws/client.go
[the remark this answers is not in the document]
den-rgb: Sorry I interpreted that wrong, let me have a look

cmd/create/cluster/cmd.go
```diff
@@ -3651,21 +3651,23 @@ func getExpectedResourceIDForAccRole(hostedCPPolicies bool, roleARN string, role
...
 	if err != nil {
 		return initialValidSubnets, err
 	}
+
 	for _, subnet := range validSubnets {
 		hasRHManaged := tags.Ec2ResourceHasTag(subnet.Tags, tags.RedHatManaged, strconv.FormatBool(true))
 		if !hasRHManaged {
-			initialValidSubnets = append(initialValidSubnets, subnet)
+			s := subnet
+			initialValidSubnets = append(initialValidSubnets, &s)
```
oriAdler, line 3668: We are converting the subnets to pointers, and then in the `create cluster` we convert them back to subnets -> `awssdk.ToString(subnet.SubnetId)`
It's better to return a slice of subnets instead.

cmd/create/cluster/cmd.go
```diff
@@ -2001,7 +2001,7 @@ func run(cmd *cobra.Command, _ []string) {
 				!serviceNetwork.Contains(subnetIP) {
 				subnets = append(subnets, subnet)
 			} else {
-				excludedSubnetsDueToCidr = append(excludedSubnetsDueToCidr, awssdk.StringValue(subnet.SubnetId))
+				excludedSubnetsDueToCidr = append(excludedSubnetsDueToCidr, awssdk.ToString(subnet.SubnetId))
```
oriAdler, line 2004: Returning a slice of subnets from `getInitialValidSubnets` will make this conversion redundant.

cmd/create/machinepool/helper.go
```diff
@@ -73,7 +73,8 @@ func getSubnetOptions(r *rosa.Runtime, cluster *cmv1.Cluster) ([]string, error)
 	// Format subnet options
 	var subnetOptions []string
 	for _, subnet := range privateSubnets {
-		subnetOptions = append(subnetOptions, aws.SetSubnetOption(subnet))
+		s := subnet
+		subnetOptions = append(subnetOptions, aws.SetSubnetOption(&s))
```
oriAdler, line 77: Can we please change the functions to use subnets instead of pointers? 
As AWS SDK previously returned a slice of pointers to subnets and now returns a slice of subnets, they changed the output of `DescribeSubnets`.

cmd/create/cluster/cmd.go
den-rgb: When changing it to a slice it still requires it to use `ToString`

PR #1565, opened 2023-10-30
OCM-1443 | fix: disable node drain grace period for hosted clusters

philipwu08:
Jira: https://issues.redhat.com/browse/OCM-1443

- report error and exit if node-drain-grace-period flag is specifed for hosted cluster upgrade.
- updating node drain grade period in cluster should only be performed for classic clusters


andreadecorte:
lgtm

PR #1566, opened 2023-10-30
OCM-4506 | fix: adjust min format so it allows nightlies

gdbranco:
Related issue: [OCM-4506](https://issues.redhat.com//browse/OCM-4506)

### DIFF
--- a/cmd/create/machinepool/machinepool.go
+++ b/cmd/create/machinepool/machinepool.go
@@ -70,8 +70,13 @@ func addMachinePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster
 			os.Exit(1)
 		}
 		if !isVersionCompatibleComputeSgIds {
+			formattedVersion, err := versions.FormatMajorMinorPatch(ocm.MinVersionForAdditionalComputeSecurityGroupIdsDay2)
+			if err != nil {
+				r.Reporter.Errorf(versions.MajorMinorPatchFormattedErrorOutput, err)
+				os.Exit(1)
+			}
 			r.Reporter.Errorf("Parameter '%s' is not supported prior to version '%s'",
-				securityGroupIdsFlag, ocm.MinVersionForAdditionalComputeSecurityGroupIdsDay2)
+				securityGroupIdsFlag, formattedVersion)
 			os.Exit(1)
 		}
 	}
--- a/pkg/helper/versions/helpers.go
+++ b/pkg/helper/versions/helpers.go
@@ -12,7 +12,8 @@ import (
 )
 
 const (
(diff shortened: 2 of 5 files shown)


pkg/helper/versions/helpers_test.go
```diff
@@ -72,7 +71,18 @@ var _ = Describe("Version Helpers", Ordered, func() {
 				"4.12.0-0.a",
 			),
 		)
-
 	})
+})
 
+var _ = Describe("Validates Format Major Minor Patch", func() {
+	DescribeTable("",
```
hunterkepley, line 78: I think even though it's the only block in this `Describe` block, we should have a title/description for the table block

pkg/helper/versions/helpers.go
```diff
@@ -114,3 +114,23 @@ func IsGreaterThanOrEqual(version1, version2 string) (bool, error) {
...
+func FormatMajorMinorPatch(version string) (string, error) {
+	major, minor, patch, err := getVersionSegments(version)
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("%d.%d.%d", major, minor, patch), nil
+}
+
+func getVersionSegments(rawVersionID string) (major, minor, patch int, err error) {
+	v, err := ver.NewVersion(rawVersionID)
```
hunterkepley, line 127: NP: I think elad has brought up before that we should be verbose when possible with variable names -- not sure if it matters here as this is a simple helper function, but going through other helper functions in our codebase it seems we generally use full names rather than single letters

cmd/create/machinepool/machinepool.go
```diff
@@ -70,8 +70,13 @@ func addMachinePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster
 			os.Exit(1)
 		}
 		if !isVersionCompatibleComputeSgIds {
+			formattedVersion, err := versions.FormatMajorMinorPatch(ocm.MinVersionForAdditionalComputeSecurityGroupIdsDay2)
+			if err != nil {
+				r.Reporter.Errorf("An error occurred formatting the version for output: %v", err)
+				os.Exit(1)
+			}
```
hunterkepley, line 77: Would it be wise to move these into a function as this code is used 3 times? I know its basically just moving a function call + error check into a function, but at the very least we should make the error string a constant/variable

gdbranco: I agree with the constant, but having a wrapper around the error would mean returning an error as well and doing something similar as I don't think the os.Exit should happen within the function. Unfortunately it is more related to how golang currently deal with errors I think

PR #1570, opened 2023-10-31
Warn users if not using latest release.

AlexVulaj:
[OCM-4624](https://issues.redhat.com//browse/OCM-4624) | feat: Show a warning when a user attempts to run a command on a version other than the latest version from mirror.

Sample run from a local build:
```bash
$ rosa create cluster
E: The current version (1.2.27) is not up to date with latest released version (1.2.28).
E: It is recommended that you update to the latest version.
...
```


cmd/rosa/main.go
```diff
@@ -105,3 +113,27 @@ func main() {
...
 	}
 }
+
+func versionCheck() {
+	rprtr := reporter.CreateReporterOrExit()
+
+	currVersion, err := gversion.NewVersion(info.Version)
+	if err != nil {
+		_ = rprtr.Errorf("There was a problem verifying the current version of rosa. Please be aware that you are possibly running an outdated version: %w", err)
+		os.Exit(1)
```
AlexVulaj, line 123: Some other thoughts I had here - instead of `os.Exit(1)` here, would we rather either of the following?

a.) Immediately prompt to continue and skip the mirror check
b.) Continue without any further prompts

cmd/rosa/main.go
```diff
@@ -105,3 +113,27 @@ func main() {
...
+	}
+
+	latestVersionFromMirror, err := rosa.RetrieveLatestVersionFromMirror()
+	if err != nil {
+		_ = rprtr.Errorf("There was a problem retrieving latest version from mirror. Please be aware that you are possibly running an outdated version: %w", err)
+		os.Exit(1)
```
AlexVulaj, line 129: Similar to the above comment - would we rather either of the following?

a.) Prompt to continue and skip the version comparison
b.) Continue without any further prompts.

cmd/rosa/main.go
```diff
@@ -105,3 +113,27 @@ func main() {
...
+	}
+
+	if !currVersion.Equal(latestVersionFromMirror) {
```
AlexVulaj, line 132: I went with an equality comparison here, but I could also see a use case for _only_ checking `currVersion.LessThan` like in /cmd/verify/rosa/cmd.go#L61 . One pain point I could see is that comparing only for equality could lead to a prompt every time someone runs a command from a development build.

In osdctl, we solved this problem by adding a `-S` flag to skip the version check altogether - I could go either way. Let me know if you'd like me to handle this differently.

cmd/rosa/main.go
ciaranRoche: I would lean towards option `b` as if there is an error with the check I dont think we would want to block. I would push to just warn the user that we could not verify and there is the risk they have an outdated version. To manually check a user can run `rosa verify rosa` or `rosa version`

cmd/rosa/main.go
ciaranRoche: I think I would lean toward `LessThan` to avoid noise from dev builds. As this will be the case for roughly 2 weeks every sprint during the release process.

AlexVulaj:
/hold

Waiting for input from UX

cmd/rosa/main.go
```diff
@@ -105,3 +113,32 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func versionCheck() {
+	rprtr := reporter.CreateReporterOrExit()
+
+	currVersion, err := gversion.NewVersion(info.Version)
+	if err != nil {
+		_ = rprtr.Errorf("There was a problem verifying the current version of rosa.")
```
wolfeallison, line 122: I'd recommend changing this to "Could not verify the current version of ROSA." or "Could not check ROSA version." making sure ROSA is capitalized

cmd/rosa/main.go
```diff
@@ -105,3 +113,32 @@ func main() {
...
+		_ = rprtr.Errorf("Please be aware that you are possibly running an outdated version: %v", err)
```
wolfeallison, line 123: We avoid using "Please" because it can be mistranslated. I would change this to "You might be running on an outdated version. Make sure you are using the current version of ROSA."

andreadecorte:
should we have some local caching first? This is doing a call for version check at every single rosa command which may add a significant traffic on the mirror, but versions don't change every second

thomasmckay:
> should we have some local caching first? This is doing a call for version check at every single rosa command which may add a significant traffic on the mirror, but versions don't change every second

After team discussion, we'd prefer to not cache until it's a known problem (we believe that is unlikely). In addition, it would be interesting/beneficial to track which commands are being run.

AlexVulaj:
@ciaranRoche There's still a prow build failing that looks strange to me, any thoughts on what might be going on?

cmd/rosa/main.go
```diff
@@ -105,3 +112,29 @@ func main() {
...
+	if err != nil {
+		_ = rprtr.Errorf("Could not verify the current version of ROSA.")
+		_ = rprtr.Errorf("You might be running on an outdated version. Make sure you are using the current version of ROSA.")
+		return
+	}
+
+	latestVersionFromMirror, err := rosa.RetrieveLatestVersionFromMirror()
+	if err != nil {
+		_ = rprtr.Errorf("There was a problem retrieving the latest version of ROSA.")
+		_ = rprtr.Errorf("You might be running on an outdated version. Make sure you are using the current version of ROSA.")
```
cben, line 129: Consider making the above 2 cases a warning rather than error.  You don't know there is a problem.  And if for some reason user is stuck in that situation persistently (e.g. mirror outage), they'll be seeing errors on _every_ command.

cben:
> After team discussion, we'd prefer to not cache until it's a known problem (we believe that is unlikely).

For me, `rosa verify rosa-client` alone takes ~0.9sec.
(`ping mirror.openshift.com` is 65~75ms RTT for me, so does it take nearly a second for the HTTP server to send the directory listing?)

So with this change, every command would become slightly, but perceptibly slower.
I'd not love that even for interactive use.
It can get worse, both for speed and traffic, when scripted in a loop... :arrows_clockwise: 

- Worth confirming with mirror.openshift.com admins they don't mind getting all that traffic?

- This header is interesting (using #1640, 100% reproducible)
  ```
  time=2023-12-05T20:30:43+02:00 level=debug msg=Response header 'X-Cache' is 'Miss from cloudfront'
  ```
  In theory, a directory listing could be much faster if cached at CDN level.  Is it un-cachable deliberately (invalidation is hard) or just oversight?

cmd/rosa/main.go
pvasant: I dont think we should error for all scenarios without understanding what is the type of the error otherwise all rosa commands will start to fail.

robpblake:
See #1777

cmd/rosa/main.go
```diff
@@ -105,3 +112,30 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func versionCheck() {
```
robpblake, line 116: @AlexVulaj to @cben point, I don't think we want this on `rosa version` or `rosa verify rosa`? It's gonna print double the warnings for commands that already include a version check in their output.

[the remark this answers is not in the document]
AlexVulaj: Can you elaborate a bit more here? It seems odd to me that we'd change functionality of the CLI based on the output format the user requests.

pkg/version/version.go
```diff
@@ -62,8 +63,8 @@ type rosaVersion struct {
 	retriever Retriever
 }
 
-func (v rosaVersion) IsLatest(latestVersion string) (*goVer.Version, bool, error) {
-	currentVersion, err := goVer.NewVersion(latestVersion)
+func (v rosaVersion) IsLatest(version string) (*goVer.Version, bool, error) {
```
AlexVulaj, line 66: This is an opinionated change more than a functional one - the variable name `latestVersion` confused me for a bit here since it's not actually the latest version, it's what we're checking to _see_ if is the latest version.

cmd/rosa/main.go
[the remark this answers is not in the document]
AlexVulaj: Changes made after our conversation in Slack.

AlexVulaj: I've updated our code to ignore version checks on specific commands.

ciaranRoche (commented):
LGTM

Thanks for you patience on this one Alex, glad to have this check in place when it lands.

@robpblake can you give this a final ack before we merge :pray:

robpblake (approved):
/lgtm

PR #1577, opened 2023-11-03
OCM-4315 | 'go mod vendor' to fix build

thomasmckay:


ciaranRoche:
@thomasmckay can we update the commit message to a correct JIRA, or even the release JIRA :pray:

Issue #1578, opened 2023-11-03
Index out of range error while creating a cluster


chentex:
We have automated test running that are testing performance on rosa, we have the following command to create the cluster(redacted values):

`rosa create cluster -y --sts --mode auto --cluster-name ci-rosa-s-3893 --region us-west-2 --version 4.15.0-0.nightly-2023-11-03-082930 --channel-group nightly --compute-machine-type m5.xlarge --tags prowci:ci-rosa-s-3893 --role-arn XXXX --support-role-arn XXXX --worker-iam-role XXXX --controlplane-iam-role XXXX --ec2-metadata-http-tokens optional --multi-az --enable-autoscaling --min-replicas 3 --max-replicas 24  --etcd-encryption `

The command fails after 14 seconds with this error

```
panic: runtime error: index out of range [0] with length 0
goroutine 1 [running]:
github.com/openshift/rosa/cmd/create/cluster.run(0x33e8b80?, {0x1dfc449?, 0x21?, 0x24?})
	/go/src/github.com/openshift/rosa/cmd/create/cluster/cmd.go:2417 +0xdfd0
github.com/spf13/cobra.(*Command).execute(0x33e8b80, {0xc00092f200, 0x21, 0x24})
	/go/src/github.com/openshift/rosa/vendor/github.com/spf13/cobra/command.go:944 +0x847
github.com/spf13/cobra.(*Command).ExecuteC(0x33fbc40)
	/go/src/github.com/openshift/rosa/vendor/github.com/spf13/cobra/command.go:1068 +0x3bd
github.com/spf13/cobra.(*Command).Execute(...)
	/go/src/github.com/openshift/rosa/vendor/github.com/spf13/cobra/command.go:992
main.main()
	/go/src/github.com/openshift/rosa/cmd/rosa/main.go:100 +0x8b
````

The lines of code point to the `subnetIDs` object.

https://github.com/openshift/rosa/blob/5997e5a4819d00e40435fbdf0d08b989490f61b5/cmd/create/cluster/cmd.go#L2416-L2417

PR #1579, opened 2023-11-03
OCM-4602 | fix: adjust subnet ids passed

gdbranco:
https://github.com/openshift/rosa/issues/1578


gdbranco:
Merging without approval as it is a simple fix, plus not a lot of devs currently online anymore

PR #1581, opened 2023-11-05
OCM-4622 | fix: Create HCP account roles with unmanaged policies

oriAdler:
If the users provides the `managed-policies` flag set to false, alongside the `hosted-cp` flag, exit with the provided error message.

```
oadler@fedora:rosa (OCM-4632)$ rosa create account-roles --hosted-cp --managed-policies=false
I: Logged in as 'oadler.openshift' on 'http://localhost:9000'
E: Setting `hosted-cp` as unmanaged policies is not supported
```


thomasmckay:
lgtm

PR #1585, opened 2023-11-06
OCM-3854 | feat: Added `rosa create kubeletconfig` sub-command

robpblake:
This PR adds support for the `rosa create kubeletconfig` sub-command.

### DIFF
--- a/cmd/create/kubeletconfig/cmd.go
+++ b/cmd/create/kubeletconfig/cmd.go
@@ -1,0 +1,147 @@
+/*
+Copyright (c) 2023 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


cmd/create/kubeletconfig/cmd.go
```diff
@@ -0,0 +1,147 @@
...
+
+	if args.podPidsLimit == unassigned && !interactive.Enabled() {
+		interactive.Enable()
+		r.Reporter.Infof("Enabling interactive mode")
+	}
+
+	if interactive.Enabled() {
+		args.podPidsLimit, err = interactive.GetInt(interactive.Input{
+			Question: "Pod Pids Limit?",
+			Help:     "Set the Pod Pids Limit field to a value between 4096 and 16,384",
```
ciaranRoche, line 102: Not a blocker for this, but I wonder can we link to documentation on the limits :thinking: my thinking here is if the limits do change we will require a ROSA release to update the help :thinking: 

Or I wonder can we provide an API to get help messages from the backend WDYT?

pkg/kubeletconfig/consts.go
```diff
@@ -0,0 +1,5 @@
+package kubeletconfig
+
+const (
+	MinPodPidsLimit = 4096
```
ciaranRoche, line 4: Same as the comment above, I wonder can we provide a way to fetch the limits from the API :thinking:

ciaranRoche (approved):
Comments in line are not a blocker. Lets get this through QE

PR #1590, opened 2023-11-08
OCM-4714 | include --properties in create cluster output

thomasmckay:


cmd/create/cluster/cmd.go
```diff
@@ -3555,6 +3556,9 @@ func buildCommand(spec ocm.Spec, operatorRolesPrefix string,
 			securitygroups.SgKindFlagMap["Control Plane"], strings.Join(spec.AdditionalControlPlaneSecurityGroupIds, ","))
 	}
 
+	if len(properties) > 0 {
+		command += fmt.Sprintf(" --properties %s", strings.Join(properties, ","))
```
andreadecorte, line 3560: shouldn't this be multiple --properties invocations? I am not sure it works this way

thomasmckay: you are correct! let me fix that

PR #1592, opened 2023-11-08
OCM-4675 | feat: show dynamically technology preview messages for hcp

lucasponce:
This PR shows the technology preview messages for hcp dynamically.

Two open questions:
- I don't find a good solution for the "Tech preview" messages in the init(), so, I guess the simple solution is to remove them as it is expected this feature will be part of the 1.2.30 release.
- I think we have to decide if the technology preview should be "explicit" meaning that if there is no entry, it can be considered GA (that would simplify the logic).

cc @andreadecorte

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -652,7 +652,7 @@ func init() {
 		&args.hostedClusterEnabled,
 		"hosted-cp",
 		false,
-		"Technology Preview: Enable the use of Hosted Control Planes",
+		"Enable the use of Hosted Control Planes",
 	)
 
 	flags.StringVar(&args.machinePoolRootDiskSize,
@@ -912,11 +912,15 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
-	// FIXME: Remove before GA
 	if isHostedCP && r.Reporter.IsTerminal() {
-		//nolint
-		r.Reporter.Infof("NOTE: Hosted control planes are currently in Technology Preview (https://access.redhat.com/support/offerings/techpreview)." +
-			" Any Technology Preview clusters will need to be destroyed and recreated prior to general availability.")
+		techPreviewMsg, err := r.OCMClient.GetTechnologyPreviewMessage(ocm.HcpProduct, time.Now())
+		if err != nil {
+			r.Reporter.Errorf("%s", err)
+			os.Exit(1)
+		}
+		if techPreviewMsg != "" {
+			r.Reporter.Infof(techPreviewMsg)
+		}
 	}
 
 	if isHostedCP && args.ec2MetadataHttpTokens != "" {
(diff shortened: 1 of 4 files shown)


pkg/ocm/products.go
```diff
@@ -0,0 +1,20 @@
...
+	response, err := c.ocm.ClustersMgmt().V1().
+		Products().Product(RosaProductId).
+		TechnologyPreviews().TechnologyPreview(id).
+		Get().
+		Send()
+	if response.Status() == 404 {
+		return nil, false, nil
+	}
+	if err != nil {
+		return nil, false, handleErr(response.Error(), err)
```
andreadecorte, line 17: this doesn't return the error to the user, I think we should still check in the caller since it may be something else

andreadecorte:
1. looks good to me
2. I think the only constraint would be to get the API populated before next week, it is reasonable
3. can we add a couple of tests of this behavior?

lucasponce:
Updated, please @andreadecorte, could you give it another pass? thanks

pkg/ocm/products_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+		message, err := ocmClient.GetTechnologyPreviewMessage("hcp", afterRelease)
+		Expect(err).ToNot(HaveOccurred())
+
+		Expect(message).To(BeEmpty())
+	})
+
+	It("Expects no message if no technology preview", func() {
+		apiServer.AppendHandlers(
+			RespondWithJSON(
```
andreadecorte, line 102: thanks, I would just add another test with a 400 to test error handling, otherwise looks good!

andreadecorte:
lgtm

cmd/create/cluster/cmd.go
```diff
@@ -912,11 +912,15 @@ func run(cmd *cobra.Command, _ []string) {
...
-		r.Reporter.Infof("NOTE: Hosted control planes are currently in Technology Preview (https://access.redhat.com/support/offerings/techpreview)." +
-			" Any Technology Preview clusters will need to be destroyed and recreated prior to general availability.")
+		techPreviewMsg, err := r.OCMClient.GetTechnologyPreviewMessage(ocm.HcpProduct, time.Now())
+		if err != nil {
+			r.Reporter.Errorf("%s", err)
+			os.Exit(1)
+		}
+		if techPreviewMsg != "" {
+			r.Reporter.Infof(techPreviewMsg)
+		}
```
chenz4027, line 923: Just wondering, what if the techPreviewMessage = ""? I guess we just skip that scenario?

lucasponce: Yes, if the message is empty, there is no print, then the ROSA CLI output won't have that extra line.

ciaranRoche (approved):
/lgtm

PR #1593, opened 2023-11-09
OCM-4771 | fix: Added -y flag to rosa edit kubeletconfig command

robpblake:
This PR adds the `-y` flag to `rosa edit` commands (in particular the `rosa edit kubeletconfig` command) which allows users to automatically accept prompts rather than having to manually acknowledge them.


ciaranRoche (approved):
/lgtm

PR #1596, opened 2023-11-10
OCM-4784 | fix: Ensure we validate maximum pids limit when creating/editing KubeletConfig

robpblake:
This PR introduces validation on the maximum value for `--pod-pids-limit` on the `rosa create kubeletconfig` and `rosa edit kubeletconfig` commands.

The maximum value for `--pod-pids-limit` varies depending on whether or not the organization has the capability to exceed the default limits. We therefore check for the capability in this PR and adjust validation limits (and help messages) accordingly.

Prior to this we relied on the backend for validation, but that would allow the user to do things like overflow the maximum value for `int32` which didn't give a very good error message. This is a good UX improvement to provide user with context on the maximum they are allowed to use and also ensure that they receive clear validation mesaages.


cmd/create/kubeletconfig/cmd.go
```diff
@@ -93,19 +93,15 @@ func run(_ *cobra.Command, _ []string) {
 		r.Reporter.Infof("Enabling interactive mode")
 	}
 
+	maxPidsLimit, err := GetMaxPidsLimit(r.OCMClient)
```
den-rgb, line 96: Could It be beneficial to put the changes in this file into a helper file considering both create and edit are using the same exact code?

pkg/kubeletconfig/config.go
```diff
@@ -0,0 +1,53 @@
...
+//go:generate mockgen -source=config.go -package=kubeletconfig -destination=mock_capability_checker.go
+type CapabilityChecker interface {
+	IsCapabilityEnabled(capability string) (bool, error)
+}
+
+// GetMaxPidsLimit - returns the maximum pids limit for the current organization
+// the maximum is varied depending on whether the current organizaton has
+// the capability.organization.bypass_pids_limit capability
+func GetMaxPidsLimit(client CapabilityChecker) (int, error) {
+	enabled, err := client.IsCapabilityEnabled(ByPassPidsLimitCapability)
```
gdbranco, line 20: We usually leave the capability check to CS. I do see some UX improvements having it under CLI for this scenario. Just not sure... I also see there's a different check for capability elsewhere. Wouldn't block the PR due to this

robpblake: @gdbranco I think ideally we'd have an API on CS that allows the CLI to retrieve the min/max for `--pod-pids-limit` (and allowable values for other fields from `KubeletConfig` as we expose them), but that's beyond the scope of this ticket and XCMSTRAT-110 tbh. 

My feeling this is probably a reasonable compromise until we have that API because I can see feedback from customers that are allowed to exceed the limits such as "I thought we could go beyond 16,384..why does the CLI tell  me we can't?"

WDYT?

cheers,

Rob

cmd/create/kubeletconfig/cmd.go
robpblake: @den-rgb Let me see what I can do 👍

pkg/kubeletconfig/config.go
robpblake: @gdbranco And for clarity: CS checks the capability when validating the submitted ranges. This is really for customer UX when using the CLI in terms of validation and error messaging.

cmd/edit/kubeletconfig/cmd.go
```diff
@@ -97,19 +97,15 @@ func run(_ *cobra.Command, _ []string) {
...
-			Options:  nil,
-			Default:  kubeletconfig.PodPidsLimit(),
-			Required: true,
-			Validators: []interactive.Validator{
-				// We only validate the minimum as some customers have capability to exceed maximum
-				// We rely on backend validation for that
-				interactive.Min(MinPodPidsLimit),
-			},
-		})
+		args.podPidsLimit, err = interactive.GetInt(GetInterativeInput(maxPidsLimit, kubeletconfig))
```
hunterkepley, line 108: GetIntera`c`tiveInput

cmd/edit/kubeletconfig/cmd.go
```diff
@@ -124,6 +120,12 @@ func run(_ *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	if args.podPidsLimit > maxPidsLimit {
+		r.Reporter.Errorf("The maximum value for --pod-pids-limit is '%d'. You have supplied '%d'",
+			maxPidsLimit, args.podPidsLimit)
+		os.Exit(1)
+	}
```
hunterkepley, line 127: Will this error/return ever be reached? The interactive prompt you made in [this file](https://github.com/openshift/rosa/blob/b1410776cbd63618511efbab573dd7bdd079264a/pkg/kubeletconfig/config.go) has validations for this same scenario

cmd/edit/kubeletconfig/cmd.go
```diff
@@ -97,19 +97,15 @@ func run(_ *cobra.Command, _ []string) {
 		r.Reporter.Infof("Enabling interactive mode")
 	}
 
+	maxPidsLimit, err := GetMaxPidsLimit(r.OCMClient)
+	if err != nil {
+		r.Reporter.Errorf("Failed to check maximum allowed Pids limit for for cluster '%s'",
```
hunterkepley, line 102: NP: `Pid limit`

cmd/create/kubeletconfig/cmd.go
```diff
@@ -121,6 +117,12 @@ func run(_ *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	if args.podPidsLimit > maxPidsLimit {
+		r.Reporter.Errorf("The maximum value for --pod-pids-limit is '%d'. You have supplied '%d'",
+			maxPidsLimit, args.podPidsLimit)
+		os.Exit(1)
+	}
```
hunterkepley, line 124: Does this block get reached? There is validation for this scenario in `GetInterativeInput`

pkg/kubeletconfig/config.go
```diff
@@ -0,0 +1,53 @@
...
+		return MaxUnsafePodPidsLimit, nil
+	}
+	return MaxPodPidsLimit, nil
+}
+
+func GetInteractiveMaxPidsLimitHelp(maxPidsLimit int) string {
+	return fmt.Sprintf(InteractivePodPidsLimitHelp, maxPidsLimit)
+}
+
+func GetInterativeInput(maxPidsLimit int, kubeletConfig *v1.KubeletConfig) interactive.Input {
```
hunterkepley, line 35: GetIntera`c`tiveInput

cmd/edit/kubeletconfig/cmd.go
hunterkepley: Or `Limit of Pids` would work :smile:

pkg/kubeletconfig/consts.go
```diff
@@ -1,12 +1,13 @@
...
+	MinPodPidsLimit                = 4096
+	MaxPodPidsLimit                = 16384
+	MaxUnsafePodPidsLimit          = 3694303
+	PodPidsLimitOption             = "pod-pids-limit"
+	PodPidsLimitOptionUsage        = "Sets the requested pod_pids_limit for your custom KubeletConfig."
 	PodPidsLimitOptionDefaultValue = -1
-
-	InteractivePodPidsLimitPrompt = "Pod Pids Limit?"
-	InteractivePodPidsLimitHelp   = "Set the Pod Pids Limit field to a value between 4096 and 16,384"
+	InteractivePodPidsLimitPrompt  = "Pod Pids Limit?"
```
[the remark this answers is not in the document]
hunterkepley: If plural is wanted, `--limit-of-pod-pids`

cmd/edit/kubeletconfig/cmd.go
hunterkepley: This is obviously a nitpick, but my reasoning is that `allowed pids` is plural, but `allowed pid limit` should be singular. Much like your `mile limit` before your next oil change is not said `miles limit` -- or `car limit` in a parking garage vs `cars limit`

pkg/kubeletconfig/config.go
ciaranRoche: I think its okay to use capability checks in the CLI, its definitely feature flags we want to avoid in the CLI as the life span of a capability is generally life long, where as a feature flag is meant to be temp

I agree, we need  to move away and allow config be pulled from the API.

pkg/kubeletconfig/consts.go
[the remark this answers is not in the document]
robpblake: @hunterkepley The field in [KubeletConfig](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#kubelet-config-k8s-io-v1beta1-KubeletConfiguration) is actually called `podPidsLimit`  (notice the extra 's' on the PID), so whilst I pretty much agree with you, I think it's best for us to be consistent with upstream to remove any doubt in users minds exactly what this field does.

cheers,
Rob

cmd/edit/kubeletconfig/cmd.go
robpblake: Eyes of a 🥷  Will get it fixed.

cheers,
Rob

cmd/edit/kubeletconfig/cmd.go
robpblake: @hunterkepley See my previous comment about the upstream field already being named `podPidsLimit`

cmd/edit/kubeletconfig/cmd.go
robpblake: @hunterkepley Yes, it's reached if I don't use interactive mode and simply pass the `--pod-pids-limit` flag e.g `rosa create/edit kubeletconfig -c rblake-test --pod-pids-limit=5000`

cmd/create/kubeletconfig/cmd.go
robpblake: Yes. See my previous comment.

pkg/interactive/validation.go
```diff
@@ -124,6 +124,25 @@ func IsCIDR(val interface{}) error {
 	return fmt.Errorf("can only validate strings, got %v", val)
 }
 
+func Max(max int) Validator {
+	return func(ans interface{}) error {
+		if i, ok := ans.(string); ok {
```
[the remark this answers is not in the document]
robpblake: Sure 👍

pkg/interactive/validation.go
```diff
@@ -124,6 +124,25 @@ func IsCIDR(val interface{}) error {
 	return fmt.Errorf("can only validate strings, got %v", val)
 }
 
+func Max(max int) Validator {
```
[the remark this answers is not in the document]
robpblake: Thanks. Renamed the function to be a bit more descriptive of the validator returned e.g `MaxValue`.

cmd/create/kubeletconfig/cmd.go
robpblake: @den-rgb Refactored as per your suggestion into the `ValidateOrPromptForRequestedPidsLimit` function. Looks much cleaner, so thanks for that!

cheers,
Rob

den-rgb:
lgtm

ciaranRoche (approved):
/lgtm

PR #1597, opened 2023-11-10
OCM-4792 | fix: do not ask billingAccount for hcp before GA

lucasponce:
This is a hotfix for the ROSA CLI version.

Before the HCP GA, the "--billing-account" should be optional and not requested.

Note that this logic is not removed from master branch, as the 1.2.31 version should be available for HCP GA.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -659,6 +659,7 @@ func init() {
 		"",
 		"Account used for billing subscriptions purchased via the AWS marketplace",
 	)
+	flags.MarkHidden("billing-account")
 
 	flags.BoolVar(
 		&args.createAdminUser,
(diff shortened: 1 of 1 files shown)


andreadecorte:
so we just discard it? Do we want to inform user of it or we just do it silently?
Otherwise looks good

cristianoveiga:
LGTM

lucasponce:
> so we just discard it? Do we want to inform user of it or we just do it silently? Otherwise looks good

Inform the user that is required?

andreadecorte:
> > so we just discard it? Do we want to inform user of it or we just do it silently? Otherwise looks good
> 
> Inform the user that is required?

that it will be required in the next versions but as of today it does nothing? It is a new flag, so maybe nobody uses it yet, still silently ignoring is weird to me. But not a blocker

andreadecorte:
discussed directly, this removes the interactive + mark hidden, so it makes sense.
/lgtm

lucasponce:
> that it will be required in the next versions but as of today it does nothing? 

Ah, thanks, I get the context now.

Well, that will be part of how the BU wants to communicate this new feature, too.

PR #1599, opened 2023-11-13
OCM-4801 | feat: Link the enforce of billing account hcp for GA

lucasponce:
Discussed linking the enforcement of the request of the billing account flow for HCP once that GA is released.

Use the new technology preview API for coordination and flexibility.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -1002,83 +1002,93 @@ func run(cmd *cobra.Command, _ []string) {
 		outputClusterAdminDetails(r, isClusterAdmin, clusterAdminPassword)
 	}
 
+	if isHostedCP && cmd.Flags().Changed("default-mp-labels") {
+		r.Reporter.Errorf("Setting the worker machine pool labels is not supported for hosted clusters")
+		os.Exit(1)
+	}
+
 	// Billing Account
 	billingAccount := args.billingAccount
 	if isHostedCP {
-		if billingAccount != "" && !ocm.IsValidAWSAccount(billingAccount) {
-			r.Reporter.Errorf("Billing account is invalid. Run the command again with a valid billing account.")
-			os.Exit(1)
-		}
-
-		cloudAccounts, err := r.OCMClient.GetBillingAccounts()
+		isHcpBillingTechPreview, err := r.OCMClient.IsTechnologyPreview(ocm.HcpBillingAccount, time.Now())
 		if err != nil {
 			r.Reporter.Errorf("%s", err)
 			os.Exit(1)
 		}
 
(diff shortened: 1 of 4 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -1002,83 +1002,93 @@ func run(cmd *cobra.Command, _ []string) {
 		outputClusterAdminDetails(r, isClusterAdmin, clusterAdminPassword)
 	}
 
+	if isHostedCP && cmd.Flags().Changed("default-mp-labels") {
```
ciaranRoche, line 1005: Is this change releated to OCM-4801 or do you have a different card for this?

pkg/ocm/products.go
```diff
@@ -23,12 +23,23 @@ func (c *Client) GetTechnologyPreview(id string) (*cmv1.ProductTechnologyPreview
...
+	return false, nil
+}
+
 func (c *Client) GetTechnologyPreviewMessage(id string, forTime time.Time) (string, error) {
 	techPreview, exists, err := c.GetTechnologyPreview(id)
 	if err != nil {
 		return "", err
 	}
-	// It assumes that the technology_preview endpoint will be available in CS before the ROSA CLI 1.2.30 is out
+	// It assumes that the technology_preview endpoint will be available in CS before the ROSA CLI HCP GA
```
andreadecorte, line 42: let's remove this line, endpoint is in production already

ciaranRoche (commented):
Do we want to handle the case if its tech preview and someone passes --billing-account From what i can see here we will just silently ignore it :thinking: should we warn the user if they pass the param

cmd/create/cluster/cmd.go
lucasponce: It's a minor refactor, this logic was inside the billing block.

As it's unrelated I've extracted for readability.

(No new changes already)

lucasponce:
> Do we want to handle the case if its tech preview and someone passes --billing-account From what i can see here we will just silently ignore it 🤔 should we warn the user if they pass the param

For this case there is a backend capability per org:

```
capability.organization.enforce_rosa_hcp_billing
```

So, that capability enforces if the billing account should be validated (at backend place).

cmd/create/cluster/cmd.go
lucasponce: https://github.com/openshift/rosa/pull/1599/files#diff-943c0ab57b82c7028652df55e1d49e4f4e4dc41fa094dc4a9b3f2c35887d44b5L1073

andreadecorte:
lgtm

ciaranRoche (approved):
/lgtm

PR #1600, opened 2023-11-13
OCM-2453 | fix: FedRAMP create accountroles fix

hunterkepley:
Issue was that the default flow for creating account roles became `doubleRolesCreator` -- meaning it created both unmanaged & managed HCP roles. For govcloud, we just want unmanaged, not both. This MR aims to accomplish this.

### DIFF
--- a/cmd/create/accountroles/creators.go
+++ b/cmd/create/accountroles/creators.go
@@ -20,6 +20,11 @@ type creator interface {
 
 func initCreator(r *rosa.Runtime, managedPolicies bool, classic bool, hostedCP bool, isClassicValueSet bool,
 	isHostedCPValueSet bool) (creator, bool) {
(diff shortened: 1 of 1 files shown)


cmd/create/accountroles/creators.go
```diff
@@ -38,7 +38,7 @@ func initCreator(r *rosa.Runtime, managedPolicies bool, classic bool, hostedCP b
 	}
 
 	// Classic ROSA unmanaged policies
-	if classic {
+	if classic || r.Creator.IsGovcloud {
```
gdbranco, line 41: Is this check needed? Can we not force classic to be true if we are on gov cloud instead?

oriAdler:
@hunterkepley are managed policies supported for Gov Cloud?

PR #1606, opened 2023-11-15
OCM-3875 | fix: only output follow-up commands if necessary

thomasmckay:


cmd/create/cluster/cmd.go
```diff
@@ -3093,20 +3093,20 @@ func run(cmd *cobra.Command, _ []string) {
...
-			if permissionsBoundary != "" {
-				rolesCMD = fmt.Sprintf("%s --permissions-boundary %s", rolesCMD, permissionsBoundary)
-			}
-			oidcCMD := "rosa create oidc-provider"
-			oidcCMD = fmt.Sprintf("%s --cluster %s", oidcCMD, clusterName)
-			output += "\nRun the following commands to continue the cluster creation:\n\n"
-			output = fmt.Sprintf("%s\t%s\n", output, rolesCMD)
-			output = fmt.Sprintf("%s\t%s\n", output, oidcCMD)
-			r.Reporter.Infof(output)
+			if !cluster.AWS().STS().OidcConfig().Reusable() {
```
gdbranco, line 3096: It is not guaranteed that the user has created the resources prior when using a reusable oidc configuration

cmd/create/cluster/cmd.go
```diff
@@ -3093,20 +3093,20 @@ func run(cmd *cobra.Command, _ []string) {
...
+				rolesCMD := fmt.Sprintf("rosa create operator-roles --cluster %s", clusterName)
+				if permissionsBoundary != "" {
+					rolesCMD = fmt.Sprintf("%s --permissions-boundary %s", rolesCMD, permissionsBoundary)
+				}
+				output += "\nRun the following commands to continue the cluster creation:\n\n"
+				output = fmt.Sprintf("%s\t%s\n", output, rolesCMD)
+				if oidcConfig == nil {
```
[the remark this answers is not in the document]
thomasmckay: When you can, take a look at the operator-roles. I'll do similar for oidc-config

cmd/create/cluster/cmd.go
```diff
@@ -3079,34 +3101,39 @@ func run(cmd *cobra.Command, _ []string) {
...
 	if isSTS {
 		if mode != "" {
-			if !output.HasFlag() || r.Reporter.IsTerminal() {
-				r.Reporter.Infof("Preparing to create operator roles.")
-			}
-			err := operatorroles.Cmd.RunE(operatorroles.Cmd, []string{clusterName, mode, permissionsBoundary})
-			if err != nil {
-				r.Reporter.Errorf("There was a problem creating operator roles: %v", err)
-				os.Exit(1)
+			if len(operatorRoles) == 0 {
```
gdbranco, line 3104: I recall that this is handled as part of the call for creating operator roles command :thinking: . Not sure if required

cmd/create/cluster/cmd.go
```diff
@@ -3079,34 +3101,39 @@ func run(cmd *cobra.Command, _ []string) {
...
+				if permissionsBoundary != "" {
+					rolesCMD = fmt.Sprintf("%s --permissions-boundary %s", rolesCMD, permissionsBoundary)
+				}
+				output = fmt.Sprintf("%s\t%s\n", output, rolesCMD)
 			}
-			rolesCMD := fmt.Sprintf("rosa create operator-roles --cluster %s", clusterName)
-			if permissionsBoundary != "" {
-				rolesCMD = fmt.Sprintf("%s --permissions-boundary %s", rolesCMD, permissionsBoundary)
+			if cluster.AWS().STS().OidcConfig() == nil || !cluster.AWS().STS().OidcConfig().Reusable() {
+				oidcCMD := "rosa create oidc-provider"
```
gdbranco, line 3128: A similar treatment might be required here where you first check if the oidc provider is already present in the customer account instead of the oidc config condition

thomasmckay: How would I check if the provider is already present? I thought the call here [handleOidcConfigOptions()](https://github.com/openshift/rosa/blob/master/cmd/create/cluster/cmd.go#L1664) was checking and if provider was present (or entered) then it would be added to the cluster.

Or do you mean I should make the call that `rosa list oidc-provider` is making? If so how do I know which of the providers returned would satisfy the need not to have to run `rosa create oidc-provider`? The call takes a _very_ long time for me (48s), if that matters.

thomasmckay: This call requires the oidcConfig.IssuerUrl() but that is nil, right?

gdbranco: Shouldn't be nil after the cluster create request is through

thomasmckay: thanks!

PR #1609, opened 2023-11-16
OCM-4877 | fix: Return versions when create cluster in fedramp

hunterkepley:
Fedramp was filtering out both non-HCP and HCP versions when using stable as the channel-group

This fix only filters out non-HCP when not creating an HCP cluster

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -1161,7 +1161,7 @@ func run(cmd *cobra.Command, _ []string) {
 	// OpenShift version:
 	version := args.version
 	channelGroup := args.channelGroup
-	versionList, err := versions.GetVersionList(r, channelGroup, isSTS, isHostedCP, true, true)
+	versionList, err := versions.GetVersionList(r, channelGroup, isSTS, isHostedCP, isHostedCP, true)
 	if err != nil {
 		r.Reporter.Errorf("%s", err)
 		os.Exit(1)


pkg/helper/versions/helpers.go
```diff
@@ -34,14 +34,12 @@ func GetVersionList(r *rosa.Runtime, channelGroup string, isSTS bool, isHostedCP
 		if isSTS && !ocm.HasSTSSupport(v.RawID(), v.ChannelGroup()) {
 			continue
 		}
-		if filterHostedCP {
```
lucasponce, line 37: This block looks valid if you change the "filterHostedCP" in the caller.

I guess if you don't want versions for HCP, you wouldn't be required to check if they have hosted cp support.

cmd/create/cluster/cmd.go
```diff
@@ -1161,7 +1161,11 @@ func run(cmd *cobra.Command, _ []string) {
 	// OpenShift version:
 	version := args.version
 	channelGroup := args.channelGroup
-	versionList, err := versions.GetVersionList(r, channelGroup, isSTS, isHostedCP, true, true)
+	filterHostedCP := true
+	if !isHostedCP {
+		filterHostedCP = false
+	}
+	versionList, err := versions.GetVersionList(r, channelGroup, isSTS, isHostedCP, filterHostedCP, true)
```
andreadecorte, line 1168: `filterHostedCP` is not the same as passing `isHostedCP`?

lucasponce, line 1168: Isn't the same value isHostedCP == filterHostedCP ?

andreadecorte, line 1168: we could probably just pass isHostedCP instead of creating a new variable

ciaranRoche, line 1168: can we just pass in 
```suggestion
	versionList, err := versions.GetVersionList(r, channelGroup, isSTS, isHostedCP, isHostedCP, true)
```

pkg/helper/versions/helpers.go
```diff
@@ -34,14 +34,12 @@ func GetVersionList(r *rosa.Runtime, channelGroup string, isSTS bool, isHostedCP
...
-			valid, err := ocm.HasHostedCPSupport(v)
-			if err != nil {
-				return versionList, fmt.Errorf("failed to check HostedCP support: %v", err)
-			}
-			if !valid {
-				continue
-			}
+		valid, err := ocm.HasHostedCPSupport(v)
```
ciaranRoche, line 37: I dont think this change is needed, we dont need to call HasHostedCPSupport if we dont need to filter on hosted CP

lucasponce:
TL;DR

Shouldn't be just

```
versionList, err := versions.GetVersionList(r, channelGroup, isSTS, isHostedCP, true, true
```

to 

```
versionList, err := versions.GetVersionList(r, channelGroup, isSTS, isHostedCP, isHostedCP, true)
```

cmd/create/cluster/cmd.go
hunterkepley: I'm a little bit confused about this, do you mean using the equality of those 2 args in place of themselves in the func call?

andreadecorte:
lgtm

ciaranRoche (approved):
/lgtm

PR #1614, opened 2023-11-20
OCM-4898 | fix: create HCP operator roles by prefix - filter by policy

oriAdler:
When the user creates operator roles using prefix and provides an installer role ARN in interactive mode, display only account roles with HCP-managed policies.

```
oadler@fedora:rosa (OCM-4898)$ rosa create operator-roles --oidc-config-id 27jpdr1kb66f6fkpkql6b384ckpqs62q --prefix oadler-nov-20 --hosted-cp
? Role creation mode: auto
? Operator roles prefix: oadler-nov-20
? Create hosted control plane operator roles: Yes
W: More than one Installer role found
? Installer role ARN:  [Use arrows to move, type to filter, ? for more help]
  arn:aws:iam::765374464689:role/ming-local-HCP-ROSA-Installer-Role
  arn:aws:iam::765374464689:role/ming-staging-HCP-ROSA-Installer-Role
  arn:aws:iam::765374464689:role/new-prefix-HCP-ROSA-Installer-Role
> arn:aws:iam::765374464689:role/oadler-aug-24-HCP-ROSA-Installer-Role
  arn:aws:iam::765374464689:role/oadler-HCP-ROSA-Installer-Role
  arn:aws:iam::765374464689:role/oadler-nov-14-HCP-ROSA-Installer-Role
  arn:aws:iam::765374464689:role/oadler-nov-16-HCP-ROSA-Installer-Role
```

### DIFF
--- a/cmd/create/oidcconfig/cmd.go
+++ b/cmd/create/oidcconfig/cmd.go
@@ -215,7 +215,7 @@ func run(cmd *cobra.Command, argv []string) {
 			}
 			if mode == aws.ModeAuto && (interactive.Enabled() || (confirm.Yes() && args.installerRoleArn == "")) {
 				args.installerRoleArn = interactiveRoles.
-					GetInstallerRoleArn(r, cmd, args.installerRoleArn, MinorVersionForGetSecret)
+					GetInstallerRoleArn(r, cmd, args.installerRoleArn, MinorVersionForGetSecret, r.AWSClient.FindRoleARNs)
 			}
 			if interactive.Enabled() {
 				prefix, err := interactive.GetString(interactive.Input{
--- a/cmd/create/operatorroles/by_prefix.go
+++ b/cmd/create/operatorroles/by_prefix.go
@@ -58,7 +58,13 @@ func handleOperatorRolesPrefixOptions(r *rosa.Runtime, cmd *cobra.Command) {
 		os.Exit(1)
 	}
 	args.hostedCp = isHostedCP
-	args.installerRoleArn = interactiveRoles.GetInstallerRoleArn(r, cmd, args.installerRoleArn, "")
+	if args.hostedCp {
+		args.installerRoleArn = interactiveRoles.GetInstallerRoleArn(r, cmd, args.installerRoleArn, "",
+			r.AWSClient.FindRoleARNsHostedCp)
+	} else {
+		args.installerRoleArn = interactiveRoles.GetInstallerRoleArn(r, cmd, args.installerRoleArn, "",
+			r.AWSClient.FindRoleARNsClassic)
+	}
 }
 
 func handleOperatorRoleCreationByPrefix(r *rosa.Runtime, env string,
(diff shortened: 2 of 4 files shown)


cmd/register/oidcconfig/cmd.go
```diff
@@ -120,7 +120,7 @@ func run(cmd *cobra.Command, argv []string) {
 
 	if !cmd.Flags().Changed(InstallerRoleArnFlag) && (interactive.Enabled() || confirm.Yes()) {
 		args.installerRoleArn = interactiveRoles.
-			GetInstallerRoleArn(r, cmd, args.installerRoleArn, MinorVersionForGetSecret)
+			GetInstallerRoleArn(r, cmd, args.installerRoleArn, MinorVersionForGetSecret, r.AWSClient.FindRoleARNs)
```
oriAdler, line 123: @gdbranco Same here, it can be either classic or HCP for the purpose of the OIDC config? as the description says `STS Role ARN with get secrets permission`.

cmd/create/operatorroles/by_prefix.go
```diff
@@ -58,7 +58,13 @@ func handleOperatorRolesPrefixOptions(r *rosa.Runtime, cmd *cobra.Command) {
 		os.Exit(1)
 	}
 	args.hostedCp = isHostedCP
-	args.installerRoleArn = interactiveRoles.GetInstallerRoleArn(r, cmd, args.installerRoleArn, "")
+	if args.hostedCp {
```
den-rgb, line 61: Will this not always be true?

oriAdler: The user can set the value to `false` in interactive mode.
Alternatively, if the flag isn't provided the default value will be `false`.

den-rgb: I see thank you

cmd/create/oidcconfig/cmd.go
```diff
@@ -215,7 +215,7 @@ func run(cmd *cobra.Command, argv []string) {
 			}
 			if mode == aws.ModeAuto && (interactive.Enabled() || (confirm.Yes() && args.installerRoleArn == "")) {
 				args.installerRoleArn = interactiveRoles.
-					GetInstallerRoleArn(r, cmd, args.installerRoleArn, MinorVersionForGetSecret)
+					GetInstallerRoleArn(r, cmd, args.installerRoleArn, MinorVersionForGetSecret, r.AWSClient.FindRoleARNs)
```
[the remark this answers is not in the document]
gdbranco: That's correct. The point is just to have the perms to open the secret, both do

cmd/register/oidcconfig/cmd.go
gdbranco: Yeah, same idea. The only requirement is to have the perms. It doesn't really need to be the installer, it's just that we know the installer works for sure

PR #1616, opened 2023-11-21
OCM-3963 v2 changes

den-rgb:
Changes for [OCM-3963](https://github.com/openshift/rosa/pull/1564/) excluding the vendoring


pkg/aws/api_interface/cloudformation_api_client.go
```diff
@@ -0,0 +1,91 @@
+package aws
```
robpblake, line 1: @den-rgb As a way of making the review smaller, everything in this package and the tests could be introduced as a separate PR. We'd just drop in the new implementations but not update anything to use them yet. Just a thought.

pkg/aws/client.go
```diff
@@ -202,16 +200,16 @@ type ClientBuilder struct {
...
-	iamClient           iamiface.IAMAPI
-	ec2Client           ec2iface.EC2API
-	orgClient           organizationsiface.OrganizationsAPI
-	s3Client            s3iface.S3API
-	smClient            secretsmanageriface.SecretsManagerAPI
-	stsClient           stsiface.STSAPI
-	cfClient            cloudformationiface.CloudFormationAPI
-	servicequotasClient servicequotasiface.ServiceQuotasAPI
-	awsSession          *session.Session
+	iamClient           f.IamApiClient
```
robpblake, line 205: @den-rgb Another idea could be a per-field-PR here. So we'd gradually swap over the `client.go` implementation to use the v2 API. There would be a window where it was mix of `v1` and `v2` and I would expect the first PR to be larger than the follow on ones.

The only goal I have is to reduce the cognitive load for reviewers, as I'm personally finding there are too many changes to review this PR as fully as it deserves - I don't have the review super-powers that @oriAdler clearly has :-)

oriAdler: Indeed multiple changes, I'm reviewing this PR in chunks:)

oriAdler: Reviewing this PR https://github.com/openshift/rosa/pull/1564, separated into two commits.

robpblake: @oriAdler I personally think we should decompose this into much smaller PRs. It's too big a change to be reviewed as thoroughly as it deserves. As a team we should recognise when a PR is too big IMHO.

PR #1626, opened 2023-11-24
OCM-807 | fix: do not allow fips flag for HCP

gdbranco:
Related issue: [OCM-807](https://issues.redhat.com//browse/OCM-807)


chenz4027 (approved):
/lgtm
@gdbranco Thanks for the quick fix!

PR #1629, opened 2023-11-27
OCM-4883 | fix: fetch passed arn in all roles

chenz4027:
https://issues.redhat.com/browse/OCM-4883

There was an issue where if a use creates a cluster using classic role, but using operator roles that uses hcp roles. There will be a mismatch between managed/unmanaged policies, thus upgrade will fail.

This change will allow the user to create operator roles with classic role if passes '--installer-role-arn string'


oriAdler:
LGTM, @gdbranco @lucasponce can you please take a look as well?

gdbranco:
lgtm as well, any reason to hold @chenz4027 ?

chenz4027:
@gdbranco Ah thanks for the heads up, I just wanted to make sure that it was reviewed by Ori first. 
/unhold

PR #1635, opened 2023-11-29
OCM-3963 | Updating ROSA CLI to use aws sdk go v2

den-rgb:
## Describe your changes

This pr is to update the `aws sdk go` version from `v1` to `v2`, the majority of these changes are refactoring the code for it to work with the updated library.

### JIRA link
[JIRA](https://issues.redhat.com/browse/OCM-3963)

### What's new?

- Refactoring of functions to work with the updated library
- Updated test cases
- Added interfaces and associated functions from the v2 [repo](https://github.com/aws/aws-sdk-go-v2/tree/main/service)
- Added a new custom [waiter function](https://github.com/den-rgb/rosa/blob/OCM-3963-x/pkg/aws/helpers.go#L1014) due to it being removed in v2
- Refactored code to include `GetPartition()` changes due to the method of retrieving partitions in v2 is now not viable as it was done in v1. Partitions are now retrieved via the Creator in runtime.

### How to test?

- All test cases should pass within the repo
- All basic commands should work as prior
- QE testing pipelines pass

## PR Author Check List

* [x]  All commit messages adhere to the project [standard](https://github.com/openshift/rosa/blob/master/CONTRIBUTING.md) 

* [ ]  All commits are squashed into a single commit `git rebase -i HEAD~<number of commits>`

* [x]  All code changes have unit test coverage

* [ ]  All changes have been tested locally and the steps documented in the `How to Test?` section


oriAdler:
@den-rgb can you please add a description to the PR? highlight the changes, and guide us on which changes should get more attention in the review.
For example, what we were discussing about the **waitForStackToBeReady** function which isn't available in V2.

den-rgb:
@oriAdler I reached out to the aws support in regards to the waiter function and I was told that at the moment the way I have it implemented is good and that there isn't much alternative to how to implement it currently.

pkg/aws/client.go
```diff
@@ -423,32 +415,40 @@ func (b *ClientBuilder) Build() (Client, error) {
 	return c, err
 }
 
-func (c *awsClient) GetIAMCredentials() (credentials.Value, error) {
-	return c.awsSession.Config.Credentials.Get()
+func (c *awsClient) GetIAMCredentials() (aws.Credentials, error) {
+	cfg, err := config.LoadDefaultConfig(context.TODO())
```
oriAdler, line 419: Do we need to load the config? or can we use the one from the `awsClient` struct?

pkg/aws/client.go
```diff
@@ -423,32 +415,40 @@ func (b *ClientBuilder) Build() (Client, error) {
...
+	if err != nil {
+		return aws.Credentials{}, err
+	}
+
+	creds, err := cfg.Credentials.Retrieve(context.TODO())
```
oriAdler, line 424:
```suggestion
	return cfg.Credentials.Retrieve(context.TODO())
```

pkg/aws/client.go
```diff
@@ -546,27 +546,28 @@ func (c *awsClient) FilterVPCsPrivateSubnets(subnets []*ec2.Subnet) ([]*ec2.Subn
...
-func (c *awsClient) getSubnetRouteTable(subnetID *string, routeTables []*ec2.RouteTable) (*ec2.RouteTable, error) {
+func (c *awsClient) getSubnetRouteTable(subnetID *string,
+	routeTables []ec2types.RouteTable) (*ec2types.RouteTable, error) {
 	// Subnet route table — A route table that's associated with a subnet
 	for _, routeTable := range routeTables {
 		for _, association := range routeTable.Associations {
-			if aws.StringValue(association.SubnetId) == aws.StringValue(subnetID) {
-				return routeTable, nil
+			if aws.ToString(association.SubnetId) == aws.ToString(subnetID) {
+				return &routeTable, nil
```
oriAdler, line 570: Were you able to test this change? previously we returned `routeTable` and now `&routeTable`

pkg/aws/client.go
```diff
@@ -741,7 +745,11 @@ func (c *awsClient) GetAWSAccessKeys() (*AccessKey, error) {
 }
 
 func (c *awsClient) GetLocalAWSAccessKeys() (*AccessKey, error) {
-	creds, err := c.awsSession.Config.Credentials.Get()
+	cfg, err := config.LoadDefaultConfig(context.TODO())
```
oriAdler, line 748: Do we have `cfg` stored in the `awsClient` struct? or do we need to retrieve it?

pkg/aws/client.go
den-rgb: yes we can use the one from the struct

cmd/create/cluster/cmd.go
```diff
@@ -3264,21 +3265,25 @@ func handleOidcConfigOptions(r *rosa.Runtime, cmd *cobra.Command, isSTS bool, is
 	return oidcConfig
 }
 
-func filterPrivateSubnets(initialSubnets []*ec2.Subnet, r *rosa.Runtime) []*ec2.Subnet {
+func filterPrivateSubnets(initialSubnets []ec2types.Subnet, r *rosa.Runtime) []ec2types.Subnet {
+	initialSubnetPointers := make([]*ec2types.Subnet, len(initialSubnets))
```
oriAdler, line 3269: If we change the signature of `FetchPublicSubnetMap` `FetchPublicSubnetMap(subnets []ec2types.Subnet)`  I think we won't need the casting here.
Inside the function we are using only the subnet IDs, therefore, we can also send a slice of subnets, WDYT?

pkg/aws/client.go
```diff
@@ -67,6 +68,7 @@ import (
 const (
 	AdminUserName        = "osdCcsAdmin"
 	OsdCcsAdminStackName = "osdCcsAdminIAMUser"
+	rosaHeaderValue      = "rosa.ROSAVersionUserAgentHandler"
```
oriAdler, line 71: Unused constant

pkg/aws/client.go
```diff
@@ -1213,62 +1212,40 @@ func (c *awsClient) DeleteSecretInSecretsManager(secretArn string) error {
...
-		logger.Warn("Throttling Rate limit exceeded. Retrying the request again")
+	for _, secGroup := range resp.SecurityGroups {
+		if c.Ec2ResourceHasTag(secGroup.Tags, tags.RedHatManaged, strconv.FormatBool(true)) {
+			continue
+		}
+		if aws.ToString(secGroup.GroupName) == "default" {
+			continue
+		}
+		sg := secGroup
+		securityGroups = append(securityGroups, &sg)
```
oriAdler, line 1238: It seems like `DescribeSecurityGroups` returns a slice of `[]SecurityGroup`, do we still need the casting to pointers here? or can we change the signature of the function to return a slice of `SecurityGroup`?

pkg/aws/client.go
```diff
@@ -293,29 +291,72 @@ func (b *ClientBuilder) UseLocalCredentials(value bool) *ClientBuilder {
...
+		config.WithAPIOptions([]func(stack *middleware.Stack) error{
+			smithyhttp.AddHeaderValue("User-Agent",
+				strings.Join([]string{"ROSACLI", info.Version}, ";")),
+		}),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return &cfg, nil
```
oriAdler, line 319: `BuildSession` triggers both `BuildSessionWithOptionsCredentials` and `BuildSessionWithOptions`, we return `*aws.config`, then inside `Build()` we reference the pointer when building the aws client:
```
	c := &awsClient{
		cfg:                 *cfg,
```
It would be better to return `aws.config` instead of the back and forth, to improve readability and reduce pointers, WDYT?

pkg/aws/client.go
```diff
@@ -799,7 +793,7 @@ func (c *awsClient) GetAWSAccessKeys() (*AccessKey, error) {
 }
 
 func (c *awsClient) GetLocalAWSAccessKeys() (*AccessKey, error) {
-	creds, err := c.awsSession.Config.Credentials.Get()
+	creds, err := c.cfg.Credentials.Retrieve(context.TODO())
```
oriAdler, line 796: NP
I couldn't find a major difference between `context.TODO()` and `context.Background()` in our case, as we don't set a deadline for the request.
I wonder if we should use `context.Background()`, just for the sake of consistency, not a must though.

pkg/aws/cloudformation.go
```diff
@@ -114,83 +115,67 @@ func (c *awsClient) EnsureOsdCcsAdminUser(stackName string, adminUserName string
...
-	// Wait until cloudformation stack creates
-	err = c.cfClient.WaitUntilStackCreateComplete(&cloudformation.DescribeStacksInput{
-		StackName: aws.String(stackName),
-	})
+	err = waitForStackCreateComplete(context.Background(), c.cfClient, stackName)
 	if err != nil {
-		switch typed := err.(type) {
-		case awserr.Error:
-			// Waiter reached maximum attempts waiting for the resource to be ready
-			if typed.Code() == request.WaiterResourceNotReadyErrorCode {
```
oriAdler, line 130: Should we keep the error type check and the logging for the max retries scenario?

pkg/aws/helpers.go
```diff
@@ -1000,3 +991,63 @@ func IsHostedCPManagedPolicies(cluster *cmv1.Cluster) bool {
 func IsHostedCP(cluster *cmv1.Cluster) bool {
 	return cluster.Hypershift().Enabled()
 }
+
+func ConvertToTagPointers(tags []iamtypes.Tag) []*iamtypes.Tag {
```
oriAdler, line 995: Do we need this conversation? 
Previously `ListPolicyTagsInput` returned a slice of pointers to tags, with V2 it returns a slice of tags.
I wonder if we should change the signature of `isAccountRoleVersionCompatible` and `IamResourceHasTag` to get a slice of tags, and then we can save the conversion and reduce the number of pointers in the code.

pkg/aws/helpers.go
```diff
@@ -1000,3 +991,63 @@ func IsHostedCPManagedPolicies(cluster *cmv1.Cluster) bool {
...
+
+func waitForStackCreateComplete(ctx context.Context, cfClient client.CloudFormationApiClient, stackName string) error {
+	waiter := cloudformation.NewStackCreateCompleteWaiter(cfClient)
+
+	maxWaitDur := 5 * time.Minute // Max Wait time
+
+	params := buildDescribeStacksInput(stackName)
+
+	// You can also use WaitForOutput if you need the output
+	err := waiter.Wait(ctx, params, maxWaitDur, func(o *cloudformation.StackCreateCompleteWaiterOptions) {
```
oriAdler, line 1018:
```suggestion
	return waiter.Wait(ctx, params, maxWaitDur, func(o *cloudformation.StackCreateCompleteWaiterOptions) {
```

pkg/aws/helpers.go
```diff
@@ -1000,3 +991,63 @@ func IsHostedCPManagedPolicies(cluster *cmv1.Cluster) bool {
...
+
+func waitForStackUpdateComplete(ctx context.Context, cfClient client.CloudFormationApiClient, stackName string) error {
+	waiter := cloudformation.NewStackUpdateCompleteWaiter(cfClient)
+
+	maxWaitDur := 5 * time.Minute // Max Wait time
+
+	params := buildDescribeStacksInput(stackName)
+
+	// You can also use WaitForOutput if you need the output
+	err := waiter.Wait(ctx, params, maxWaitDur, func(o *cloudformation.StackUpdateCompleteWaiterOptions) {
```
oriAdler, line 1033:
```suggestion
	return waiter.Wait(ctx, params, maxWaitDur, func(o *cloudformation.StackUpdateCompleteWaiterOptions) {
```

pkg/aws/idps.go
```diff
@@ -34,72 +35,64 @@ const (
...
 	})
 	if err != nil {
-		if aerr, ok := err.(awserr.Error); ok {
-			switch aerr.Code() {
-			case iam.ErrCodeNoSuchEntityException:
-				return fmt.Errorf("OIDC provider '%s' does not exists",
-					oidcProviderARN)
-			}
+		if awserr.IsNoSuchEntityException(err) {
+			return fmt.Errorf("The OIDC provider does not exist")
```
oriAdler, line 95: Should we include the OIDC provider ARN as we had previously in the error message?

pkg/aws/helpers.go
```diff
@@ -1000,3 +991,63 @@ func IsHostedCPManagedPolicies(cluster *cmv1.Cluster) bool {
...
+		// Optionally set MinDelay, MaxDelay, and other options here
+	})
+
+	return err
+}
+
+func waitForStackUpdateComplete(ctx context.Context, cfClient client.CloudFormationApiClient, stackName string) error {
+	waiter := cloudformation.NewStackUpdateCompleteWaiter(cfClient)
+
+	maxWaitDur := 5 * time.Minute // Max Wait time
```
robpblake, line 1028: @den-rgb Shall we make this a `const` and then simply refer to it in the three functions that have the 5 minute wait?

Are these functions tested anywhere?

pkg/aws/client.go
robpblake: @oriAdler I think `BuildSession` only triggers one or the other as it returns the result of the function in the `if` statement e.g:

```
if b.credentials != nil {
		return b.BuildSessionWithOptionsCredentials(b.credentials, logLevel)
	}
```

Am I correct?

pkg/aws/idps.go
robpblake: +1 to including the `arn` in the error message as before.

pkg/aws/permissions.go
```diff
@@ -18,8 +19,10 @@ type SimulateParams struct {
 func (c *awsClient) ValidateSCP(target *string, policies map[string]*cmv1.AWSSTSPolicy) (bool, error) {
 	policyDetails := GetPolicyDetails(policies, "osd_scp_policy")
 
+	region := c.GetRegion()
```
robpblake, line 22: @den-rgb I don't think there's any need to assign the `region` variable. The `c.getRegion()` function call can be moved onto the definition of the `SimulateParams` struct e.g:

```
sParams := &SimulateParams{
		Region: c.GetRegion(),
	}
```

pkg/aws/sts.go
```diff
@@ -79,7 +80,7 @@ func (c *awsClient) ValidateRoleARNAccountIDMatchCallerAccountID(roleARN string)
 }
 
 func (c *awsClient) HasPermissionsBoundary(roleName string) (bool, error) {
-	output, err := c.iamClient.GetRole(&iam.GetRoleInput{
+	output, err := c.iamClient.GetRole(context.Background(), &iam.GetRoleInput{
```
robpblake, line 83: @den-rgb Personally, I think we should go with `context.TODO()` rather than `context.Background()` as it feels most appropriate given the context API docs:

```
TODO returns a non-nil, empty Context. Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).
```
https://pkg.go.dev/context#TODO

pkg/aws/client.go
robpblake: @oriAdler @den-rgb I think we should use `context.TODO()` consistently. From the Go API docs for `context.TODO()` :

```
TODO returns a non-nil, empty Context. Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).
```

I think that describes our situation. We don't have a parent context being passed down through, so we're not sure which context to use.

pkg/aws/client.go
den-rgb: yes it only returns one or the other based on the if statement

pkg/aws/client.go
```diff
@@ -1213,62 +1212,40 @@ func (c *awsClient) DeleteSecretInSecretsManager(secretArn string) error {
...
 
-func buildCustomRetryer() CustomRetryer {
-	return CustomRetryer{
-		DefaultRetryer: client.DefaultRetryer{
-			NumMaxRetries:    12,
-			MinRetryDelay:    1 * time.Second,
-			MinThrottleDelay: 5 * time.Second,
-			MaxThrottleDelay: 5 * time.Second,
-		},
+func (c *awsClient) Ec2ResourceHasTag(tags []ec2types.Tag, tagName, tagValue string) bool {
```
robpblake, line 1244: @den-rgb Could you add unit tests for this new function please.

pkg/aws/client.go
robpblake: @oriAdler @den-rgb FYI: https://issues.redhat.com/browse/OCM-5364

pkg/aws/policies.go
```diff
@@ -144,25 +146,30 @@ var roleTypeMap = map[string]string{
...
 	})
 	if err != nil {
-		if aerr, ok := err.(awserr.Error); ok {
-			switch aerr.Code() {
-			case iam.ErrCodeNoSuchEntityException:
-				return c.createRole(name, policy, permissionsBoundary, tagList, path)
-			default:
-				return "", err
-			}
+		if awserr.IsNoSuchEntityException(err) {
```
oriAdler, line 153: Nice refactor! much better than the previous switch statement:)

den-rgb: Thanks

pkg/aws/policies.go
```diff
@@ -634,69 +634,75 @@ func isAccountRoleVersionCompatible(tagsList []*iam.Tag, roleType string,
...
+	roles := []*iamtypes.Role{}
+	paginator := iam.NewListRolesPaginator(c.iamClient, &iam.ListRolesInput{})
+	for paginator.HasMorePages() {
+		output, err := paginator.NextPage(context.TODO())
+		if err != nil {
+			return nil, err
+		}
+		for _, role := range output.Roles {
+			r := role
+			roles = append(roles, &r)
```
oriAdler, line 647: Should we return a slice of roles like the AWS SDK is now returning? instead of pointers

pkg/aws/policies.go
```diff
@@ -1069,15 +1082,13 @@ func (c *awsClient) DeleteOperatorRole(roleName string, managedPolicies bool) er
...
 	if err != nil {
 		if err != nil {
-			if aerr, ok := err.(awserr.Error); ok {
-				switch aerr.Code() {
-				case iam.ErrCodeNoSuchEntityException:
-					return fmt.Errorf("operator role '%s' does not exists.skipping",
-						role)
-				}
+			if awserr.IsNoSuchEntityException(err) {
+				return fmt.Errorf("operator role '%s' does not exists.skipping",
```
oriAdler, line 1090:
```suggestion
				return fmt.Errorf("operator role '%s' does not exist, skipping",
```

pkg/aws/policies.go
```diff
@@ -1171,21 +1177,21 @@ func (c *awsClient) detachAttachedRolePolicies(role *string) error {
...
 	}
 	for _, policyName := range listRolePolicyOutput.PolicyNames {
-		_, err = c.iamClient.DeleteRolePolicy(&iam.DeleteRolePolicyInput{
-			PolicyName: policyName,
-			RoleName:   aws.String(role),
-		})
+		pn := policyName
+		_, err = c.iamClient.DeleteRolePolicy(context.Background(),
+			&iam.DeleteRolePolicyInput{
+				PolicyName: &pn,
```
oriAdler, line 1189: Would that work? or it won't be allowed to return the address of an entry?
```suggestion
				PolicyName: aws.string(policyName),
```

pkg/aws/policies.go
```diff
@@ -1679,36 +1661,30 @@ func (c *awsClient) GetOpenIDConnectProviderByOidcEndpointUrl(oidcEndpointUrl st
...
+		role, err := c.iamClient.GetRole(context.Background(), &iam.GetRoleInput{
 			RoleName: aws.String(roleName),
 		})
-		if aerr, ok := err.(awserr.Error); ok {
-			switch aerr.Code() {
-			case iam.ErrCodeNoSuchEntityException:
-				return "", errors.NotFound.Errorf("Roles with the prefix'%s' not found", prefix)
-			}
+		if awserr.IsNoSuchEntityException(err) {
+			return "", errors.NotFound.Errorf("Roles with the prefix'%s' not found", prefix)
```
oriAdler, line 1668:
```suggestion
			return "", errors.NotFound.Errorf("Roles with the prefix '%s' not found", prefix)
```

pkg/aws/helpers.go
den-rgb: should the same be done for: `HasCompatibleVersionTags` ?

pkg/logging/aws_logger.go
```diff
@@ -38,9 +37,6 @@ type AWSLogger struct {
 	logger *logrus.Logger
 }
 
-// Make sure that we implement the OCM logger interface.
-var _ aws.Logger = &AWSLogger{}
```
oriAdler, line 42: We don't need this anymore?

pkg/ocm/helpers.go
```diff
@@ -684,23 +681,13 @@ func ValidateHostedClusterSubnets(awsClient aws.Client, isPrivate bool, subnetID
 		return 0, vpcSubnetsErr
 	}
 
-	var subnets []*ec2.Subnet
-	for _, subnet := range vpcSubnets {
-		for _, subnetId := range subnetIDs {
```
oriAdler, line 689: It seems like we are filtering through the user choices before calling `FilterVPCsPrivateSubnets`, why are we removing this part?

pkg/aws/helpers.go
den-rgb: They're tested as part of this test case: https://github.com/den-rgb/rosa/blob/OCM-3963-x/pkg/aws/client_test.go#L68

den-rgb: I've also done manual testing verifying they work

pkg/aws/client.go
oriAdler: @robpblake yes, thanks for the clarification.
We can still change the return value of the function to `aws.config`, to simplify the code.

pkg/aws/sts.go
den-rgb: Given the nature of IAM role management and that the operations are routine without specific timing requirements, using context.Background() seems appropriate in this context

den-rgb: although I doubt it'll make much of a difference whichever is used

pkg/aws/helpers.go
oriAdler: Since `HasCompatibleVersionTags` is already in the ocm-common, to avoid the delay of this PR, I assume it's ok to refactor it in a follow-up PR.

pkg/aws/sts.go
oriAdler: +1 I agree.

pkg/aws/policies.go
den-rgb: no it requires a string pointer

pkg/logging/aws_logger.go
den-rgb: aws sdk v2 doesn't implement a logger

pkg/aws/helpers.go
robpblake: Thanks @den-rgb

pkg/ocm/helpers.go
den-rgb: mistake on my part, I thought it was just an uneccessary conversion to a pointer

pkg/aws/cloudformation.go
oriAdler: Since we are only logging the error in case of this error type, and there isn't a special handling for that, I think it's okay to omit it with V2 if there isn't an equivalent error code.

cmd/create/cluster/cmd.go
oriAdler: @den-rgb can we change `FetchPublicSubnetMap` signature and remove the casting?

pkg/aws/helpers.go
oriAdler: @den-rgb we can now remove `ConvertToTagPointers` as it's an unused function.

oriAdler (commented):
Left two comments, other than that LGTM, well done!

cmd/create/cluster/cmd.go
den-rgb: no we dont i have removed it now

PR #1650, opened 2023-12-06
OCM-1079 | fix: only show private subnets when private link cluster is selected

gdbranco:
Related issue: [OCM-1079](https://issues.redhat.com//browse/OCM-1079)

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -1036,7 +1036,9 @@ func run(cmd *cobra.Command, _ []string) {
 				if helper.ContainsPrefix(billingAccounts, awsCreator.AccountID) {
 					billingAccount = awsCreator.AccountID
 					r.Reporter.Infof("Using '%s' as billing account", billingAccount)
-					r.Reporter.Infof("To use a different billing account, add --billing-account xxxxxxxxxx to previous command")
+					r.Reporter.Infof(
+						"To use a different billing account, add --billing-account xxxxxxxxxx to previous command",
+					)
 				} else {
 					r.Reporter.Errorf("A billing account is required for Hosted Control Plane clusters.")
 				}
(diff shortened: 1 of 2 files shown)


pkg/aws/client.go
```diff
@@ -431,6 +435,47 @@ func (c *awsClient) GetRegion() string {
 	return aws.StringValue(c.awsSession.Config.Region)
 }
 
+func (c *awsClient) IsPublicSubnet(subnet *ec2.Subnet) (bool, error) {
+	routeTablesResp, err := c.ec2Client.DescribeRouteTables(&ec2.DescribeRouteTablesInput{
```
oriAdler, line 439: I wonder if we should fetch all route tables at once to reduce API calls, then we can have a subnet -> route table map, WDYT?

cmd/create/cluster/cmd.go
```diff
@@ -2011,26 +2018,47 @@ func run(cmd *cobra.Command, _ []string) {
...
 		for _, subnet := range initialSubnets {
+			skip := false
 			subnetIP, subnetNetwork, err := net.ParseCIDR(*subnet.CidrBlock)
 			if err != nil {
 				r.Reporter.Errorf("Unable to parse subnet CIDR")
 				os.Exit(1)
 			}
 
-			if machineNetwork.Contains(subnetIP) &&
+			if !(machineNetwork.Contains(subnetIP) &&
```
oriAdler, line 2030: Should it be `||` here?
```suggestion
			if !(machineNetwork.Contains(subnetIP) ||
```
Maybe we can simplify the condition to improve readability.

oriAdler: Maybe we can create a new function `isValidCidr` and then encapsulate the previous conditions we had in the function?

pkg/aws/client.go
gdbranco: I'm actually not sure if takes longer to retrieve all route tables or filter through subnets :thinking: . Will test a bit before going back to this

oriAdler: It could be a premature optimization, if it takes a reasonable time to check one by one, we can continue with the current approach.

oriAdler: Oh, I meant retrieve all route tables providing all the provided subnets, not all the route tables in the AWS account.
The subnet input field is of type `[]*string`, we can pass multiple subnets to filter.

gdbranco: Feels faster! Thanks for the input

pkg/aws/client.go
```diff
@@ -431,6 +436,54 @@ func (c *awsClient) GetRegion() string {
 	return aws.StringValue(c.awsSession.Config.Region)
 }
 
+func (c *awsClient) IsPublicSubnet(subnets []*ec2.Subnet) (map[string]bool, error) {
```
oriAdler, line 439: Maybe we can go for something like `getPublicSubnetMap` or `generateSubnetToPublicMap`? as the function doesn't check if a single subnet is public but generates a mapping.

oriAdler:
Left a small comment, other than that LGTM:)

PR #1651, opened 2023-12-07
OCM-3875 | fix: don't show create operator roles midflow

thomasmckay:


cmd/create/cluster/cmd.go
```diff
@@ -1615,12 +1615,7 @@ func run(cmd *cobra.Command, _ []string) {
 		if len(operatorRoles) != 0 {
 			r.Reporter.Infof("Operator roles found matching prefix '%s'.", operatorRolesPrefix)
 		} else {
-			rolesCMD := fmt.Sprintf("rosa create operator-roles --cluster %s", clusterName)
-			if permissionsBoundary != "" {
-				rolesCMD = fmt.Sprintf("%s --permissions-boundary %s", rolesCMD, permissionsBoundary)
-			}
-			r.Reporter.Infof("No operator roles found matching prefix '%s'. "+
-				"Run the follow command to create them:\n\t%s", operatorRolesPrefix, rolesCMD)
+			r.Reporter.Infof("No operator roles found matching prefix '%s'.", operatorRolesPrefix)
```
gdbranco, line 1618: I actually don't think we need to output anything here. Could be a debug in case we want to investigate one day. But if there are no operator roles pre created yet means the user is going with the old flow or forgot to pre create. So they will use the output from after the post request for the cluster

thomasmckay: done, thanks!

PR #1655, opened 2023-12-11
OCM-3594 | Feat | Add rhRegion flag to rosa login command

tirthct:
Issue : https://issues.redhat.com/browse/OCM-3594

**Changed**
- Add hidden flag rhRegion to rosa login command
- Add list rh-region command to list regions
- Add unit tests

**Tested**
This MR tweaks the `rosa login --env` resolution algorithm to try to reuse `cfg.URL` before falling back to "api.openshift.com". New `--env` resolution algorithm (from highest priority to lowest priority):

1. runtime `--env` cli arg (key found in `URLAliases`)
2. runtime `--env` cli arg (non-empty string)
3. [NEW] config file `URL` value (non-empty string)
4. sdk.DefaultURL

`rosa list rh-region`:
* show discovery URL used
* add optional `--discovery-url` flag with the same url resolution behavior as `ocm login`
* display region name and its associated url rather than just region name

### DIFF
--- a/cmd/list/cmd.go
+++ b/cmd/list/cmd.go
@@ -34,6 +34,7 @@ import (
 	"github.com/openshift/rosa/cmd/list/oidcprovider"
 	"github.com/openshift/rosa/cmd/list/operatorroles"
 	"github.com/openshift/rosa/cmd/list/region"
+	"github.com/openshift/rosa/cmd/list/rhRegion"
 	"github.com/openshift/rosa/cmd/list/service"
 	"github.com/openshift/rosa/cmd/list/tuningconfigs"
 	"github.com/openshift/rosa/cmd/list/upgrade"
@@ -71,6 +72,7 @@ func init() {
 	Cmd.AddCommand(tuningconfigs.Cmd)
 	Cmd.AddCommand(oidcprovider.Cmd)
 	Cmd.AddCommand(dnsdomains.Cmd)
+	Cmd.AddCommand(rhRegion.Cmd)
 	Cmd.AddCommand(externalauthprovider.Cmd)
 	flags := Cmd.PersistentFlags()
 	arguments.AddProfileFlag(flags)
--- a/cmd/list/rhRegion/cmd.go
+++ b/cmd/list/rhRegion/cmd.go
@@ -1,0 +1,88 @@
+package rhRegion
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"text/tabwriter"
+
+	sdk "github.com/openshift-online/ocm-sdk-go"
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/rosa/pkg/config"
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+var (
+	writer = tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', tabwriter.TabIndent)
+	args   struct {
+		discoveryURL string
+	}
+)
+
+var Cmd = NewRhRegionCommand()
+
+func NewRhRegionCommand() *cobra.Command {
+	Cmd := &cobra.Command{
+		Use:   "rh-regions",
+		Short: "List available OCM regions",
+		Long:  "The command lists available OpenShift Cluster Manager regions.",
+		Example: `  # List all supported OCM regions 
+ocm list rh-regions`,
+		Run:    run,
+		Hidden: true,
+		Args:   cobra.NoArgs,
+	}
+
+	flags := Cmd.Flags()
+	flags.StringVar(
+		&args.discoveryURL,
+		"discovery-url",
+		"",
+		"URL of the OCM API gateway. If not provided, will reuse the URL from the configuration "+
+			"file or "+sdk.DefaultURL+" as a last resort. The value should be a complete URL "+
+			"or a valid URL alias: "+strings.Join(ocm.ValidOCMUrlAliases(), ", "),
+	)
+	return Cmd
+}
+
+func run(cmd *cobra.Command, argv []string) {
+	r := rosa.NewRuntime()
+
+	err := ListRhRegions(args.discoveryURL, r)
+	if err != nil {
+		r.Reporter.Errorf("Failed to determine gateway URL: %v", err)
+		os.Exit(1)
+	}
+}
+
+func ListRhRegions(discoveryURL string, r *rosa.Runtime) error {
+	cfg, err := config.Load()
+	if err != nil {
+		return fmt.Errorf("Failed to load config file: %v", err)
+	}
+
+	gatewayURL, err := ocm.ResolveGatewayUrl(discoveryURL, cfg)
+	if err != nil {
+		return fmt.Errorf("Failed to determine gateway URL: %v", err)
+	}
+
+	fmt.Fprintf(writer, "Discovery URL: %s\n\n", gatewayURL)
+	regions, err := sdk.GetRhRegions(gatewayURL)
+	if err != nil {
+		return fmt.Errorf("Failed to get OCM regions: %v", err)
+	}
+
+	// If there are no regions, print a warning message and return early
+	if len(regions) == 0 {
+		r.Reporter.Warnf("No regions found")
+		return nil
+	}
+	fmt.Fprintf(writer, "RH Region\t\tGateway URL\n")
+	for regionName, region := range regions {
+		fmt.Fprintf(writer, "%s\t\t%v\n", regionName, region.URL)
+	}
+	writer.Flush()
+	return nil
+}
--- a/cmd/login/cmd.go
+++ b/cmd/login/cmd.go
@@ -57,6 +57,7 @@ var args struct {
 	insecure      bool
 	useAuthCode   bool
 	useDeviceCode bool
+	rhRegion      string
 }
 
 var Cmd = &cobra.Command{
@@ -109,14 +110,15 @@ func init() {
 		"OpenID scope. If this option is used it will completely replace the default "+
 			"scopes. Can be repeated multiple times to specify multiple scopes.",
 	)
+	flags.SetNormalizeFunc(arguments.NormalizeFlags)
 	flags.StringVar(
 		&args.env,
-		"env",
-		sdk.DefaultURL,
+		arguments.NewEnvFlag,
+		"",
 		"Environment of the API gateway. The value can be the complete URL or an alias. "+
 			"The valid aliases are 'production', 'staging' and 'integration'.",
 	)
-	flags.MarkHidden("env")
+	flags.MarkHidden(arguments.NewEnvFlag)
 	flags.StringVarP(
 		&args.token,
 		"token",
@@ -146,6 +148,15 @@ func init() {
 			"This should only be used for remote hosts and containers where browsers are "+
 			"not available. Use auth code for all other scenarios.",
 	)
+	flags.StringVar(
+		&args.rhRegion,
+		"rh-region",
+		"",
+		"OCM data sovereignty region identifier. --env will be used to initiate a service discovery "+
+			"request to find the region URL matching the provided identifier. Use `rosa list rh-regions` "+
+			"to see available regions.",
+	)
+	flags.MarkHidden("rh-region")
 	arguments.AddRegionFlag(flags)
 	fedramp.AddFlag(flags)
 }
@@ -160,10 +171,6 @@ func run(cmd *cobra.Command, argv []string) {
 
 	// Check mandatory options:
 	env := args.env
-	if env == "" {
-		r.Reporter.Errorf("Option '--env' is mandatory")
-		os.Exit(1)
-	}
 
 	// Confirm that token is not passed with auth code flags
 	if (args.useAuthCode || args.useDeviceCode) && args.token != "" {
@@ -238,8 +245,8 @@ func run(cmd *cobra.Command, argv []string) {
 		config.IsEncryptedToken(token) {
 		fedramp.Enable()
 		// Always default to prod
-		if env == sdk.DefaultURL {
-			env = "production"
+		if env == sdk.DefaultURL || env == "" {
+			env = ocm.Production
 		}
 		if fedramp.HasAdminFlag(cmd) {
 			uiTokenPage = fedramp.AdminLoginURLs[env]
@@ -296,11 +303,13 @@ func run(cmd *cobra.Command, argv []string) {
 		fedramp.Enable()
 	}
 
-	// Apply the default configuration details if not explicitly provided by the user:
-	gatewayURL, ok := ocm.URLAliases[env]
-	if !ok {
-		gatewayURL = env
+	gatewayURL, err := ocm.ResolveGatewayUrl(env, cfg)
+	if err != nil {
+		r.Reporter.Errorf("Failed to resolve gateway URL: %v", err)
+		os.Exit(1)
 	}
+
+	var ok bool
 	tokenURL := sdk.DefaultTokenURL
 	if args.tokenURL != "" {
 		tokenURL = args.tokenURL
@@ -338,6 +347,26 @@ func run(cmd *cobra.Command, argv []string) {
 				clientID = args.clientID
 			}
 		}
+	}
+
+	// If an --rh-region is provided, gatewayURL is resolved from ResolveGatewayUrl() and then used to initiate
+	// service discovery for the environment gatewayURL is a part of, but it (and
+	// ultimately the cfg.URL) is then updated to the URL of the matching --rh-region:
+	//   1. resolve the gatewayURL as above
+	//   2. fetch a well-known file from sdk.GetRhRegion
+	//   3. update the gatewayURL to the region URL matching args.rhRegion
+	//
+	// So `--env=https://api.stage.openshift.com --rh-region=ap-southeast-1` might result in
+	// gatewayURL/cfg.URL being mutated to "https://api.aws.ap-southeast-1.stage.openshift.com"
+	//
+	// See ocm-sdk-go/rh_region.go for full details on how service discovery works.
+	if args.rhRegion != "" {
+		regValue, err := sdk.GetRhRegion(gatewayURL, args.rhRegion)
+		if err != nil {
+			r.Reporter.Errorf("Can't find region: %v", err)
+			os.Exit(1)
+		}
+		gatewayURL = fmt.Sprintf("https://%s", regValue.URL)
 	}
 
 	// Update the configuration with the values given in the command line:
@@ -467,7 +496,7 @@ func tokenType(jwtToken *jwt.Token) (typ string, err error) {
 }
 
 func Call(cmd *cobra.Command, argv []string, reporter *rprtr.Object) error {
-	loginFlags := []string{"token-url", "client-id", "client-secret", "scope", "env", "token", "insecure"}
+	loginFlags := []string{"token-url", "client-id", "client-secret", "scope", arguments.NewEnvFlag, "token", "insecure"}
 	hasLoginFlags := false
 	// Check if the user set login flags
 	for _, loginFlag := range loginFlags {
--- a/pkg/arguments/normalize.go
+++ b/pkg/arguments/normalize.go
@@ -13,6 +13,8 @@ const (
 	NewControlPlaneIAMRole         = "controlplane-iam-role-arn"
 	DeprecatedWorkerIAMRole        = "worker-iam-role"
 	NewWorkerIAMRole               = "worker-iam-role-arn"
+	DeprecatedEnvFlag              = "env"
+	NewEnvFlag                     = "url"
 )
 
 func NormalizeFlags(_ *pflag.FlagSet, name string) pflag.NormalizedName {
@@ -25,6 +27,8 @@ func NormalizeFlags(_ *pflag.FlagSet, name string) pflag.NormalizedName {
 		name = NewWorkerIAMRole
 	case DeprecatedControlPlaneIAMRole:
 		name = NewControlPlaneIAMRole
+	case DeprecatedEnvFlag:
+		name = NewEnvFlag
 	}
 	return pflag.NormalizedName(name)
 }
--- a/pkg/ocm/config.go
+++ b/pkg/ocm/config.go
@@ -21,7 +21,10 @@ package ocm
 
 import (
 	"fmt"
+	"net/url"
 	"strings"
+
+	sdk "github.com/openshift-online/ocm-sdk-go"
 
 	"github.com/openshift/rosa/pkg/config"
 	"github.com/openshift/rosa/pkg/fedramp"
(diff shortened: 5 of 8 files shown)


cmd/login/cmd.go
```diff
@@ -121,6 +122,15 @@ func init() {
 		"Enables insecure communication with the server. This disables verification of TLS "+
 			"certificates and host names.",
 	)
+	flags.StringVar(
+		&args.rhRegion,
+		"rh-region",
+		"",
+		fmt.Sprintf(
```
robpblake, line 129: @tirthct You don't need the `fmt.Sprintf()` here as there is no formatting on the string.

tirthct: Ah, yes. Removed.

cmd/login/cmd.go
```diff
@@ -121,6 +122,13 @@ func init() {
 		"Enables insecure communication with the server. This disables verification of TLS "+
 			"certificates and host names.",
 	)
+	flags.StringVar(
+		&args.rhRegion,
+		"rh-region",
+		"",
+		"OCM region identifier.",
```
robpblake, line 129: Is `OCM` a known acronym to our customers or is it better to use the full `Openshift Cluster Manager` description?

robpblake:
@tirthct What's the status of this one please? Do you intend to rebase and try to merge?

cheers,
Rob

tirthct:
> @tirthct What's the status of this one please? Do you intend to rebase and try to merge?
> 
> cheers, Rob

Hey @robpblake, we have a PR under review https://github.com/openshift-online/ocm-cli/pull/599. Since the implementation would be similar here as well, I'll update this draft PR once that is approved/merged.

tirthct:
> > @tirthct What's the status of this one please? Do you intend to rebase and try to merge?
> > cheers, Rob
> 
> Hey @robpblake, we have a PR under review [openshift-online/ocm-cli#599](https://github.com/openshift-online/ocm-cli/pull/599). Since the implementation would be similar here as well, I'll update this draft PR once that is approved/merged.

@robpblake This PR is ready for review

macgregor:
fyi this MR now maintains feature parity with what was recently added to the ocm-cli in https://issues.redhat.com/browse/OCM-3226. If we need to make rosa specific adaptations thats fine, but we'd like to keep the usage and implementation as similar as possible.

cmd/login/cmd.go
```diff
@@ -111,7 +112,7 @@ func init() {
 	flags.StringVar(
 		&args.env,
```
macgregor, line 113: rather than env, can we add a `--url` flag and make `--env` an alias of it. that gives people the option of which to use. we can deprecate `--env` eventually once people are used to it.

pkg/ocm/config.go
```diff
@@ -39,6 +42,46 @@ var URLAliases = map[string]string{
...
+	if err != nil {
+		return "", fmt.Errorf(
+			"%w\n\nURL Source: %s\nExpected an absolute URI/path (e.g. %s) or a case-sensitive alias, one of: [%s]",
+			err, source, sdk.DefaultURL, strings.Join(ValidOCMUrlAliases(), ", "))
+	}
+
+	return url.String(), nil
+}
+
 func GetEnv() (string, error) {
```
macgregor, line 85: how is this used in rosa? it probably needs to be made compatible with the `--rh-region` url override. I.e. what happens when `GetEnv()` is called after someone does `rosa login --rh-region=ap-southeast-1` and the URL in config lookes like "https://api.aws.ap-southeast-1.stage.openshift.com" im pretty sure it will return nil/err and whatever command is relying on it will break.

macgregor: in otherwords im saying this function needs to be modified so that cfg.URL=https://api.aws.ap-southeast-1.stage.openshift.com/ returns "stage", etc.

pkg/ocm/config.go
```diff
@@ -39,6 +42,46 @@ var URLAliases = map[string]string{
...
+//
+// Finally, it will try to url.ParseRequestURI the resolved URL to make sure it's a valid URL.
+func ResolveGatewayURL(optionalParsedCliFlagValue string, optionalParsedConfig *config.Config) (string, error) {
+	gatewayURL := sdk.DefaultURL
+	source := "default"
+	if optionalParsedCliFlagValue != "" {
+		gatewayURL = optionalParsedCliFlagValue
+		source = "flag"
+		if _, ok := URLAliases[optionalParsedCliFlagValue]; ok {
+			gatewayURL = URLAliases[optionalParsedCliFlagValue]
```
macgregor, line 67: i noticed in `GetEnv()` theres a `fedramp.URLAliases`? So i guess we need to support that too?

cmd/list/cmd.go
```diff
@@ -69,6 +70,7 @@ func init() {
 	Cmd.AddCommand(tuningconfigs.Cmd)
 	Cmd.AddCommand(oidcprovider.Cmd)
 	Cmd.AddCommand(dnsdomains.Cmd)
+	Cmd.AddCommand(rhRegion.Cmd)
```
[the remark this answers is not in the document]
tirthct: Yes, hidden flag is added in the Cmd itself (cmd/list/rhRegion/cmd.go line 33)

cmd/login/cmd.go
tirthct: I agree with the consistency. However, I'll let @ciaranRoche or @robpblake or the ROSA team decide if this change should be made and if so, how and when to deprecate --env.
In any case, maybe a separate MR if we decide to go through with this?

pkg/ocm/config.go
tirthct: Do we expect this change in FedRAMP as well?
cc: @tylercreller

tylercreller: If resolve gateway URL is going to be a main entry point for any authentication paths, we will need to take into account FedRAMP URLs.

If this is specifically only for region discovery flows, it wouldn't apply.

macgregor: I have not heard plans to deploy regionalized services to fedramp, however i do expect it to happen at some point. So whether we need to worry about rosa's fedramp concept or not is more about not introducing regressions. I would look at how `rosa login --govcloud` works today and think through how our changes to env, the new rh-region flag and behavior change of falling back to cfg.URL when none is provided (simplified precise behavior there for brevity) might interact with it. 

Same with the `GetEnv()` function.

cmd/list/cmd.go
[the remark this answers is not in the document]
macgregor: oh, I overlooked that, thank you :)

cmd/login/cmd.go
macgregor: i suspect it will never change if we dont do it now :smile: and honestly I'd pref to just rip the bandaid off and change it to `--url` here and now. I only offered the `--url`/`--env` as synonymous flags as a compromise position. Not a deal breaker if they want to keep it as `--env`, i would just prefer getting things in sync with ocm cli now while we are here.

ciaranRoche: Right now we do not have a dep policy, so we need to maintain backwards compatibility. 

I agree on the point if its not done now it will never be done. Given env is hidden and only intended for internal use, and we plan to make url open to end users. I would recommend for now following the pattern here - https://github.com/openshift/rosa/pull/1720 to support both `--env` and `--url` and keep url hidden until such time we want to expose this to end users

pkg/ocm/config.go
ciaranRoche: ACK, we need to ensure there is an alias for every url

cmd/list/rhRegion/cmd.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+)
+
+var Cmd = NewRhRegionCommand()
+
+func NewRhRegionCommand() *cobra.Command {
+	Cmd := &cobra.Command{
+		Use:   "rh-regions",
+		Short: "List available OCM regions",
+		Long:  "List available OCM regions",
```
ciaranRoche, line 29: NP for the long can we expand the acronym `OCM` for the most part of ROSA customer OCM is abstracted, so I would like to expand the acronym and even if possible add a doc or flesh out the what/why in the long for this command.

cmd/list/rhRegion/cmd.go
```diff
@@ -0,0 +1,71 @@
...
+	r := rosa.NewRuntime()
+	defer r.Cleanup()
+	cfg, _ := config.Load()
+	gatewayURL, err := ocm.ResolveGatewayURL(args.discoveryURL, cfg)
+	if err != nil {
+		r.Reporter.Errorf("Failed to determine gateway URL: %v", err)
+		os.Exit(1)
+	}
+
+	fmt.Fprintf(os.Stdout, "Discovery URL: %s\n\n", gatewayURL)
```
ciaranRoche, line 58: is there a reason to use fmt.Fprintf over using the reporter?

cmd/login/cmd.go
```diff
@@ -111,7 +112,7 @@ func init() {
...
 		"env",
-		sdk.DefaultURL,
+		"",
```
ciaranRoche, line 115: We can not remove this default, as right now its not needed and we default to prod, we should maintain this not to break backward compatibility

macgregor: its not removed, rather moved to pkg/ocm/config.go ResolveGatewayURL. there is still a default and for backwards compatability purposes its still `sdk.DefaultURL` (slightly oversimplified). We need the default cli arg to be "" so we can tell if they actually supplied the flag or not. 

Analyze `ResolveGatewayURL` if you want to double check me.

pkg/ocm/config.go
macgregor: Just to be clear, region discovery means we cant hard code aliases for every regional url. the point is for us to be able to add regions without requiring clients to update their cli version. However, we can map regions to known env aliases using naming conventions, e.g. https://api.aws.ap-southeast-1.stage.openshift.com is in the "stage" env. We actually already basically do this in the ocm-sdk when we are determining which global gateway url to use for region discovery: https://github.com/openshift-online/ocm-sdk-go/blob/main/rh_region.go#L48

Similar tactic should work in GetEnv()

cmd/list/rhRegion/cmd.go
ciaranRoche: Ideally yes, unless it requires some specific formatting not in the reporter

cmd/login/cmd.go
tirthct: Thanks @ciaranRoche. Added `--url` as an alternative to `--env` as per the suggested MR ref.

pkg/ocm/config.go
tirthct: Added EnvAliases to ensure we are able to discover regionalized environments. However, haven't added production URL (`openshift.com`) yet, since we don't have shards there. Also, comparing and returning **production** env based on just the suffix `openshift.com` does not seem like a concrete check for now (since `openshift.com` is a suffix for **int**(`integration.openshift.com`) and **stage**(`stage.openshift.com`) environments as well). 
We do need to tackle this though. My suggestion is to wait till we define regionalized production URLs and have a unique signature there that we can use here. We might also consider defaulting a production env but that seems a bit risky.
Thoughts @macgregor?

pkg/ocm/config.go
tirthct: If FedRAMP is enabled, we use the env variable there (as opposed to gatewyURL here), so this should not affect FedRAMP. For an edge case where env is not provided (since we made it optional now), I've added the previous implementation (i.e. defaulting to prod env) back. 
As for GetEnv(), FedRAMP aliases are different than normal ones, so that would also work as expected.

cmd/list/rhRegion/cmd.go
tirthct: Added the acronym for OCM. I'm not sure if we already have the docs available for this. @macgregor can you please share any docs for OCM regionalization, if present?

cmd/list/rhRegion/cmd.go
tirthct: Added writer

pkg/arguments/normalize.go
```diff
@@ -25,6 +27,8 @@ func NormalizeFlags(_ *pflag.FlagSet, name string) pflag.NormalizedName {
 		name = NewWorkerIAMRole
 	case DeprecatedControlPlaneIAMRole:
 		name = NewControlPlaneIAMRole
+	case DeprecatedEnvFlag:
```
robpblake, line 30: @tirthct Can you please update the unit tests for the `NormalizeFlags` function to account for your changes. Thanks!

pkg/ocm/config.go
robpblake: Can't we just push this into the `ocm-shards.json` as another field alongside `url` so we have a single source of truth on the environment `name`, `url` and `alias`?

cmd/list/rhRegion/cmd.go
```diff
@@ -0,0 +1,72 @@
...
+		"discovery-url",
+		"",
+		"URL of the OCM API gateway. If not provided, will reuse the URL from the configuration "+
+			"file or "+sdk.DefaultURL+" as a last resort. The value should be a complete URL "+
+			"or a valid URL alias: "+strings.Join(ocm.ValidOCMUrlAliases(), ", "),
+	)
+	return Cmd
+}
+
+func run(cmd *cobra.Command, argv []string) {
```
robpblake, line 49: @tirthct Could we follow the same pattern we used for `get`/`set` config where we have a function that takes in parameters and returns the `error` with a single `os.Exit` call. That makes this function testable. Bonus points for that as well 👍

tirthct:
https://github.com/openshift/rosa/pull/1655#discussion_r1514389843
I think we don't have the shards file finalized yet, so we probably can. But that would require changes in ocm-sdk and ocm-cli.
Thoughts @macgregor ?

pkg/arguments/normalize.go
tirthct: Updated

cmd/list/rhRegion/cmd.go
tirthct: Added this in the separate method. Also added a really small test case to confirm error doesn't exist. Since we are using tabwriter, could not check the output

pkg/ocm/config.go
```diff
@@ -50,12 +98,20 @@ func GetEnv() (string, error) {
...
 
+	// Check for URL aliases
 	for env, api := range urlAliases {
 		if api == strings.TrimSuffix(cfg.URL, "/") {
 			return env, nil
 		}
 	}
 
+	// In case of regionalized URLs, check for env aliases
+	for env, api := range EnvAliases {
```
macgregor, line 109: no. we already solved this when determining what global gateway to query for ocm-shards.json: https://github.com/openshift-online/ocm-sdk-go/blob/main/rh_region.go#L48

Its the same thing here, we just need an "env alias" instead of the gateway url. heres your pseudo code:
```
//
discoveryGatewayUrl, _ := URL.Parse(DetermineRegionDiscoveryUrl(cfg.URL))
for env, api := range EnvAlias {
  if strings.Contains(api, discoveryGatewayUrl.Host){
    return env, nil
  }
}
```

pkg/ocm/config.go
macgregor: @tirthct we dont need to wait on anything, ocm-shards.json exists in prod, its just an empty json structure. ocm-sdk-go and ocm-cli both handle it fine:
```
> ocm list rh-regions                     
Discovery URL: https://api.openshift.com

RH Region    Gateway URL
```

macgregor: > Can't we just push this into the ocm-shards.json as another field alongside url so we have a single source of truth on the environment name, url and alias?

1. rosa is the only thing that cares about the environment name. 
2. to get ocm-shards.json you necessarily need to know which environment to ask for the discovery file:
  * https://api.integration.openshift.com/static/ocm-shards.json
  * https://api.stage.openshift.com/static/ocm-shards.json
  * https://api.openshift.com/static/ocm-shards.json
3. as ive mentioned a few times, we already have a shared way of determining what gateway/discovery file should be used from an arbitrary url: https://github.com/openshift-online/ocm-sdk-go/blob/main/rh_region.go#L48. I am not changing the ocm-shards.json schema right now to add the name of the environment it was fetched from since the client (rosa) already knows it by then. 

So the answer here is to use the shared DetermineRegionDiscoveryUrl from ocm-sdk to go from arbitrary ocm url (global, regional, int, stage, prod, whatever) to global discovery gateway url (i.e. the int, stage, prod ocm urls you know today). Display it or refer to that as "prod" or "api.openshift.com" makes no difference and is trivial to map client side, and what we will do here: https://github.com/openshift/rosa/pull/1655/files#r1516218065

pkg/ocm/config.go
macgregor: hmmm. actually i take that back @tirthct , i think what you have is fine actually. Nice job. Sorry, pre-coffee :)

cmd/login/cmd.go
```diff
@@ -328,6 +342,27 @@ func run(cmd *cobra.Command, argv []string) {
...
+	//   2. fetch a well-known file from sdk.GetRhRegion
+	//   3. update the gatewayURL to the region URL matching args.rhRegion
+	//
+	// So `--env=https://api.stage.openshift.com --rh-region=ap-southeast-1` might result in
+	// gatewayURL/cfg.URL being mutated to "https://api.aws.ap-southeast-1.stage.openshift.com"
+	//
+	// See ocm-sdk-go/rh_region.go for full details on how service discovery works.
+	if args.rhRegion != "" {
+		regValue, err := sdk.GetRhRegion(gatewayURL, args.rhRegion)
+		fmt.Println(gatewayURL, args.rhRegion, regValue)
```
macgregor, line 358: this println feels out of place to me

pkg/ocm/config.go
macgregor: actually i retract my retraction. i really need coffee before i code review. 

We need to use `DetermineRegionDiscoveryUrl` to normalize cfg.URL to the discorvery url to handle the production use case.:
```
cfg.URL=api.aws.ap-southeast-1.stage.openshift.com
urlAliases = {"stage" : "api.stage.openshift.com"}
api.aws.ap-southeast-1.stage.openshift.com => https://api.stage.openshift.com/static/ocm-shards.json => api.stage.openshift.com => stage

cfg.URL=api.aws.ap-southeast-1.openshift.com
urlAliases = {"prod" : "api.openshift.com"}
api.aws.ap-southeast-1.openshift.com => https://api.openshift.com/static/ocm-shards.json => api.openshift.com => prod
```

tirthct: As per the discussion in https://redhat-internal.slack.com/archives/C0421HEGLSW/p1709230320256999, fixed the logic

cmd/login/cmd.go
tirthct: Removed

macgregor:
lgtm

cmd/list/rhRegion/cmd.go
```diff
@@ -0,0 +1,83 @@
...
+	args   struct {
+		discoveryURL string
+	}
+)
+
+var Cmd = NewRhRegionCommand()
+
+func NewRhRegionCommand() *cobra.Command {
+	Cmd := &cobra.Command{
+		Use:   "rh-regions",
```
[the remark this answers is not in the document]
robpblake: After speaking with @macgregor , this is being used elsewhere and has been agreed as the preferred terminology currently.

cmd/list/rhRegion/cmd.go
macgregor: https://docs.google.com/document/d/1WqGRdkeSWd4WEe6S2JubgrA9ESJd0n8RgSrnY9OhxEY/edit#heading=h.bupciudrwmna

robpblake (approved):

pkg/ocm/config.go
```diff
@@ -39,6 +42,46 @@ var URLAliases = map[string]string{
...
+}
+
+// URL Precedent (from highest priority to lowest priority):
+//  1. runtime `--env` cli arg (key found in `urlAliases`)
+//  2. runtime `--env` cli arg (non-empty string)
+//  3. config file `URL` value (non-empty string)
+//  4. sdk.DefaultURL
+//
+// Finally, it will try to url.ParseRequestURI the resolved URL to make sure it's a valid URL.
+func ResolveGatewayURL(optionalParsedCliFlagValue string, optionalParsedConfig *config.Config) (string, error) {
```
hunterkepley, line 60: This is a nitpick, but generally in CS/Rosa, we now use CamelCase for acronyms. So this would be `ResolveGatewayUrl`

pkg/ocm/config.go
```diff
@@ -50,9 +93,17 @@ func GetEnv() (string, error) {
...
 	}
 
-	for env, api := range urlAliases {
-		if api == strings.TrimSuffix(cfg.URL, "/") {
-			return env, nil
+	// Check for OCM environments (including regionalized URLs)
+	if strings.HasSuffix(strings.TrimSuffix(cfg.URL, "/"), "openshift.com") {
+		regionDisvocveryURL, err := sdk.DetermineRegionDiscoveryUrl(cfg.URL)
+		if err == nil {
+			discoveryGatewayUrl, _ := url.Parse(regionDisvocveryURL)
```
hunterkepley, line 100: `regionDiscoveryUrl` (typo - casing change optional)

cmd/login/cmd.go
```diff
@@ -296,11 +303,18 @@ func run(cmd *cobra.Command, argv []string) {
...
+		r.Reporter.Errorf("Failed to resolve gateway URL: %v", err)
+		os.Exit(1)
 	}
+
+	// Make sure env is present in case of fedRAMP
+	if env == "" {
+		env = args.env
+	}
+
+	var ok bool
```
hunterkepley, line 317: Could you explain this to me? Fedramp has the same environments so i'm not sure of the reason we need this logic specifically for fedramp

hunterkepley: For example, line 247 in this command we use the env to get the right `LoginURL` for fedramp

hunterkepley (commented):
overall, LGTM, just a couple nitpicks/one question

hunterkepley:
Not sure why approval was removed by @robpblake - hopefully he can let us know when he's back online

robpblake:
> Not sure why approval was removed by @robpblake - hopefully he can let us know when he's back online

@hunterkepley I just wanted another member of the team to add their review. My +1 set it to auto-merge, so I removed approval to prevent that. 

cheers,

Rob

cmd/login/cmd.go
tirthct: Hmm, right. Added check in the fedramp condition on line 247 itself.
We still need an empty check here because env would not be mandatory with this change and is defaulted to empty string

cmd/list/rhRegion/cmd.go
```diff
@@ -0,0 +1,72 @@
...
+	fmt.Fprintf(writer, "Discovery URL: %s\n\n", gatewayURL)
+	regions, err := sdk.GetRhRegions(gatewayURL)
+	if err != nil {
+		r.Reporter.Errorf("Failed to get OCM regions: %v", err)
+		os.Exit(1)
+	}
+
+	fmt.Fprintf(writer, "RH Region\t\tGateway URL\n")
+	for regionName, region := range regions {
+		fmt.Fprintf(writer, "%s\t\t%v\n", regionName, region.URL)
```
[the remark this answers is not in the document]
tirthct: Added a warning check for empty regions list

pkg/ocm/config.go
```diff
@@ -50,9 +93,17 @@ func GetEnv() (string, error) {
 		urlAliases = fedramp.URLAliases
 	}
 
-	for env, api := range urlAliases {
-		if api == strings.TrimSuffix(cfg.URL, "/") {
-			return env, nil
+	// Check for OCM environments (including regionalized URLs)
+	if strings.HasSuffix(strings.TrimSuffix(cfg.URL, "/"), "openshift.com") {
```
gdbranco, line 97: This does not work with localhost urls

@tirthct

tirthct: Hmm, seems like a bug. Created [OCM-6926](https://issues.redhat.com/browse/OCM-6926) and starting to fix the same.

PR #1656, opened 2023-12-11
OCM-5231 | fix: use spaces instead of \t

gdbranco:
Related issue: [OCM-5231](https://issues.redhat.com//browse/OCM-5231)


gdbranco:
Taking some liberty to merge to unblock QE tracks as there are no functionality changes

PR #1657, opened 2023-12-11
OCM-3570 | feat: Add FedRAMP Prod Sector Env aliases

hunterkepley:
We need to update ROSA CLI FedRAMP environment aliases to ensure a consistent experience with the FedRAMP Keycloak go-live.


pkg/fedramp/config.go
```diff
@@ -77,7 +77,7 @@ const keycloakURL = "realms/redhat-external/protocol/openid-connect/token"
 
 // TokenURLs allows the value of the `--env` option to map to the various AWS Cognito token URLs.
 var TokenURLs = map[string]string{
-	"production":  fmt.Sprintf("https://ocm-ra-production-domain.%s", cognitoURL),
+	"production":  fmt.Sprintf("https://sso.stage.openshiftusgov.com/%s", cognitoURL),
```
tonytheleg, line 80: the production url here is sso.openshiftusgov.com for prod and this would be keycloak instead of cognito

`"production":    fmt.Sprintf("https://sso.openshiftusgov.com/%s", keycloakURL),`

tylercreller:
LGTM

PR #1659, opened 2023-12-14
OCM-5621 | feat: Add codecov to CI pipeline for ROSA

robpblake:
This PR is the first step for adding CodeCov coverage checking to the ROSA CI pipeline. In this PR I'm simply introducing a `Makefile` target and a bash script to upload results to CodeCov. The bash script has been copied from other projects uploading to CodeCov within the `openshift` org, and then slightly adapated to match our project.

Next steps will be to update our `prow` configuration to run this on each PR and then on merges to `master` so that we have a record of our coverage statistics over time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ test:
 
 .PHONY: coverage
 coverage:
-	go test -coverprofile=cover.out  ./...
+	go test -coverprofile=cover.out -covermode=atomic -p 4 ./...
 
 .PHONY: install
 install:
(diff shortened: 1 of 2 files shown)


davidleerh:
Aside from the move of the command to create the temporary coverage file from within the shell script to the makefile,
confirming script is identical to coverage upload scripts used in other openshift projects:
ex - https://github.com/openshift/hive/blob/e5f809fca22b9629c325c2ef88d384764645bf89/hack/codecov.sh#L8

robpblake:
@davidleerh I wanted us to be able to run coverage checks independently of codecov, so I kept that part in the `Makefile` as a separate target 👍

PR #1660, opened 2023-12-15
OCM-5313 | fix: chunk slice before sending request to AWS

gdbranco:
Related issue: [OCM-5313](https://issues.redhat.com//browse/OCM-5313)

### DIFF
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -75,6 +75,8 @@ const (
 	Attached      = "attached"
 
 	govPartition = "aws-us-gov"
+
+	awsMaxFilterLength = 200
 )
 
 // addROSAVersionToUserAgent is a named handler that will add ROSA CLI
@@ -195,7 +197,7 @@ type Client interface {
 	GetDefaultPolicyDocument(policyArn string) (string, error)
 	GetAccountRoleByArn(roleArn string) (*Role, error)
 	GetSecurityGroupIds(vpcId string) ([]*ec2.SecurityGroup, error)
-	FetchPublicSubnetMap(subnet []*ec2.Subnet) (map[string]bool, error)
+	FetchPublicSubnetMap(subnets []*ec2.Subnet) (map[string]bool, error)
 }
 
 // ClientBuilder contains the information and logic needed to build a new AWS client.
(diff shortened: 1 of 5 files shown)


pkg/helper/helpers.go
```diff
@@ -241,3 +241,17 @@ func KeysByValue(m map[string]string, value string) []string {
 	}
 	return keys
 }
+
+func ChunkSlice[T any](slice []T, chunkSize int) [][]T {
```
robpblake, line 245: @gdbranco Quick unit test for this function would be good 👍

pkg/aws/client.go
```diff
@@ -441,42 +443,45 @@ func (c *awsClient) FetchPublicSubnetMap(subnets []*ec2.Subnet) (map[string]bool
...
-			for _, route := range routes.Routes {
-				if strings.HasPrefix(aws.StringValue(route.GatewayId), "igw") {
-					// There is no direct way in the AWS API to determine if a subnet is public or private.
-					// A public subnet is one which has an internet gateway route
-					// we look for the gatewayId and make sure it has the prefix of igw to differentiate
-					// from the default in-subnet route which is called "local"
-					// or other virtual gateway (starting with vgv)
-					// or vpc peering connections (starting with pcx).
-					mapSubnetIdToPublic[subnetAssociation] = true
+	for _, curChunk := range chunks {
```
robpblake, line 449: @gdbranco Same comment here re: unit tests. Looks like it should be doable in the `client_test.go` as we're already using mocks there.

pkg/aws/client.go
```diff
@@ -441,42 +443,45 @@ func (c *awsClient) FetchPublicSubnetMap(subnets []*ec2.Subnet) (map[string]bool
...
+				{
+					Name:   aws.String("association.subnet-id"),
+					Values: subnetIds,
+				},
+			},
+		})
+		if err != nil {
+			return mapSubnetIdToPublic, err
+		}
+		if routeTablesResp == nil {
```
oriAdler, line 465: Should we continue to the next chunk in this case?

gdbranco: I've thought about it, I'm not so sure. If we skip it might be confusing for the customer too, I'd rather we merge as is and check on user behavior :thinking:

pkg/aws/client.go
```diff
@@ -441,42 +443,45 @@ func (c *awsClient) FetchPublicSubnetMap(subnets []*ec2.Subnet) (map[string]bool
...
+			return mapSubnetIdToPublic, fmt.Errorf(
+				"No route table found for associated subnets '%s'",
```
robpblake, line 467: @gdbranco Can we give the customer a bit more help to resolve things? From a technical perspective the error message is correct, but how does the customer proceed from here?

pkg/aws/client.go
robpblake: I think if we can help the user with a slightly more informative error message that would be good. I kinda feel like most customers are going to react with "And what do I do now?" when they see a message about there being no route table associated with a subnet. 

Is this a problem they can fix with ROSA CLI or something they need to talk to their AWS account owner about for example.

oriAdler: I checked and it seems like " Each subnet in your VPC must be associated with a route table, which controls the routing for the subnet (subnet route table)" so I think it's ok to return an error here after a second thought. https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Route_Tables.html

oriAdler:
LGTM

PR #1663, opened 2023-12-20
OCM-4646 | fix: better messages for options prompts

gdbranco:
Related issue: [OCM-4646](https://issues.redhat.com//browse/OCM-4646)

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -1179,6 +1179,7 @@ func run(cmd *cobra.Command, _ []string) {
 			Help:     cmd.Flags().Lookup("version").Usage,
 			Options:  versionList,
 			Default:  version,
+			Required: true,
 		})
 		if err != nil {
 			r.Reporter.Errorf("Expected a valid OpenShift version: %s", err)
@@ -2848,7 +2849,7 @@ func run(cmd *cobra.Command, _ []string) {
 			routeSelector = args.defaultIngressRouteSelectors
 		} else if interactive.Enabled() && !isHostedCP && shouldAskCustomIngress {
 			routeSelectorArg, err := interactive.GetString(interactive.Input{
-				Question: "Route Selector for ingress (e.g. 'route=external')",
+				Question: "Router Ingress Sharding: Route Selector (e.g. 'route=external')",
 				Help:     cmd.Flags().Lookup(ingress.DefaultIngressRouteSelectorFlag).Usage,
 				Default:  args.defaultIngressRouteSelectors,
 				Validators: []interactive.Validator{
@@ -2878,7 +2879,7 @@ func run(cmd *cobra.Command, _ []string) {
 			excludedNamespaces = args.defaultIngressExcludedNamespaces
 		} else if interactive.Enabled() && !isHostedCP && shouldAskCustomIngress {
 			excludedNamespacesArg, err := interactive.GetString(interactive.Input{
-				Question: "Excluded namespaces for ingress",
+				Question: "Router Ingress Sharding: Namespace exclusion",
 				Help:     cmd.Flags().Lookup(ingress.DefaultIngressExcludedNamespacesFlag).Usage,
 				Default:  args.defaultIngressExcludedNamespaces,
 			})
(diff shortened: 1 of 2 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -1179,6 +1179,7 @@ func run(cmd *cobra.Command, _ []string) {
 			Help:     cmd.Flags().Lookup("version").Usage,
 			Options:  versionList,
 			Default:  version,
+			Required: true,
```
oriAdler, line 1182: Does this change related to OCM-4646?

cmd/create/cluster/cmd.go
```diff
@@ -1198,21 +1199,23 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	httpTokens := args.ec2MetadataHttpTokens
+	if httpTokens == "" {
```
oriAdler, line 1202: Does this change is needed for OCM-4646?

oriAdler (commented):
Ingress changes LGTM, do you have a ticket for the other changes in this PR?

cmd/create/cluster/cmd.go
gdbranco: Not directly, we could add something to UX improvements I suppose. The current prompt is a string which allows to enter anything and then we validate. I just don't think it makes sense when we have clear options for it

cmd/create/cluster/cmd.go
gdbranco: This one is required due to the changes to how the options prompt works. So that we may not skip the version selection as it was before

cmd/create/cluster/cmd.go
gdbranco: Separated into a different PR https://github.com/openshift/rosa/pull/1664

pkg/interactive/interactive.go
```diff
@@ -177,12 +177,30 @@ func GetOption(input Input) (a string, err error) {
...
-		dflt = consts.SkipSelectionOption
+		if dflt == "" {
+			dflt = consts.SkipSelectionOption
+		} else {
+			optionalMessage += ". The default value will be provided"
+		}
+	}
+	if optionalMessage != "" || defaultMessage != "" {
+		question += " ("
+		question = fmt.Sprintf("%s%s", question, optionalMessage)
```
mnecas, line 197: To the question is added `" ("` and right after that overwritten with the question and optional message. Does not seem right

pkg/interactive/interactive.go
```diff
@@ -177,12 +177,30 @@ func GetOption(input Input) (a string, err error) {
...
+		separator := ""
+		if optionalMessage != "" {
+			separator = "; "
+		}
+		question = fmt.Sprintf("%s%s%s", question, separator, defaultMessage)
```
mnecas, line 202: And then overwritten again

pkg/interactive/interactive.go
```diff
@@ -177,12 +177,30 @@ func GetOption(input Input) (a string, err error) {
...
+		question += ")"
```
mnecas, line 203: Could be part of `Sprintf` above

pkg/interactive/interactive.go
```diff
@@ -177,12 +177,30 @@ func GetOption(input Input) (a string, err error) {
...
-			"The default value will be supplied.)", question, consts.SkipSelectionOption)
+	optionalMessage := ""
+	if !input.Required {
+		optionalMessage = fmt.Sprintf("optional, choose '%s' to skip selection", consts.SkipSelectionOption)
 		input.Options = append([]string{consts.SkipSelectionOption}, input.Options...)
-		dflt = consts.SkipSelectionOption
+		if dflt == "" {
+			dflt = consts.SkipSelectionOption
+		} else {
+			optionalMessage += ". The default value will be provided"
```
mnecas, line 192: Why is the dot in the beginning of the string, shouldn't it be moved to above?

pkg/interactive/interactive.go
gdbranco: It does not override, it appends to the original variable

pkg/interactive/interactive.go
gdbranco: It does not override

pkg/interactive/interactive.go
gdbranco: Can't really argue here, it could be

pkg/interactive/interactive.go
gdbranco: It is below as I do not want to have a final dot if we do not have more to add :)

oriAdler:
LGTM

pkg/interactive/interactive.go
mnecas: ahh, the combination of using `Sprintf` and `+=` confused me

mnecas:
ahh, the combination of using Sprintf and += confused me, but yeah looks better now, thanks!

PR #1668, opened 2023-12-21
OCM-5358 | fix: add a warning message when using best effort deletion

oriAdler:
Notify the user about cloud resources that may left behind when using the 'best effort' flag.

### DIFF
--- a/cmd/dlt/cluster/cmd.go
+++ b/cmd/dlt/cluster/cmd.go
@@ -79,6 +79,11 @@ func run(cmd *cobra.Command, _ []string) {
 
 	clusterKey := r.GetClusterKey()
 
+	if args.bestEffort {
+		r.Reporter.Warnf("Deleting cluster '%s' with 'best effort' means that certain resources may be left behind"+
+			" in AWS account '%s'. These resources will need to be deleted manually.", clusterKey, r.Creator.AccountID)
+	}
+
 	if !confirm.Confirm("delete cluster %s", clusterKey) {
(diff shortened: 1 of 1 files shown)


cmd/dlt/cluster/cmd.go
```diff
@@ -85,6 +85,11 @@ func run(cmd *cobra.Command, _ []string) {
 
 	cluster := r.FetchCluster()
 
+	if args.bestEffort {
+		r.Reporter.Warnf("Deleting a cluster with 'best effort' means that certain resources may be left behind " +
```
gdbranco, line 89: Maybe a prompt might be more effective :thinking:

robpblake, line 89: @oriAdler Wondering if we should include the clusterKey and AWS account details in this message so something like:

```
r.Reporter.Warnf("Deleting cluster '%s' with 'best effort' means that certain resources may be left behind in AWS account '%s'. These resources will need to be deleted manually", r.GetClusterKey(),r.Creator.AccountID)
 ```

den-rgb: +1

oriAdler: Good point, added, thanks.

oriAdler: Makes sense, since we already have a prompt `delete cluster <cluster_key>?` I swapped the order, so the warning is displayed before the confirmation, WDYT?

PR #1670, opened 2023-12-21
OCM-4783 | feat: display warnings after cluster creation

renan-campos:


pkg/rosa/runtime.go
```diff
@@ -97,3 +98,17 @@ func (r *Runtime) FetchCluster() *cmv1.Cluster {
 	r.Cluster = cluster
 	return cluster
 }
+
+func (r *Runtime) DisplayClusterWarnings(clusterId string) error {
+	serviceLogs, err := r.OCMClient.ListServiceLogs(clusterId)
+	if err != nil {
+		return err
+	}
+	serviceLogs.Items().Each(func(entry *slv1.LogEntry) bool {
```
robpblake, line 107: @renan-campos I think we need to provide the user with more context here on what these warnings are, and what they need to do about them (if anything). 

For example, I'd expect _something_ like:

```
The following warnings were generated on creation of your cluster:

- Lack of Jam
 - There was a serious lack of Jam when creating your cluster. Buy some more jam (see example.com/buy-jam)

```

renan-campos: Hi @robpblake the user context would be provided by the serviceLog's summary and description. 

For a little background on this PR: This warning mechanism was needed for a preflight displayed by the OCM cli. In the OCM cli, this would be the way that warning would be displayed:
```
⚠️ WARNING:
Please enable the Org Policy API for the GCP project 'REDACTED'
Without the GCP Org Policy API enabled, OCM is unable to determine whether the GCP project contains any policies that would affect the in
stallation
```

renan-campos: @robpblake happy new year! With fresh eyes I had a better understanding of your comment. I added context to the warnings as requested. PTAL

pkg/rosa/runtime.go
```diff
@@ -97,3 +98,20 @@ func (r *Runtime) FetchCluster() *cmv1.Cluster {
 	r.Cluster = cluster
 	return cluster
 }
+
+func (r *Runtime) DisplayClusterWarnings(clusterId string) error {
```
robpblake, line 102: @renan-campos Why did you decide to put this function on `Runtime`?

pkg/ocm/service_logs.go
```diff
@@ -0,0 +1,9 @@
+package ocm
+
+import (
+	slv1 "github.com/openshift-online/ocm-sdk-go/servicelogs/v1"
+)
+
+func (c *Client) ListServiceLogs(clusterId string) (*slv1.ClustersClusterLogsListResponse, error) {
```
robpblake, line 7: @renan-campos I think it's worth us making this function take a variadic parameter of `SeverityWarning` that it should search for so that we can make the backend do the work of filtering out events and avoid fetching unnecessary data as at the minute we just pull all events regardless and then iterate over them locally.

So it would be something like:

```
func (c *Client) ListServiceLogs(clusterId string, matchSeverities... ) (*slv1.ClustersClusterLogsListResponse, error)
```

If the user supplies anything, then we use the `Search()` function and only fetch the events matching the given severity. Otherwise, we fetch all events.

pkg/ocm/service_logs.go
```diff
@@ -0,0 +1,9 @@
+package ocm
+
+import (
+	slv1 "github.com/openshift-online/ocm-sdk-go/servicelogs/v1"
+)
+
```
robpblake, line 6: @renan-campos I'd also like to see a unit test for this as I'm trying to get us to the point where we're not adding new code that isn't tested. Take a look at `kubeletconfig_test` in the `ocm` package for how we're currently testing interactions with the OCM API.

renan-campos:
Shift in priorities makes this pr no longer needed at this time.

PR #1672, opened 2023-12-26
OCM-5060 | fix:Validate min replicas for hosted cp clusters

oriAdler:
When creating a hosted CP cluster, the user must set the nodes/replicas to a minimal value of 2.
Improve the UX and provide feedback to the client.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -3348,6 +3348,11 @@ func minReplicaValidator(multiAZ bool, isHostedCP bool, privateSubnetsCount int)
 			return err
 		}
 
+		if isHostedCP && minReplicas < 2 {
+			return fmt.Errorf("hosted Control Plane clusters require a minimum of 2 nodes, "+
+				"but %d was requested", minReplicas)
+		}
+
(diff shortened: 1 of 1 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -3348,6 +3348,11 @@ func minReplicaValidator(multiAZ bool, isHostedCP bool, privateSubnetsCount int)
 			return err
 		}
 
+		if isHostedCP && minReplicas < 2 {
+			return fmt.Errorf("hosted Control Plane clusters require a minimum of 2 nodes, "+
```
gdbranco, line 3352: The minimum of nodes varies from single to multi AZ, and the min replicas variable usually holds the value for that. Should this be validating what's supplied in args.computeNodes instead?

oriAdler: While it applies for Hive, the MinReplicaValidator doesn't check multiAZ minimal number of replicas for hosted CP https://github.com/openshift-online/ocm-common/blob/main/pkg/cluster/validations/cluster_node.go#L22.
We could validate args.computeNodes, but then it will be possible to supply 0 replicas in the interactive mode resulting in the same bug.

gdbranco: I see, thanks!

PR #1680, opened 2024-01-04
OCM-1405 | feat: include 'Display Name' in cluster description

thomasmckay:


cmd/describe/cluster/cmd.go
```diff
@@ -70,6 +70,12 @@ func run(cmd *cobra.Command, argv []string) {
 	cluster := r.FetchCluster()
 	isHypershift := cluster.Hypershift().Enabled()
 
+	displayName := ""
+	subscription, subscriptionExists, _ := r.OCMClient.GetSubscriptionByID(cluster.ID())
```
ciaranRoche, line 74: Would it be worth handling the error, even just in debug mode? Not worth exiting if there is an error on this, but at least we can be sure to print some logs in debug mode.

pkg/ocm/clusters.go
```diff
@@ -392,6 +392,23 @@ func (c *Client) getSubscriptionByExternalID(externalID string) (*amv1.Subscript
 	return response.Items().Slice()[0], true, nil
 }
 
+func (c *Client) GetSubscriptionByID(id string) (*amv1.Subscription, bool, error) {
+	query := fmt.Sprintf("cluster_id = '%s'", id)
+	response, err := c.ocm.AccountsMgmt().V1().Subscriptions().List().
```
ciaranRoche, line 397: Should we use get :thinking: as we have the subscription id in the cluster spec we can pass it in here? WDYT?

cmd/describe/cluster/cmd.go
cben: --debug will log the request & response anyway, no?

cmd/describe/cluster/cmd.go
```diff
@@ -676,12 +684,14 @@ func formatCluster(cluster *cmv1.Cluster, scheduledUpgrade *cmv1.UpgradePolicy,
 		upgrade["nextRun"] = scheduledUpgrade.NextRun().Format("2006-01-02 15:04 MST")
 		ret["scheduledUpgrade"] = upgrade
 	}
+	ret["displayName"] = displayName
```
oriAdler, line 687: Do we need to check if `displayName != ""`?

thomasmckay: I considered that but would prefer to show it as blank if it is

thomasmckay: This is the UI logic. I think we should do the same. Thoughts?
```
const getClusterName = (cluster: ClusterFromSubscription): string => {
  if (!cluster) {
    return '';
  }

  let clusterName = cluster.subscription?.display_name;
  if (!clusterName || clusterName === cluster.subscription?.external_cluster_id) {
    clusterName = cluster.name || cluster.external_id;
  }

  if (clusterName === undefined) {
    if (cluster.subscription?.status === 'Deprovisioned') {
      const subscriptionId = cluster.subscription?.id;
      if (subscriptionId) {
        return subscriptionId;
      }
    }
    return 'Unnamed Cluster';
  }
  return clusterName;
};
```

cben: As I mentioned on internal Slack, a lot of this is historic baggage that may be irrelevant; we just never had incentives to simplify it.  And it's so complex I'd hate to enshrine it as the official way one should consume the API :-1:.

For example the `display_name === subscription?.external_cluster_id` case falling back to `cluster.name || cluster.external_id` — I've been on the cluster_mgmt backend + UI team from the very start yet I can't explain in what scenario that is supposed to help... :laughing:  
Normal OCP clusters have cluster.name === cluster.display_name === cluster.external_id === subscription.display_name === subscription.external_cluster_id, so you'll get that same UUID anyway (until user edits display_name).

But here the decision is simpler: You're going to show the fields separately:
```
		"Name:                       %s\n"+
		"Display Name:               %s\n"+
                ...
		"External ID:                %s\n"+
```
so if display_name is absent/blank, IMHO the appropriate thing is just reflect API reality and show it empty.  Let user figure out if other fields contain relevant info but don't lie to them that it has a display_name when it doesn't.

----

I do see some OCP & OCP-AssistedInstall subscriptions in prod with `display_name = '' OR display_name IS NULL`, and I'm gonna reply here with more details soon.  
But none for ROSA — all MOA & MOA-HostedControlPlane clusters have a non-empty display_name :heavy_check_mark:

oriAdler: I tend to agree with showing a blank display name if not available, to keep the logic in the CLI simple.

pkg/ocm/clusters.go
thomasmckay: done!

oriAdler:
Other than this comment https://github.com/openshift/rosa/pull/1680#discussion_r1442825329, LGTM

cmd/describe/cluster/cmd.go
cben: [just following up as promised, but nothing here is relevant to `rosa`]

So I didn't find a single clear pattern to OCP & OCP-AssistedInstall subscriptions with null blank (67 cases) display_name; many Archived & old but not all, various provenance, various states... some recent, including 1 Active now.  
8 even had non-UUID _cluster.display_name_!  But those were all old <=2021.

What I did find is API allows users to set a blank display_name:
```
echo '{"display_name": ""}' | ocm patch sub 2aiKK76moYZywHSpfwV4m0BzvEZ
```
upon which the field disappears from the API (omitempty), until you set it non-blank again.  This could be what happened to the 1 Active?

It can't explain the IS NULL ones.

* Well 334 of the 335 IS NULL ones are all OCP that also lack subscription.status (match `status = ''`), which is extra weird :question:.  
  The one exception (OCP-AssistedInstall, Reserved) seems to have been some manual test (has "XXX" at end of cluster_id and external_cluster_id) so can be ignored.

cmd/describe/cluster/cmd.go
```diff
@@ -703,6 +713,7 @@ func formatClusterHypershift(cluster *cmv1.Cluster,
...
```
cben, line 716: This (and L687 above) are for --output=json and --output=yaml, right?

That output contains tons of snake_case field names, and very few lowerCamelCase names:
```yaml
...
delete_protection:
  enabled: false
disable_user_workload_monitoring: true
displayName: dt-hcp-test-4  ⬅️
dns:
  base_domain: jw4b.p3.openshiftapps.com
etcd_encryption: false
...
```
Is it desirable to mix naming conventions, and to rename a field that already exists in API as `display_name`? :thinking:  Or is that idea to set apart all fields that are not part of _same_ API?

your call, just asking.  LGTM either way

cben:
LGTM :100:

cmd/describe/cluster/cmd.go
thomasmckay: moved to snake_case, good call

cmd/describe/cluster/cmd.go
thomasmckay: added debug output, thanks!

cmd/describe/cluster/cmd.go
```diff
@@ -70,6 +70,15 @@ func run(cmd *cobra.Command, argv []string) {
 	cluster := r.FetchCluster()
 	isHypershift := cluster.Hypershift().Enabled()
 
+	displayName := ""
+	subscription, subscriptionExists, err := r.OCMClient.GetSubscriptionBySubscriptionID(cluster.Subscription().ID())
+	if err != nil {
+		r.Reporter.Debugf("Failed to get subscription by ID: %s", err)
+	}
+	if subscriptionExists {
+		displayName = subscription.DisplayName()
```
gdbranco, line 79: I don't mind doing this in the CLIs/UI. But if all clients will eventually be expected to do so, isn't it easier to retrieve as part of the ocm cluster spec in the response?

cben: - Would that require an extra CS->AMS call overhead?
  Especially when you **list** clusters in CS, does it already ask AMS for all matching subscriptions?

  fwiw, UI talks to both services anyway, for many fields.  (It can only talk to AMS for Deprovisioned clusters.). So it'd only help the CLI, UI wouldn't gain, and if it costs CS->AMS call(s), the system could become slower.

- If you add a field to CS api, ideally it should support `search` by it (very hard if source of truth remains AMS),
- and to patch it (doable).

These are not mandatory of course, but if you don't then was it worth it?  It'd remain simpler to use the more complete AMS API for all needs.

----

I think many of these assumptions are gonna be re-evaluated for XCM effort; until then this PR is good step imho.

thomasmckay: You mean have CS retrieve from AMS? I think @cben has context about the history of this idea.

ciaranRoche: Going to resolve this for now, to merge this in. I would suggest having this conversation on a DDR to capture how to handle this for all clients. As this has and will be an on going topic of conversation. Would be good to have it documents and agreed too :crossed_fingers:

ciaranRoche (approved):
/lgtm

PR #1687, opened 2024-01-08
OCM-3178 | fix: handle bad disk size input

thomasmckay:


pkg/ocm/helpers.go
```diff
@@ -67,6 +67,8 @@ const (
 
 	HcpProduct        = "hcp"
 	HcpBillingAccount = "hcp-billing"
+
+	maxPositiveInt = 1<<63 - 1
```
robpblake, line 71: @thomasmckay What's the purpose of this definition and why is it set to this number?

pkg/ocm/helpers.go
```diff
@@ -937,11 +939,19 @@ func ParseDiskSizeToGigibyte(size string) (int, error) {
 		return 0, fmt.Errorf("invalid disk size: %w", err)
 	}
 
+	if qty.Value() < 0 {
+		return 0, fmt.Errorf("invalid disk size: %s. positive size required", size)
```
robpblake, line 943: @thomasmckay I think we should add `'` around the placeholders in the string formats in this function to bring is into alignment with other functions. It also helps to catch whitespace problems. So ^ would become:

```
return 0, fmt.Errorf("invalid disk size: '%s'. positive size required", size)
```

pkg/ocm/helpers.go
thomasmckay: This is the largest int. When the size specified overflows in the k8s.io package, this is the value returned. Unfortunately the package doesn't throw an error in this case.

pkg/ocm/helpers.go
thomasmckay: I'd be +1 on that but patterned this after the format just a few lines up. What do you think?

pkg/ocm/helpers.go
robpblake: Can we use `math.MaxInt` instead and I'd personally find your explanation useful as a comment if I was looking at this code?

pkg/ocm/helpers.go
robpblake: I'd say let's make the other format line also use `'` around the placeholders, or does it look weird?

pkg/ocm/helpers.go
thomasmckay: I should have looked harder for that const, thanks!

pkg/ocm/helpers.go
thomasmckay: done, thanks!

pkg/ocm/helpers.go
```diff
@@ -937,11 +937,20 @@ func ParseDiskSizeToGigibyte(size string) (int, error) {
 		return 0, fmt.Errorf("invalid disk size: %w", err)
 	}
```
robpblake, line 938: @thomasmckay Can you also add the `'` to these as well please so that we're consistent throughout the function?

cheers,

Rob

thomasmckay: I think I got them all now

ciaranRoche (approved):
/lgtm

PR #1689, opened 2024-01-09
OCM-4178 | fix: ROSA CLI N/W Verifier Error Message for IPI VPC Clusters

willkutler:
`rosa network verify` on a non-BYO VPC cluster will error with "no subnets on cluster xyz" which could be a confusing output. This PR adds a check for BYO VPC and a corresponding error, as well as moving a "IsBYOVPC" function to a common file to avoid code duplication and improve readability. There is also a unit test to verify the fix.

https://issues.redhat.com/browse/OCM-4178

### DIFF
--- a/cmd/create/machinepool/machinepool.go
+++ b/cmd/create/machinepool/machinepool.go
@@ -43,7 +43,7 @@ func addMachinePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster
 
 	// Validate flags that are only allowed for BYOVPC cluster
 	isSubnetSet := cmd.Flags().Changed("subnet")
-	isByoVpc := isBYOVPC(cluster)
+	isByoVpc := helper.IsBYOVPC(cluster)
 	if !isByoVpc && isSubnetSet {
 		r.Reporter.Errorf("Setting the `subnet` flag is only allowed for BYO VPC clusters")
 		os.Exit(1)
@@ -129,7 +129,7 @@ func addMachinePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster
 
 	// Allow the user to select subnet for a single AZ BYOVPC cluster
 	var subnet string
-	if !cluster.MultiAZ() && isBYOVPC(cluster) {
+	if !cluster.MultiAZ() && isByoVpc {
 		subnet = getSubnetFromUser(cmd, r, isSubnetSet, cluster)
 	}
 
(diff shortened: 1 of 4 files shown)


cmd/verify/network/cmd.go
```diff
@@ -155,10 +156,14 @@ func runWithRuntime(r *rosa.Runtime, cmd *cobra.Command) error {
 
 	if !cmd.Flags().Changed(subnetIDsFlag) {
 		if cluster != nil {
-			args.subnetIDs = cluster.AWS().SubnetIDs()
-			if len(args.subnetIDs) == 0 {
-				return fmt.Errorf("No subnets on cluster '%s'", cluster.ID())
+			isByoVpc := helper.IsBYOVPC(cluster)
+
+			if !isByoVpc {
```
oriAdler, line 161:
```suggestion
			if !helper.IsBYOVPC(cluster) {
```

cmd/create/machinepool/machinepool.go
```diff
@@ -129,7 +129,7 @@ func addMachinePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster
 
 	// Allow the user to select subnet for a single AZ BYOVPC cluster
 	var subnet string
-	if !cluster.MultiAZ() && isBYOVPC(cluster) {
+	if !cluster.MultiAZ() && helper.IsBYOVPC(cluster) {
```
gdbranco, line 132:
```suggestion
	if !cluster.MultiAZ() &&  isByoVpc {
```

cmd/create/machinepool/machinepool.go
```diff
@@ -160,7 +160,7 @@ func addMachinePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster
 
 		if !multiAZMachinePool {
 			// Allow to create a single AZ machine pool providing the subnet
-			if isBYOVPC(cluster) && args.availabilityZone == "" {
+			if helper.IsBYOVPC(cluster) && args.availabilityZone == "" {
```
gdbranco, line 163:
```suggestion
			if isByoVpc && args.availabilityZone == "" {
```

cmd/verify/network/cmd.go
```diff
@@ -155,10 +156,12 @@ func runWithRuntime(r *rosa.Runtime, cmd *cobra.Command) error {
 
 	if !cmd.Flags().Changed(subnetIDsFlag) {
 		if cluster != nil {
-			args.subnetIDs = cluster.AWS().SubnetIDs()
-			if len(args.subnetIDs) == 0 {
-				return fmt.Errorf("No subnets on cluster '%s'", cluster.ID())
+			if !helper.IsBYOVPC(cluster) {
+				return fmt.Errorf(
+					"Running the network verifier is only supported for clusters created on existing VPCs")
```
gdbranco, line 161:
```suggestion
					"Running the network verifier is only supported for clusters created on pre existing VPCs prior to cluster creation")
```

Just cause the VPCs do exist for non BYO as well, they just didn't pre existing prior to cluster creation

gdbranco: Or just mention only supported for BYO VPC clusters, might be clearer :)

PR #1691, opened 2024-01-10
OCM-5542 | feat: add `--client` argument to `rosa version` command

jfchevrette:
https://issues.redhat.com/browse/OCM-5542

In some restricted environments in might not be desirable to have the version command do a network call

This adds a `--client` argument to `rosa version`. The naming of the argument align with other tools like kubectl/oc. In order to preserve current behavior this defaults to `false`.

### DIFF
--- a/cmd/version/cmd.go
+++ b/cmd/version/cmd.go
@@ -26,6 +26,10 @@ import (
 	"github.com/openshift/rosa/pkg/info"
 )
 
+var args struct {
+	clientOnly bool
+}
+
 var Cmd = &cobra.Command{
 	Use:   "version",
 	Short: "Prints the version of the tool",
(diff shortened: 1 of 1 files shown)


cmd/version/cmd.go
```diff
@@ -26,14 +26,31 @@ import (
...
 	Use:   "version",
 	Short: "Prints the version of the tool",
 	Long:  "Prints the version number of the tool.",
 	Run:   run,
 }
 
+func init() {
+	flags := Cmd.Flags()
+
+	flags.BoolVar(
```
ciaranRoche, line 43: Is this something we want to expose to our users? 

I am wondering if we want to use `MarkHidden` for now, to unblock your work, and we can check with BU if this is something we want to expose.

jfchevrette: I don't see any harm or risk in letting a user print a CLI tool version while explicitly disabling an unneeded external network call. Can you explain?

ciaranRoche: typically if we are not ready to support a new flag, we MarkHidden from the user. Functionality is still there. So I am wondering are we better in this case to do the same to unblock your effort. And if the need or ask from customers for this ability, we we can un-hide it.

gdbranco: I'd say there might definitely be customers that use it within an environment without connection. I see no harm in moving this through. Your call :)

ciaranRoche (approved):
/lgtm

PR #1693, opened 2024-01-12
OCM-5530 | feat: add new parameter to disable CNI at cluster creation for HCP

andreadecorte:
In Hosted Control Planes, we can disable CNI at creation to let user bring their own.

This change exposes the relevant parameter with a new boolean flag.

There is an existing --network-type parameter, but decision was taken not to reuse this as it is going to be deprecated moving forward and the dedicated parameter is clearer in terms of usage.

Additional validation added in order to ensure the 2 parameters don't collide and are correctly validated.

No breaking change expected compared to the current behavior.

Related: [OCM-5530](https://issues.redhat.com//browse/OCM-5530)

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -176,6 +176,7 @@ var args struct {
 	// Hypershift options:
 	hostedClusterEnabled bool
 	billingAccount       string
+	noCni                bool
 
 	// Cluster Admin
 	createAdminUser      bool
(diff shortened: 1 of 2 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -3186,6 +3217,25 @@ func run(cmd *cobra.Command, _ []string) {
...
+	var networkType string
+	if args.networkType == "" {
+		// Parameter not specified, nothing to do
+		return networkType
+	}
+	for _, validType := range ocm.NetworkTypes {
+		if args.networkType == validType {
+			networkType = args.networkType
+		}
+	}
```
gdbranco, line 3231:
```suggestion
	helper.Contains(ocm.NetworkTypes, args.networkType)
```

PR #1695, opened 2024-01-12
OCM-4963 | Feat | Add oauth login using PKCE

tirthct:
**Changed**
- Add oauth login using PKCE
- go mod vendor packages

**Tested**
- `--oauth-login` flag should redirect to Red Hat SSO and get the token from there upon login : tested


go.mod
```diff
@@ -75,18 +75,24 @@ require (
...
 	gopkg.in/square/go-jose.v2 v2.6.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	k8s.io/utils v0.0.0-20221107191617-1a15be271d1d // indirect
 )
 
-replace github.com/golang/glog => github.com/kubermatic/glog-logrus v0.0.0-20180829085450-3fa5b9870d1d
+replace (
+	github.com/golang/glog => github.com/kubermatic/glog-logrus v0.0.0-20180829085450-3fa5b9870d1d
+	github.com/openshift-online/ocm-sdk-go v0.1.391 => ../ocm-sdk-go
```
tirthct, line 97: TODO: replace this with actual sdk version when https://github.com/openshift-online/ocm-sdk-go/pull/885 is merged

tirthct: Removed

cmd/login/cmd.go
```diff
@@ -38,6 +39,8 @@ import (
 // #nosec G101
 var uiTokenPage string = "https://console.redhat.com/openshift/token/rosa"
 
+const oauthClientId = "ocm-cli"
```
gdbranco, line 42: Does it matter to reuse the ID from the ocm-cli?

tirthct: The clientId (ocm-cli) is provided by the IT team, specifically for the CLI usage. So we can reuse this in ocm-cli as well as rosa

ciaranRoche (approved):
/lgtm

ciaranRoche:
@tirthct can you follow our commit message format so we can merge this :pray: 

https://github.com/openshift/rosa/blob/master/CONTRIBUTING.md#contributing-code

go.mod
```diff
@@ -1,10 +1,10 @@
 module github.com/openshift/rosa
 
-go 1.20
+go 1.21
```
robpblake, line 3: @tirthct If you require ROSA to be using `1.21` of go, then that should be done under a separate JIRA first so that we can ensure the CI pipeline is working as expected. If not, then please revert this change to stick with `1.20`.

cheers,

Rob

robpblake (changes requested):
If `1.21` of go is a requirement, then let's do that first, otherwise please revert to `1.20` of go.

go.mod
tirthct: Reverted back to 1.20

PR #1702, opened 2024-01-17
OCM-1366 | fix: reword usage help text for delete and link ocm role c…

ciaranRoche:
Updates wording for link and delete ocm roles help text

JIRA - https://issues.redhat.com/browse/OCM-1366

### DIFF
--- a/cmd/link/cmd.go
+++ b/cmd/link/cmd.go
@@ -28,8 +28,8 @@ import (
 var Cmd = &cobra.Command{
 	Use:     "link",
 	Aliases: []string{"associate"},
-	Short:   "Link a ocm/user role from stdin",
-	Long:    "Link a ocm/user role from stdin",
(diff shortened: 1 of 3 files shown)


cmd/link/cmd.go
```diff
@@ -28,8 +28,8 @@ import (
 var Cmd = &cobra.Command{
 	Use:     "link",
 	Aliases: []string{"associate"},
-	Short:   "Link a ocm/user role from stdin",
-	Long:    "Link a ocm/user role from stdin",
+	Short:   "Link ocm role to specific OCM organization",
```
robpblake, line 31: @ciaranRoche We're using `ocm` and `OCM` in the same sentence. Which one is our preferred?

ciaranRoche: Good catch, updated to `OCM`

robpblake (approved):

PR #1710, opened 2024-01-18
OCM-5626 | feat: Added fix for showing empty account roles

den-rgb:
[JIRA](https://issues.redhat.com/browse/OCM-5626)

### DIFF
--- a/pkg/aws/policies.go
+++ b/pkg/aws/policies.go
@@ -833,16 +833,18 @@ func (c *awsClient) mapToAccountRole(version string, role iamtypes.Role) (Role, 
 
 func (c *awsClient) mapToAccountRoles(version string, roles []iamtypes.Role) ([]Role, error) {
 
-	accountRoles := []Role{}
+	var accountRoles []Role
 	for _, role := range roles {
-
+		if !checkIfAccountRole(role.RoleName) {
+			continue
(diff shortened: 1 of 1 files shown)


pkg/aws/policies.go
```diff
@@ -841,6 +841,10 @@ func (c *awsClient) mapToAccountRoles(version string, roles []iamtypes.Role) ([]
 			return accountRoles, err
 		}
 
+		if !checkIfAccountRole(role.RoleName) {
+			return []Role{}, nil
```
ciaranRoche, line 845: I dont think this has the intended outcome we want. It will always return empty list if it finds one role that is not an account role. 

I would think maybe something like 
```
var accountRoles []Role
	for _, role := range roles {
		if !checkIfAccountRole(role.RoleName) {
			continue
		}
		accountRole, err := c.mapToAccountRole(version, role)
		if err != nil {
			return accountRoles, err
		}
		accountRoles = append(accountRoles, accountRole)
	}
	
If we move the check before the map to role, and just continue if its not an account role, should do the trick

den-rgb: you're right thanks for the spot

ciaranRoche (approved):
/lgtm

PR #1714, opened 2024-01-18
OCM-5676 | feat: Make ID in UX fix for 'rosa edit addon' nonpositional

hunterkepley:
https://issues.redhat.com/browse/OCM-5676

Fixes issue with only being able to provide ID in `rosa edit addon` as the first argument (`rosa edit addon ID`)

With this fix, the help message is fixed (https://issues.redhat.com/browse/OCM-1305), to no longer show invalid flags, as well as allow ID in all places

[Test data can be found here, testing all different examples with this command](https://issues.redhat.com/browse/OCM-5676) (**in the comments**)


hunterkepley:
need to gofmt

cmd/edit/addon/cmd.go
```diff
@@ -46,7 +46,20 @@ var Cmd = &cobra.Command{
...
 		}
-		if len(argv) > 0 && (strings.HasPrefix(argv[0], "--") || strings.HasPrefix(argv[0], "-")) {
+		
+		idMissing := true
+		if len(argv) > 0 {
+			for i, arg := range argv {
+				if i == 0 && !(strings.HasPrefix(arg, "--") || strings.HasPrefix(arg, "-")) {
+					idMissing = false
+				}
+				if !strings.HasPrefix(arg, "--") && !strings.HasPrefix(arg, "-") &&
```
gdbranco, line 56: I think you could just loop through and find an argument that doesn't start with '-'

hunterkepley: There is an issue with this solution, for example `-c hkepley-cluster` will return true in this case as the args are separated by spaces not `--flag arg`

hunterkepley: where as it should be false unless `-c hkepley-cluster test-id`

pkg/arguments/arguments.go
```diff
@@ -185,6 +188,88 @@ func ParseKnownFlags(cmd *cobra.Command, argv []string, failOnUnknown bool) erro
 	return nil
 }
 
+// PreprocessUnknownFlagsWithId Parses known and unknown flags will take the command line arguments and map the ones
+// that fit with known flags.
+func PreprocessUnknownFlagsWithId(cmd *cobra.Command, argv []string) error {
```
ciaranRoche, line 193: Can we get unit tests for this func, a lot going on here think it would help to have some unit tests on it :pray:

pkg/arguments/arguments_test.go
```diff
@@ -0,0 +1,48 @@
...
+	Context("Test PreprocessUnknownFlagsWithId func", func() {
+		BeforeEach(func() {
+			cmd = &cobra.Command{
+				Use:     "addon ID",
+				Aliases: []string{"addons", "add-on", "add-ons"},
+				Short:   "Edit add-on installation parameters on cluster",
+				Long:    "Edit the parameters on installed Red Hat managed add-ons on a cluster",
+				Example: `  # Edit the parameters of the Red Hat OpenShift logging operator add-on installation
+  rosa edit addon --cluster=mycluster cluster-logging-operator`,
+				DisableFlagParsing: true,
```
ciaranRoche, line 27: NP : Can you add some dummy uses and stuff to this cmd, just want to ensure it is not confused in the future as a test for addon command. 

Also might be worth breaking into a helper function to allow us to easily call this for the other funcs in the arguments file that are missing tests :smile:

robpblake:
@hunterkepley A few comments for you to address on this one by the looks of things. Any chance we can do that and then get it merged?

cheers,
Rob

pkg/arguments/arguments.go
```diff
@@ -24,13 +24,16 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	errors "github.com/zgalor/weberr"
```
ciaranRoche, line 27: Can you remove this, just use `fmt.errors` weberr should only be used in CS

pkg/arguments/arguments_test.go
```diff
@@ -0,0 +1,48 @@
+package arguments
+
+import (
+	"github.com/spf13/cobra"
+	errors "github.com/zgalor/weberr"
```
ciaranRoche, line 5: same comment as above

hunterkepley:
Codecov seems to be checking unrelated lines here, is this intentional?

pkg/arguments/arguments_test_suite.go
ciaranRoche, line 1: Why the file name change we normally do <pkg_suite_test.go> 

Its good to keep test files post fixed with `test`

pkg/arguments/arguments.go
```diff
@@ -185,6 +187,91 @@ func ParseKnownFlags(cmd *cobra.Command, argv []string, failOnUnknown bool) erro
...
+	var validArgs []string
+	var upcomingValue bool
+
+	foundId := false
+	for _, arg := range argv {
+		switch {
+		// Upcoming value from a space-separated value
+		case upcomingValue:
+			if strings.HasPrefix(arg, "-") {
+				return fmt.Errorf("No value given for flag '%s'", arg)
```
robpblake, line 204: @hunterkepley Does this error need to refer to the previous value of `arg`? If I follow correctly, you expect `arg` to be the value of the previous flag, but here we determine that `arg` is another flag. So there is no value given for the previous `arg` and not the current `arg`.  So as an example:

```
./rosa edit addon -foo= -bar=baz
```

Where we expect `-foo` and `-bar` to have values. I _think_ the error statement here will log:

`No value given for the flag '-bar'` as you're relying on the current value of `arg`.

cmd/edit/addon/cmd.go
```diff
@@ -42,11 +42,9 @@ var Cmd = &cobra.Command{
 	Run:                run,
 	DisableFlagParsing: true,
 	Args: func(cmd *cobra.Command, argv []string) error {
-		err := cmd.ParseFlags(argv)
+
+		err := arguments.PreprocessUnknownFlagsWithId(cmd, argv)
```
robpblake, line 46: @hunterkepley We're discarding the `err` returned from your function. Is there anyway we can wrap it so that we can more easily debug failures? It also will help the user fix any problems with the way they are using the command.

pkg/arguments/arguments.go
```diff
@@ -185,6 +187,91 @@ func ParseKnownFlags(cmd *cobra.Command, argv []string, failOnUnknown bool) erro
...
+	}
+
+	err := flags.Parse(validArgs)
+	if err != nil {
+		return err
+	}
+
+	// If help is called, regardless of other flags, return we want help.
+	// Also say we need help if the command isn't runnable.
+	helpVal, err := cmd.Flags().GetBool("help")
```
robpblake, line 257: @hunterkepley Should this check be first? If the user is asking for help, can we simply bypass all the parsing of args and return early?

pkg/arguments/arguments.go
```diff
@@ -185,6 +187,91 @@ func ParseKnownFlags(cmd *cobra.Command, argv []string, failOnUnknown bool) erro
...
+	if err != nil {
+		// should be impossible to get here as we always declare a help
+		// flag in InitDefaultHelpFlag()
+		cmd.Println("\"help\" flag declared as non-bool. Please correct your code")
```
robpblake, line 261: @hunterkepley If we don't expect this to happen and cannot recover from it, shouldn't we `panic`? I feel that's much more likely to get peoples attention rather than a log message.

pkg/arguments/arguments.go
```diff
@@ -185,6 +187,91 @@ func ParseKnownFlags(cmd *cobra.Command, argv []string, failOnUnknown bool) erro
...
+		return err
+	}
+	if helpVal {
+		return pflag.ErrHelp
+	}
+
+	if !foundId {
+		return fmt.Errorf("ID not found in command")
```
robpblake, line 269: @hunterkepley At the minute we're swallowing the returned `error`. I don't think we should do that, so I think we'll need a clearer error message here. As a suggestion:

```
ID argument not found in list of arguments passed to command
```

But your call 👍

pkg/arguments/arguments_test.go
```diff
@@ -0,0 +1,43 @@
...
+		It("Returns without error", func() {
+			err := PreprocessUnknownFlagsWithId(cmd, []string{"test", "-c", "test-cluster"})
+			Expect(err).ToNot(HaveOccurred())
+		})
+		It("Returns error with no ID", func() {
+			err := PreprocessUnknownFlagsWithId(cmd, []string{"-c", "test-cluster"})
+			Expect(err).To(HaveOccurred())
+		})
+		It("Returns error with flag that has no value", func() {
+			err := PreprocessUnknownFlagsWithId(cmd, []string{"test", "-c", "-c", "-c", "-c"})
```
robpblake, line 39: @hunterkepley See my comment above about testing the actual error message that we have here as I think we might be logging out the incorrect argument.

robpblake (commented):
Thanks @hunterkepley . Some comments from me.

cheers,
Rob

pkg/arguments/arguments.go
hunterkepley: This is the case, but i ended up removing this check as it is not needed, cobra will give a proper error anyway without this :+1:

robpblake (approved):
Minor comment from me @hunterkepley , but looks good overall.

robpblake (approved):
/lgtm

PR #1716, opened 2024-01-19
Ocm 5613

den-rgb:
[JIRA](https://issues.redhat.com/browse/OCM-5613)

Added fix to list additional security group prompt when creating a cluster using interactive mode

### DIFF
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -193,7 +193,7 @@ type Client interface {
 	ValidateAccountRoleVersionCompatibility(roleName string, roleType string, minVersion string) (bool, error)
 	GetDefaultPolicyDocument(policyArn string) (string, error)
 	GetAccountRoleByArn(roleArn string) (Role, error)
-	GetSecurityGroupIds(vpcId string) ([]ec2types.SecurityGroup, error)
+	GetSecurityGroups(vpcId string) ([]ec2types.SecurityGroup, error)
 	FetchPublicSubnetMap(subnets []ec2types.Subnet) (map[string]bool, error)
 }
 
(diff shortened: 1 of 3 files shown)


pkg/aws/client.go
```diff
@@ -1212,30 +1212,42 @@ func (c *awsClient) DeleteSecretInSecretsManager(secretArn string) error {
 }
 
 func (c *awsClient) GetSecurityGroupIds(vpcId string) ([]ec2types.SecurityGroup, error) {
```
robpblake, line 1214: @den-rgb It would be nice to see a unit test for this. The EC2 client is mocked already in the `client_test.go` so it shouldn't be too complicated.

den-rgb: done

robpblake (approved):

gdbranco:
Not sure why prow is not running on this PR. Side note, could you also adjust the PR title?

robpblake:
@gdbranco I think it's because the PR is not targeting `master`. It is targeting `aws-sdk` branch.

oriAdler:
@den-rgb can you please add a description to the PR?

pkg/aws/client.go
```diff
@@ -1212,27 +1212,39 @@ func (c *awsClient) DeleteSecretInSecretsManager(secretArn string) error {
...
```
ciaranRoche, line 1214: Would it be worth updating the naming of this function, as this returns a list of security groups not just their id's

den-rgb: updated

oriAdler:
LGTM

ciaranRoche (approved):
/lgtm

PR #1724, opened 2024-01-23
OCM-5715 | fix: add error message when supplying worker disk size on unsupported version

gdbranco:
Related issue: [OCM-5715](https://issues.redhat.com//browse/OCM-5715)


marcolan018:
changes look good. for me
just wonder whether we can move the common func such as IsSettingRootDiskSizeSupported() into ocm-common so that it can be used in both client/server side.

gdbranco:
> /lgtm changes look good. for me just wonder whether we can move the common func such as IsSettingRootDiskSizeSupported() into ocm-common so that it can be used in both client/server side.

Considering we have multiple common functionality around this I'd say we should elect what needs to be transferred and capture it in an epic so we can have better tracking

PR #1725, opened 2024-01-23
OCM-1388 | fix: add progress indicator for account role policy upgrades

davidleerh:


pkg/helper/helpers.go
```diff
@@ -258,3 +258,13 @@ func ChunkSlice[T any](slice []T, chunkSize int) [][]T {
 func IsBYOVPC(cluster *cmv1.Cluster) bool {
 	return len(cluster.AWS().SubnetIDs()) > 0
 }
+
+func WithSpinner(spin *spinner.Spinner, fn func() error) error {
```
gdbranco, line 262: Not sure if it would end up cleaner or not. But I thought about setting this up in rosa runtime to make it less cumbersome for the developer. Not having to setup the spinner etc, just rely on calling runtime with spinner or something and supply the function

davidleerh: Updated! Ptal

PR #1726, opened 2024-01-23
OCM-1092 | fix: improve error message for invalid credentials

willkutler:
Makes the error message for invalid AWS creds more readable and suggests a fix for users. Also adds a unit test for the `ValidateCredentials` function to verify the fix.

https://issues.redhat.com/browse/OCM-1092

### DIFF
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -705,6 +705,12 @@ func (c *awsClient) ValidateCredentials() (bool, error) {
 	// token
 	_, err := c.stsClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
 	if err != nil {
+		if strings.Contains(fmt.Sprintf("%s", err), "InvalidClientTokenId") {
+			return false, fmt.Errorf(
+				"Invalid AWS Credentials. For help configuring your credentials, see " +
+					"https://docs.openshift.com/rosa/rosa_install_access_delete_clusters/rosa_getting_started_iam/" +
+					"rosa-config-aws-account.html#rosa-configuring-aws-account_rosa-config-aws-account")
+		}
 		return false, err
 	}
 
--- a/pkg/aws/client_test.go
+++ b/pkg/aws/client_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -31,6 +32,7 @@ var _ = Describe("Client", func() {
 		mockIamAPI            *mocks.MockIAMAPI
 		mockS3API             *mocks.MockS3API
 		mockSecretsManagerAPI *mocks.MockSecretsManagerAPI
+		mockSTSApi            *mocks.MockSTSAPI
 	)
 
 	BeforeEach(func() {
@@ -40,6 +42,7 @@ var _ = Describe("Client", func() {
 		mockEC2API = mocks.NewMockEC2API(mockCtrl)
 		mockS3API = mocks.NewMockS3API(mockCtrl)
 		mockSecretsManagerAPI = mocks.NewMockSecretsManagerAPI(mockCtrl)
+		mockSTSApi = mocks.NewMockSTSAPI(mockCtrl)
 		client = aws.New(
 			logrus.New(),
 			mockIamAPI,
@@ -47,7 +50,7 @@ var _ = Describe("Client", func() {
 			mocks.NewMockOrganizationsAPI(mockCtrl),
 			mockS3API,
 			mockSecretsManagerAPI,
-			mocks.NewMockSTSAPI(mockCtrl),
+			mockSTSApi,
 			mockCfAPI,
 			mocks.NewMockServiceQuotasAPI(mockCtrl),
 			&session.Session{},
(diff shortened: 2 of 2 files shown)


cmd/initialize/cmd.go
```diff
@@ -140,7 +140,12 @@ func run(cmd *cobra.Command, argv []string) {
 		if strings.Contains(fmt.Sprintf("%s", err), "STS") {
 			r.OCMClient.LogEvent("ROSAInitCredentialsSTS", nil)
 		}
-		r.Reporter.Errorf("Error creating AWS client: %v", err)
+		if strings.Contains(fmt.Sprintf("%s", err), "InvalidClientTokenId") {
```
robpblake, line 143: @willkutler Rather than duplicating this logic throughout the code, how about a function in the `aws` package that handles client build errors? So _something_ like:

```
function HandleAWSClientBuildError(runtime *Runtime, err error) {
        if strings.Contains(fmt.Sprintf("%s", err), "STS") {
	r.OCMClient.LogEvent("ROSAInitCredentialsSTS", nil)
		}
		r.Reporter.Errorf("Error creating AWS client: %v", err)
		if strings.Contains(fmt.Sprintf("%s", err), "InvalidClientTokenId") {
			r.Reporter.Errorf(
				"Error creating AWS client: Invalid AWS Credentials. Update credentials in ~/.aws/credentials")
		} else {
			r.Reporter.Errorf("Error creating AWS client: %v", err)
		}
}
```

And then, where we fail to build the client:

```
if err != nil {
   aws.HandleAWSClientBuildError(r, err)
   os.Exit(1)
}
```

You could also write a unit test for the `HandleAWSClientBuildError` function 👍

cmd/verify/quota/cmd.go
```diff
@@ -67,7 +67,12 @@ func run(cmd *cobra.Command, _ []string) (err error) {
 		if strings.Contains(fmt.Sprintf("%s", err), "STS") {
 			r.OCMClient.LogEvent("ROSAInitCredentialsSTS", nil)
 		}
-		r.Reporter.Errorf("Error creating AWS client: %v", err)
+		if strings.Contains(fmt.Sprintf("%s", err), "InvalidClientTokenId") {
+			r.Reporter.Errorf(
+				"Error creating AWS client: Invalid AWS Credentials. Update credentials in ~/.aws/credentials")
```
robpblake, line 72: @willkutler Does it make more sense to link to the ROSA docs on how to configure your AWS client? https://docs.openshift.com/rosa/rosa_install_access_delete_clusters/rosa_getting_started_iam/rosa-config-aws-account.html#rosa-configuring-aws-account_rosa-config-aws-account

`~/.aws/credentials` may not be how the customer is configuring their AWS credentials e.g they have set the environment variables `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`...

cmd/verify/permissions/cmd.go
```diff
@@ -69,7 +69,12 @@ func run(cmd *cobra.Command, _ []string) {
 		if strings.Contains(fmt.Sprintf("%s", err), "STS") {
 			r.OCMClient.LogEvent("ROSAInitCredentialsSTS", nil)
 		}
-		r.Reporter.Errorf("Error creating AWS client: %v", err)
+		if strings.Contains(fmt.Sprintf("%s", err), "InvalidClientTokenId") {
```
robpblake, line 72: @willkutler What about code that is reliant on calling `Runtime.WithAWS()` i.e code that isn't building the AWS client directly? Are we handling the error consistently there as well?

cmd/initialize/cmd.go
robpblake: @willkutler Alternatively, can we hook this into the `build` method of the aws.ClientBuilder and wrap the `error` we return to have the friendly message about updating your AWS credentials? Following the code in the `build` method, we call `awsClient.ValidateCredentials()` via `getClientDetails()` function. This would mean that we have a single update and then anyone using the builder benefits from your new error message.

willkutler: moved logic to this function per your request and verified it works for all aws client build()

willkutler: moved logic to this function per your request and verified it works for all aws client build()

cmd/verify/permissions/cmd.go
willkutler: runtime.WithAWS calls Build() so the new placement of the error handler will resolve this as well

willkutler:
@robpblake thanks for the suggestions, moved the error handling into the builder and it is much cleaner now. Here is a sample of the output:

```
willkutler@Wills-MacBook-Pro rosa % ./rosa create cluster --profile=fake
E: Failed to create AWS client: Error creating AWS client: Invalid AWS Credentials. For help configuring your credentials, see https://docs.openshift.com/rosa/rosa_install_access_delete_clusters/rosa_getting_started_iam/rosa-config-aws-account.html#rosa-configuring-aws-account_rosa-config-aws-account
willkutler@Wills-MacBook-Pro rosa % ./rosa verify permissions --region=us-west-2 --profile=fake
E: Error creating AWS client: Error creating AWS client: Invalid AWS Credentials. For help configuring your credentials, see https://docs.openshift.com/rosa/rosa_install_access_delete_clusters/rosa_getting_started_iam/rosa-config-aws-account.html#rosa-configuring-aws-account_rosa-config-aws-account
```

robpblake:
> @robpblake thanks for the suggestions, moved the error handling into the builder and it is much cleaner now. Here is a sample of the output:
> 
> ```
> ./rosa verify permissions --region=us-west-2 --profile=fake
> E: Error creating AWS client: Invalid AWS Credentials. For help configuring your credentials, see https://docs.openshift.com/rosa/rosa_install_access_delete_clusters/rosa_getting_started_iam/rosa-config-aws-account.html#rosa-configuring-aws-account_rosa-config-aws-account
> ```

@willkutler Nice change!

Let's add some unit tests for the `ValidateCredentials` function whilst we're here to improve our coverage and verify your change. How about tests for:

1. We wrap the error from AWS with your message and return `false` on the `InvalidClientTokenId` error from AWS
2. We do not wrap the error from AWS, but return `false` on any other error
3. If no error from AWS, we return `nil` and `true` e.g the credentials are valid

Take a look at the `pkg/aws/client_test.go` file. You'll see that the STS client is available as a mock using gomock. You should be able to mock the behaviour of the client to give you want you need before calling the `ValidateCredentials` function. As an example:

```
It("Wraps InvalidClientTokenId to get user login information", func() {

	err := fmt.Errorf("InvalidClientTokenId: bad credentials")
	mockSTSAPI.EXPECT().GetCallerIdentity(&sts.GetCallerIdentityInput{}).Return(nil, err)
 	
        valid, err := client.ValidateCredentials()
	Expect(valid).To(BeFalse())
	Expect(err.Error()).To(ContainSubstring(
		"https://docs.openshift.com/rosa/rosa_install_access_delete_clusters/rosa_getting_started_iam/"))
}) 
```
Drop me a message on Slack if you need any more help.

cheers,

Rob

robpblake (commented):
Left a comment about unit testing.

willkutler:
@robpblake unit tests are in! PR should be good to go now

robpblake:
Nice work @willkutler 👍

robpblake (approved):
/lgtm

PR #1727, opened 2024-01-23
OCM-1225 | fix: validate machinepool labels for /

thomasmckay:


cmd/create/machinepool/cmd.go
```diff
@@ -232,6 +233,14 @@ func run(cmd *cobra.Command, _ []string) {
 	val, ok := cluster.Properties()[properties.UseLocalCredentials]
 	useLocalCredentials := ok && val == "true"
 
+	if cmd.Flags().Changed("labels") {
```
gdbranco, line 236: Does it needs the same treatment on node pools?

davidleerh: +1

thomasmckay: There is an if-statement at the bottom of each of these files that directs to either machine- or node- pools. Which file(s) were you thinking also need changes?

gdbranco: Oh if it comes later, that's alright then!

davidleerh: Sorry, I thought that the labels were parsed / evaluated in the individual files as well (np vs mp) as well.
I think we can set this as resolved~

ciaranRoche (approved):

PR #1729, opened 2024-01-24
OCM-1234 | feat: Improve error message for `rosa link ocm-role`

oriAdler:
In the case of no admin permission error, include the OCM account ID in the error message.

e.g. 
```
E: OCM account '21PZhUMJwhZLwlf8iXVB4hqyT2m' has no permission for this command, Only organization admin can run this command. Please ask someone with the organization admin role to run the following command 

         rosa link ocm-role --role-arn arn:aws:iam::765374464689:role/oadler-jan-24-OCM-Role-12541229 --organization-id 1jIHnIbrnLH9kQD57W0BuPm78f1
```

### DIFF
--- a/cmd/link/ocmrole/cmd.go
+++ b/cmd/link/ocmrole/cmd.go
@@ -14,6 +14,7 @@ limitations under the License.
 package ocmrole
 
 import (
+	"fmt"
 	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


cmd/link/ocmrole/cmd.go
```diff
@@ -133,9 +133,16 @@ func run(cmd *cobra.Command, argv []string) (err error) {
 	linked, err := r.OCMClient.LinkOrgToRole(orgAccount, roleArn)
 	if err != nil {
 		if errors.GetType(err) == errors.Forbidden || strings.Contains(err.Error(), "ACCT-MGMT-11") {
-			r.Reporter.Errorf("Only organization admin can run this command. "+
+			ocmAccount, localErr := r.OCMClient.GetCurrentAccount()
+			if localErr != nil {
+				r.Reporter.Errorf("Error getting OCM account: %v", localErr)
+				return localErr
```
oriAdler, line 139: Should we return an error here? or include the account ID in the error message if available.
I guess the API call should work, we rely on it in multiple places.

cmd/link/ocmrole/cmd.go
```diff
@@ -133,9 +133,16 @@ func run(cmd *cobra.Command, argv []string) (err error) {
...
+			}
+
+			r.Reporter.Errorf("OCM account '%s' has no permission for this command, "+
```
oriAdler, line 142: We use the term `current account` in other places in the error message, I wonder if it's worth changing all of them to `OCM account`?
Not sure if the user is familiar with the term OCM, on the other hand, `current account` can be confusing with the current AWS account.

cmd/link/ocmrole/cmd.go
marcolan018: imho, log a warning message is bettern than return an error directly

cmd/link/ocmrole/cmd.go
robpblake: @oriAdler Just wondering if the output is easier to understand if we display the username instead of the account id? e.g I feel this is more readable:

```
Your account 'rblake.openshift' has no permission for this command. Only an organization admin can run this command. Please ask an organization admin to run the following command:
```

davidleerh: I agree that current account can be confused with user aws account user.
Can we word it along the lines of, "OCM account of the user" so that it is more descriptive to the user?

robpblake: > We use the term `current account` in other places in the error message, I wonder if it's worth changing all of them to `OCM account`? Not sure if the user is familiar with the term OCM, on the other hand, `current account` can be confusing with the current AWS account.

Red Hat Account maybe 🤔 ?

cmd/link/ocmrole/cmd.go
oriAdler: Changed, thanks.

cmd/link/ocmrole/cmd.go
oriAdler: Changed to `Red Hat account` with the user name included, thanks.

robpblake (approved):
Works for me @oriAdler . Thanks.

PR #1732, opened 2024-01-26
OCM-5461 | feat: use default version

chenz4027:
https://issues.redhat.com/browse/OCM-5461

### DIFF
--- a/cmd/create/machinepool/nodepool.go
+++ b/cmd/create/machinepool/nodepool.go
@@ -72,7 +72,7 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
 		clusterVersion := cluster.Version().RawID()
 		// This is called in HyperShift, but we don't want to exclude version which are HCP disabled for node pools
 		// so we pass the relative parameter as false
-		versionList, err := versions.GetVersionList(r, channelGroup, true, true, false, false)
+		_, versionList, err := versions.GetVersionList(r, channelGroup, true, true, false, false)
 		if err != nil {
 			r.Reporter.Errorf("%s", err)
 			os.Exit(1)
--- a/pkg/helper/versions/helpers.go
+++ b/pkg/helper/versions/helpers.go
@@ -18,7 +18,7 @@ const (
 )
 
 func GetVersionList(r *rosa.Runtime, channelGroup string, isSTS bool, isHostedCP bool, filterHostedCP bool,
-	defaultFirst bool) (versionList []string, err error) {
+	defaultFirst bool) (defaultVersion string, versionList []string, err error) {
 	var vs []*v1.Version
 	var product string
 	if isHostedCP {
@@ -32,13 +32,14 @@ func GetVersionList(r *rosa.Runtime, channelGroup string, isSTS bool, isHostedCP
 	}
 
 	for _, v := range vs {
+		defaultVersion = getDefaultVersion(v, isHostedCP)
 		if isSTS && !ocm.HasSTSSupport(v.RawID(), v.ChannelGroup()) {
 			continue
 		}
 		if filterHostedCP {
 			valid, err := ocm.HasHostedCPSupport(v)
 			if err != nil {
-				return versionList, fmt.Errorf("failed to check HostedCP support: %v", err)
+				return defaultVersion, versionList, fmt.Errorf("failed to check HostedCP support: %v", err)
 			}
 			if !valid {
 				continue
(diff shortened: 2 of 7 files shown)


cmd/create/machinepool/nodepool.go
```diff
@@ -72,7 +72,7 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
 		clusterVersion := cluster.Version().RawID()
 		// This is called in HyperShift, but we don't want to exclude version which are HCP disabled for node pools
 		// so we pass the relative parameter as false
-		versionList, err := versions.GetVersionList(r, channelGroup, true, true, false, false)
+		_, versionList, err := versions.GetVersionList(r, channelGroup, true, true, false, false)
```
chenz4027, line 75: Should we also use the defaultVersion [here](https://github.com/openshift/rosa/blob/b6ca8f1112885cfb56cab8f4ba3a1a734005c195/cmd/create/machinepool/nodepool.go#L103)?

pkg/helper/versions/helpers.go
```diff
@@ -18,7 +18,7 @@ const (
 )
 
 func GetVersionList(r *rosa.Runtime, channelGroup string, isSTS bool, isHostedCP bool, filterHostedCP bool,
```
robpblake, line 20: @chenz4027 You're currently not meeting our requirements on adding test coverage for changes to ROSA which is why the build on your PR is failing.

From your changes, this looks like the easiest function to write a test for. If you can do that, your build should pass.

robpblake (changes requested):

pkg/helper/versions/helpers.go
chenz4027: Thanks @robpblake , I will add a test for it 👍

pkg/helper/versions/helpers.go
```diff
@@ -32,13 +32,23 @@ func GetVersionList(r *rosa.Runtime, channelGroup string, isSTS bool, isHostedCP
...
+		if isHostedCP {
+			if v.HostedControlPlaneDefault() {
+				defaultVersion = v.RawID()
+			}
+
+		} else {
+			if v.Default() {
+				defaultVersion = v.RawID()
+			}
+		}
```
mikeshng, line 44:
```
		if (isHostedCP && v.HostedControlPlaneDefault()) || v.Default() {
				defaultVersion = v.RawID()
		}
```

Minior nit. Is this logically equivalent to the if block above?

pkg/ocm/versions.go
```diff
@@ -87,6 +87,9 @@ func (c *Client) GetVersionsWithProduct(product string, channelGroup string,
 			request.Parameter("product", product)
 		}
 		response, err = request.Send()
+		// log response
+		fmt.Println("here!", response.Items())
```
mikeshng, line 91: Just a reminder to remove this.

chenz4027: Removed 👍

pkg/helper/versions/helpers.go
chenz4027: Yep I moved it to a function

pkg/helper/versions/helpers.go
```diff
@@ -55,6 +56,13 @@ func GetVersionList(r *rosa.Runtime, channelGroup string, isSTS bool, isHostedCP
 	return
 }
 
+func GetDefaultVersion(version *v1.Version, isHostedCP bool) string {
```
mikeshng, line 59:
```suggestion
func getDefaultVersion(version *v1.Version, isHostedCP bool) string {
```

If it doesn't need to be exposed then let's make it a private helper function.

cmd/create/machinepool/nodepool.go
andreadecorte: for node pool default version is the cluster version (aka control plane). So no need to change it.

pkg/helper/versions/helpers.go
mikeshng: @chenz4027 please feel free to resolve. You have addressed my comment. Thanks.

pkg/ocm/versions.go
```diff
@@ -275,10 +275,6 @@ func GetVersionMinorList(ocmClient *Client) (versionList []string, err error) {
 	return
 }
 
-func (c *Client) GetDefaultVersion(channelGroup string) (version string, err error) {
```
andreadecorte, line 278: unused?

chenz4027: Yep

andreadecorte:
lgtm

PR #1733, opened 2024-01-26
[DO-NOT-MERGE]: Proof: factor `rosa create cluster` for reuse.

stevekuznetsov:
cmd/create/cluster: add a test for the help-text output

We're about to do some refactoring for the way flags are bound and
consumed, so we need to be certain that none of the flag names or their
default values have changed during that process. Since `--help` prints
all of this information in an easily-digestible way, we can test that
the output has not changed during our refactor.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

cmd/create/cluster: move code around to new locations

None of this is strictly necessary but a little organization can't hurt,
and we need to start thinking in the Options flow.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

cmd/create/cluster: allow user-driven defaults for flags

When we're factoring the codebase to allow programmatic re-use of the
Options concept between different commands, across projects and in
tests, we need to be considerate of the fact that different consumers
will have different opinions on defaults. Furthermore, consumers may
want to operate over options only, instead of messing around with
programmatically setting flag values. Allowing them to pass in their
data directly makes that easy.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

*: don't assume a *cobra.Command in library code

If multiple callers are to re-use the functionality of a CLI command,
they need to be able to programmatically configure it and run the work
the command would have done, all the while perhaps not even exposing
this in a CLI, or with a *cobra.Command. Factoring helpers to be
ambivalent about where the data comes from and not requiring a
*cobra.Command makes that much easier. A more involved approach here
would have us refactoring all of these helpers to not assume a flag set,
either, but that's a good follow-up for another day.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

cmd/create/cluster: use normal control flow

Instead of having global variables that hold our state, create structs
and pass them around as necessary. Similarly, return errors when they
occur and allow the top-level caller the ability to determine what to do
with those errors. A top-level caller in a CLI may print to the console
and use os.Exit(). A controller would not. Using os.Exit deep inside of
helpers and library code makes it impossible to re-use that code for
anything else, or run unit tests on it, for that matter.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

cmd/create/cluster: update tests

This function now is explicit about its inputs, as opposed to reading
from globals. We can even run the tests in parallel now.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---


stevekuznetsov:
Fun! A bunch of the linters enabled for this repository don't seem to run on `var foo struct{}` declarations - in the `master` branch these are all true on the `cmd/create/cluster/cmd.go` but don't show up until the refactor :D

```
 cmd/create/cluster/options.go:81:2: field `region` is unused (unused)
	region                    string
	^
cmd/create/cluster/options.go:140:2: field `enableProxy` is unused (unused)
	^
cmd/create/cluster/options.go:167:2: field `defaultIngressClusterRoutesHostname` is unused (unused)
	defaultIngressClusterRoutesHostname     string
	^
cmd/create/cluster/options.go:168:2: field `defaultIngressClusterRoutesTlsSecretRef` is unused (unused)
	defaultIngressClusterRoutesTlsSecretRef string
	^
cmd/create/cluster/options.go:1914:4: SA9003: empty branch (staticcheck)
			if ok := confirm.Prompt(false, "Continue with default? A new RH Managed VPC will be created for your cluster"); !ok {
			^ 
```

robpblake:
@stevekuznetsov I'm going to close this PR, but we'll most likely refer back to it as we start to try and drive some refactors.
cheers,
Rob

PR #1738, opened 2024-01-29
OCM-5815 | feat: Add Device Code Flow

tylercreller:
- Adds device code flow login
- Minor cleanup to auth code flow login

### DIFF
--- a/cmd/login/cmd.go
+++ b/cmd/login/cmd.go
@@ -18,9 +18,12 @@ package login
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"strings"
+	"time"
 
+	"github.com/briandowns/spinner"
 	"github.com/golang-jwt/jwt/v4"
 	sdk "github.com/openshift-online/ocm-sdk-go"
 	"github.com/openshift-online/ocm-sdk-go/authentication"
@@ -32,6 +35,7 @@ import (
 	"github.com/openshift/rosa/pkg/fedramp"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/output"
 	rprtr "github.com/openshift/rosa/pkg/reporter"
 	"github.com/openshift/rosa/pkg/rosa"
 )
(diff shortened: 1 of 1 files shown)


cmd/login/cmd.go
```diff
@@ -149,14 +161,40 @@ func run(cmd *cobra.Command, argv []string) {
...
+			r.Reporter.Errorf("An error occurred while retrieving the token: %v", err)
 			os.Exit(1)
 		}
 		args.token = token
-		fmt.Println("Token received successfully")
+		args.clientID = oauthClientId
+		r.Reporter.Infof("Token received successfully")
+	}
+
+	if args.useDeviceCode {
```
marcolan018, line 177: should we consider the corner case that user provides both args: useAuthCode and useDeviceCode?

cmd/login/cmd.go
```diff
@@ -149,14 +161,40 @@ func run(cmd *cobra.Command, argv []string) {
...
+		deviceAuthConfig := &authentication.DeviceAuthConfig{
+			ClientID: oauthClientId,
+		}
+		_, err := deviceAuthConfig.InitiateDeviceAuth(ctx)
+		if err != nil || deviceAuthConfig == nil {
+			r.Reporter.Errorf("An error occurred while initiating device auth: %v", err)
```
marcolan018, line 183: this will lead to confusing message when err==nil

cmd/login/cmd.go
tylercreller: Updated so that the flows are mutually exclusive

cmd/login/cmd.go
tylercreller: Good catch, removed the deviceAuthConfig condition. This will never be true.

cmd/login/cmd.go
```diff
@@ -149,14 +161,38 @@ func run(cmd *cobra.Command, argv []string) {
 	}
 
 	if args.useAuthCode {
-		fmt.Println("You will now be redirected to Red Hat SSO login")
+		r.Reporter.Infof("You will now be redirected to Red Hat SSO login")
+		// Short wait for a less jarring experience
+		time.Sleep(2 * time.Second)
```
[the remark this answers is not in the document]
tylercreller: updated!

tylercreller:
@ciaranRoche PTAL again

ciaranRoche (approved):
/lgtm

PR #1739, opened 2024-01-30
cmd/create/cluster: reduce use of globals

stevekuznetsov:
Instead of using package-global variables and init() to drive control flow, use values of types.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -85,7 +85,7 @@ const (
 	createVpcForHcpDoc = "https://docs.openshift.com/rosa/rosa_hcp/rosa-hcp-sts-creating-a-cluster-quickly.html#rosa-hcp-creating-vpc"
 )
 
-var args struct {
+type args struct {
 	// Watch logs during cluster installation
 	watch bool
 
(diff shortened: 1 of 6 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -807,6 +803,15 @@ func networkTypeCompletion(cmd *cobra.Command, args []string, toComplete string)
 }
 
 func run(cmd *cobra.Command, _ []string) {
+	flags := cmd.Flags()
+	flags.SortFlags = false
+	args := &args{}
+	args.bind(flags)
+
+	cmd.RegisterFlagCompletionFunc("network-type", networkTypeCompletion)
+	aws.AddModeFlag(cmd)
```
stevekuznetsov, line 812: note: fan-out of the changes proposed in this PR would also end up changing the other libraries, but that can come in individual PRs as follow-ups to keep the scope small here.

stevekuznetsov:
The PR exposes errors that the linters could not catch in the anonymous struct before - please let me know what you'd like me to do with these:

```
 cmd/create/cluster/cmd.go:111:2: field `region` is unused (unused)
	region                    string
	^
cmd/create/cluster/cmd.go:170:2: field `enableProxy` is unused (unused)
	^
cmd/create/cluster/cmd.go:197:2: field `defaultIngressClusterRoutesHostname` is unused (unused)
	defaultIngressClusterRoutesHostname     string
	^
cmd/create/cluster/cmd.go:198:2: field `defaultIngressClusterRoutesTlsSecretRef` is unused (unused)
	defaultIngressClusterRoutesTlsSecretRef string
	^ 
```

robpblake:
@stevekuznetsov I'm going to close this PR, but we'll likely refer back to it as we try and drive some refactors.

cheers,

Rob

PR #1743, opened 2024-01-31
Ocm 1060

den-rgb:


cmd/create/ocmrole/cmd.go
```diff
@@ -101,6 +103,13 @@ func init() {
 		"Attach Classic ROSA AWS managed policies to the account roles. This is an alias for --managed-policies")
 	flags.MarkHidden("mp")
 
+	flags.BoolVar(
```
gdbranco, line 106: Can we not use the same from the create command?
https://github.com/openshift/rosa/blob/06a141ef9258ebe956428dd1eff33bc4fc88cdc3/cmd/create/ocmrole/cmd.go#L115

I feel like if the user supplies this we can just pass it along

gdbranco: There's already a flag for "-y" which gets included through confirm.AddFlag, would it be possible to just pass that value along instead of having a new argument?

den-rgb: Okay I think I gt what you mean PTAL

cmd/link/ocmrole/cmd.go
```diff
@@ -28,8 +28,9 @@ import (
 )
 
 var args struct {
-	roleArn        string
-	organizationID string
+	roleArn         string
+	organizationID  string
+	autoConfirmLink bool
```
ciaranRoche, line 33: Is this being used?

cmd/link/ocmrole/cmd.go
```diff
@@ -127,8 +128,10 @@ func run(cmd *cobra.Command, argv []string) (err error) {
 		os.Exit(1)
 	}
 
-	if !confirm.Prompt(true, "Link the '%s' role with organization '%s'?", roleArn, orgAccount) {
-		os.Exit(0)
+	if cmd.Flags().Lookup("yes").Value.String() == "false" {
```
ciaranRoche, line 131: I dont think this change is needed. It is implicitly handled in `confirm.Prompt`

cmd/create/ocmrole/cmd.go
```diff
@@ -332,7 +336,7 @@ func run(cmd *cobra.Command, argv []string) {
 }
 
 func buildCommands(prefix string, roleName string, rolePath string, permissionsBoundary string,
-	accountID string, env string, isAdmin bool, managedPolicies bool,
+	accountID string, env string, isAdmin bool, managedPolicies bool, autoConfirmLink pflag.Value,
```
ciaranRoche, line 339:
```suggestion
	accountID string, env string, isAdmin bool, managedPolicies bool, autoConfirmLink bool,
```
Would suggest to bump this to a bool instead of a flag

cmd/create/ocmrole/cmd.go
```diff
@@ -317,6 +320,7 @@ func run(cmd *cobra.Command, argv []string) {
 			env,
 			isAdmin,
 			managedPolicies,
+			autoLink,
```
ciaranRoche, line 323:
```suggestion
			confirm.Yes()
```
Suggestion if you move this to accept a bool you can just pass in `confirm.Yes()`

cmd/create/ocmrole/cmd.go
```diff
@@ -307,6 +309,7 @@ func run(cmd *cobra.Command, argv []string) {
 			r.Reporter.Infof("All policy files saved to the current directory")
 			r.Reporter.Infof("Run the following commands to create the ocm role and policies:\n")
 		}
+		autoLink := cmd.Flags().Lookup("yes").Value
```
ciaranRoche, line 312: You can remove this if you update the buildCommands to take in a bool

cmd/create/ocmrole/cmd.go
```diff
@@ -433,6 +437,9 @@ func buildCommands(prefix string, roleName string, rolePath string, permissionsB
 
 	linkRole := fmt.Sprintf("rosa link ocm-role --role-arn %s",
 		aws.GetRoleARN(accountID, roleName, rolePath))
+	if autoConfirmLink.String() == "true" {
+		linkRole += " -y"
+	}
```
ciaranRoche, line 442:
```suggestion
	if autoConfirmLink {
		linkRole += " -y"
	}
```
You can clean this up if you take in a bool instead of a flag

cmd/create/ocmrole/cmd.go
```diff
@@ -41,6 +42,7 @@ var args struct {
 	admin               bool
 	path                string
 	managed             bool
+	autoConfirmLink     bool
```
ciaranRoche, line 45: Is this needed?

cmd/link/ocmrole/cmd.go
```diff
@@ -38,7 +38,7 @@ var Cmd = &cobra.Command{
 	Short:   "Link OCM role to specific OCM organization.",
 	Long:    "Link OCM role to specific OCM organization before you create your cluster.",
 	Example: ` # Link OCM role
-  rosa link ocm-role --role-arn arn:aws:iam::123456789012:role/ManagedOpenshift-OCM-Role`,
+  rosa link ocm-role --role-arn arn:aws:iam::123456789012:role/ManagedOpenshift-OCM-Role -y`,
```
gdbranco, line 41: Any reason to change the example?

den-rgb: Just to let users know there is a -y option

ciaranRoche (approved):
/lgtm

PR #1744, opened 2024-01-31
OCM-2780 | feat: Bump SDK release to 0.1.398

hunterkepley:
Fixes issue with 204s checking for content-type which was causing errors in the public FedRAMP environments


ciaranRoche (approved):
/lgtm

PR #1745, opened 2024-01-31
pkg/ocm: remove dependency on AWS client for cluster fetch

stevekuznetsov:
It looks like the AWS-account-specific filters that are exposed in the GetClusterByID() may be a historical side-effect of a previous filtering approach. No futher filtering is necessary today when fetching a cluster by ID.

### DIFF
--- a/pkg/ocm/clusters.go
+++ b/pkg/ocm/clusters.go
@@ -444,9 +444,8 @@ func (c *Client) GetCluster(clusterKey string, creator *aws.Creator) (*cmv1.Clus
 	return c.getCluster(clusterKey, creator)
 }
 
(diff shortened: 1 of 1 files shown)


pkg/ocm/clusters.go
```diff
@@ -740,6 +739,24 @@ func (c *Client) DeleteCluster(clusterKey string, bestEffort bool,
 	return cluster, nil
 }
```
robpblake, line 741: @stevekuznetsov We will need some unit test coverage on these new functions for our CI to pass. We don't yet have a concrete pattern for unit testing when the OCM SDK is being used, but there are reasonable examples of it being done in `pkg/ocm/kubeletconfig_test.go` 

cheers,
Rob

stevekuznetsov:
@ciaranRoche believes this to be a regression and the AWS filters to be required, so I'm closing this PR.

/close

PR #1748, opened 2024-02-01
OCM-5862 | fix: default version computation should be done only once

andreadecorte:
Fix a panic caused by the default version being reinitialized at every loop. Add also a safety fallback in case the default version can't be computed.

Related: [OCM-5862](https://issues.redhat.com//browse/OCM-5862)


chenz4027:
Ah I don't have the permission to add "approved" label. @robpblake Could you help tag that label?

ciaranRoche (approved):
/lgtm

PR #1751, opened 2024-02-01
OCM-5827 | fix: revert throttle counter message

davidleerh:


pkg/aws/client.go
```diff
@@ -1262,8 +1262,7 @@ func (r CustomRetryer) ShouldRetry(req *request.Request) bool {
 	}
 	logger := logging.NewLogger()
 	if strings.Contains(req.Error.Error(), "Throttling") {
-		logger.Warn(fmt.Sprintf("Throttling Rate limit exceeded. Retrying Attempt: '%v' of '%v'",
-			req.RetryCount+1, r.MaxRetries()))
+		logger.Warn("Throttling Rate limit exceeded. Retrying the request again.")
```
robpblake, line 1265: @davidleerh Can you add a unit test for the `ShouldRetry` function. That will allow the CodeCov check to pass and increase our overall coverage. 

cheers!

Rob

PR #1755, opened 2024-02-02
 build: *: use bingo for versioning binary deps

stevekuznetsov:
[HOSTEDCP-1412](https://issues.redhat.com//browse/HOSTEDCP-1412) | ci: Makefile: add a verify target for formatting and modules

We need to ensure that our modules are correctly noted in go.sum and
that the codebase is formatted correctly.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

[HOSTEDCP-1412](https://issues.redhat.com//browse/HOSTEDCP-1412) | build: *: use bingo for versioning binary deps

A number of the tools that we use for ancillary tasks - linting,
formatting, code-generators, etc - are written in Go. There's an ad-hoc
method of managing these tools in the correct versions, but recent
developments in the space include tools like `bingo` which make the
whole process much, much easier.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

[HOSTEDCP-1412](https://issues.redhat.com//browse/HOSTEDCP-1412) | ci: *: fix linting errors

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---


cmd/create/idp/openid.go
```diff
@@ -49,8 +49,6 @@ func buildOpenidIdp(cmd *cobra.Command,
 	}
 
 	if interactive.Enabled() {
-		instructionsURL := "https://docs.openshift.com/dedicated/identity_providers/" +
```
gdbranco, line 52: This is guiding towards openID and the other is towards ldap, they are different :)

stevekuznetsov: Yikes - since the line length linter broke these up into separate "constants" it flagged the two, when only the first half was shared :facep

cmd/create/cluster/cmd.go
```diff
@@ -3218,16 +3218,20 @@ func GenerateContractDisplay(contract *accountsv1.Contract) string {
 	dimensions := contract.Dimensions()
 
 	numberOfVCPUs, numberOfClusters := ocm.GetNumsOfVCPUsAndClusters(dimensions)
-	prePurchaseInfo := fmt.Sprintf("   | Number of vCPUs:    |'%s'             | \n"+
```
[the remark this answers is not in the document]
stevekuznetsov: The info was just inlined below - the test shows that the overall output did not change

cmd/create/idp/openid.go
stevekuznetsov: (I'd personally suggest just removing the `goconst` linter, but that's a bigger change than obeying it for now)

cmd/create/cluster/cmd.go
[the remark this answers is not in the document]
gdbranco: Oh hadn't noticed. My bad

robpblake (approved):
/lgtm

PR #1763, opened 2024-02-06
OCM-5958 | chore: Move release scripts from Wiki to Git repository

robpblake:
This PR moves the scripts used to manage parts of the ROSA release from our Wiki into the Git repository for the project. Once merged, I will update the wiki to reference these scripts instead.

All scripts have been named (or renamed) to follow the pattern `release-<function>.sh`

Legacy scripts that are no longer required have been removed.


thomasmckay:
i'm +1 on this but at least an initial note in hack's README.md would be good imo

robpblake:
> i'm +1 on this but at least an initial note in hack's README.md would be good imo

@thomasmckay Good catch. I had completely missed that.

ciaranRoche (approved):
/lgtm

PR #1772, opened 2024-02-08
OCM-1109 | fix: Check if op roles exist always again

hunterkepley:
Originally moved all logic into auto mode that didn't have to be in manual. Now you can generate commands to make op roles even if roles exist. This MR moves the code that checks for already existing op roles back into general flow (manual & auto modes)

### DIFF
--- a/cmd/create/operatorroles/by_clusterkey.go
+++ b/cmd/create/operatorroles/by_clusterkey.go
@@ -28,6 +28,17 @@ func handleOperatorRoleCreationByClusterKey(r *rosa.Runtime, env string,
 		os.Exit(1)
 	}
 
+	// Check to see if IAM operator roles have already created
+	missingRoles, err := validateOperatorRoles(r, cluster)
+	if err != nil {
+		if strings.Contains(err.Error(), "AccessDenied") {
+			r.Reporter.Debugf("Failed to verify if operator roles exist: '%v'", err)
+		} else {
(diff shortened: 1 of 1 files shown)


cmd/create/operatorroles/by_clusterkey.go
```diff
@@ -28,6 +28,17 @@ func handleOperatorRoleCreationByClusterKey(r *rosa.Runtime, env string,
 		os.Exit(1)
 	}
 
+	// Check to see if IAM operator roles have already created
+	missingRoles, err := validateOperatorRoles(r, cluster)
+	if err != nil {
+		if strings.Contains(err.Error(), "AccessDenied") {
```
robpblake, line 34: @hunterkepley I know you're just moving the code around, but any idea why this is like this? If the error contains `AccessDenied`, we just debug log it and move on, but otherwise we `os.Exit()`?

robpblake (approved):
/lgtm

cmd/create/operatorroles/by_clusterkey.go
hunterkepley: @robpblake I think, in the case there are no permissions, we want to attempt to move on and run manual mode and return commands as the user may still want the commands. But, when auto mode, it will fail regardless later if no permissions.

In other cases, such as the normal 'roles exist' error, it will os.Exit @robpblake

marcolan018: @hunterkepley imho, it's better to add this as the code comment

PR #1777, opened 2024-02-12
OCM-4624 | feat: add cache for fetching versions from mirror

ciaranRoche:
Adds a cache for fetching versions from mirror. 

Cache is stored in `/tmp/ocm/cache.gob`, this can be overriden by the env var OCM_CONFIG, 
We had to use the gob file due to the limitations of cobra and storing state between consecutive command runs.

This will allow us to implement [OCM-4624](https://issues.redhat.com//browse/OCM-4624) across all commands with out overloading requests  to the mirror

### Validation 
 Check out branch and update `DefaultCacheExpiration` to a shorter time example 
 ```
 var DefaultCacheExpiration = time.Now().Add(5 * time.Second)
 ```
 run 
 ``` 
 rosa version --debug
```
notice `GET` to mirror on first call, run again within the `DefaultCacheExpiration` and note the `GET` is not made as it is using the cache

### Verification
Run without OCM_CONFIG set it will store in /tmp/ocm 
```
╭─croche@fedora ~/Work/ocm/rosa ‹[OCM-4624](https://issues.redhat.com//browse/OCM-4624)-cache●› 
╰─$ make rosa && sudo mv rosa /usr/bin
go build -ldflags="-X github.com/openshift/rosa/pkg/info.Build=73c84a2a" ./cmd/rosa
╭─croche@fedora ~/Work/ocm/rosa ‹[OCM-4624](https://issues.redhat.com//browse/OCM-4624)-cache●› 
╰─$ echo $ROSA_CONFIG_DIR

╭─croche@fedora ~/Work/ocm/rosa ‹[OCM-4624](https://issues.redhat.com//browse/OCM-4624)-cache●› 
╰─$ rosa version
I: 1.2.39
I: Your ROSA CLI is up to date.
╭─croche@fedora ~/Work/ocm/rosa ‹[OCM-4624](https://issues.redhat.com//browse/OCM-4624)-cache●› 
╰─$ rosa verify rosa
I: Your ROSA CLI is up to date.
╭─croche@fedora ~/Work/ocm/rosa ‹[OCM-4624](https://issues.redhat.com//browse/OCM-4624)-cache●› 
╰─$ ls /tmp/ocm
ocm-cache.gob
```
Run with OCM_CONFIG set 
```
╭─croche@fedora ~/Work/ocm/rosa ‹[OCM-4624](https://issues.redhat.com//browse/OCM-4624)-cache●› 
╰─$ ls ~/.config/ocm 
ocm-cache.gob  ocm.json
╭─croche@fedora ~/Work/ocm/rosa ‹[OCM-4624](https://issues.redhat.com//browse/OCM-4624)-cache●› 
╰─$ rm ~/.config/ocm/ocm-cache.gob 
╭─croche@fedora ~/Work/ocm/rosa ‹[OCM-4624](https://issues.redhat.com//browse/OCM-4624)-cache●› 
╰─$ echo $ROSA_CONFIG_DIR
/home/croche/.config/ocm/
╭─croche@fedora ~/Work/ocm/rosa ‹[OCM-4624](https://issues.redhat.com//browse/OCM-4624)-cache●› 
╰─$ rosa version    
I: 1.2.39
I: Your ROSA CLI is up to date.
╭─croche@fedora ~/Work/ocm/rosa ‹[OCM-4624](https://issues.redhat.com//browse/OCM-4624)-cache●› 
╰─$ ls ~/.config/ocm
ocm-cache.gob  ocm.json
```
### JIRA
https://issues.redhat.com/browse/OCM-4624

### CODE REVIEW

File: pkg/cache/cache.go [hunk 1]
```diff
@@ -0,0 +1,92 @@
...
+			if v.Expired() {
+				continue
+			}
+		}
+		m[k] = v
+	}
+	return m
+}
+
+func (c *rosaCache) Dir() (string, error) {
```
robpblake, line 86: @ciaranRoche I'd be tempted for us to introduce support for an ENV variable here that allows for the cache directory to be overridden, something like `ROSA_CONFIG_DIR`. So the flow would be:

- Check `ROSA_CONFIG_DIR` exists, if yes, use that; Otherwise
- Use `~/.config/ocm/cache.gob`; Otherwise
- Use `~/tmp/ocm/cache.gob`

That type of variable tends to be useful in CI runs where the home directory of a container is ephemeral between runs, but we can bind mount some non-ephemeral storage onto the `ROSA_CONFIG_DIR` location. It's also just useful for folks who don't want this stuff in their home directory.

If none of those work, then caching is disabled. I think otherwise we run the risk of leaving `cache.gob` files in directories where the command was run from, which I'm not sure feels like being a good citizen.

WDYT?

pkg/cache/utils_test.go
```diff
@@ -0,0 +1,76 @@
...
+			expectedPath := filepath.Join("/home/testuser", ".config", "ocm", "cache.gob")
+
+			path := joinFilePath(testUser)
+
+			Expect(path).To(Equal(expectedPath))
+		})
+	})
+
+	When("when given a nil user", func() {
+		It("should return an empty string", func() {
```
robpblake, line 68: @ciaranRoche This would leave the cache file in the $(PWD) of the command?

cmd/verify/rosa/cmd.go
```diff
@@ -109,10 +117,31 @@ func retrievePossibleVersionsFromMirror() ([]string, error) {
...
 		})
 	})
+	if err := c.Set(cache.VersionCacheKey, possibleVersions); err != nil {
+		logger.Debugf("Failed to set possible versions in cache : %v", err)
+	}
 	logger.Debugf("Versions available for download: %v", possibleVersions)
 	return possibleVersions, nil
 }
 
+func retrievePossibleVersionsFromCache(c cache.RosaCacheService, logger *logrus.Logger) ([]string, bool) {
```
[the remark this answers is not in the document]
ciaranRoche: Yeah that was the plan, i moved the todo comments from this. Would be better to move a lot of the logic from this `cmd` to a `pkg`

robpblake:
@ciaranRoche Shall we try and get this one in, as we can then close out #1570 ?

Lemme know if you want some help to get it over the line.

cheers,
Rob

ciaranRoche:
@robpblake i have updated now, across several commits, took inspiration from your unit test doc, take a look and maybe we can work on this PR, and use it as an example of the changes needed for unit testing. 

Coverage is low because of the mock files :cry:

pkg/cache/cache.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		return "", fmt.Errorf("error checking if path exists: %v", err)
+	}
+	if ocmConfigPathExists {
+		return ocmConfigPath, nil
+	}
+
+	tmpDir, err := getAndEnsureOCMDirectoryExists()
+	if err != nil {
+		return "", fmt.Errorf("error ensuring tmp OCM directory exists: %v", err)
```
robpblake, line 120: @ciaranRoche We seem to mix ROSA and OCM interchangeably. Which one is the most appropriate? I think I'd find it a bit confusing to get an error message about an `OCM` directory when using a tool called ROSA. WDYT?

cmd/version/cmd.go
```diff
@@ -17,77 +17,104 @@ limitations under the License.
...
-	r := rosa.NewRuntime()
-	defer r.Cleanup()
-	err := runWithRuntime(r, cmd)
-	if err != nil {
-		r.Reporter.Errorf(err.Error())
-		os.Exit(1)
-	}
-}
+func (o *RosaVersionOptions) Version() error {
+	o.reporter.Infof("%s (Build: %s)", info.Version, info.Build)
```
robpblake, line 106: @ciaranRoche Just a reminder that we've removed the `(Build: %s)` string for now until we can work out how to apply it consistently in CPaaS builds.

cmd/version/cmd.go
```diff
@@ -17,77 +17,104 @@ limitations under the License.
...
+func NewRosaVersionOptions() (*RosaVersionOptions, error) {
+	verifyRosa, err := verify.NewVerifyRosaOptions()
+	if err != nil {
+		return nil, fmt.Errorf("failed to build rosa verify options : %v", err)
 	}
 
-	Cmd             = makeCmd()
-	delegateCommand = verify.Cmd.Run // used in testing
-)
+	rpt, err := reporter.CreateReporter()
```
robpblake, line 48: @ciaranRoche I'm actually not sure why the `reporter.Builder.Build()` method returns an `error`. It just returns a new pointer to the `Object`. I think we can drop the `error` return from the method signature which greatly cleans things up in terms of having to propagate that `error` everywhere. 

Probably worth doing that under another PR to keep the changes in this one focused. Let me take a look 👍

cmd/version/cmd.go
```diff
@@ -17,77 +17,104 @@ limitations under the License.
...
+	if err != nil {
+		return nil, fmt.Errorf("these was a problem creating the reporter: %v", err)
+	}
+
+	return &RosaVersionOptions{
+		verifyRosa: verifyRosa,
+		reporter:   rpt,
```
robpblake, line 55: @ciaranRoche This _feels_ like a duplication of what we have on `Runtime` in terms of the reporter and moves us away a little bit from the centralized instantiation of things. Is the alternative for us to pass the `Runtime` around as method parameters?

cmd/verify/cmd.go
```diff
@@ -37,5 +40,10 @@ func init() {
 	Cmd.AddCommand(oc.Cmd)
 	Cmd.AddCommand(permissions.Cmd)
 	Cmd.AddCommand(quota.Cmd)
-	Cmd.AddCommand(rosa.Cmd)
+	r, err := rosa.VerifyRosaCmd()
```
robpblake, line 43: @ciaranRoche I think we can tidy this up by dropping the `error` from the signature of the `reporter.Builder.Build` function. Lemme take a look 👍

pkg/cache/cache.go
```diff
@@ -0,0 +1,158 @@
...
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+)
+
+var DefaultCacheExpiration = time.Now().Add(30 * time.Minute)
+
+const (
+	RosaConfigDir = "ROSA_CONFIG_DIR"
```
robpblake, line 15: @ciaranRoche Just wondering if we should reuse `OCM_CONFIG` environment variable or do you see value in having a separate `ROSA_CONFIG_DIR` variable? I wasn't aware of the `OCM_CONFIG` variable until taking a look at the `rosa edit/get config` commands.

robpblake (commented):
@ciaranRoche I think this is a great step in the right direction for us. I have left some comments for you for further discussion. We don't need to nail it on the first attempt, so happy to iterate over this on multiple PRs. 

cheers,

Rob

pkg/cache/cache.go
ciaranRoche: Agree, we do have a mixture, Will update to ROSA

cmd/version/cmd.go
```diff
@@ -17,77 +17,104 @@ limitations under the License.
...
 	"github.com/openshift/rosa/pkg/info"
+	"github.com/openshift/rosa/pkg/reporter"
 	"github.com/openshift/rosa/pkg/rosa"
 )
 
-var (
-	args struct {
-		clientOnly bool
-		verbose    bool
+var args struct {
```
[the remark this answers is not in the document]
ciaranRoche: Yeah sounds good, will update that

cmd/version/cmd.go
ciaranRoche: :see_no_evil:

cmd/version/cmd.go
ciaranRoche: Agree, lets coordinate on what would be the easiest option, and less of a rebase hell :laughing: We can follow up with the reporter change after this lands as part of the continuous improvement

ciaranRoche: Just seen the PR, will merge that and rebase on this

pkg/cache/cache.go
ciaranRoche: Currently uses the OCM Config, so either we stick with that and use the OCM_CONFIG or we create a specific rosa config. I am leading on the former, what do you think?
Both have pro's and cons

robpblake:
@ciaranRoche Shall we try and get this one merged?

cheers,
Rob

cmd/version/cmd.go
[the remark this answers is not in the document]
ciaranRoche: Can you elaborate on this, as we need the args as a var to pass to the flags

cmd/verify/rosa/cmd.go
```diff
@@ -17,123 +17,94 @@ limitations under the License.
...
+	Verify() error
+}
+
+var _ VerifyRosa = &VerifyRosaOptions{}
+
+func NewVerifyRosaOptions() (VerifyRosa, error) {
+	v, err := version.NewRosaVersion()
 	if err != nil {
-		return []string{}, weberr.Wrapf(err, "Error parsing response body")
+		return nil, fmt.Errorf("there was a problem creating version: %v", err)
```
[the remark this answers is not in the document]
ciaranRoche: yeah good point, updated to just return the bare error as its wrapped in the runner method

cmd/version/cmd.go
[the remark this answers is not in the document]
robpblake: I was thinking if we can get rid of the global args variable as it feels a bit out-of-place that everything else is well encapsulated. 

Open to suggestions here, but I was thinking something like this:

Define a struct for the options the user can pass with a function to create a new, default set of these options:

```
type RosaVersionUserOptions struct {
	clientOnly bool
	verbose    bool
}

func NewRosaVersionUserOptions() RosaVersionUserOptions {
	return RosaVersionUserOptions{}
}
```

These options are then bound to the flagset and passed into the Runner, like so:

```
func NewRosaVersionCommand() *cobra.Command {

	options := NewRosaVersionUserOptions()

	cmd := &cobra.Command{
		Use:   use,
		Short: short,
		Long:  long,
		Args:  cobra.NoArgs,
		Run:   rosa.DefaultRunner(rosa.RuntimeWithOCM(), RosaVersionRunner(options)),
	}

	cmd.Flags().SortFlags = false
	cmd.Flags().BoolVar(
		&options.clientOnly,
		"client",
		false,
		"Client version only (no remote version check)",
	)
	cmd.Flags().BoolVarP(
		&options.verbose,
		"verbose",
		"v",
		false,
		"Display verbose version information, including download locations",
	)
	return cmd
}
```

Finally, the `RosaVersionOptions` has a method that lets it bind from the user options e.g:

```
func (o *RosaVersionOptions) BindAndValidate(options RosaVersionUserOptions) {
	o.verbose = options.verbose
	o.clientOnly = options.clientOnly
}
```

and the runner is called like this:

```
func RosaVersionRunner(userOptions RosaVersionUserOptions) rosa.CommandRunner {
	return func(ctx context.Context, runtime *rosa.Runtime, command *cobra.Command, args []string) error {
		options, err := NewRosaVersionOptions()
		options.BindAndValidate(userOptions)
		if err != nil {
			return fmt.Errorf("there was a problem creating version options: %v", err)
		}
		return options.Version()
	}
}
```

In that model, the runner expects a set of user options, but how we bind the values into it...not a concern of the runner. For testing we just pass in the struct directly, otherwise we lean on Cobra to do it for us.

WDYT?

pkg/cache/cache.go
robpblake: @ciaranRoche Coming back to this, I think we should use `OCM_CONFIG` for the time being rather than introducing another variable just for this.

cmd/version/cmd.go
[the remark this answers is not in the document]
ciaranRoche: Yeah sounds good to me, @hunterkepley can you take a look at this pattern and we follow the same in your refactor of the describe machine pool

robpblake (approved):
/lgtm

PR #1779, opened 2024-02-12
OCM-5829 | feat: Supressing help message when error is thrown

den-rgb:
[JIRA](https://issues.redhat.com/browse/OCM-5829) Suppressing help message when error is shown from linking multiple roles to the same org

### DIFF
--- a/cmd/link/ocmrole/cmd.go
+++ b/cmd/link/ocmrole/cmd.go
@@ -39,7 +39,7 @@ var Cmd = &cobra.Command{
 	Long:    "Link OCM role to specific OCM organization before you create your cluster.",
 	Example: ` # Link OCM role
   rosa link ocm-role --role-arn arn:aws:iam::123456789012:role/ManagedOpenshift-OCM-Role`,
-	RunE: run,
+	Run: run,
 }
 
 func init() {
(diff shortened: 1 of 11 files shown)


hunterkepley:
Ahhh I always wondered how to do this, nice

den-rgb:
> Ahhh I always wondered how to do this, nice

Yeah It's a simple enough solution haha

cmd/link/ocmrole/cmd.go
```diff
@@ -151,6 +151,7 @@ func run(cmd *cobra.Command, argv []string) (err error) {
 		}
 		r.Reporter.Errorf("Unable to link role arn '%s' with the organization id : '%s' : %v",
 			roleArn, orgAccount, err)
+		cmd.SilenceUsage = true
```
ciaranRoche, line 154: Is it worth setting this when defining the cmd :thinking: just wondering what is the use case for when we  do not want to silence the usage :thinking:

den-rgb: It'll only be silenced for that error the help prompt should still show up when putting in invalid params

robpblake (approved):
Nicely done @den-rgb . Thanks!

AlexVulaj:
I believe this PR introduced a bug now where any commands that used to call `.RunE` on these commands will error. For example, the following line will now try to call a nil function. https://github.com/openshift/rosa/blob/master/cmd/create/accountroles/cmd.go#L220

mikeshng:
I think this PR might have introduced a bug for cluster upgrade. Please see comment in https://issues.redhat.com/browse/OCM-6619 for more details. CC @den-rgb @AlexVulaj

PR #1780, opened 2024-02-12
OCM-5804 | feat: Add token and config commands

tirthct:
**Changed**
- Add token command (similar to [OCM CLI](https://github.com/openshift-online/ocm-cli/blob/main/cmd/ocm/token/cmd.go))
- Add config command with get and set subcommands (similar to [OCM CLI](https://github.com/openshift-online/ocm-cli/blob/main/cmd/ocm/config/cmd.go))

### DIFF
--- a/cmd/config/cmd.go
+++ b/cmd/config/cmd.go
@@ -1,0 +1,76 @@
+/*
+Copyright (c) 2024 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/rosa/cmd/config/get"
+	"github.com/openshift/rosa/cmd/config/set"
+	"github.com/openshift/rosa/pkg/config"
+)
+
+func configVarDocs() (ret string) {
+	configType := reflect.ValueOf(config.Config{}).Type()
+	fieldHelps := make([]string, configType.NumField())
+	for i := 0; i < len(fieldHelps); i++ {
+		tag := configType.Field(i).Tag
+		name := strings.Split(tag.Get("json"), ",")[0]
+		doc := tag.Get("doc")
+		fieldHelps[i] = fmt.Sprintf("\t%-15s%s", name, doc)
+	}
+	ret = strings.Join(fieldHelps, "\n")
+	return
+}
+
+func longHelp() (ret string) {
+	loc, err := config.Location()
+	if err != nil {
+		// I think this only happens if homedir.Dir() fails, which is unlikely.
+		loc = fmt.Sprintf("UNKNOWN (%s)", err)
+	}
+	ret = fmt.Sprintf(`Get or set variables from a configuration file.
+
+The location of the configuration file is gleaned from the 'OCM_CONFIG' environment variable,
+or ~/.ocm.json if that variable is unset. Currently using: %s
+
+The following variables are supported:
+
+%s
+
+Note that "rosa config get access_token" gives whatever the file contains - may be missing or expired;
+you probably want "rosa token" command instead which will obtain a fresh token if needed.
+`, loc, configVarDocs())
+	return
+}
+
+func NewConfigCommand() *cobra.Command {
+	Cmd := &cobra.Command{
+		Use:   "config COMMAND VARIABLE",
+		Short: "get or set configuration variables",
+		Long:  longHelp(),
+	}
+	Cmd.AddCommand(get.Cmd)
+	Cmd.AddCommand(set.Cmd)
+	return Cmd
+}
+
+var Cmd = NewConfigCommand()
--- a/cmd/config/cmd.go
+++ b/cmd/config/cmd.go
@@ -28,7 +28,7 @@ import (
 	"github.com/openshift/rosa/pkg/config"
 )
 
-func configVarDocs() (ret string) {
+func configVarDocs() string {
 	configType := reflect.ValueOf(config.Config{}).Type()
 	fieldHelps := make([]string, configType.NumField())
 	for i := 0; i < len(fieldHelps); i++ {
@@ -37,17 +37,16 @@ func configVarDocs() (ret string) {
 		doc := tag.Get("doc")
 		fieldHelps[i] = fmt.Sprintf("\t%-15s%s", name, doc)
 	}
-	ret = strings.Join(fieldHelps, "\n")
-	return
+	return strings.Join(fieldHelps, "\n")
 }
 
-func longHelp() (ret string) {
+func longHelp() string {
 	loc, err := config.Location()
 	if err != nil {
 		// I think this only happens if homedir.Dir() fails, which is unlikely.
 		loc = fmt.Sprintf("UNKNOWN (%s)", err)
 	}
-	ret = fmt.Sprintf(`Get or set variables from a configuration file.
+	return fmt.Sprintf(`Get or set variables from a configuration file.
 
 The location of the configuration file is gleaned from the 'OCM_CONFIG' environment variable,
 or ~/.ocm.json if that variable is unset. Currently using: %s
@@ -59,7 +58,6 @@ The following variables are supported:
 Note that "rosa config get access_token" gives whatever the file contains - may be missing or expired;
 you probably want "rosa token" command instead which will obtain a fresh token if needed.
 `, loc, configVarDocs())
-	return
 }
 
 func NewConfigCommand() *cobra.Command {
--- a/cmd/config/get/get.go
+++ b/cmd/config/get/get.go
@@ -1,0 +1,107 @@
+/*
+Copyright (c) 2024 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package get
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/rosa/pkg/config"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+var (
+	Writer io.Writer = os.Stdout
+	args   struct {
+		debug bool
+	}
+)
+
+var Cmd = NewConfigCommand()
+
+func NewConfigCommand() *cobra.Command {
+	Cmd := &cobra.Command{
+		Use:   "get [flags] VARIABLE",
+		Short: "Prints the value of a config variable",
+		Long:  "Prints the value of a config variable. See 'rosa config --help' for supported config variables.",
+		Args:  cobra.ExactArgs(1),
+		Run:   run,
+	}
+	flags := Cmd.Flags()
+	flags.BoolVar(
+		&args.debug,
+		"debug",
+		false,
+		"Enable debug mode.",
+	)
+	return Cmd
+}
+
+func run(cmd *cobra.Command, argv []string) {
+	r := rosa.NewRuntime()
+
+	err := PrintConfig(argv[0])
+	if err != nil {
+		r.Reporter.Errorf(err.Error())
+		os.Exit(1)
+	}
+}
+
+func PrintConfig(arg string) error {
+	// Load the configuration file:
+	cfg, err := config.Load()
+	if err != nil {
+		err := fmt.Errorf("Can't load config file: %v", err)
+		return err
+	}
+
+	// If the configuration file doesn't exist yet assume that all the configuration settings
+	// are empty:
+	if cfg == nil {
+		err := fmt.Errorf("Config file doesn't exist yet")
+		return err
+	}
+
+	// Print the value of the requested configuration setting:
+	switch arg {
+	case "access_token":
+		fmt.Fprintf(Writer, "%s\n", cfg.AccessToken)
+	case "client_id":
+		fmt.Fprintf(Writer, "%s\n", cfg.ClientID)
+	case "client_secret":
+		fmt.Fprintf(Writer, "%s\n", cfg.ClientSecret)
+	case "insecure":
+		fmt.Fprintf(Writer, "%v\n", cfg.Insecure)
+	case "refresh_token":
+		fmt.Fprintf(Writer, "%s\n", cfg.RefreshToken)
+	case "scopes":
+		fmt.Fprintf(Writer, "%s\n", cfg.Scopes)
+	case "token_url":
+		fmt.Fprintf(Writer, "%s\n", cfg.TokenURL)
+	case "url":
+		fmt.Fprintf(Writer, "%s\n", cfg.URL)
+	case "fedramp":
+		fmt.Fprintf(Writer, "%v\n", cfg.FedRAMP)
+	default:
+		err := fmt.Errorf("Unknown setting")
+		return err
+	}
+	return nil
+}
--- a/cmd/config/get/get.go
+++ b/cmd/config/get/get.go
@@ -29,29 +29,18 @@ import (
 
 var (
 	Writer io.Writer = os.Stdout
-	args   struct {
-		debug bool
-	}
 )
 
-var Cmd = NewConfigCommand()
+var Cmd = NewConfigGetCommand()
 
-func NewConfigCommand() *cobra.Command {
-	Cmd := &cobra.Command{
+func NewConfigGetCommand() *cobra.Command {
+	return &cobra.Command{
 		Use:   "get [flags] VARIABLE",
 		Short: "Prints the value of a config variable",
 		Long:  "Prints the value of a config variable. See 'rosa config --help' for supported config variables.",
 		Args:  cobra.ExactArgs(1),
 		Run:   run,
 	}
-	flags := Cmd.Flags()
-	flags.BoolVar(
-		&args.debug,
-		"debug",
-		false,
-		"Enable debug mode.",
-	)
-	return Cmd
 }
 
 func run(cmd *cobra.Command, argv []string) {
@@ -68,15 +57,17 @@ func PrintConfig(arg string) error {
 	// Load the configuration file:
 	cfg, err := config.Load()
 	if err != nil {
-		err := fmt.Errorf("Can't load config file: %v", err)
-		return err
+		return fmt.Errorf("Failed to load config file: %v", err)
 	}
 
 	// If the configuration file doesn't exist yet assume that all the configuration settings
 	// are empty:
 	if cfg == nil {
-		err := fmt.Errorf("Config file doesn't exist yet")
-		return err
+		loc, err := config.Location()
+		if err != nil {
+			return fmt.Errorf("Failed to find config file location: %v", err)
+		}
+		return fmt.Errorf("Config file '%s' does not exist", loc)
 	}
 
 	// Print the value of the requested configuration setting:
@@ -100,8 +91,7 @@ func PrintConfig(arg string) error {
 	case "fedramp":
 		fmt.Fprintf(Writer, "%v\n", cfg.FedRAMP)
 	default:
-		err := fmt.Errorf("Unknown setting")
-		return err
+		return fmt.Errorf("'%s' is not a supported setting", arg)
 	}
 	return nil
 }
--- a/cmd/config/set/set.go
+++ b/cmd/config/set/set.go
@@ -1,0 +1,120 @@
+/*
+Copyright (c) 2024 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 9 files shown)


cmd/config/cmd.go
```diff
@@ -0,0 +1,73 @@
+/*
+Copyright (c) 2019 Red Hat, Inc.
```
tylercreller, line 2: Copyright should be 2024 in all new files

pkg/ocm/client.go
```diff
@@ -169,6 +169,6 @@ func (c *Client) GetConnectionURL() string {
 	return c.ocm.URL()
 }
 
-func (c *Client) GetConnectionTokens() (string, string, error) {
+func (c *Client) GetConnectionTokens(expiresIn ...time.Duration) (string, string, error) {
```
tylercreller, line 172: Is the expiresIn parameter used?

cmd/config/cmd.go
```diff
@@ -0,0 +1,73 @@
...
+limitations under the License.
+*/
+
+package config
+
+import (
+	"fmt"
+	"github.com/openshift/rosa/cmd/config/get"
+	"github.com/openshift/rosa/cmd/config/set"
+	"github.com/openshift/rosa/pkg/config"
```
ciaranRoche, line 23: Can you run `make fmt` to fix these imports to the correct order please :pray:

ciaranRoche (commented):
Can you add some tests to this PR please :pray:

cmd/config/cmd.go
```diff
@@ -0,0 +1,73 @@
...
+	"reflect"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+func configVarDocs() (ret string) {
```
robpblake, line 30: @tirthct I think a comment explaining this function would be helpful.

Is there an update planned for the `Config` struct as I don't see any `doc` tags on it? Also a unit test would be awesome 👍

cmd/config/cmd.go
```diff
@@ -0,0 +1,73 @@
...
+
+%s
+
+Note that "rosa config get access_token" gives whatever the file contains - may be missing or expired;
+you probably want "rosa token" command instead which will obtain a fresh token if needed.
+`, loc, configVarDocs())
+	return
+}
+
+var Cmd = &cobra.Command{
```
robpblake, line 64: @tirthct We're going to try and move away from global variables, so would you mind turning this into a function that returns the `*cobra.Command` e.g:

```
func NewConfigCommand() *cobra.Command {
return &cobra.Command {
  ...
}
```

cmd/config/cmd.go
```diff
@@ -0,0 +1,73 @@
...
+	Use:   "config COMMAND VARIABLE",
+	Short: "get or set configuration variables",
+	Long:  longHelp(),
+}
+
+func init() {
```
robpblake, line 70: @tirthct It would also be great if you could move this into the function that returns the command struct and remove the use of the `init` function 👍

cmd/config/get/get.go
```diff
@@ -0,0 +1,94 @@
...
+	flags.BoolVar(
+		&args.debug,
+		"debug",
+		false,
+		"Enable debug mode.",
+	)
+}
+
+func run(cmd *cobra.Command, argv []string) {
+	r := rosa.NewRuntime().WithOCM()
```
robpblake, line 52: @tirthct I don't think you need the `WithOCM()` call here unless you're using the OCM API?

cmd/config/get/get.go
```diff
@@ -0,0 +1,94 @@
...
+		fmt.Fprintf(writer, "%s\n", cfg.Scopes)
+	case "token_url":
+		fmt.Fprintf(writer, "%s\n", cfg.TokenURL)
+	case "url":
+		fmt.Fprintf(writer, "%s\n", cfg.URL)
+	case "fedramp":
+		fmt.Fprintf(writer, "%v\n", cfg.FedRAMP)
+	default:
+		r.Reporter.Errorf("Unknown setting")
+		os.Exit(1)
```
robpblake, line 92: @tirthct As a suggestion, how about we move this into a `ConfigService` that has a `PrintConfig` function? That would mean we keep Cobra as a thin wrapper around our actual logic and we can significantly improve our test coverage.

As an example, assuming an interface that looks like this:

```
type ConfigService interface {
  PrintConfig(writer Writer) error
}
```

This command could be re-written as follows

```
  r := rosa.NewRuntime()
  cfgService := config.NewConfigService()
  err := execute(cfgService)
  if err != nil {
    r.reporter.Errorf(err)
    os.Exit(1)
  }
}

func execute(cfgService config.ConfigService) {
  return cfgService.PrintConfig(writer)
}

``` 

We could then add unit tests for the `execute` function by passing in a `mock` of the `ConfigService` interface, and also for your `ConfigService` implementation.

WDYT?

cmd/config/get/get.go
```diff
@@ -0,0 +1,94 @@
...
+	"text/tabwriter"
+
+	"github.com/spf13/cobra"
+)
+
+var args struct {
+	debug bool
+}
+
+var Cmd = &cobra.Command{
```
robpblake, line 33: @tirthct Same comment re: global arg vs function.

cmd/config/set/set.go
```diff
@@ -0,0 +1,108 @@
...
+	flags := Cmd.Flags()
+	flags.BoolVar(
+		&args.debug,
+		"debug",
+		false,
+		"Enable debug mode.",
+	)
+}
+
+func run(cmd *cobra.Command, argv []string) {
```
robpblake, line 50: @tirthct Same suggestion here to make use of a `ConfigService` that can `Save` the config given a `Config` struct e.g

```
type ConfigService interface {
  SaveConfig(config Config) error
}
```

cmd/token/cmd.go
```diff
@@ -0,0 +1,168 @@
...
+	flags.BoolVar(
+		&args.generate,
+		"generate",
+		false,
+		"Generate a new token.",
+	)
+	output.AddFlag(Cmd)
+}
+
+func run(cmd *cobra.Command, argv []string) {
```
robpblake, line 68: @tirthct Same suggestion here about defining a service that performs the logic and simply calling into it from the `cobra.Command`.

robpblake (commented):
Hi @tirthct 

I've a few suggestions in terms of layout that I think will make it easier for us to write unit tests for these additions.

Let me know if anything I've said doesn't make sense or you need to discuss it further.

cheers,

Rob

robpblake:
@tirthct Could you also please squash your commits. Cheers!

Rob

pkg/ocm/client.go
tylercreller: Hmm I don't see any usage in the function itself though on 173

cmd/config/cmd.go
tirthct: Hey @robpblake, sorry I couldn't understand what you meant by this : Is there an update planned for the Config struct as I don't see any doc tags on it?
Also, this is mostly a documentation function, so adding a few comments here, but not sure if we can write tests for this function specifically.

pkg/ocm/client.go
tylercreller: resolved

cmd/config/set/set.go
```diff
@@ -0,0 +1,110 @@
...
+	debug bool
+}
+
+var Cmd = NewConfigCommand()
+
+func NewConfigCommand() *cobra.Command {
+	Cmd := &cobra.Command{
+		Use:   "set [flags] VARIABLE VALUE",
+		Short: "Sets the variable's value",
+		Long:  "Sets the value of a config variable. See 'ocm config --help' for supported config variables.",
```
tylercreller, line 39: Help text contains "OCM"

cmd/config/get/get.go
```diff
@@ -0,0 +1,96 @@
...
+	debug bool
+}
+
+var Cmd = NewConfigCommand()
+
+func NewConfigCommand() *cobra.Command {
+	Cmd := &cobra.Command{
+		Use:   "get [flags] VARIABLE",
+		Short: "Prints the value of a config variable",
+		Long:  "Prints the value of a config variable. See 'ocm config --help' for supported config variables.",
```
tylercreller, line 40: Help text contains "OCM"

cmd/config/cmd.go
tylercreller: @tirthct The doc tags can be found here in OCM https://github.com/openshift-online/ocm-cli/blob/main/pkg/config/config.go#L43-L53

@robpblake is asking for an update to the ROSA config `doc:` tags since we are attempting to leverage them in this function.

tirthct:
> Can you add some tests to this PR please 🙏

Added tests for config and token commands

cmd/config/cmd.go
tirthct: Added doc tags to config

cmd/config/cmd.go
tirthct: Changed to a function

cmd/config/cmd.go
tirthct: Removed init() function

cmd/config/get/get.go
tirthct: Removed WithOCM()

cmd/config/get/get.go
tirthct: @robpblake I am guessing the purpose here was to have as much code coverage as possible, and we have added tests to make sure we can check the entire config save/print functionality, instead of mocking these methods. Hence, we have not added the interface as of now.
@tylercreller

cmd/config/get/get.go
tirthct: Added function

cmd/config/set/set.go
tirthct: Same as above, we tried covering the method without the mock behavior

cmd/token/cmd.go
tirthct: Added code coverage without mocking or using the service

cmd/config/set/set.go
tirthct: Changed

cmd/config/get/get.go
tirthct: Changed

tirthct:
> squa

Squashed commits

cmd/config/cmd.go
```diff
@@ -0,0 +1,76 @@
...
+	configType := reflect.ValueOf(config.Config{}).Type()
+	fieldHelps := make([]string, configType.NumField())
+	for i := 0; i < len(fieldHelps); i++ {
+		tag := configType.Field(i).Tag
+		name := strings.Split(tag.Get("json"), ",")[0]
+		doc := tag.Get("doc")
+		fieldHelps[i] = fmt.Sprintf("\t%-15s%s", name, doc)
+	}
+	ret = strings.Join(fieldHelps, "\n")
+	return
```
ciaranRoche, line 41: Having named return values is not something we encourage in the CLI. 

I dont think it adds any benefit to this func in particular, when we can just `return strings.Join(fieldHelps, "\n")

cmd/config/cmd.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func longHelp() (ret string) {
```
ciaranRoche, line 44: Same comment as above, to remove the named return value :pray:

cmd/config/get/get.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/rosa/pkg/config"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+var (
+	Writer io.Writer = os.Stdout
+	args   struct {
```
ciaranRoche, line 32: Is this being used?

cmd/config/get/get.go
```diff
@@ -0,0 +1,107 @@
...
+func NewConfigCommand() *cobra.Command {
+	Cmd := &cobra.Command{
+		Use:   "get [flags] VARIABLE",
+		Short: "Prints the value of a config variable",
+		Long:  "Prints the value of a config variable. See 'rosa config --help' for supported config variables.",
+		Args:  cobra.ExactArgs(1),
+		Run:   run,
+	}
+	flags := Cmd.Flags()
+	flags.BoolVar(
```
ciaranRoche, line 48: Debug is a persistant global flag, no need to add it here.

cmd/config/get/get.go
```diff
@@ -0,0 +1,107 @@
...
+		fmt.Fprintf(Writer, "%s\n", cfg.Scopes)
+	case "token_url":
+		fmt.Fprintf(Writer, "%s\n", cfg.TokenURL)
+	case "url":
+		fmt.Fprintf(Writer, "%s\n", cfg.URL)
+	case "fedramp":
+		fmt.Fprintf(Writer, "%v\n", cfg.FedRAMP)
+	default:
+		err := fmt.Errorf("Unknown setting")
+		return err
```
ciaranRoche, line 104:
```suggestion
		return fmt.Errorf("Unknown setting")
```

cmd/config/set/set.go
```diff
@@ -0,0 +1,120 @@
...
+func NewConfigCommand() *cobra.Command {
+	Cmd := &cobra.Command{
+		Use:   "set [flags] VARIABLE VALUE",
+		Short: "Sets the variable's value",
+		Long:  "Sets the value of a config variable. See 'rosa config --help' for supported config variables.",
+		Args:  cobra.ExactArgs(2),
+		Run:   run,
+	}
+	flags := Cmd.Flags()
+	flags.BoolVar(
```
ciaranRoche, line 46: Same as above, can remove the debug flag

cmd/config/set/set.go
```diff
@@ -0,0 +1,120 @@
...
+		cfg.URL = value
+	case "fedramp":
+		cfg.FedRAMP, err = strconv.ParseBool(value)
+		if err != nil {
+			err := fmt.Errorf("Failed to set fedramp: %v", value)
+			return err
+		}
+	default:
+		err := errors.New("Unknown setting")
+		return err
```
ciaranRoche, line 109:
```suggestion
		return fmt.Errorf("Unknown setting")
```

cmd/config/set/set.go
```diff
@@ -0,0 +1,120 @@
...
+*/
+
+package set
+
+import (
+	"fmt"
+	"os"
+	"strconv"
+
+	"github.com/pkg/errors"
```
ciaranRoche, line 24: Do we need to use this pkg, or can we use `fmt.Error`?

cmd/config/set/set.go
```diff
@@ -0,0 +1,120 @@
...
+		return err
+	case "token_url":
+		cfg.TokenURL = value
+	case "url":
+		cfg.URL = value
+	case "fedramp":
+		cfg.FedRAMP, err = strconv.ParseBool(value)
+		if err != nil {
+			err := fmt.Errorf("Failed to set fedramp: %v", value)
+			return err
```
ciaranRoche, line 105:
```suggestion
			return fmt.Errorf("Failed to set fedramp: %v", value)
			
```

cmd/config/set/set.go
```diff
@@ -0,0 +1,120 @@
...
+		cfg.Insecure, err = strconv.ParseBool(value)
+		if err != nil {
+			err := fmt.Errorf("Failed to set insecure: %v", value)
+			return err
+		}
+	case "refresh_token":
+		cfg.RefreshToken = value
+	case "scopes":
+		err := errors.New("Setting scopes is unsupported")
+		return err
```
ciaranRoche, line 96:
```suggestion
		return fmt.Errorf("Setting scopes is unsupported")
```
There is also a mismatch between the use of fmt.Errorf and errors.New. We should align these

cmd/config/set/set.go
```diff
@@ -0,0 +1,120 @@
...
+		cfg.AccessToken = value
+	case "client_id":
+		cfg.ClientID = value
+	case "client_secret":
+		cfg.ClientSecret = value
+	case "insecure":
+		cfg.Insecure, err = strconv.ParseBool(value)
+		if err != nil {
+			err := fmt.Errorf("Failed to set insecure: %v", value)
+			return err
```
ciaranRoche, line 90: Same as other comments, we can just return the error

cmd/config/set/set.go
```diff
@@ -0,0 +1,120 @@
...
+		err := errors.New("Unknown setting")
+		return err
+	}
+
+	// Save the configuration:
+	err = config.Save(cfg)
+	if err != nil {
+		err := fmt.Errorf("Can't save config file: %v", err)
+		return err
+	}
```
ciaranRoche, line 117: We can shorten this up a bit
```suggestion
	if err = config.Save(cfg); err != nil {
		return fmt.Errorf("Can't save config file: %v", err)
	}
```

cmd/token/cmd.go
```diff
@@ -0,0 +1,173 @@
...
+			os.Exit(1)
+		}
+	} else {
+		// Get the tokens:
+		accessToken, refreshToken, err = r.OCMClient.GetConnectionTokens()
+		if err != nil {
+			r.Reporter.Errorf("Can't get token: %v", err)
+			os.Exit(1)
+		}
+	}
```
ciaranRoche, line 114: Can we flatten this out a bit, and move to its own function and return early
```suggestion
	accessToken, refreshToken, err = getAccessTokens(r, args.generate)
	if err != nil {
		r.Reporter.Errorf("Can't get token: %v", err)
		os.Exit(1)
	}
```
The func can be something like : 
```
func getAccessTokens(r *rosa.Runtime, generate bool) (string, string, error) {
	if generate {
		return r.OCMClient.GetConnectionTokens(15 * time.Minute)
	}
	return r.OCMClient.GetConnectionTokens()
}
```

cmd/token/cmd.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// Select the token according to the options:
+	selectedToken := accessToken
+	if args.refresh {
+		selectedToken = refreshToken
+	}
+
+	// Parse the token:
+	parser := new(jwt.Parser)
+	fmt.Println(refreshToken)
```
ciaranRoche, line 124: Is this expected to be printed or left over debugging?

cmd/token/cmd.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	// Save the configuration:
+	cfg.AccessToken = accessToken
+	cfg.RefreshToken = refreshToken
+	err = config.Save(cfg)
+	if err != nil {
+		r.Reporter.Errorf("Can't save config file: %v", err)
+		os.Exit(1)
+	}
```
ciaranRoche, line 172: NP 
```suggestion
	if err = config.Save(cfg); err != nil {
		r.Reporter.Errorf("Can't save config file: %v", err)
		os.Exit(1)
	}
```

cmd/config/get/get.go
```diff
@@ -0,0 +1,107 @@
...
+		r.Reporter.Errorf(err.Error())
+		os.Exit(1)
+	}
+}
+
+func PrintConfig(arg string) error {
+	// Load the configuration file:
+	cfg, err := config.Load()
+	if err != nil {
+		err := fmt.Errorf("Can't load config file: %v", err)
```
robpblake, line 71: @tirthct Can be shortened to just `return fmt.Errorf("Failed to load config file: %v", err)`

cmd/config/get/get.go
```diff
@@ -0,0 +1,107 @@
...
+		return err
+	}
+
+	// If the configuration file doesn't exist yet assume that all the configuration settings
+	// are empty:
+	if cfg == nil {
+		err := fmt.Errorf("Config file doesn't exist yet")
```
robpblake, line 78: @tirthct Should we include the path of the config file in the error message to make it easier to debug? e.g 

```
return fmt.Errorf("Config file '%s' does not exist", pathToFile)
```

cmd/config/get/get.go
robpblake: I think lets include the name of the `arg` in the error message and make it a bit clearer e.g:

```
return fmt.Errorf("'%s' is not a supported setting", arg)

File: cmd/token/cmd.go (hunk 11 above)
Comment (robpblake) [reply cmd/token/cmd.go:74]: @tirthct Thank you! 

Could you please change this to be the same format as the `get` / `set` commands where we bubble the error upwards and not have lots of `os.Exit(1)` statements e.g:

```
r := rosa.NewRuntime()

err := CreateToken(args)
if err != nil {
	r.Reporter.Errorf(err.Error())
	os.Exit(1)
}
```

cmd/token/cmd.go
robpblake: See my comment about bubbling the `err` upwards rather than `os.Exit(1`) in the function itself to align with the style of the `set` / `get` commands

robpblake (commented):
Some minor feedback @tirthct . Looking good and I think we can get this merged soon.

cmd/config/cmd.go
tirthct: Returned as suggested

cmd/config/cmd.go
tirthct: Returned as suggested

cmd/config/get/get.go
tirthct: It was only used for debug flag. Removed.

cmd/config/get/get.go
tirthct: Removed

cmd/config/get/get.go
tirthct: Changed

cmd/config/set/set.go
tirthct: Removed

cmd/config/set/set.go
tirthct: Changed

cmd/config/set/set.go
tirthct: Removed since added consistency with fmt.Error

cmd/config/set/set.go
tirthct: Changed

cmd/config/set/set.go
tirthct: Aligned

cmd/config/set/set.go
tirthct: Changed

cmd/config/set/set.go
tirthct: Changed

cmd/token/cmd.go
tirthct: Changed

cmd/token/cmd.go
tirthct: Bubbled up and also added inline if

cmd/config/get/get.go
tirthct: Changed

cmd/config/get/get.go
tirthct: Added config.Location() call to get file path and included in error message

robpblake (approved):
/lgtm

PR #1781, opened 2024-02-13
OCM-4459 | feature: CRUD for external auth providers

chenz4027:


cmd/describe/externalauthproviders/cmd.go
```diff
@@ -0,0 +1,126 @@
...
+	isHypershift := cluster.Hypershift().Enabled()
+
+	if isHypershift {
+		return describeExternalAuthProviders(r, cluster, clusterKey)
+	} else {
+		return fmt.Errorf("External authentication provider is only supported for Hosted Control Planes.")
+	}
+}
+
+func describeExternalAuthProviders(r *rosa.Runtime, cluster *cmv1.Cluster, clusterKey string) error {
```
ciaranRoche, line 83: Id suggest a wrapper to this function that returns the string to be displayed to make it easier to test

cmd/describe/externalauthproviders/cmd.go
```diff
@@ -0,0 +1,126 @@
...
+	externalAuthId := "abc"
+	r.Reporter.Debugf("Fetching external authentication providers '%s' for cluster '%s'", externalAuthId, clusterKey)
+	externalAuthProviders, exists, err := r.OCMClient.GetExternalAuthProviders(cluster.ID(), externalAuthId)
+
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return fmt.Errorf("External authentication provider '%s' not found", externalAuthId)
```
ciaranRoche, line 92: NP : error strings should not be capitalized as it can be wrapped and could end up in some weird grammatical issues.

cmd/dlt/externalauthproviders/cmd.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func run(_ *cobra.Command, _ []string) {
+	r := rosa.NewRuntime().WithOCM()
+	defer r.Cleanup()
+
+	clusterKey := r.GetClusterKey()
+	cluster := r.FetchCluster()
+
+	r.Reporter.Debugf("Deleting External Authentication Providers for cluster '%s'", clusterKey)
```
ciaranRoche, line 48: NP : should we have this debug here, best to only print if cluster is hcp IMO

cmd/dlt/externalauthproviders/cmd.go
```diff
@@ -0,0 +1,55 @@
...
+
+	if cluster.Hypershift().Enabled() {
+		r.OCMClient.DeleteExternalAuthProviders(clusterKey, "test")
+	} else {
+		r.Reporter.Errorf("External authentication provider is only supported for Hosted Control Planes.")
+	}
```
ciaranRoche, line 54: Suggest to move this to a func that returns an error IMO, so it can be easier to test.

cmd/edit/externalauthproviders/cmd.go
```diff
@@ -0,0 +1,162 @@
...
+
+package externalauthproviders
+
+import (
+	"os"
+
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/rosa"
+	"github.com/spf13/cobra"
```
ciaranRoche, line 25: Will need to run `make fmt` to sort the imports

cmd/edit/externalauthproviders/cmd.go
```diff
@@ -0,0 +1,162 @@
...
+var args struct {
+	name                             string
+	issuerName                       string
+	issuerAudiences                  []string
+	issuerUrl                        string
+	issuerCaFile                     string
+	claimMappingGroupsClaim          string
+	claimMappingUsernameClaim        string
+	claimValidationRuleClaim         string
+	claimValidationRuleRequiredValue string
```
ciaranRoche, line 37: The args are the same for both create and edit. I would suggest moving these to a `pkg` to be shared between cmd's

cmd/edit/externalauthproviders/cmd.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func run(cmd *cobra.Command, argv []string) {
+	r := rosa.NewRuntime().WithAWS().WithOCM()
+	defer r.Cleanup()
+
+	clusterKey := r.GetClusterKey()
+	cluster := r.FetchCluster()
+
+	if cluster.Hypershift().Enabled() {
```
ciaranRoche, line 127: NP : Invert this check, it will remove the need for the `else` and we can return early if the cluster is not HCP. 

Potentially might be worth moving this logic to a helper function for easier testing and improved clarity.

cmd/edit/externalauthproviders/cmd.go
```diff
@@ -0,0 +1,162 @@
...
+				cluster.ID(), err)
+			os.Exit(1)
+		}
+
+		r.Reporter.Infof("Successfully updated custom external authentication providers for cluster '%s'", clusterKey)
+		os.Exit(0)
+
+	} else {
+		r.Reporter.Errorf("External authentication provider is only supported for Hosted Control Planes.")
+	}
```
ciaranRoche, line 161: IMO would be best to move this to a new func that returns an error to make it easier to test :pray:

chenz4027:
Thanks @ciaranRoche for all the suggestions, will refactor those for better enhancements 👍

cmd/edit/externalauthproviders/cmd.go
chenz4027: 🙏 To pass the coverage

chenz4027:
@ciaranRoche I refactored the code. Will plan to add more validations to certain params while waiting for the API, let me know if you see any additional concern - more than happy to improve it 😄

chenz4027:
Not sure why but this pr is not syncing with branch https://github.com/chenz4027/rosa/tree/OCM-4459, closing it.

chenz4027:
Turns out that it was a problem with git, fun time LOL
image

PR #1782, opened 2024-02-13
HOSTEDCP-1412 | refactor: pkg/aws: expose the Creator transform from …

stevekuznetsov:
…caller identity

### DIFF
--- a/cmd/create/cluster/spec_test.go
+++ b/cmd/create/cluster/spec_test.go
@@ -12,7 +12,7 @@ import (
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/properties"
 	"github.com/openshift/rosa/pkg/rosa"
-	"github.com/openshift/rosa/pkg/test"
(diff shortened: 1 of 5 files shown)


cmd/create/cluster/spec_test.go
```diff
@@ -50,7 +50,7 @@ var _ = Describe("Cluster Configuration Creation", func() {
 	) {
 		out, err := clusterConfigFor(r.Reporter, in, creator, aws)
 		Expect(err).To(BeNil())
-		Expect(out).To(test.MatchExpected(expected, cmpopts.IgnoreUnexported(cmv1.ExternalAuthConfig{})))
+		Expect(out).To(matchers.MatchExpected(expected, cmpopts.IgnoreUnexported(cmv1.ExternalAuthConfig{})))
```
stevekuznetsov, line 53: I had to move this helper from the `pkg/test` package since that imports `pkg/aws` and would otherwise have created an import cycle.

ciaranRoche (approved):
/lgtm

PR #1783, opened 2024-02-14
OCM-5830 | feat: removing duplicate err msg output

den-rgb:
[JIRA](https://issues.redhat.com/browse/OCM-5830) Removing duplicate err output

### DIFF
--- a/cmd/upgrade/roles/cmd.go
+++ b/cmd/upgrade/roles/cmd.go
@@ -713,7 +713,7 @@ func upgradeOperatorPolicies(
 	case aws.ModeAuto:
 		if !confirm.Prompt(true, "Upgrade each operator role policy to latest version (%s)?", defaultPolicyVersion) {
 			if args.isInvokedFromClusterUpgrade {
-				return r.Reporter.Errorf("Operator roles need to be upgraded to proceed")
+				return fmt.Errorf("operator roles need to be upgraded to proceed")
 			}
 			return nil
 		}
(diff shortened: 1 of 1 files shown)


cmd/upgrade/roles/cmd.go
```diff
@@ -713,7 +713,8 @@ func upgradeOperatorPolicies(
 	case aws.ModeAuto:
 		if !confirm.Prompt(true, "Upgrade each operator role policy to latest version (%s)?", defaultPolicyVersion) {
 			if args.isInvokedFromClusterUpgrade {
-				return r.Reporter.Errorf("Operator roles need to be upgraded to proceed")
+				err := fmt.Errorf("Operator roles need to be upgraded to proceed")
+				return err
```
ciaranRoche, line 717:
```suggestion
				return fmt.Errorf("operator roles need to be upgraded to proceed")
```

ciaranRoche: Update to just return `fmt.Errorf` 

Also its best practice not to return errors that start capitalized. This is because they can be wrapped and we can run into some funky formatting. Best to keep them lower case IMO.

ciaranRoche (commented):
Can you update all the returns in this function to either return `nil` or `fmt.Errorf` :pray:

ciaranRoche (commented):
Just bumping my earlier comment, 

Can you update the following lines 
- 739
- 746
- 768
- 773

This whole func, needs to follow the same pattern and return an error not a `r.reporter` error

den-rgb:
> Just bumping my earlier comment,
> 
> Can you update the following lines
> 
>     * 739
> 
>     * 746
> 
>     * 768
> 
>     * 773
> 
> 
> This whole func, needs to follow the same pattern and return an error not a `r.reporter` error

Is this not a redundant change due to the changes agreed for this ticket: https://github.com/openshift/rosa/pull/1779

ciaranRoche:
> Is this not a redundant change due to the changes agreed for this ticket: #1779

No this is the func sig
```
func upgradeOperatorPolicies(
	mode string,
	r *rosa.Runtime,
	isAccountRoleUpgradeNeed bool,
	policies map[string]*v1.AWSSTSPolicy,
	env string,
	defaultPolicyVersion string,
	credRequests map[string]*v1.STSOperator,
	cluster *v1.Cluster,
	operatorRolePolicyPrefix string,
) error
```

It returns an error. Currently this is not the case, some cases it reports and error, some cases it reports and exits, and in your change it returns an error. 

This should be aligned, it should only return an error and it should leave it up to the caller on how it wants to handle that error.

ciaranRoche (approved):
/lgtm

PR #1784, opened 2024-02-15
HOSTEDCP-1412 | Extract hypershift version validation to a separate function

muraee:
- Extract hypershift version validation to a separate function `ValidateHypershiftVersion()`
- Update `LowestHostedCpSupport` to 4.14.0


den-rgb:
Please include a test for the new function for the codecov to pass

muraee:
@den-rgb shouldn't it be covered by https://github.com/openshift/rosa/blob/master/cmd/create/cluster/cmd_test.go#L148-L159?

den-rgb:
@muraee I think it's giving out because it is function specific, and the tests cover `ValidateVersion` as a whole, I think adding a couple of tests that test the list response should fix it.

pkg/helper/versions/helpers_test.go
```diff
@@ -65,16 +65,16 @@ var _ = Describe("Version Helpers", Ordered, func() {
...
-				"4.14.0-0.nightly-2023-02-27-084419",
-				"4.12.0",
+				"4.16.0-0.nightly-2023-02-27-084419",
+				"4.14.0",
 			),
 			Entry("Current control plane",
-				"4.12.5",
-				"4.12.0-0.a",
+				"4.14.5",
+				"4.14.0",
```
stevekuznetsov, line 77: Side-comment for the maintainers, this seems like a source of infinite toil as the minimum supported version keeps creeping up in the service. Perhaps possible to factor this some other way so the tests remain valid but the work is not necessary on every bump.

muraee:
@den-rgb done

muraee:
@den-rgb @robpblake could you approve if it looks good to you, thanks!

PR #1785, opened 2024-02-15
OCM-6120 | fix: mark --no-cni flag as hidden

andreadecorte:
Mark the --no-cni as hidden as we want to expose it gradually to customers


ciaranRoche (approved):
/lgtm

PR #1786, opened 2024-02-15
OCM-6119 | fix: block users from passing region flag when creating an oidc provider

davidleerh:
image

Indicate to users that passing / using the region flag is not allowed when creating the OIDC provider.

### DIFF
--- a/cmd/create/oidcprovider/cmd.go
+++ b/cmd/create/oidcprovider/cmd.go
@@ -77,6 +77,12 @@ func run(cmd *cobra.Command, argv []string) {
 	r := rosa.NewRuntime().WithAWS().WithOCM()
 	defer r.Cleanup()
 
+	if cmd.Flags().Changed("region") {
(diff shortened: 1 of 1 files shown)


cmd/create/oidcprovider/cmd.go
```diff
@@ -69,6 +70,9 @@ func init() {
 	ocm.AddOptionalClusterFlag(Cmd)
 	aws.AddModeFlag(Cmd)
 
+	arguments.AddRegionFlag(flags)
+	flags.MarkHidden("region")
```
ciaranRoche, line 74: Should we use `MarkDeprecated` instead of `MarkHidden`

gdbranco: We had [this solution](https://github.com/openshift/rosa/blob/5cc4797726a3ab39f94753d76af555b22c61a030/cmd/create/cmd.go#L71) in the past which also marks hidden. I'd say we should stick to it and add the command to the list :thinking:

PR #1789, opened 2024-02-20
OCM-6244 | chore: update scripts to use fix version instead of labels

ciaranRoche:
We have got releases in JIRA and can move away from the release label to using fix versions instead. 
This change allows us to filter on fix versions instead. It also makes changes to update to remove `signed-off-by` from the release notes. 

JIRA - https://issues.redhat.com/browse/OCM-6244


robpblake (approved):
/lgtm

robpblake:
@ciaranRoche I think a rebase will fix the codecov failure. I'll take a task to investigate why it's considering these files as impacting coverage 🤔

PR #1791, opened 2024-02-20
OCM-1012 | feat: Add filtering by config ID for `list oidc-providers`

hunterkepley:
https://issues.redhat.com/browse/OCM-1012

Lists OIDC providers by config ID (as long as that config exists in OCM)

Below are examples of the new flag:

image

### DIFF
--- a/cmd/list/oidcprovider/cmd.go
+++ b/cmd/list/oidcprovider/cmd.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/briandowns/spinner"
+	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 
 	"github.com/openshift/rosa/pkg/aws"
@@ -41,11 +42,23 @@ var Cmd = &cobra.Command{
 	Run: run,
 }
 
+var args struct {
+	oidcConfigId string
+}
+
 func init() {
 	flags := Cmd.Flags()
 	flags.SortFlags = false
 	output.AddFlag(Cmd)
 	ocm.AddOptionalClusterFlag(Cmd)
+
+	flags.StringVarP(
+		&args.oidcConfigId,
+		"oidc-config-id",
+		"i",
+		"",
+		"Filter by OIDC Config ID, returns one provider linked to the config ID.",
+	)
 }
 
 func run(cmd *cobra.Command, _ []string) {
@@ -66,8 +79,16 @@ func run(cmd *cobra.Command, _ []string) {
 		clusterKey := r.GetClusterKey()
 		clusterId = clusterKey
 	}
-
-	providers, err := r.AWSClient.ListOidcProviders(clusterId)
+	var config *v1.OidcConfig
+	var err error
+	if args.oidcConfigId != "" {
+		config, err = r.OCMClient.GetOidcConfig(args.oidcConfigId)
+		if err != nil {
+			r.Reporter.Errorf("Failed to get OIDC config: %v", err)
+			os.Exit(1)
+		}
+	}
+	providers, err := r.AWSClient.ListOidcProviders(clusterId, config)
 
 	if spin != nil {
 		spin.Stop()
@@ -117,17 +138,21 @@ func run(cmd *cobra.Command, _ []string) {
 	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
 	fmt.Fprintf(writer, "OIDC PROVIDER ARN\tCluster ID\tIn Use\n")
 	for _, provider := range providers {
-		providerInUse := "No"
-		if ok := providersInUse[provider.Arn]; ok {
-			providerInUse = "Yes"
-		}
-		fmt.Fprintf(
-			writer,
-			"%s\t%s\t%v\n",
-			provider.Arn,
-			provider.ClusterId,
-			providerInUse,
-		)
+		printProvider(writer, providersInUse, provider)
 	}
 	writer.Flush()
+}
+
+func printProvider(writer *tabwriter.Writer, providersInUse map[string]bool, provider aws.OidcProviderOutput) {
+	providerInUse := "No"
+	if ok := providersInUse[provider.Arn]; ok {
+		providerInUse = "Yes"
+	}
+	fmt.Fprintf(
+		writer,
+		"%s\t%s\t%v\n",
+		provider.Arn,
+		provider.ClusterId,
+		providerInUse,
+	)
 }
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -137,7 +137,7 @@ type Client interface {
 	ListOCMRoles() ([]Role, error)
 	ListAccountRoles(version string) ([]Role, error)
 	ListOperatorRoles(version string, clusterID string) (map[string][]OperatorRoleDetail, error)
-	ListOidcProviders(targetClusterId string) ([]OidcProviderOutput, error)
+	ListOidcProviders(targetClusterId string, config *cmv1.OidcConfig) ([]OidcProviderOutput, error)
 	GetRoleByARN(roleARN string) (*iam.Role, error)
 	DeleteOperatorRole(roles string, managedPolicies bool) error
 	GetOperatorRolesFromAccountByClusterID(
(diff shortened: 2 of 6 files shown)


cmd/list/oidcprovider/cmd.go
```diff
@@ -113,21 +126,41 @@ func run(cmd *cobra.Command, _ []string) {
...
 	// Create the writer that will be used to print the tabulated results:
 	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
 	fmt.Fprintf(writer, "OIDC PROVIDER ARN\tCluster ID\tIn Use\n")
 	for _, provider := range providers {
-		providerInUse := "No"
-		if ok := providersInUse[provider.Arn]; ok {
-			providerInUse = "Yes"
+		if !filterProviders {
+			printProvider(writer, providersInUse, provider)
+		} else {
```
robpblake, line 140: @hunterkepley How about we extend or add another function to `AWSClient` that allows you to provide a filter when retrieving the list of OIDC providers? We can then move the filtering logic there rather than in the command implementation and we continue to have Cobra as a thin wrapper around our APIs.

It would also make it easier for you to write a unit test for the filtering logic :-)

pkg/aws/sts.go
```diff
@@ -262,10 +263,23 @@ func (c *awsClient) ListOidcProviders(targetClusterId string) ([]OidcProviderOut
...
-			providers = append(providers, OidcProviderOutput{
-				Arn:       *provider.Arn,
-				ClusterId: clusterId,
-			})
+			if idFilter != "" {
+				resourceId, err := GetResourceIdFromOidcProviderARN(*provider.Arn)
+				if err != nil {
+					return nil, fmt.Errorf("unable to get resource ID from OIDC Provider's ARN. Error: '%v'", err)
+				}
+				if strings.Contains(resourceId, idFilter) {
```
gdbranco, line 271: You could move this prior to the skip and consider this is a skip true when it doesn't contain instead. So that you can keep appending only once at the end of the flow

robpblake (approved):
/lgtm

hunterkepley:
Had to run mockgen to get ci to pass

pkg/aws/helpers.go
```diff
@@ -864,6 +864,13 @@ func GetResourceIdFromARN(stringARN string) (string, error) {
 	return parsedARN.Resource[index+1:], nil
 }
 
+func MockOidcConfig(id string, issuerUrl string) (*cmv1.OidcConfig, error) {
```
robpblake, line 867: @hunterkepley If this is only used for tests, then I'd suggest we create `test_helpers.go` file for clarity 👍

pkg/aws/client.go
```diff
@@ -93,6 +93,8 @@ var addROSAVersionToUserAgent = request.NamedHandler{
 }
 
 // Client defines a client interface
+//
+//go:generate mockgen -source=client.go -package=mocks -destination=mocks/client.go
```
robpblake, line 97: @hunterkepley If you rebase on `master`, I think a `mock` got merged as part of #1742

cmd/list/oidcprovider/cmd.go
```diff
@@ -41,11 +41,23 @@ var Cmd = &cobra.Command{
...
 	flags := Cmd.Flags()
 	flags.SortFlags = false
 	output.AddFlag(Cmd)
 	ocm.AddOptionalClusterFlag(Cmd)
+
+	flags.StringVarP(
+		&args.oidcConfigId,
+		"oidc-config-id",
+		"i",
+		"",
```
robpblake, line 58: @hunterkepley Do we want to include a `usage` string here to help the users?

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/AlecAivazis/survey/v2 v2.2.15
 	github.com/PuerkitoBio/goquery v1.8.1
 	github.com/aws/aws-sdk-go v1.45.26
+	github.com/aws/aws-sdk-go-v2 v1.24.1
```
robpblake, line 11: @hunterkepley Is this intentional?

robpblake:
@hunterkepley You have an import cycle in your test. I'm seeing this error:

```
cmd/list/instancetypes/cmd.go:1: : # github.com/openshift/rosa/cmd/list/instancetypes [github.com/openshift/rosa/cmd/list/instancetypes.test]
cmd/list/instancetypes/cmd_test.go:229:17: cannot use awsClient (variable of type *"github.com/openshift/rosa/pkg/aws".MockClient) as "github.com/openshift/rosa/pkg/aws".Client value in assignment: *"github.com/openshift/rosa/pkg/aws".MockClient does not implement "github.com/openshift/rosa/pkg/aws".Client (wrong type for method ListOidcProviders)
		have ListOidcProviders(string) ([]"github.com/openshift/rosa/pkg/aws".OidcProviderOutput, error)
		want ListOidcProviders(string, string, *"github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1".OidcConfig) ([]"github.com/openshift/rosa/pkg/aws".OidcProviderOutput, error) (typecheck)
/*
pkg/aws/client.go:1: : import cycle not allowed in test (typecheck)
/*
make: *** [Makefile:57: lint] Error 1
```

If you rebase on `master` and then use the `mock` introduced as part of #1742, I think that will solve it for you.

cheers,

Rob

go.mod
hunterkepley: nope :thinking: not sure why thats there

cmd/list/oidcprovider/cmd.go
hunterkepley: Oh, there should be one, on line 59

robpblake: My mistake. You're absolutely right...not sure how I missed that!

cmd/list/oidcprovider/cmd.go
```diff
@@ -66,8 +78,12 @@ func run(cmd *cobra.Command, _ []string) {
 		clusterKey := r.GetClusterKey()
 		clusterId = clusterKey
 	}
-
-	providers, err := r.AWSClient.ListOidcProviders(clusterId)
+	config, err := r.OCMClient.GetOidcConfig(args.oidcConfigId)
+	if err != nil {
+		r.Reporter.Errorf("Failed to get OIDC configs: %v", err)
```
gdbranco, line 83:
```suggestion
		r.Reporter.Errorf("Failed to get OIDC config: %v", err)
```

pkg/aws/sts.go
```diff
@@ -259,6 +261,15 @@ func (c *awsClient) ListOidcProviders(targetClusterId string) ([]OidcProviderOut
 					return providers, nil
 				}
 			}
+			if idFilter != "" {
```
gdbranco, line 264: Would it work to supply the config only and check if it's nil instead? One less param

pkg/aws/client_test.go
```diff
@@ -528,4 +533,65 @@ var _ = Describe("Client", func() {
...
+				config1, err := MockOidcConfig("config1", "http://oidc.test1/123123123123")
+				Expect(err).ToNot(HaveOccurred())
+				config2, err := MockOidcConfig("config2", "http://oidc.test2/234234234234")
+				Expect(err).ToNot(HaveOccurred())
+				config3, err := MockOidcConfig("config1", "http://oidc.test3/345345345345")
+				Expect(err).ToNot(HaveOccurred())
+				mockedOidcConfigs = []*cmv1.OidcConfig{config1, config2, config3}
+				mockIamAPI.EXPECT().ListOpenIDConnectProviders(gomock.Any()).Return(&iam.
+					ListOpenIDConnectProvidersOutput{OpenIDConnectProviderList: []*iam.OpenIDConnectProviderListEntry{
+					{Arn: aws.String(mockedOidcProviderArns[0])},
```
robpblake, line 555: @hunterkepley I think we can use `awsSdk.String`  which will remove the dependency on the AWS SDK v2 which you're currently importing as `aws.

pkg/aws/client_test.go
```diff
@@ -528,4 +533,65 @@ var _ = Describe("Client", func() {
...
+					{Arn: aws.String(mockedOidcProviderArns[1])},
+					{Arn: aws.String(mockedOidcProviderArns[2])},
+					{Arn: aws.String(mockedOidcProviderArns[3])},
+				}}, nil)
+				mockIamAPI.EXPECT().ListOpenIDConnectProviderTags(gomock.Any()).Return(
+					&iam.ListOpenIDConnectProviderTagsOutput{IsTruncated: aws.Bool(false), Marker: aws.String(""),
```
robpblake, line 561: @hunterkepley Same comment here for `aws.Bool`

robpblake (commented):
@hunterkepley Minor comment about removing dependency on v2 of the AWS SDK.

robpblake (approved):
/lgtm

PR #1795, opened 2024-02-21
OCM-1012 | feat: Oidc Config filter + unit tests

hunterkepley:
https://issues.redhat.com/browse/OCM-1012

Filters based on OCM UUID if managed, and if unmanaged, filters based on issuer URL

Automatically determines which one to filter depending on each individual oidc config, so you can search both managed and unmanaged with the same flag.

### DIFF
--- a/pkg/ocm/oidc_config.go
+++ b/pkg/ocm/oidc_config.go
@@ -15,6 +15,7 @@ package ocm
 
 import (
 	"fmt"
+	"strings"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 )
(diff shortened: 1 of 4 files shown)


pkg/ocm/oidc_config.go
```diff
@@ -40,7 +41,21 @@ func (c *Client) ListOidcConfigs(awsAccountId string) ([]*cmv1.OidcConfig, error
...
+
+func filterOidcConfigs(configs []*cmv1.OidcConfig, idFilter string) []*cmv1.OidcConfig {
+	var returnSlice []*cmv1.OidcConfig
+	if idFilter == "" {
+		return configs
+	}
+	for _, oidcConfig := range configs {
+		if (oidcConfig.Managed() && strings.Contains(oidcConfig.ID(), idFilter)) ||
+			(!oidcConfig.Managed() && strings.Contains(oidcConfig.IssuerUrl(), idFilter)) {
+			returnSlice = append(returnSlice, oidcConfig)
```
gdbranco, line 55: Could break just to not go over all

PR #1799, opened 2024-02-26
OCM-6279 | fix: Temporarily remove build output from rosa version command

robpblake:
This PR temporarily removes the `(Build: $GIT_SHA)` output from the `rosa version` command due to the fact that we are not reliably setting this on internal build systems for productised releases. We need to align on our approach throughout all build systems to ensure this is applied consistently, but for now, we remove the output from `rosa version` to reduce confusion for our customers.


ciaranRoche (approved):
/lgtm

PR #1803, opened 2024-02-26
OCM-1388 | fix: revert spinner on upgrade account role policies

davidleerh:


marcolan018:
why do we revert this change?

davidleerh:
We are reverting a portion due to trying to keep consistency where the spinner appears next to the throttling message.
We have decided not to pursue adding the spinner to individual aws api calls (requests) at the moment.
https://issues.redhat.com/browse/OCM-1388

PR #1804, opened 2024-02-26
OCM-1469 | fix: Fail commands when unknown arguments are passed

robpblake:
This PR updates the `cobra.Command` definition of all commands within the ROSA CLI to enforce the correct handling of positional arguments. In particular:

- Commands where positional args are not supported, now define `cobra.NoArgs` as their arg handler
- Commands where a particular number of args are supported, now define `cobra.MaximumNArgs` as their arg handler.
- Commands that handle arg parsing themselves, now define `cobra.ArbitraryArgs`

When passing unknown positional args to commands now, it will fail (printing help) with the following example message:

```
Failed to execute root command: unknown command "foo" for "rosa create cluster"
```

Additionally, I have added a test for the root command that ensures that all commands are well defined in that they:

- Define their usage (long and short) and provide an example
- Correctly set a `command.Args` function
- If a leaf command, set a `command.Run` function

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,7 @@ verify: fmt
 clean:
 	rm -rf \
 		./cover.out \
+		rosa \
 		rosa-darwin-amd64 \
 		rosa-darwin-arm64 \
 		rosa-linux-amd64 \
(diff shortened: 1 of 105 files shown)


ciaranRoche (commented):
There is a lot of changes here, over all LGTM I think it is worth merging ASAP so we can get it in the hands of QE to run regression on.

PR #1806, opened 2024-02-26
Change 'RunE' calls to just 'Run'

AlexVulaj:
https://github.com/openshift/rosa/pull/1779 changed any `RunE` calls to just use `Run`, which means that any other commands calling those `RunE` functions would now panic due to calling a nil function. This PR changes those calls to use the new `Run`.


willkutler (approved):

ciaranRoche (commented):
/unhold

mikeshng:
I think this PR might be involved in a bug related to cluster upgrade. Please see https://issues.redhat.com/browse/OCM-6619 for more details CC @AlexVulaj

PR #1810, opened 2024-02-29
OCM-6435 | fix: Bound math.MaxInt64 to avoid overflow on non-64bit arch

damdo:
math.MaxInt64 needs to be bound by int64() as that can otherwise overflow if the helpers are compiled for non-64bit architectures (e.g. armv7l) as per https://github.com/golang/go/issues/23086.

Context:
Now that ROSA control planes are supported in the upstream Cluster API Provider AWS, the code used in the provider needs to also be compatible with arm architectures (which are a valid CAPA target arch)
Ref: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4835

As such it is desirable for the ROSA helpers to be compilable for arm. The issue this PR fixes prevents that from happening. Thus this PR.


ciaranRoche (approved):
/approve

PR #1812, opened 2024-02-29
OCM-4613 | feat: support customize username for cluster admin creation

marcolan018:

### DIFF
--- a/cmd/create/admin/cmd.go
+++ b/cmd/create/admin/cmd.go
@@ -17,6 +17,7 @@ limitations under the License.
 package admin
 
 import (
+	"fmt"
 	"os"
 
 	idputils "github.com/openshift-online/ocm-common/pkg/idp/utils"
@@ -30,6 +31,8 @@ import (
 )
 
 const ClusterAdminUsername = "cluster-admin"
+const ClusterAdminGroupname = "cluster-admins"
+const DedicatedAdminGroupname = "dedicated-admins"
 const ClusterAdminIDPname = "cluster-admin"
 const GeneratingRandomPasswordString = "Generating random password"
 const MaxPasswordLength = 23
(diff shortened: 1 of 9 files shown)


davidleerh:
hi @marcolan018 
Just wanted to confirm, 
is the custom name only available for "create cluster" command / step at this time? (via your implementation)
It seems the create, delete and describe admin commands target the htpassword with name cluster-admin but doesn't target custom names. Please advise if I am looking at this wrong.

marcolan018:
@davidleerh 
yes, the `admin` related rosa commands only handle the `cluster-admin` user

cmd/create/admin/cmd.go
```diff
@@ -176,7 +188,25 @@ func run(cmd *cobra.Command, _ []string) {
 	r.Reporter.Infof("It may take several minutes for this access to become active.")
 }
 
-func FindExistingClusterAdminIDP(cluster *cmv1.Cluster, r *rosa.Runtime) (
+// find the htpasswd idp "cluster-admin"
+func FindClusterAdminIDP(cluster *cmv1.Cluster, r *rosa.Runtime) (htpasswdIDP *cmv1.IdentityProvider) {
```
robpblake, line 192: @marcolan018 Could you add a unit test for this function and I think the use of named returns is not something we encourage on this project.

marcolan018: @robpblake 
added the unit test and made refactoring on the returns for the functions
please help to review again, thanks

cmd/create/admin/cmd.go
```diff
@@ -176,8 +188,25 @@ func run(cmd *cobra.Command, _ []string) {
...
 }
 
-func FindExistingClusterAdminIDP(cluster *cmv1.Cluster, r *rosa.Runtime) (
-	htpasswdIDP *cmv1.IdentityProvider, userList *cmv1.HTPasswdUserList) {
+// find the htpasswd idp "cluster-admin"
+func FindClusterAdminIDP(cluster *cmv1.Cluster, r *rosa.Runtime) *cmv1.IdentityProvider {
+	idps, err := r.OCMClient.GetIdentityProviders(cluster.ID())
+	if err != nil {
+		r.Reporter.Errorf("Failed to get identity providers for cluster '%s': %v", r.ClusterKey, err)
+		os.Exit(1)
```
robpblake, line 196: @marcolan018 I'd personally suggest returning the  `error` from this function rather than `os.Exit`, then you can test the error flow as well.

cmd/create/admin/cmd.go
```diff
@@ -72,16 +72,18 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
-	// Try to find an existing cluster admin IDP
-	existingClusterAdminIDP, _ := FindExistingClusterAdminIDP(cluster, r)
-	if existingClusterAdminIDP != nil {
-		r.Reporter.Errorf("Cluster '%s' already has an admin", clusterKey)
+	adminUser, err := r.OCMClient.GetUser(cluster.ID(), "cluster-admins", ClusterAdminUsername)
```
robpblake, line 75: @marcolan018 Is it worth defining a `const` for the `cluster-admin` group? I see we have one for `ClusterAdminUsername`.

cmd/create/admin/cmd.go
marcolan018: changed 
@robpblake please help to re-review the latest commits, thanks

cmd/create/admin/cmd.go
marcolan018: added new const

robpblake:
Thanks for the additional review @davidleerh

PR #1818, opened 2024-03-04
OCM-6467 | fix: shouldn't automatically start an interactive mode for domain prefix as it's optional

machi1990:
Initially, the flow envisaged for the domain prefix was the same as the name, however this was wrong since the flag is meant to be optional and it turned out that this is going to block QE since as the flag is non required. Removing launching the interactive mode automatically when the flag is not supplied; the implication here is that the backend will generate the this prefix or use the name if it's less than 15 chars. 

I want to get this in to unblock the current QE flows, and we'll further iterate on the flow for the domain prefix once QE have run through this and we've received internal feedbacks from demos.

JIRA: OCM-6467


ciaranRoche (approved):
Will merge this now to unblock QE pipelines

machi1990:
@ciaranRoche I've pushed the changes with the confirmation when name > 15 chars and domain_prefix not supplied.
Can you give this another look? Thanks

ciaranRoche (approved):
/lgtm

cmd/create/cluster/cmd.go
```diff
@@ -996,6 +991,18 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
+	if domainPrefix == "" && len(clusterName) > ocm.MaxClusterDomainPrefixLength {
```
machi1990, line 994: looks like this made codecov unhappy. I'll try to update the tests

machi1990: Extracted this logic onto something that I can test. Let's see if that will make CI happy.

machi1990: done

ciaranRoche (approved):
/lgtm

PR #1820, opened 2024-03-04
OCM-6484 | fix: generates a valid operator role prefix from the cluster name

machi1990:
A valid operator role prefix has at most 32 characters, but the cluster name can have up to 54 characters this was causing errors as the one reported in https://issues.redhat.com/browse/OCM-6484. To fix this issue, we use only the first 27 characters of the name if the name is longer than 27 characters.

### DIFF
--- a/pkg/helper/roles/helpers.go
+++ b/pkg/helper/roles/helpers.go
@@ -20,8 +20,21 @@ import (
 )
 
 const (
-	RosaUpgradeAccRolesModeAuto = "ROSAUpgradeAccountRolesModeAuto"
+	RosaUpgradeAccRolesModeAuto            = "ROSAUpgradeAccountRolesModeAuto"
+	maxClusterNameLengthToUseForRolePrefix = 27
 )
+
+// GeOperatorRolePrefixFromClusterName returns a valid operator role prefix from the cluster name
+// An operator role prefix is considered valid if it's length is less than or equal to 32 chars.
+// A random 4 characters label is attached to the cluster name to reduce chances of collision.
+// The cluster name and the random label are separate by '-'.
+// If the cluster name is longer than 27 characters, only the first 27 characters will be used.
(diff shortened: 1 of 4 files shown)


pkg/helper/roles/helpers.go
```diff
@@ -23,6 +23,19 @@ const (
 	RosaUpgradeAccRolesModeAuto = "ROSAUpgradeAccountRolesModeAuto"
 )
 
+// GeOperatorRolePrefixFromClusterName returns a valid operator role prefix from the cluster name
+// An operator role prefix is considered valid if it's length is less than or equal to 32 chars.
+// A random 4 characters label is attached to the cluster name to reduce chances of collision.
+// The cluster name and the random label are separate by '-'.
+// If the cluster name is longer than 27 characters, only the first 27 characters will be used.
+func GeOperatorRolePrefixFromClusterName(clusterName string) string {
+	maxClusterLengthToUse := 27
```
robpblake, line 32: @machi1990 Nit: Shall we make this a `const`?

pkg/helper/roles/helpers.go
```diff
@@ -23,6 +23,19 @@ const (
 	RosaUpgradeAccRolesModeAuto = "ROSAUpgradeAccountRolesModeAuto"
 )
 
+// GeOperatorRolePrefixFromClusterName returns a valid operator role prefix from the cluster name
```
robpblake, line 26: @machi1990 Thanks for the GoDoc. We definitely need more of this 👍

pkg/helper/roles/helpers.go
machi1990: done

pkg/helper/roles/helpers.go
machi1990: 👍 Yes.

robpblake (approved):

PR #1821, opened 2024-03-04
OCM-4969 | feat: Add deprecation warning for login using offline token and disable oauth tokens for FedRAMP

tirthct:
**Changed**
- Add deprecation warning for login using offline token
- Disable oauth tokens for FedRAMP

**Tested**
- Warning message should be displayed in case of offline tokens : tested

### DIFF
--- a/cmd/login/cmd.go
+++ b/cmd/login/cmd.go
@@ -169,19 +169,21 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
-	// Load the configuration file:
-	cfg, err := config.Load()
-	if err != nil {
-		r.Reporter.Errorf("Failed to load config file: %v", err)
+	// Confirm that token is not passed with auth code flags
+	if (args.useAuthCode || args.useDeviceCode) && args.token != "" {
+		r.Reporter.Errorf("Token cannot be passed with '--use-auth-code' or '--use-device-code' commands")
 		os.Exit(1)
-	}
-	if cfg == nil {
-		cfg = new(config.Config)
 	}
 
-	if (cfg.FedRAMP || fedramp.HasFlag(cmd) || fedramp.IsGovRegion(arguments.GetRegion())) && (args.useAuthCode || args.useDeviceCode) {
+	// FedRAMP does not support oauth code flow login yet
+	if fedramp.HasFlag(cmd) && (args.useAuthCode || args.useDeviceCode) {
 		r.Reporter.Errorf("This login method is currently not supported with FedRAMP")
 		os.Exit(1)
+	}
+
+	// Show deprecation message if offline token is passed
+	if !fedramp.HasFlag(cmd) && args.token != "" {
+		r.Reporter.Warnf(offlineTokenDeprecationMessage)
 	}
 
 	if args.useAuthCode {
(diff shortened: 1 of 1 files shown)


cmd/login/cmd.go
```diff
@@ -164,6 +169,21 @@ func run(cmd *cobra.Command, argv []string) {
...
+	cfg, err := config.Load()
+	if err != nil {
+		r.Reporter.Errorf("Failed to load config file: %v", err)
+		os.Exit(1)
+	}
+	if cfg == nil {
+		cfg = new(config.Config)
+	}
+
+	if (cfg.FedRAMP || fedramp.HasFlag(cmd) || fedramp.IsGovRegion(arguments.GetRegion())) && (args.useAuthCode || args.useDeviceCode) {
```
tylercreller, line 182: We want to avoid using the stored configuration for this check. It will prevent someone from switching between govcloud and commercial environments if `"fedramp": "true"` is stored in their configuration.

cmd/login/cmd.go
```diff
@@ -325,6 +335,10 @@ func run(cmd *cobra.Command, argv []string) {
 				clientID = args.clientID
 			}
 		}
+	} else {
+		if !args.useDeviceCode && !args.useDeviceCode {
```
tylercreller, line 339: We should specifically target displaying this message only if the offline token is provided. In this case we would display the warning message for client credentials, which we don't want.

cmd/login/cmd.go
```diff
@@ -164,6 +169,21 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
+	// Load the configuration file:
+	cfg, err := config.Load()
+	if err != nil {
+		r.Reporter.Errorf("Failed to load config file: %v", err)
+		os.Exit(1)
+	}
+	if cfg == nil {
```
robpblake, line 178: @tirthct You don't need this check as the `config.Load()` method will always return an empty `Config` if one doesn't exist.

cmd/login/cmd.go
robpblake: @tirthct @tylercreller Can we combine this statement into a function and write a unit test for it e.g something like `fedramp.IsLoginMethodSupported()`

cheers,

Rob

tirthct: @robpblake turns out we don't need the extra conditions except for fedramp.HasFlag() check. So just kept that instead of creating a whole new method.
Fixed the condition

cmd/login/cmd.go
tirthct: Ack. Added offline token condition.

cmd/login/cmd.go
tirthct: I don't think that check would be working properly. I removed it and it stopped me from logging in:
image

Not looking into that in this PR this change is already in the master branch for long, but def something to look into.

tirthct: PS. ignore the double warnings. For some reason my machine sometimes does that. 
@tylercreller has verified and confirmed the message is only shown once

tylercreller:
Hold this PR until we get the green-light from PM

tirthct:
> Hold this PR until we get the green-light from PM

@tylercreller : Converted to draft

robpblake:
@tirthct What's the status of this PR please?

cheers,
Rob

tirthct:
@tylercreller should we merge this?

tylercreller:
@tirthct @robpblake We are still waiting for the appropriate time to merge this. There are internal processes that need to happen before we can ship this.

robpblake:
@tirthct / @tylercreller What's our status on this one please?
cheers,
Rob

tylercreller:
@robpblake Still pending green-light from PM - expecting more information following the disablement of internal offline tokens.

tylercreller:
@tirthct Let's go ahead and close this. I'll be refactoring this work into all commands at the client builder level, rather than just on log in. Appreciate your work on this.

PR #1822, opened 2024-03-05
OCM-6496 | chore: Remove error from reporter.Builder.Build method signature

robpblake:
This PR removes the `error` from the `reporter.Builder.Build()` method signature as it was not used and lead to unnecessary error handling and propagation. Instead, replaced with a single `reporter.CreateReporter` function and removed the `Builder` functionality completely.

### DIFF
--- a/pkg/reporter/reporter_test.go
+++ b/pkg/reporter/reporter_test.go
@@ -1,0 +1,178 @@
+package reporter
+
+import (
+	"io"
+	"os"
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/openshift/rosa/pkg/color"
+	"github.com/openshift/rosa/pkg/debug"
+)
+
+func TestReporter(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "reporter testing")
+}
+
+var _ = Describe("Reporter Tests", func() {
+
+	var reporter *Object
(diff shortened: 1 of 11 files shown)


robpblake:
This PR is a follow up on the discussions on #1777

robpblake:
> /approve cancel

Removing the `approved` label from my own PR to ensure it gets a proper review.

ciaranRoche (approved):
/lgtm

pkg/reporter/reporter_test.go
```diff
@@ -0,0 +1,178 @@
...
+			stdOut, stdErr := captureStdOutAndStdError(func() {
+				reporter.Debugf("Hello World")
+			})
+			Expect(stdOut).To(Equal(infoColorPrefix + "Hello World\n"))
+			Expect(stdErr).To(BeEmpty())
+		})
+	})
+})
+
+func captureStdOutAndStdError(function func()) (string, string) {
```
ciaranRoche, line 154: Leaving a note, we probably need to move funcs like this to a helper dir, might be worth capturing in the testing DDR

ciaranRoche (approved):
/lgtm

pkg/reporter/reporter_test.go
ciaranRoche: excluded or tested :laughing:

robpblake: It's test all the way down.....

PR #1825, opened 2024-03-06
OCM-4966 | feat: Add keyring support for configuration

tylercreller:
[OCM-4966](https://issues.redhat.com//browse/OCM-4966)

- Added keyring support for configuration storage via `OCM_KEYRING`
   - Added README docs
   - Added keyring support for config saves, loads, removal
   - Removes configuration from keyring on logout
- Added basic login tests
- Added keyring tests for config

**Other**
- Fixed support for client-id / client-secret login
- Added support for `preferred_username` claim, required for organizational service accounts


cmd/config/get/get.go
```diff
@@ -54,20 +55,26 @@ func run(_ *cobra.Command, argv []string) {
...
-	if cfg == nil {
-		loc, err := config.Location()
+	cfg := &config.Config{}
+	var err error
+	if !skipConfigLoadMap[arg] {
+		// Load the configuration:
+		cfg, err = config.Load()
 		if err != nil {
-			return fmt.Errorf("Failed to find config file location: %v", err)
+			return fmt.Errorf("can't load config: %v", err)
```
tirthct, line 70: @robpblake suggested to show the config file location in the error message (https://github.com/openshift/rosa/pull/1780#discussion_r1505723401). Should we keep it that way?

cmd/config/get/get.go
```diff
@@ -54,20 +55,26 @@ func run(_ *cobra.Command, argv []string) {
...
+		}
+		// If the configuration file doesn't exist yet assume that all the configuration settings
+		// are empty:
+		if cfg == nil {
+			fmt.Printf("\n")
```
tirthct, line 75: Is this needed? If so, we can keep it consistent with other print functions (fmt.Fprintf)

cmd/config/get/get.go
```diff
@@ -90,8 +97,24 @@ func PrintConfig(arg string) error {
 		fmt.Fprintf(Writer, "%s\n", cfg.URL)
 	case "fedramp":
 		fmt.Fprintf(Writer, "%v\n", cfg.FedRAMP)
+	case "keyrings":
+		keyrings, err := getKeyrings()
+		if err != nil {
+			return err
+		}
+		for _, keyring := range keyrings {
+			fmt.Fprintf(os.Stdout, "%s\n", keyring)
```
tirthct, line 106: Might be easier if we can print using %v? Removed the need for looping

cmd/login/cmd.go
```diff
@@ -136,23 +138,32 @@ func init() {
...
+			"not available. See --use-auth-code for all other scenarios.",
 	)
 	arguments.AddRegionFlag(flags)
 	fedramp.AddFlag(flags)
 }
 
 func run(cmd *cobra.Command, argv []string) {
-	ctx := cmd.Context()
 	r := rosa.NewRuntime()
+	defer r.Cleanup()
```
tirthct, line 157: I think we need to cleanup only in case of `WithOCM()`. We can remove this.

cmd/login/cmd.go
```diff
@@ -345,15 +358,14 @@ func run(cmd *cobra.Command, argv []string) {
 			// If a token has been provided parse it:
 			jwtToken, err := config.ParseToken(token)
 			if err != nil {
-				r.Reporter.Errorf("Failed to parse token: %v", err)
-				os.Exit(1)
+				fmt.Println("yep: ", token)
```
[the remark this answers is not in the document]
tylercreller: yep 😄

cmd/config/get/get.go
tylercreller: `config.Load() -> loadFromFile` will return the file path if it fails to load from a file. I am removing cases of the word "file" where errors can come from the keyring OR the file now.

pkg/config/config.go
```diff
@@ -84,15 +123,27 @@ func Save(cfg *Config) error {
 	if err != nil {
 		return err
 	}
+
+	data, err := json.MarshalIndent(cfg, "", "  ")
+	if err != nil {
+		return fmt.Errorf("can't marshal config: %v", err)
+	}
+
+	keyring := os.Getenv("RH_KEYRING")
```
robpblake, line 132: @tylercreller How about a function `IsKeyRingManaged` (or something) that returns a `bool`. Means we have a single location that checks the value of the `RH_KEYRING` env variable.

cmd/login/cmd.go
```diff
@@ -161,8 +172,16 @@ func run(cmd *cobra.Command, argv []string) {
 	// Check mandatory options:
 	env := args.env
 	if env == "" {
-		r.Reporter.Errorf("Option '--env' is mandatory")
-		os.Exit(1)
+		return fmt.Errorf("Option '--env' is mandatory")
+	}
+
+	// Fail fast if RH_KEYRING is provided and invalid
+	keyring := os.Getenv("RH_KEYRING")
```
robpblake, line 179: @tylercreller Shall we introduce a `const` for `RH_KEYRING` as we use it quite a bit.

pkg/config/config.go
tylercreller: Good call-outs, thanks Rob. Updated.

cmd/login/cmd.go
tylercreller: Updated

cmd/config/get/get.go
```diff
@@ -90,8 +97,24 @@ func PrintConfig(arg string) error {
 		fmt.Fprintf(Writer, "%s\n", cfg.URL)
 	case "fedramp":
 		fmt.Fprintf(Writer, "%v\n", cfg.FedRAMP)
+	case "keyrings":
```
robpblake, line 100: @tylercreller This isn't really reading the users config right? It's listing available keyrings that the user can use. That's a small, but different change in functionality to all of the other `rosa config get` commands and  I kinda feel like you've covered that in the `Prints the value of a config variable. See 'rosa config --help' for supported config variables.` output of this command if I run `rosa help config get`.

My personal expectation is that a `rosa config get keyring` would return the keyring I'm currently using (if that makes sense), but I would expect to read the list of available keyrings from docs or `rosa config help` command.

WDYT and also would be good to get some other opinions here rather than just me :-)

robpblake: @tylercreller Thanks for the explanation and I understand the intent.

Did you consider `rosa config set keyring $KEYSTORE` versus the environment variable implementation? It would restrict `rosa` to managing a single external environment variable (`OCM_CONFIG`) and might offer another option here in terms of helping the user see the list of supported keyrings e.g:

* `rosa config set keyring`
  * Outputs: "Selected keyring is not valid. Valid options are: "foo", "bar" 
* `rosa config set keyring doesntExist` 
  * Outputs: "Selected keyring 'doesntExist' is not valid. Valid options are: "foo", "bar" 
* `rosa config set keyring keychain`
* `rosa config get keyring`

This would leave the config file managing _only_ the selected keyring. The other values in the config are completely ignored. 

I don't want to over-engineering this/get too hung up on it. So let's get other opinions from other members of the team and make a decision.

tylercreller: Always appreciate your thoughtful input @robpblake. We did originally consider an implementation like this. We opted to not go this route for similar consistency reasons. The keyring selection cannot be stored in the config file itself since the keyring stores the config data as an alternative to the file. This would be equivalent of storing the config file path in the config file itself. `rosa config set keyring` would still need to store that value in another file/variable/etc.

To maintain consistency we opted for the keyring env variable to match parity with standards set from `OCM_CONFIG` since they are mutually exclusive and handle similar operations.

robpblake: Thanks @tylercreller 

My current vote then based upon your explanation is to drop the `rosa config get keyrings` and rely on  documentation. I'd prefer consistency of the command semantics e.g `rosa config get` reads your actual applied config versus potential options for you.

If we get feedback that folks want to see available options for config values, then I think would could look at something like `rosa config defaults` or `rosa config describe $config` or some other route.

cheers,

Rob

tylercreller: Updated based on our discussions @robpblake

PR #1827, opened 2024-03-06
OCM-6515 | fix: Removed output flag from rosa token command

tirthct:
**Changed**
- Removed output flag from rosa token command since we only have string output in this case

https://issues.redhat.com/browse/OCM-6515


robpblake (approved):
/lgtm

PR #1829, opened 2024-03-06
OCM-6106 | node drain grace period for hosted control planes

philipwu08:
Support node drain grace period for NodePools of HyperShift-enabled clusters. The node drain grace period is supported for these operations:
- Create machinepool (nodepool)
- Update machinepool
- Describe machinepool

https://issues.redhat.com/browse/OCM-6106

### DIFF
--- a/cmd/describe/machinepool/cmd_test.go
+++ b/cmd/describe/machinepool/cmd_test.go
@@ -31,8 +31,9 @@ Subnet:
 Version:                               4.12.24
 Autorepair:                            No
 Tuning configs:                        
-Message:                               
 Additional security group IDs:         
+Node drain grace period:               1 minute
+Message:                               
 `
 	describeStringWithUpgradeOutput = `
 ID:                                    nodepool85
@@ -48,8 +49,9 @@ Subnet:
 Version:                               4.12.24
 Autorepair:                            No
 Tuning configs:                        
-Message:                               
 Additional security group IDs:         
+Node drain grace period:               1 minute
+Message:                               
 Scheduled upgrade:          scheduled 4.12.25 on 2023-08-07 15:22 UTC
 `
 
@@ -59,6 +61,9 @@ aws_node_pool:
   kind: AWSNodePool
 id: nodepool85
 kind: NodePool
+node_drain_grace_period:
+  unit: minute
+  value: 1
 scheduledUpgrade:
   nextRun: 2023-08-07 15:22 UTC
   state: scheduled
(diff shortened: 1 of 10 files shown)


pkg/helper/machinepools/helpers.go
```diff
@@ -221,3 +222,37 @@ func HostedClusterOnlyFlag(r *rosa.Runtime, cmd *cobra.Command, flagName string)
...
+	valueBuilder := cmv1.NewValue()
+	if nodeDrainGracePeriod == "" {
+		return valueBuilder
+	}
+
+	nodeDrainParsed := strings.Split(nodeDrainGracePeriod, " ")
+	if len(nodeDrainParsed) > 2 {
+		r.Reporter.Errorf("Invalid value for node drain grace period. Expects only the duration and " +
+			"the unit (minute|minutes|hour|hours).")
+		os.Exit(1)
```
robpblake, line 236: @philipwu08 Can you please make this function return an `error` and not `os.Exit(1)` so it can be tested. Thanks!

pkg/ocm/output/nodepools.go
```diff
@@ -91,3 +91,15 @@ func PrintNodePoolTuningConfigs(tuningConfigs []string) string {
 	}
 	return strings.Join(tuningConfigs, ",")
 }
+
+func PrintNodeDrainGracePeriod(period *cmv1.Value) string {
```
robpblake, line 95: @philipwu08 Could you add a unit test for this please

cmd/edit/machinepool/cmd.go
```diff
@@ -132,6 +135,17 @@ func init() {
 			"This list will overwrite any modifications made to node tuning configs on an ongoing basis.",
 	)
 
+	flags.StringVar(&args.nodeDrainGracePeriod,
+		"node-drain-grace-period",
+		"",
+		"You may set a grace period for how long Pod Disruption Budget-protected workloads will be "+
+			"respected when the NodePool is replaced or upgraded.\nAfter this grace period, all remaining workloads "+
+			"will be forcibly evicted.\n"+
+			"Valid value is from 0 to 1 week (10080 minutes), and the supported units are 'minute|minutes' or "+
```
robpblake, line 144: @philipwu08 Are there any plans to support `day|days|week|weeks` as units?

cmd/describe/machinepool/cmd_test.go
```diff
@@ -48,8 +49,9 @@ Subnet:
 Version:                               4.12.24
 Autorepair:                            No
 Tuning configs:                        
-Message:                               
 Additional security group IDs:         
+Node drain grace period:
```
robpblake, line 53: @philipwu08 We should extend this test to ensure that the `Node drain grace period` is correctly printed when it is set 👍

cmd/edit/machinepool/cmd.go
philipwu08: @robpblake Not at the moment. The reason is to be consistent with our existing node drain grace period on the cluster, which supports minutes and hours only.

philipwu08:
@robpblake Thanks for the review. I think I've addressed all the comments, but please let me know if I've missed something.

chenz4027 (approved):
/lgtm
Let's see what's Rob's feedback next week

cmd/edit/machinepool/cmd.go
robpblake: Thanks for the explanation @philipwu08

robpblake (approved):

PR #1831, opened 2024-03-07
OCM-6516 | fix: Improved help text for config get/set commands

tirthct:
**Changed**
- Improved help text for config get/set commands

https://issues.redhat.com/browse/OCM-6516

### DIFF
--- a/cmd/config/get/get.go
+++ b/cmd/config/get/get.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 
@@ -37,9 +38,10 @@ func NewConfigGetCommand() *cobra.Command {
 	return &cobra.Command{
 		Use:   "get [flags] VARIABLE",
 		Short: "Prints the value of a config variable",
-		Long:  "Prints the value of a config variable. See 'rosa config --help' for supported config variables.",
-		Args:  cobra.ExactArgs(1),
-		Run:   run,
+		Long: fmt.Sprintf("Prints the value of a config variable. Supported variables are:\n%s",
+			strings.Join(config.GetAllConfigProperties(), "\n")),
+		Args: cobra.ExactArgs(1),
+		Run:  run,
 	}
 }
 
--- a/cmd/config/set/set.go
+++ b/cmd/config/set/set.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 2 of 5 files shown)


pkg/config/config.go
```diff
@@ -46,6 +48,31 @@ type Config struct {
 	FedRAMP      bool     `json:"fedramp,omitempty" doc:"Indicates FedRAMP."`
 }
 
+var DisallowedSetConfigProperties = []string{"scopes"}
+
+func ConfigPropertiesNamesAndDocs() ([]string, []string) {
```
robpblake, line 53: @tirthct Could we add a unit test for this please

cmd/config/get/get.go
```diff
@@ -37,12 +38,18 @@ func NewConfigGetCommand() *cobra.Command {
...
-		Args:  cobra.ExactArgs(1),
-		Run:   run,
+		Long: fmt.Sprintf("Prints the value of a config variable. Supported variables are: %s.",
+			strings.Join(GetAllConfigProperties(), ", ")),
+		Args: cobra.ExactArgs(1),
+		Run:  run,
 	}
 }
 
+func GetAllConfigProperties() []string {
```
robpblake, line 48: @tirthct Shall we move this into `config.go` so it's closer to the `ConfigPropertiesNamesAndDocs()` function and the pattern remains that the `command` implementations simply call a function and do minimal logic themselves?

cmd/config/set/set.go
```diff
@@ -33,10 +36,22 @@ func NewConfigSetCommand() *cobra.Command {
...
-		Args:  cobra.ExactArgs(2),
-		Run:   run,
+		Long: fmt.Sprintf("Sets the value of a config variable. Supported variables are: %s.",
+			strings.Join(getAllowedConfigProperties(), ", ")),
+		Args: cobra.ExactArgs(2),
+		Run:  run,
+	}
+}
+
+func getAllowedConfigProperties() []string {
```
robpblake, line 46: @tirthct Same comment here about moving this into `config.go`. And also a unit test.

cmd/config/get/get.go
```diff
@@ -37,12 +38,18 @@ func NewConfigGetCommand() *cobra.Command {
 	return &cobra.Command{
 		Use:   "get [flags] VARIABLE",
 		Short: "Prints the value of a config variable",
-		Long:  "Prints the value of a config variable. See 'rosa config --help' for supported config variables.",
-		Args:  cobra.ExactArgs(1),
-		Run:   run,
+		Long: fmt.Sprintf("Prints the value of a config variable. Supported variables are: %s.",
+			strings.Join(GetAllConfigProperties(), ", ")),
```
robpblake, line 42: @tirthct The general pattern seems to be to print supported values on a new line e.g `rosa help config`

pkg/config/config.go
tirthct: Added tests

cmd/config/get/get.go
tirthct: Makes sense. Moved

cmd/config/set/set.go
tirthct: Moved.

cmd/config/get/get.go
tirthct: Changed to print in new line

robpblake (approved):
/lgtm

PR #1833, opened 2024-03-11
OCM-4969 | fix: Add validation to not pass token or fedramp flag for OAuth login

tirthct:
**Changed**
- Add validation to not pass token or fedramp flag for OAuth login
https://issues.redhat.com/browse/OCM-4969


tylercreller:
LGTM

PR #1790, opened 2024-02-20
OCM-1473 | fix: filter out local/wavelength zone in cluster creation

marcolan018:
https://issues.redhat.com/browse/OCM-1473
Signed-off-by: marcolan018 <llan@redhat.com>

PR #1834, opened 2024-03-11
OCM-6592 | fix: filter out local-zone and wave-length zone associated subnets during 'rosa create machinepool (HCP)'

davidleerh:
BEFORE: (shows local-zone subnets)
image

AFTER: (local-zone subnets filtered away)
image

### DIFF
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -77,6 +77,7 @@ const (
 	Inline        = "inline"
 	Attached      = "attached"
 
+	standardZone   = "availability-zone"
 	LocalZone      = "local-zone"
 	WavelengthZone = "wavelength-zone"
 
@@ -116,8 +117,9 @@ type Client interface {
 	GetSubnetAvailabilityZone(subnetID string) (string, error)
 	GetAvailabilityZoneType(availabilityZoneName string) (string, error)
 	GetVPCSubnets(subnetID string) ([]*ec2.Subnet, error)
-	GetVPCPrivateSubnets(subnetID string) ([]*ec2.Subnet, error)
-	FilterVPCsPrivateSubnets(subnets []*ec2.Subnet) ([]*ec2.Subnet, error)
+	GetVPCPrivateSubnets(isHostedCp bool, subnetID string) ([]*ec2.Subnet, error)
+	FilterVPCsPrivateSubnets(isHostedCp bool, subnets []*ec2.Subnet) ([]*ec2.Subnet, error)
+	FilterSubnetsWithStandardAvailabilityZones(subnets []*ec2.Subnet) ([]*ec2.Subnet, error)
 	ValidateQuota() (bool, error)
 	TagUserRegion(username string, region string) error
 	GetClusterRegionTagForUser(username string) (string, error)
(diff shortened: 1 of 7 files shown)


cmd/create/cluster/cmd_test.go
```diff
@@ -474,7 +474,7 @@ var _ = Describe("getInitialValidSubnets()", func() {
 	})
 
 	It("OK: get valid subnets", func() {
-		validSubnets, err := getInitialValidSubnets(mockClient, ids, r.Reporter)
+		validSubnets, err := getInitialValidSubnets(false, mockClient, ids, r.Reporter)
```
robpblake, line 477: @davidleerh Would it be possible to add a test for setting the `hostedCp` variable to `true`.

davidleerh: my mistake here. 
After review.
It seems it was already implemented on the "rosa create cluster" side of things via 
https://github.com/openshift/rosa/pull/1790

Will remove the change on "rosa create cluster" -> "getInitialValidSubnets".

davidleerh:
Confirmed as of 03/13/2024
Local Zone AZ Subnets are only supported or Day 2 Machine Pools (Classic Clusters).

pkg/aws/client.go
```diff
@@ -611,13 +613,53 @@ func (c *awsClient) FilterVPCsPrivateSubnets(subnets []*ec2.Subnet) ([]*ec2.Subn
...
+	}
+
+	describeAvailabilityZonesOutput, err := c.ec2Client.DescribeAvailabilityZones(
+		&ec2.DescribeAvailabilityZonesInput{ZoneNames: subnetAvailabilityZonesList})
+	if err != nil {
+		return filteredSubnets, err
+	}
+
+	for _, availabilityZone := range describeAvailabilityZonesOutput.AvailabilityZones {
+		if *availabilityZone.ZoneType == standardZone {
```
andreadecorte, line 648: any risk to do this instead of filtering out the 2 types we don't want? It means that if there is another type we risk missing that too

davidleerh: As of now, the only 3 types of az types exposed via AWS SDK are,
"availability-zone (standard)", "local-zone", "wavelength-zone".
These az types seem to be created and classified based on how close they are to edge locations.

I am thinking that we on ly expose to customers, subnets / az types that are already confirmed / tested and compatible with Openshift, vs. having the possibility of exposing new subnets / az types if they become readily available via the SDK, without our knowledge - if we don't keep track.

pkg/aws/client.go
```diff
@@ -611,13 +613,53 @@ func (c *awsClient) FilterVPCsPrivateSubnets(subnets []*ec2.Subnet) ([]*ec2.Subn
...
+	subnetAvailabilityZonesList := []*string{}
+	standardAvailabilityZonesMap := map[string]bool{}
+
+	// list of availability zones to pass as input
+	for _, subnet := range subnets {
+		subnetAvailabilityZonesList = append(subnetAvailabilityZonesList, subnet.AvailabilityZone)
+	}
+
+	describeAvailabilityZonesOutput, err := c.ec2Client.DescribeAvailabilityZones(
+		&ec2.DescribeAvailabilityZonesInput{ZoneNames: subnetAvailabilityZonesList})
```
oriAdler, line 642: What happens in case there are duplicates in the zone names? I would assume AWS will ignore the duplicates.

davidleerh: Thanks for pointing this out.
From my manual test, I duplicates seem to be ignored.

robpblake:
@davidleerh What's the status of this PR please?

Shall we rebase and get it merged or can it be closed?

cheers,

Rob

pkg/aws/client.go
andreadecorte: thanks, makes sense

andreadecorte:
> @davidleerh What's the status of this PR please?
> 
> Shall we rebase and get it merged or can it be closed?
> 
> cheers,
> 
> Rob

+1 for rebase and merge from me

davidleerh:
fixed merge conflicts in new PR:
https://github.com/openshift/rosa/pull/1939

PR #1837, opened 2024-03-12
OCM-5677 | feat: Added `rosa describe autoscaler` command

robpblake:
This PR adds `rosa describe autoscaler` to the ROSA CLI to allow customers to view details of their cluster autoscaler.

This PR attempts to introduced some of the concepts from the ROSA unit testing DDR to explore how we can make our commands more test friendly.

### DIFF
--- a/cmd/describe/autoscaler/cmd.go
+++ b/cmd/describe/autoscaler/cmd.go
@@ -1,0 +1,66 @@
+package autoscaler
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/rosa/pkg/clusterautoscaler"
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/output"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+const (
+	use     = "autoscaler"
+	short   = "Show details of the autoscaler for a cluster"
+	long    = short
+	example = ` # Describe the autoscaler for cluster 'foo'
+rosa describe autoscaler --cluster foo`
+)
+
+func NewDescribeAutoscalerCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     use,
+		Short:   short,
+		Long:    long,
+		Example: example,
+		Args:    cobra.NoArgs,
+		Run:     rosa.DefaultRunner(rosa.RuntimeWithOCM(), DescribeAutoscalerRunner()),
+	}
+
+	output.AddFlag(cmd)
+	ocm.AddClusterFlag(cmd)
+	return cmd
+}
+
+func DescribeAutoscalerRunner() rosa.CommandRunner {
+	return func(_ context.Context, runtime *rosa.Runtime, _ *cobra.Command, _ []string) error {
+		cluster, err := runtime.OCMClient.GetCluster(runtime.GetClusterKey(), runtime.Creator)
+		if err != nil {
+			return err
+		}
+
+		err = clusterautoscaler.IsAutoscalerSupported(runtime, cluster)
+		if err != nil {
+			return err
+		}
+
+		autoscaler, err := runtime.OCMClient.GetClusterAutoscaler(cluster.ID())
+		if err != nil {
+			return err
+		}
+
+		if autoscaler == nil {
+			return fmt.Errorf("No autoscaler exists for cluster '%s'", runtime.ClusterKey)
+		}
+
+		if output.HasFlag() {
+			output.Print(autoscaler)
+		} else {
+			fmt.Print(clusterautoscaler.PrintAutoscaler(autoscaler))
+		}
+		return nil
+	}
+}
--- a/cmd/upgrade/accountroles/cmd.go
+++ b/cmd/upgrade/accountroles/cmd.go
@@ -242,10 +242,8 @@ func run(cmd *cobra.Command, _ []string) {
 	case aws.ModeAuto:
 		if isUpgradeNeedForAccountRolePolicies {
 			reporter.Infof("Starting to upgrade the policies")
-			err = r.WithSpinner(func() error {
-				return upgradeAccountRolePolicies(reporter, awsClient, prefix, creator.Partition, creator.AccountID, policies,
-					policyVersion, policyPath, isVersionChosen)
-			})
+			err := upgradeAccountRolePolicies(reporter, awsClient, prefix, creator.Partition, creator.AccountID, policies,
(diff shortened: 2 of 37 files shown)


pkg/clusterautoscaler/validation.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+const (
+	NoHCPAutoscalerSupportMessage = "Hosted Control Plane clusters do not support cluster-autoscaler configuration"
+	ClusterNotReadyMessage        = "Cluster '%s' is not yet ready. Current state is '%s'"
+)
+
+func IsAutoscalerSupported(runtime *rosa.Runtime, cluster *cmv1.Cluster) error {
+	if cluster.Hypershift().Enabled() {
+		return fmt.Errorf(NoHCPAutoscalerSupportMessage)
```
gdbranco, line 18: Autoscaler is supported on HCP, but is only enabled for IBM if I'm not mistaken

robpblake: @gdbranco I replicated the logic from `create autoscaler` which doesn't seem to do any check to see if current org has capability to create autoscaler for their cluster. IIRC, IBM doesn't use the CLI, but OCM SDK directly?

gdbranco: Correct, they are using SDK from what I know

cmd/describe/autoscaler/cmd.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		autoscaler, err := runtime.OCMClient.GetClusterAutoscaler(cluster.ID())
+		if err != nil {
+			return err
+		}
+
+		if autoscaler == nil {
+			return fmt.Errorf("No autoscaler exists for cluster '%s'", runtime.ClusterKey)
+		}
```
robpblake, line 57: Does it make sense to decompose this further into a new `ClusterAutoscalerService` so that the command becomes even briefer e.g:

```
service := NewClusterAutoscalerService(runtime.OCMClient)
supported, err = service.IsAutoscalerSupported(runtime, cluster)

autoscaler, err := service.GetClusterAutoscaler(cluster.ID())
if err != nil {
return err
}
```

cmd/describe/autoscaler/cmd.go
```diff
@@ -0,0 +1,66 @@
...
+
+const (
+	use     = "autoscaler"
+	short   = "Show details of the autoscaler for a cluster"
+	long    = short
+	example = ` # Describe the autoscaler for cluster 'foo'
+rosa describe autoscaler --cluster foo`
+)
+
+func NewDescribeAutoscalerCommand() *cobra.Command {
```
robpblake, line 23: The command is returned, fully initialised from a function. There is no use of global variables (ignoring the `output.AddFlag` and `ocm.AddClusterFlag` calls which are beyond the scope of this particular refactor). There is no use of `init`

cmd/describe/autoscaler/cmd.go
```diff
@@ -0,0 +1,66 @@
...
+		Args:    cobra.NoArgs,
+		Run:     rosa.DefaultRunner(rosa.DefaultRuntime(), DescribeAutoscalerRunner()),
+	}
+
+	output.AddFlag(cmd)
+	ocm.AddClusterFlag(cmd)
+	return cmd
+}
+
+func DescribeAutoscalerRunner() rosa.CommandRunner {
```
robpblake, line 38: Command logic is executed by a `CommandRunner`. The runner does not `os.Exit`, but instead returns an `error` if something went wrong. This makes it fully testable. The passed `Runtime` can be instantiated with mocks in the tests.

cmd/describe/autoscaler/cmd.go
oriAdler: I think it's ok to keep it that way, it seems like we return nil for `StatusNotFound`, and have different error messages per command.
I wish all the commands in ROSA were as brief as this one:)

With that being said, we could return a boolean from the function for `exist` and check the boolean value instead of nil to improve readability (maybe outside of the scope of this PR).

pkg/rosa/runner.go
```diff
@@ -0,0 +1,66 @@
...
+// DefaultRuntime returns a Runtime with the most basic of setups. None of the clients are initialised.
+func DefaultRuntime() RuntimeVisitor {
+	return func(ctx context.Context, runtime *Runtime, command *cobra.Command, args []string) error {
+		return nil
+	}
+}
+
+// RuntimeWithOCM configures the Runtime with an OCM Client
+func RuntimeWithOCM() RuntimeVisitor {
+	return func(ctx context.Context, runtime *Runtime, command *cobra.Command, args []string) error {
```
oriAdler, line 46: Do we need an error return type if the function always returns nil?

robpblake: @oriAdler I was thinking about future extensions where other visitors might want to return an `error`, but we can always extend the definition of the function when that becomes the case. I'll remove it for now.

cmd/upgrade/accountroles/cmd.go
```diff
@@ -242,10 +242,8 @@ func run(cmd *cobra.Command, _ []string) {
 	case aws.ModeAuto:
 		if isUpgradeNeedForAccountRolePolicies {
 			reporter.Infof("Starting to upgrade the policies")
-			err = r.WithSpinner(func() error {
```
robpblake, line 245: Our AWS SDK v2 rebase incorrectly bought this back in, so just returning us to previous state.

pkg/ocm/flag.go
```diff
@@ -25,29 +25,28 @@ import (
...
 func AddClusterFlag(cmd *cobra.Command) {
-	cmd.Flags().StringVarP(
-		&clusterKey,
-		"cluster",
-		"c",
-		"",
-		"Name or ID of the cluster.",
-	)
-	cmd.MarkFlagRequired("cluster")
-	cmd.RegisterFlagCompletionFunc("cluster", clusterCompletion)
```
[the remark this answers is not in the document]
robpblake: @oriAdler Sorry, which function call are you referring to?

robpblake:
> LGTM, happy to merge this if we want. I think the overall struct is good, and small improvements can be followed up as we tackle a more complex command next WDYT?

@ciaranRoche That makes sense to me. +1

pkg/ocm/flag.go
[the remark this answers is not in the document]
oriAdler: My bad, I see `cmd.RegisterFlagCompletionFunc` is now called from `AddOptionalClusterFlag`.

oriAdler (approved):

PR #1841, opened 2024-03-12
OCM-6527 | feat: add describe ingress command

gdbranco:
[OCM-6527](https://issues.redhat.com//browse/OCM-6527)

### DIFF
--- a/cmd/describe/ingress/cmd.go
+++ b/cmd/describe/ingress/cmd.go
@@ -1,0 +1,73 @@
+package ingress
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/rosa/pkg/ingress"
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/output"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+const (
+	use     = "ingress"
+	short   = "Show details of the specified ingress within cluster"
+	example = `rosa describe ingress <ingress_id> -c mycluster`
+)
+
+// Regular expression to used to make sure that the identifier given by the
+// user is safe and that it there is no risk of SQL injection:
+var ingressKeyRE = regexp.MustCompile(`^[a-z0-9]{3,5}$`)
+
+func NewDescribeIngressCommand() *cobra.Command {
+	options := NewDescribeIngressUserOptions()
+	cmd := &cobra.Command{
+		Use:     use,
+		Short:   short,
+		Example: example,
+		Run:     rosa.DefaultRunner(rosa.RuntimeWithOCM(), DescribeIngressRunner(options)),
+		Args:    cobra.NoArgs,
+	}
+
+	flags := cmd.Flags()
+	flags.StringVar(
+		&options.ingress,
+		"ingress",
+		"",
+		"Ingress of the cluster to target",
+	)
+
+	ocm.AddClusterFlag(cmd)
+	output.AddFlag(cmd)
+	return cmd
+}
+
+func DescribeIngressRunner(userOptions DescribeIngressUserOptions) rosa.CommandRunner {
+	return func(_ context.Context, runtime *rosa.Runtime, cmd *cobra.Command, argv []string) error {
+		options := NewDescribeIngressOptions()
+		if len(argv) == 1 && !cmd.Flag("ingress").Changed {
+			userOptions.ingress = argv[0]
+		} else {
+			err := cmd.ParseFlags(argv)
+			if err != nil {
+				return fmt.Errorf("unable to parse flags: %v", err)
+			}
+		}
+		err := options.Bind(userOptions)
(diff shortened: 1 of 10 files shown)


cmd/describe/ingress/cmd.go
```diff
@@ -0,0 +1,185 @@
...
+			ingressID,
+		)
+		os.Exit(1)
+	}
+
+	clusterKey := r.GetClusterKey()
+	cluster := r.FetchCluster()
+
+	// Try to find the ingress:
+	r.Reporter.Debugf("Loading ingresses for cluster '%s'", clusterKey)
```
robpblake, line 63: @gdbranco How about collapsing this into a `GetIngress(clusterId string, ingressId string)` method on `OCMClient`?

cmd/describe/ingress/cmd.go
```diff
@@ -0,0 +1,185 @@
...
+	keys := helper.MapKeys(entries)
+	sort.Strings(keys)
+	minWidth := getMinWidth(keys)
+	for _, key := range keys {
+		ingressOutput += fmt.Sprintf("%s: %s\n", key, strings.Repeat(" ", minWidth-len(key))+entries[key])
+	}
+	fmt.Print(ingressOutput)
+}
+
+func getMinWidth(keys []string) int {
```
robpblake, line 117: @gdbranco Can we get some test coverage on this function please.

cheers,
Rob

cmd/describe/ingress/cmd.go
```diff
@@ -0,0 +1,185 @@
...
+	minWidth := 0
+	for _, key := range keys {
+		if len(key) > minWidth {
+			minWidth = len(key)
+		}
+	}
+	return minWidth
+}
+
+func generateEntriesOutput(cluster *cmv1.Cluster, ingress *cmv1.Ingress) map[string]string {
```
robpblake, line 127: @gdbranco Can we can some test coverage on this function please.

cheers,
Rob

cmd/describe/ingress/cmd.go
```diff
@@ -0,0 +1,185 @@
...
+		return nil
+	},
+}
+
+func init() {
+	ocm.AddClusterFlag(Cmd)
+	output.AddFlag(Cmd)
+}
+
+func run(_ *cobra.Command, argv []string) {
```
robpblake, line 46: @gdbranco I've been trying to encourage folks to delegate to another function so that the main work of their command can be tested. At the minute with the `os.Exit()` calls. We can't test this. 

Recent examples are the `config get` and `config set` commands added by Tirth. 

It's also something I'm working to make easier in #1837

cmd/describe/ingress/cmd.go
gdbranco: Yeah, still going through. I'll send a commit soon :)

cmd/describe/ingress/cmd.go
```diff
@@ -0,0 +1,87 @@
...
+			}
+		}
+		err := options.Bind(userOptions)
+		if err != nil {
+			return err
+		}
+		ingressKey := options.args.ingress
+		if !ingressKeyRE.MatchString(ingressKey) {
+			return fmt.Errorf(
+				"ingress identifier %q isn't valid: it must contain only letters or digits",
```
hunterkepley, line 75: I think we can move this validation to the bind itself, to make this a little cleaner

cmd/describe/ingress/cmd.go
```diff
@@ -0,0 +1,87 @@
+package ingress
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/spf13/cobra"
+
```
hunterkepley, line 10: looks like `fmt-imports` might need to be ran

cmd/describe/ingress/cmd.go
```diff
@@ -0,0 +1,87 @@
...
+func NewDescribeIngressCommand() *cobra.Command {
+	options := NewDescribeIngressUserOptions()
+	cmd := &cobra.Command{
+		Use:     use,
+		Short:   short,
+		Example: example,
+		Run:     rosa.DefaultRunner(rosa.RuntimeWithOCM(), DescribeIngressRunner(options)),
+		Args: func(_ *cobra.Command, argv []string) error {
+			if len(argv) != 1 {
+				return fmt.Errorf(
```
hunterkepley, line 36: do we need this check here? Later in the runner we check for length of args, might could throw that in there along with that logic to keep similar logic together

hunterkepley: We also check in `options.Bind` for an ingress ID, so that code path will probably never be reached

hunterkepley (commented):
Couple of comments, maybe we can improve test coverage too, im not great at reading codecov but seems like it's around 40% covered

cmd/describe/ingress/cmd.go
gdbranco: Oh true, will do

cmd/describe/ingress/cmd.go
gdbranco: huh, it didn't do anything :thinking:

PR #1843, opened 2024-03-13
OCM-6619 | fix: upgrade cluster should not exit prematurely after calling upgrade role

mikeshng:
In https://github.com/openshift/rosa/pull/1779 a change was made to a part of upgrade role cmd to exit instead of return. Since upgrade cluster calls upgrade role run internally, this is now causing the upgrade cluster to terminate prematurely. This fix is to return to the previous behavior to unblock upgrade cluster cmd.

Ref: https://issues.redhat.com/browse/OCM-6619


mikeshng:
CC @den-rgb

den-rgb:
I believe the same issue will occur when running the upgrade for account and operator roles due to them also using an `os.Exit(0)` in places where an info log is triggered, could it be possible to modify the rest as part of this pr in order to avoid revisiting the same issues?

cmd/upgrade/roles/cmd.go
```diff
@@ -219,7 +219,7 @@ func run(cmd *cobra.Command, argv []string) {
 		}
 		r.Reporter.Infof("Cluster '%s' operator roles have attached managed policies. "+
 			"An upgrade isn't needed", cluster.Name())
-		os.Exit(0)
+		return
```
robpblake, line 222: @den-rgb @mikeshng Should we revert to `RunE` for this command? I feel `return nil` consistently rather than a single `return` is a better outcome for clarity. It does leave this command as an outlier by using `RunE`, but we can come back to that and try and fix it. It would also require the `SilenceUsage: true` on the `Cmd` definition.

Thoughts?

mikeshng: @robpblake I think `RunE` no longer exists and will panic if called. There is another PR https://github.com/openshift/rosa/pull/1806 that removed all the `RunE` references. I was going to create a secondary ticket to deal with the error handling that seems to regressed after discussing with @den-rgb and @AlexVulaj 

My preference is to bring `RunE` back because we discovered that there are some internal cmd calls that were not analyzed previously but I don't have enough technical expertise here to make that call. WDYT?

@den-rgb nice catch. If we are going ahead with this fix approach I should fix all the places.

den-rgb: That's a possibility, we will have to revert all of the return changes in that case, due to there being some i removed being unaware `return` on its own was viable for this code like the change here: https://github.com/openshift/rosa/pull/1779/files#diff-42e429a79b1a9b4a9fbf33a5dab5ad029f8365750589cd5a1e3402e4ff3ddea1L460 and 
https://github.com/openshift/rosa/pull/1779/files#diff-42e429a79b1a9b4a9fbf33a5dab5ad029f8365750589cd5a1e3402e4ff3ddea1L386 
but I feel like it would be the same effort to just include them for this file as part of this pr as it would to come back to it,  as long as it's verified, we can create a seperate ticket for the rest of the files in the meantime?

den-rgb: It seems like the only thing being handled incorrectly is the info level warnings and basic returns, the errors regarding `RunE` were handled as part of https://github.com/openshift/rosa/pull/1806 just like Mike mentioned.

mikeshng: @den-rgb thank you for your review and feedback. I've done as you suggested. PTAL. Thanks.

den-rgb: Could we possibly include the deleted `nil returns` also removed in this pr for all the files where it was removed: https://github.com/openshift/rosa/pull/1779/files#diff-3802c119f0c619f9d349b65a5fa516ef3f7bde25defcd998d71bbe117dbb7e01L240

mikeshng: @den-rgb done as suggested. PTAL. Thanks.

mikeshng: Got lint error ``` S1023: redundant `return` statement (gosimple)``` when I did that so I reverted the changes to make the lint pass

mikeshng:
FYI @oriAdler this is the fix for the [upgrade block issue](https://issues.redhat.com/browse/OCM-6619). I think the other [error handling issue](https://issues.redhat.com/browse/OCM-6697) doesn't need to be a blocker.

PR #1844, opened 2024-03-13
OCM-5718 | feat: Display login error if `--govcloud` supplied with commercial region

hunterkepley:
[OCM-5718](https://issues.redhat.com//browse/OCM-5718) describes a bug where if a user attempts to log in supplying all commercial values, except passing in `--govcloud` flag, ROSA will be in 'fedramp mode' and will attempt to run commands on the wrong API endpoint. This MR makes sure users are meaning to log into govcloud by requiring a govcloud region upon logging in.

All other logic is kept the same, so if you _only_ supply a govcloud region, or an encrypted token, etc, you will still log into fedramp. But, the `--govcloud` flag bug is solved by requiring the region upon passing the flag. If you pass `--govcloud` without meaning to (I.E. not providing a govcloud region), you will face an error. Other safeguard in this case already exists, where you must pass in a fedramp token to log in to govcloud anyway.

### DIFF
--- a/cmd/login/cmd.go
+++ b/cmd/login/cmd.go
@@ -28,6 +28,7 @@ import (
 	sdk "github.com/openshift-online/ocm-sdk-go"
 	"github.com/openshift-online/ocm-sdk-go/authentication"
 	"github.com/spf13/cobra"
+	errors "github.com/zgalor/weberr"
 
 	"github.com/openshift/rosa/cmd/logout"
 	"github.com/openshift/rosa/pkg/arguments"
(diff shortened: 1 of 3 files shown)


cmd/login/cmd.go
```diff
@@ -236,6 +236,13 @@ func run(cmd *cobra.Command, argv []string) {
 		(cfg.FedRAMP && token == "") ||
 		fedramp.IsGovRegion(arguments.GetRegion()) ||
 		config.IsEncryptedToken(token) {
+		// Display error to user if they attempt to log into govcloud without a region specified (fixes OCM-5718)
```
davidleerh, line 239: @hunterkepley
what if users supply "cfg.FedRAMP" or config.IsEncryptedToken(token) == true?
Since only of the conditions need to truthy to be considered FedRamp, would they all require the fedramp region argument to be passed?

cmd/login/cmd.go
```diff
@@ -236,6 +236,13 @@ func run(cmd *cobra.Command, argv []string) {
...
+		if !fedramp.IsGovRegion(arguments.GetRegion()) {
+			r.Reporter.Errorf("Must supply govcloud region when using flag '--govcloud'\n" +
```
davidleerh, line 241: I'm personally thinking it'd be better if the message is more generic? Since there are different approaches to specifying whether a user wants to log into FedRAMP.

like "When logging into the FedRAMP environment, a recognized us-gov region needs to be specified. 
Example: --region us-gov-west-1"

Also, would you kindly add a unit test?
I think you can put the "or" statements into a separate function and try to emulate the error with mock arguments.

WDYT?

hunterkepley: I agree, will change the msg and add a unit test

robpblake (approved):
/lgtm

PR #1846, opened 2024-03-13
OCM-6655 | feat: Handle trying to make hcp cluster on fedramp properly

hunterkepley:
Error when creating HCP cluster using flag `--hosted-cp` while logged into fedramp

Do not prompt about HCP (and default to false) while logged into fedramp using interactive mode


gdbranco (approved):

ciaranRoche (commented):
/lgtm

PR #1848, opened 2024-03-14
OCM-6648 | fix: create node pool with security group - check version

oriAdler:
For interactive mode, check the node pool version and prompt for additional security groups accordingly.

### DIFF
--- a/cmd/create/machinepool/helper_test.go
+++ b/cmd/create/machinepool/helper_test.go
@@ -7,6 +7,7 @@ import (
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 
+	"github.com/openshift/rosa/pkg/helper/features"
 	"github.com/openshift/rosa/pkg/rosa"
 )
 
@@ -67,6 +68,27 @@ var _ = Describe("Machine pool helper", func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(len(awsNodePool.AdditionalSecurityGroupIds())).To(Equal(0))
 			Expect(awsNodePool.InstanceType()).To(Equal(instanceType))
+		})
+	})
+
+	Context("It validate version is compatible for security groups", func() {
+		It("Skips validation if the version isn't provided", func() {
+			version := ""
+			isCompatible, err := features.IsFeatureSupported(features.AdditionalDay2SecurityGroupsHcpFeature, version)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(isCompatible).To(BeTrue())
+		})
+		It("Returns false for 4.14.0", func() {
+			version := "4.14.0"
+			isCompatible, err := features.IsFeatureSupported(features.AdditionalDay2SecurityGroupsHcpFeature, version)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(isCompatible).To(BeFalse())
+		})
+		It("Returns true for 4.15.0", func() {
+			version := "4.15.0"
+			isCompatible, err := features.IsFeatureSupported(features.AdditionalDay2SecurityGroupsHcpFeature, version)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(isCompatible).To(BeTrue())
 		})
 	})
 })
--- a/cmd/create/machinepool/nodepool.go
+++ b/cmd/create/machinepool/nodepool.go
@@ -10,6 +10,7 @@ import (
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 
+	"github.com/openshift/rosa/pkg/helper/features"
 	"github.com/openshift/rosa/pkg/helper/machinepools"
 	"github.com/openshift/rosa/pkg/helper/versions"
 	"github.com/openshift/rosa/pkg/interactive"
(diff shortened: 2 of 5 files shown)


cmd/create/machinepool/nodepool.go
```diff
@@ -493,3 +499,17 @@ func createAwsNodePoolBuilder(instanceType string, securityGroupIds []string) *c
 
 	return awsNpBuilder
 }
+
+func isCompatibleVersionForAdditionalSecurityGroupsIds(version string) (bool, error) {
```
robpblake, line 503: @oriAdler How about we turn this into something generic that can be reused throughout the CLI and then easily swapped out if we move the implementation to the backend?

I was thinking something like:

```
package features

import (
	"github.com/openshift/rosa/pkg/helper/versions"
)

type OCMFeature string

const (
	AdditionalDay2SecurityGroupsHcpFeature OCMFeature = "AdditionSecurityGroupsDay2HCP"
)

var ocmFeatureVersions = map[OCMFeature]string{
	AdditionalDay2SecurityGroupsHcpFeature: "4.15.0-0.a",
}

type FeatureService interface {
	IsFeatureSupported(feature OCMFeature, cluster *cmv1.Cluster) (bool, error)
}

type FeatureServiceImpl struct {
	featureVersionMap map[OCMFeature]string
}

func NewFeatureService() FeatureService {
	return &FeatureServiceImpl{featureVersionMap: ocmFeatureVersions}
}

func (f *FeatureServiceImpl) IsFeatureSupported(feature OCMFeature, cluster *cmv1.Cluster) (bool, error) {
	supportedVersion := f.featureVersionMap[feature]
	if supportedVersion != "" {
		return versions.IsGreaterThanOrEqual(cluster.Version().RawID(), supportedVersion)
	}

	return true, nil
}
```

And then, your caller looks like:

```
featureService := NewFeatureService()
isSupported, err := featureService.IsFeatureSupported(AdditionalDay2SecurityGroupsHcpFeature, cluster)
```

WDYT?

oriAdler: Thanks a lot for the feedback.
I like the approach of a generic function to check the feature version.

With that being said, I'm not sure we use the `Service` pattern in the CLI, usually more common in the backend.
Created a new `features` package, and added a general function to the package.

Moreover, added a note to `versions.go` to add new minimal versions to the new package, we can also refactor the existing checks as part of our tech debt.

Please let me know if the solution seems valid to you, and I can create a ticket to move existing minimal version checks to use the `features` package.

pkg/ocm/versions.go
```diff
@@ -43,6 +43,7 @@ const (
 
 	MinVersionForAdditionalComputeSecurityGroupIdsDay1 = "4.14.0-0.a"
 	MinVersionForAdditionalComputeSecurityGroupIdsDay2 = "4.11.0-0.a"
+	// TODO: Add new minimal versions to the `features.go` package.
```
robpblake, line 46: @oriAdler How about we create a section in `CONTRIBUTING.md` around version-gating a feature rather than a TODO in the code? We could then link to the section in `CONTRIBUTING.md` from here.

Ideally we don't want versions in the CLI at all, but for now we can at least try and be consistent. If we update `CONTRIBUTING.md` and then include an announcement in our Slack channel so that everyone on the team is aware of this new approach?

cmd/create/machinepool/nodepool.go
robpblake: @oriAdler First implementation looks good to me 👍 

I've left a comment about us updating `CONTRIBUTING.md` and also communicating to the team about this change. We want to ensure that everyone knows this exists and this is how we want to enforce version gating of features in the CLI at this time, until we can fully remove it and rely on the backend.

+1 to creating another issue to move existing feature gates to this approach. Sounds like `good-first-issue` to me 👍 

cheers,

Rob

pkg/helper/features/features.go
```diff
@@ -0,0 +1,38 @@
...
+
+	supportedVersion := ocmFeatureVersions[feature]
+	// OCM feature isn't defined.
+	if supportedVersion == "" {
+		return true, nil
+	}
+
+	isSupported, err := versions.IsGreaterThanOrEqual(version, supportedVersion)
+	if err != nil {
+		return false, fmt.Errorf("failed to check if feautre '%s' is supported for version '%s': %v",
```
robpblake, line 33: @oriAdler Typo in `feautre` 👍

cmd/create/machinepool/helper_test.go
```diff
@@ -69,4 +70,25 @@ var _ = Describe("Machine pool helper", func() {
 			Expect(awsNodePool.InstanceType()).To(Equal(instanceType))
 		})
 	})
+
+	Context("It validate version is compatible for security groups", func() {
+		It("Skips validation if the version isn't provided", func() {
+			version := ""
+			isCompatible, err := features.IsFeatureSupported(features.AdditionalDay2SecurityGroupsHcpFeature, version)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(isCompatible).To(Equal(true))
```
ziccardi, line 79:
```suggestion
			Expect(isCompatible).To(BeTrue())
```

cmd/create/machinepool/helper_test.go
```diff
@@ -69,4 +70,25 @@ var _ = Describe("Machine pool helper", func() {
...
+		})
+		It("Returns false for 4.14.0", func() {
+			version := "4.14.0"
+			isCompatible, err := features.IsFeatureSupported(features.AdditionalDay2SecurityGroupsHcpFeature, version)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(isCompatible).To(Equal(false))
```
ziccardi, line 85:
```suggestion
			Expect(isCompatible).To(BeFalse())
```

cmd/create/machinepool/helper_test.go
```diff
@@ -69,4 +70,25 @@ var _ = Describe("Machine pool helper", func() {
...
+		})
+		It("Returns true for 4.15.0", func() {
+			version := "4.15.0"
+			isCompatible, err := features.IsFeatureSupported(features.AdditionalDay2SecurityGroupsHcpFeature, version)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(isCompatible).To(Equal(true))
```
ziccardi, line 91:
```suggestion
			Expect(isCompatible).To(BeTrue())
```

pkg/helper/features/features.go
```diff
@@ -0,0 +1,38 @@
...
+	// The version is not specified.
+	if version == "" {
+		return true, nil
+	}
+
+	supportedVersion := ocmFeatureVersions[feature]
+	// OCM feature isn't defined.
+	if supportedVersion == "" {
+		return true, nil
+	}
```
ziccardi, line 29: As this is a public helper function, we should add a comment that clearly describes what the function does. The comment should include a statement that if the version string provided is empty or if the feature is not defined, the function will return `true` nonetheless.

oriAdler: Added, thanks.

oriAdler:
@robpblake @ziccardi all comments are addressed, can you rePTAL?

robpblake (approved):
/lgtm

PR #1849, opened 2024-03-14
OCM-6687 | fix: 4.14 nightly version not compatible with ROSA HCP

davidleerh:
Issue with nightly versions being evaluated as prelease versions.
This caused cluster create issues with openshift nightly versions greater than 4.14.0.
The minimum versions for HCP required is 4.14.0

With command:
`./rosa create cluster --hosted-cp --channel-group=nightly`

Interactive List before fix (rosa create cluster) [only shows 4.15.0+ nightly versions]
image

Interactive List after fix (rosa create cluster)
image

Confirming fix includes all nightly versions meeting 4.14.0 version requirement.

### DIFF
--- a/pkg/ocm/versions.go
+++ b/pkg/ocm/versions.go
@@ -36,7 +36,7 @@ const (
 	LowestHttpTokensRequiredSupport = "4.11.0"
 	LowestSTSMinor                  = "4.7"
 
-	LowestHostedCpSupport            = "4.14.0"
+	LowestHostedCpSupport            = "4.14.0-0.a"
 	MinVersionForManagedIngressV2    = "4.14.0-0.a"
 	MinVersionForMachinePoolRootDisk = "4.10.0-0.a"
 	VersionPrefix                    = "openshift-v"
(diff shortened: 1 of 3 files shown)


lucasponce:
LGTM

robpblake (approved):
/lgtm

pkg/ocm/versions.go
```diff
@@ -36,7 +36,7 @@ const (
 	LowestHttpTokensRequiredSupport = "4.11.0"
 	LowestSTSMinor                  = "4.7"
 
-	LowestHostedCpSupport            = "4.14.0"
+	LowestHostedCpSupport            = "4.14.0-0.a"
```
lucasponce, line 39: We've noticed that this was wrongly introduced.

https://redhat-internal.slack.com/archives/C04C3G5MDKN/p1710502133096589

lucasponce: I wonder if we could use this one or another PR to address it.

I'm not sure about the implications, perhaps a separated PR is good.

But I also missed from my side, I didn't have all context.

robpblake: @lucasponce As discussed in Slack, I think I approved this just as you were commenting 😄 . Let's follow up with another PR to fix the 4.12 issue.

cheers,

Rob

PR #1851, opened 2024-03-14
OCM-5549 | feat: Disable ManagedService creation/updates in FedRAMP

hunterkepley:
https://issues.redhat.com/browse/OCM-5549

Prevents user from creating or updating managed services (leading to an error as they are not supported in fedramp) before the OCM client even begins to attempt to do so.


pkg/ocm/managedservice_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2/dsl/core"
+	. "github.com/onsi/ginkgo/v2/dsl/decorators"
+	. "github.com/onsi/gomega"
+
+	"github.com/openshift/rosa/pkg/fedramp"
+)
+
+var _ = Describe("ManagedService", Ordered, func() {
```
robpblake, line 13: @hunterkepley Does it need to be `Ordered` ?

pkg/ocm/managedservice_test.go
```diff
@@ -0,0 +1,33 @@
...
+	var client Client
+
+	When("FedRAMP is true", func() {
+		It("Can't create managedService", func() {
+			fedramp.Enable() // Simulate logging into fedramp
```
robpblake, line 18: @hunterkepley Can you please provide `BeforeEach` and `AfterEach` calls to reset the value of `fedramp.enabled` to false. Because `fedramp.enabled` is a global var, this test has the potential to impact on other test suites by leaving the value as `true`.

```
BeforeEach(func() {
	fedramp.SetEnabled(false)
})
	
AfterEach(func() {
	fedramp.SetEnabled(false)
})
```

robpblake (changes requested):
@hunterkepley Can you ensure that the test resets the value of the global `fedramp.enabled` var so that it does not impact other tests now, or in the future. 

cheers,

Rob

robpblake (approved):
/lgtm

PR #1853, opened 2024-03-15
OCM-6746 | fix: CLI support versions 4.12 for HCP cluster creation

davidleerh:
Reverting portion of changes introduced via
https://github.com/openshift/rosa/commit/3ca77edb101c9d0019959daa711c5d28efa3e265

With this change:
The CLI will gate users from using Openshift versions x < 4.12.0 during HCP cluster creation flow.

The clusters service backend will however, selectively determine and validate whether users can utilize versions between 4.12  and 4.14.


robpblake (approved):
/lgtm

PR #1855, opened 2024-03-15
OCM-4460 | feat: Add actions for break glass credential

chenz4027:
https://issues.redhat.com/browse/OCM-4460

### DIFF
--- a/cmd/create/breakglasscredential/cmd.go
+++ b/cmd/create/breakglasscredential/cmd.go
@@ -1,0 +1,85 @@
+package breakglasscredential
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/rosa/pkg/breakglasscredential"
+	"github.com/openshift/rosa/pkg/externalauthprovider"
+	"github.com/openshift/rosa/pkg/interactive"
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+var breakGlassCredentialArgs *breakglasscredential.BreakGlassCredentialArgs
+
+var Cmd = makeCmd()
+
+func makeCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:     "break-glass-credential",
+		Aliases: []string{"break-glass-credentials", "breakglasscredential", "breakglasscredentials"},
+		Short:   "Create a break glass credential for a cluster.",
+		Long:    "Create a break glass credential for a hosted control plane cluster with external authentication enabled.",
+		Example: `  # Interactively create a break glass credential to a cluster named "mycluster"
+  rosa create break-glass-credential --cluster=mycluster --interactive`,
+		Run:    run,
+		Hidden: true,
+		Args:   cobra.NoArgs,
+	}
+}
+
+func init() {
+	ocm.AddClusterFlag(Cmd)
+	interactive.AddFlag(Cmd.Flags())
+	breakGlassCredentialArgs = breakglasscredential.AddBreakGlassCredentialFlags(Cmd)
+}
+
+func run(cmd *cobra.Command, argv []string) {
+	r := rosa.NewRuntime().WithOCM()
+	defer r.Cleanup()
+	err := runWithRuntime(r, cmd, argv)
+	if err != nil {
+		r.Reporter.Errorf(err.Error())
+		os.Exit(1)
+	}
+}
+
+func runWithRuntime(r *rosa.Runtime, cmd *cobra.Command, argv []string) error {
+	clusterKey := r.GetClusterKey()
+	cluster := r.FetchCluster()
+
+	externalAuthService := externalauthprovider.NewExternalAuthService(r.OCMClient)
+	err := externalAuthService.IsExternalAuthProviderSupported(cluster, clusterKey)
+	if err != nil {
+		return err
+	}
+
+	if !breakglasscredential.IsBreakGlassCredentialSetViaCLI(cmd.Flags()) && !interactive.Enabled() {
+		interactive.Enable()
+		r.Reporter.Infof("Enabling interactive mode")
+	}
+	r.Reporter.Debugf("Creating a break glass credential for cluster '%s'", clusterKey)
+
+	args, err := breakglasscredential.GetBreakGlassCredentialOptions(
+		cmd.Flags(), breakGlassCredentialArgs)
+	if err != nil {
+		return fmt.Errorf("failed to create a break glass credential for cluster '%s': %s",
+			clusterKey, err)
+	}
+
+	credentialResponse, err := breakglasscredential.CreateBreakGlass(cluster, clusterKey, args, r)
+	if err != nil {
+		return err
+	}
+
+	kubeconfig, err := r.OCMClient.PollKubeconfig(cluster.ID(), credentialResponse.ID())
+	if err != nil {
+		return fmt.Errorf("An error occurred while polling for kubeconfig: %v", err)
+	}
+	fmt.Print(kubeconfig)
+
+	return nil
+}
--- a/cmd/describe/breakglasscredential/cmd.go
+++ b/cmd/describe/breakglasscredential/cmd.go
@@ -1,0 +1,154 @@
+/*
+Copyright (c) 2024 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package breakglasscredential
+
+import (
+	"fmt"
+	"os"
+
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/rosa/pkg/breakglasscredential"
+	"github.com/openshift/rosa/pkg/externalauthprovider"
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/output"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+var Cmd = &cobra.Command{
+	Use:     "break-glass-credential",
+	Aliases: []string{"break-glass-credentials", "breakglasscredential", "breakglasscredentials"},
+	Short:   "Show details of a break glass credential on a cluster",
+	Long:    "Show details of a break glass credential on a cluster.",
+	Example: `  # Show details of a break glass credential with ID "12345" on a cluster named "mycluster"
+  rosa describe break-glass-credential 12345 --cluster=mycluster `,
+	Run:    run,
+	Hidden: true,
+	Args:   cobra.MaximumNArgs(2),
+}
+
+var args struct {
+	id         string
+	kubeconfig bool
+}
+
+func init() {
+	flags := Cmd.Flags()
+	flags.SortFlags = false
+	ocm.AddClusterFlag(Cmd)
+	output.AddFlag(Cmd)
+	flags.StringVar(
+		&args.id,
+		"id",
+		"",
+		"Id for the break glass credential of the cluster to target",
+	)
+
+	flags.BoolVar(
+		&args.kubeconfig,
+		"kubeconfig",
+		false,
+		"Retrieve the kubeconfig from the break glass credential",
+	)
+}
+
+func run(cmd *cobra.Command, argv []string) {
+	r := rosa.NewRuntime().WithAWS().WithOCM()
+	defer r.Cleanup()
+	err := runWithRuntime(r, cmd, argv)
+	if err != nil {
+		r.Reporter.Errorf(err.Error())
+		os.Exit(1)
+	}
+}
+
+func runWithRuntime(r *rosa.Runtime, cmd *cobra.Command, argv []string) error {
+	breakGlassCredentialId := args.id
+	getKubeconfig := args.kubeconfig
+	// Allow the use also directly set the break glass credential id as positional parameter
+	if len(argv) == 1 && !cmd.Flag("id").Changed {
+		breakGlassCredentialId = argv[0]
+	}
+	if breakGlassCredentialId == "" {
+		return fmt.Errorf("you need to specify a break glass credential id with '--id' parameter")
+	}
+	clusterKey := r.GetClusterKey()
+	cluster := r.FetchCluster()
+
+	externalAuthService := externalauthprovider.NewExternalAuthService(r.OCMClient)
+	err := externalAuthService.IsExternalAuthProviderSupported(cluster, clusterKey)
+	if err != nil {
+		return err
+	}
+
+	r.Reporter.Debugf("Fetching the break glass credential '%s' for cluster '%s'", breakGlassCredentialId, clusterKey)
+	if !getKubeconfig {
+		r.Reporter.Infof(
+			"To retrieve only the kubeconfig for this credential "+
+				"use: 'rosa describe break-glass-credential %s -c %s --kubeconfig'",
+			breakGlassCredentialId, cluster.Name())
+	}
+
+	breakGlassCredentialConfig, err := r.OCMClient.GetBreakGlassCredential(cluster.ID(), breakGlassCredentialId)
+	if err != nil {
+		return err
+	}
+
+	if output.HasFlag() {
+		var formattedOutput map[string]interface{}
+		formattedOutput, err = breakglasscredential.FormatBreakGlassCredentialOutput(breakGlassCredentialConfig)
+		if err != nil {
+			return err
+		}
+		return output.Print(formattedOutput)
+	}
+
+	if getKubeconfig {
+		fmt.Print(breakGlassCredentialConfig.Kubeconfig())
+		return nil
+	}
+
+	fmt.Print(describeBreakGlassCredential(r, cluster, clusterKey, breakGlassCredentialConfig))
+
+	return nil
+}
+
+func describeBreakGlassCredential(r *rosa.Runtime, cluster *cmv1.Cluster,
+	clusterKey string, config *cmv1.BreakGlassCredential) string {
+	breakGlassCredentialOutput := fmt.Sprintf("\n"+
+		"ID:                                    %s\n"+
+		"Username:                              %s\n"+
+		"Expire at:                             %s\n"+
+		"Status:                                %s\n",
+		config.ID(),
+		config.Username(),
+		config.ExpirationTimestamp().Format("Jan _2 2006 15:04:05 MST"),
+		config.Status(),
+	)
+
+	revocationTimeStamp, ok := config.GetRevocationTimestamp()
+	if ok {
+		breakGlassCredentialOutput = fmt.Sprintf("%s"+
+			"Revoked at:                            %s\n",
+			breakGlassCredentialOutput,
+			revocationTimeStamp.Format("Jan _2 2006 15:04:05 MST"),
+		)
+	}
+
+	return breakGlassCredentialOutput
+}
--- a/cmd/list/breakglasscredential/cmd.go
+++ b/cmd/list/breakglasscredential/cmd.go
@@ -1,0 +1,87 @@
+package breakglasscredential
+
+import (
+	"fmt"
+	"os"
+	"text/tabwriter"
+
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/rosa/pkg/externalauthprovider"
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/output"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+var Cmd = &cobra.Command{
+	Use:     "break-glass-credentials",
+	Aliases: []string{"break-glass-credential", "breakglasscredential", "breakglasscredentials"},
+	Short:   "List break glass credential",
+	Long:    "List break glass credential for a cluster.",
+	Example: `  # List all break glass credentials for a cluster named 'mycluster'"
+  rosa list break-glass-credentials -c mycluster`,
+	Run:    run,
+	Args:   cobra.NoArgs,
+	Hidden: true,
+}
+
+func init() {
+	ocm.AddClusterFlag(Cmd)
+	output.AddFlag(Cmd)
+}
+
+func run(cmd *cobra.Command, _ []string) {
+	r := rosa.NewRuntime().WithAWS().WithOCM()
+	defer r.Cleanup()
+	err := runWithRuntime(r, cmd)
+	if err != nil {
+		r.Reporter.Errorf(err.Error())
+		os.Exit(1)
+	}
+}
+
+func runWithRuntime(r *rosa.Runtime, cmd *cobra.Command) error {
+	clusterKey := r.GetClusterKey()
+	cluster := r.FetchCluster()
+
+	externalAuthService := externalauthprovider.NewExternalAuthService(r.OCMClient)
+	err := externalAuthService.IsExternalAuthProviderSupported(cluster, clusterKey)
+	if err != nil {
+		return err
+	}
+
+	// Load any break glass credential for this cluster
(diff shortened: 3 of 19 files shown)


cmd/create/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,77 @@
...
+	clusterKey := r.GetClusterKey()
+	cluster := r.FetchCluster()
+
+	if !interactive.Enabled() {
+		interactive.Enable()
+		r.Reporter.Infof("Enabling interactive mode")
+	}
+	r.Reporter.Debugf("Creating a break glass credential for cluster '%s'", clusterKey)
+
+	// need to check if cluster is in ready state. I wonder if this should be the same as IsExternalAuthProviderSupported
```
andreadecorte, line 60: I would check:
- if hypershift cluster
- if ready cluster
- external auth?
All of this are also done in the backend but we usually check it before calling it

cmd/describe/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,129 @@
...
+	if breakGlassCredentialId == "" {
+		return fmt.Errorf("you need to specify a break glass credential id with '--id' parameter")
+	}
+	clusterKey := r.GetClusterKey()
+	cluster := r.FetchCluster()
+	if cluster.State() != cmv1.ClusterStateReady {
+		return r.Reporter.Errorf("Cluster '%s' is not yet ready", clusterKey)
+	}
+
+	if !cluster.Hypershift().Enabled() {
```
andreadecorte, line 85: we can unify these checks in a single place and call them in all the commands.

cmd/describe/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,129 @@
...
+
+	fmt.Print(output)
+	return nil
+}
+
+func describeBreakGlassCredential(r *rosa.Runtime, cluster *cmv1.Cluster,
+	clusterKey string, config *cmv1.BreakGlassCredential) string {
+	breakGlassCredentialOutput := fmt.Sprintf("\n"+
+		"ID:                                    %s\n"+
+		"Cluster ID:                            %s\n"+
```
andreadecorte, line 108: in similar commands (e.g. tuning configs), we don't have the cluster id, should we include it?

cmd/dlt/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,72 @@
...
+	cluster := r.FetchCluster()
+
+	if confirm.Confirm("delete all the break glass credentials on cluster %s", clusterKey) {
+		r.Reporter.Debugf("Deleting break glass credentials on cluster '%s'", clusterKey)
+		err := r.OCMClient.DeleteBreakGlassCredentials(cluster.ID())
+		if err != nil {
+			return fmt.Errorf("failed to delete break glass credentials on cluster '%s': %s",
+				clusterKey, err)
+		}
+		r.Reporter.Infof("Successfully deleted/revoked all break glass credentials from cluster '%s'",
```
andreadecorte, line 68: let's clarify which term we should use here. I tend to say we should only expose revoke

cmd/list/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,81 @@
...
+
+	if output.HasFlag() {
+		err = output.Print(breakGlassCredentials)
+		if err != nil {
+			return fmt.Errorf("%v", err)
+		}
+	}
+
+	if len(breakGlassCredentials) == 0 {
+		return fmt.Errorf("there are no break glass credentials for this cluster")
```
andreadecorte, line 64: Info or Warning usually

cmd/describe/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,129 @@
...
+	)
+
+	revocationTimeStamp, ok := config.GetRevocationTimestamp()
+	if ok {
+		breakGlassCredentialOutput = fmt.Sprintf("%s"+
+			"Revoked at:                            %s\n",
+			breakGlassCredentialOutput,
+			revocationTimeStamp.Format("Jan _2 2006 15:04:05 MST"),
+		)
+	}
```
andreadecorte, line 126: missing the kubeconfig if present

cmd/describe/breakglasscredential/cmd.go
chenz4027: Remove the clusterid

cmd/dlt/breakglasscredential/cmd.go
chenz4027: Moved to revoke

cmd/list/breakglasscredential/cmd.go
chenz4027: Hmmm this one is a bit tricky. It's inside runWithRuntime which only returns error.

cmd/describe/breakglasscredential/cmd.go
chenz4027: Discussed with @andreadecorte , added a message to extract kubeconfig

cmd/create/breakglasscredential/cmd.go
chenz4027: Used a predefined function IsExternalAuthProviderSupported that checks all three

cmd/list/breakglasscredential/cmd.go
andreadecorte:
```
r.Reporter.Infof("There are no break glass credentials for this cluster.")
return nil
```
I was thinking something like this

chenz4027: Ah I'll try that, thanks @andreadecorte

cmd/list/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// Load any break glass credential for this cluster
+	r.Reporter.Debugf("Loading break glass credentials for cluster '%s'", clusterKey)
+	breakGlassCredentials, err := r.OCMClient.GetBreakGlassCredentials(cluster.ID())
+	if err != nil {
+		return fmt.Errorf("failed to get break glass credentials for cluster '%s': %v", cluster.ID(), err)
+	}
+
+	if output.HasFlag() {
+		// -o json returns empty array, need to investigate why
```
chenz4027, line 61: Not sure what's wrong with it. 🤔

cmd/describe/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,155 @@
...
+
+	return breakGlassCredentialOutput
+}
+
+func formatBreakGlassCredentialOutput(breakGlassCredential *cmv1.BreakGlassCredential) (map[string]interface{}, error) {
```
andreadecorte, line 141: I would move this to the pkg/breakglasscredential

cmd/create/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,82 @@
...
+
+var breakGlassCredentialArgs *breakglasscredential.BreakGlassCredentialArgs
+
+const argsPrefix string = ""
+
+var Cmd = &cobra.Command{
+	Use:     "break-glass-credential",
+	Aliases: []string{"break-glass-credentials", "breakglasscredential", "breakglasscredentials"},
+	Short:   "Create a break glass credential for a cluster.",
+	Long:    "Create a break glass credential for a hosted control plane cluster with external OIDC.",
```
andreadecorte, line 24:
```suggestion
	Long:    "Create a break glass credential for a hosted control plane cluster with external authentication enabled.",
```
To double check

chenz4027: Thanks Andrea, will address those changes

andreadecorte (approved):
lgtm

cmd/create/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/openshift/rosa/pkg/breakglasscredential"
+	"github.com/openshift/rosa/pkg/externalauthprovider"
+	"github.com/openshift/rosa/pkg/interactive"
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+var breakGlassCredentialArgs *breakglasscredential.BreakGlassCredentialArgs
+
+const argsPrefix string = ""
```
robpblake, line 18: @chenz4027 I don't think this is being used. Can we remove it?

pkg/breakglasscredential/flags.go
```diff
@@ -0,0 +1,144 @@
...
+	usernameFlag   = "username"
+	expirationFlag = "expiration"
+)
+
+type BreakGlassCredentialArgs struct {
+	username           string
+	expirationDuration time.Duration
+}
+
+func AddBreakGlassCredentialFlags(cmd *cobra.Command, prefix string) *BreakGlassCredentialArgs {
```
robpblake, line 27: @chenz4027 The `prefix` parameter to this function is not being used, so I think lets remove it 👍

pkg/breakglasscredential/flags.go
```diff
@@ -0,0 +1,144 @@
...
+		&args.expirationDuration,
+		expirationFlag,
+		0,
+		fmt.Sprintf("Expire the break glass credential after a relative duration like 2h, 8h. "+
+			"The expiration duration needs to be at least 10 minutes from now and to be at maximum 24 hours."),
+	)
+	return args
+}
+
+func GetBreakGlassCredentialOptions(cmd *pflag.FlagSet, prefix string,
```
robpblake, line 47: @chenz4027 `prefix` and `confirmBeforeAllArgs` parameters are not being used, so shall we remove them?

pkg/breakglasscredential/flags.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func CreateBreakGlass(cluster *cmv1.Cluster,
+	clusterKey string,
+	args *BreakGlassCredentialArgs, r *rosa.Runtime) error {
+
+	breakGlassConfig, err := CreateBreakGlassConfig(args)
+	if err != nil {
+		return fmt.Errorf("failed to create a break glass credential for cluster '%s': %s",
+			cluster.ID(), err)
```
robpblake, line 95: @chenz4027 We should be using `clusterKey` in this error message 👍

pkg/breakglasscredential/flags.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	_, err = r.OCMClient.CreateBreakGlassCredential(cluster.ID(), breakGlassConfig)
+
+	if err != nil {
+		return fmt.Errorf("failed to create a break glass credential for cluster '%s': %s",
```
robpblake, line 101: @chenz4027 We should be using `clusterKey` in this error message 👍

cmd/create/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,82 @@
...
+	clusterKey := r.GetClusterKey()
+	cluster := r.FetchCluster()
+
+	externalAuthService := externalauthprovider.NewExternalAuthService(r.OCMClient)
+	err := externalAuthService.IsExternalAuthProviderSupported(cluster, clusterKey)
+	if err != nil {
+		return err
+	}
+
+	if !interactive.Enabled() {
```
robpblake, line 61: @chenz4027 Just checking that we always want to enable interactive mode for this command? If yes, then we can skip the `if !interactive.Enabled()` check and just call `interactive.Enable()`

cmd/describe/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,138 @@
...
+	if output.HasFlag() {
+		var formattedOutput map[string]interface{}
+		formattedOutput, err = breakglasscredential.FormatBreakGlassCredentialOutput(breakGlassCredentialConfig)
+		if err != nil {
+			return err
+		}
+		return output.Print(formattedOutput)
+	}
+
+	output := describeBreakGlassCredential(r, cluster, clusterKey, breakGlassCredentialConfig)
```
robpblake, line 109: @chenz4027 This can be shortened to:

```
fmt.Print(describeBreakGlassCredential(r, cluster, clusterKey, breakGlassCredentialConfig))
```

cmd/describe/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,138 @@
...
+	cluster := r.FetchCluster()
+
+	externalAuthService := externalauthprovider.NewExternalAuthService(r.OCMClient)
+	err := externalAuthService.IsExternalAuthProviderSupported(cluster, clusterKey)
+	if err != nil {
+		return err
+	}
+
+	r.Reporter.Debugf("Fetching the break glass credential '%s' for cluster '%s'", breakGlassCredentialId, clusterKey)
+	r.Reporter.Infof(
```
robpblake, line 91: @chenz4027 What's the use-case here? We're making an assumption that people have `jq` available to them. I personally don't think we should rely on additional tools for people to use `rosa` for key workflows. For example , does `jq` exist for Windows?

cmd/list/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,88 @@
...
+	err := externalAuthService.IsExternalAuthProviderSupported(cluster, clusterKey)
+	if err != nil {
+		return err
+	}
+
+	// Load any break glass credential for this cluster
+	r.Reporter.Debugf("Loading break glass credentials for cluster '%s'", clusterKey)
+	breakGlassCredentials, err := r.OCMClient.GetBreakGlassCredentials(cluster.ID())
+	if err != nil {
+		return fmt.Errorf("failed to get break glass credentials for cluster '%s': %v", cluster.ID(), err)
```
robpblake, line 57: @chenz4027 `clusterKey` in the error message rather than `cluster.ID()` for consistency for the customer.

cmd/revoke/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+}
+
+func runWithRuntime(r *rosa.Runtime, cmd *cobra.Command, argv []string) error {
+
+	clusterKey := r.GetClusterKey()
+
+	cluster := r.FetchCluster()
+
+	if confirm.Confirm("revoke all the break glass credentials on cluster %s", clusterKey) {
```
robpblake, line 62: @chenz4027 `'` around the `%s`

pkg/breakglasscredential/flags.go
```diff
@@ -0,0 +1,144 @@
...
+	if args.username != "" {
+		breakGlassBuilder.Username(args.username)
+	}
+
+	if args.expirationDuration != 0 {
+		expirationTimeStamp := time.Now().Add(args.expirationDuration).Round(time.Second)
+		breakGlassBuilder.ExpirationTimestamp(expirationTimeStamp)
+	}
+
+	breakGlassConfig, err := breakGlassBuilder.Build()
```
robpblake, line 120: @chenz4027 Can this be shortened to:

```
return breakGlassBuilder.Build()
```
?

cmd/describe/breakglasscredential/cmd.go
andreadecorte: chiming it as I suggested it, the use case is that this command provides a kubeconfig for the user.
Showing it in the describe is not great UX IMO.
Here we are providing it to the user in the json output and an hint on how to retrieve it with jq (but no strict dependency on it).

I was also looking at AKS, https://learn.microsoft.com/en-us/cli/azure/aks?view=azure-cli-latest#az-aks-get-credentials, they provide a much more opinionated default (writing kubeconfig to your default .kube/config) if we want to follow that.

robpblake: @chenz4027 @andreadecorte Is this exposed to customers or is Red Hat SRE the end-user?

pkg/breakglasscredential/flags.go
chenz4027: @robpblake yep I will remove them

cmd/describe/breakglasscredential/cmd.go
chenz4027: I think it's for the end-user, but will wait for @andreadecorte to confirm

pkg/breakglasscredential/flags.go
chenz4027: Will look into it & refactor the rest - Thanks @robpblake 👍

cmd/describe/breakglasscredential/cmd.go
andreadecorte: end user.
User flow is:
1. Create cluster with external authentication enabled
2. Use this command to get a kubeconfig
3. Use the kubeconfig to apply its own config for access
4. Profit

cmd/create/breakglasscredential/cmd.go
chenz4027: Good catch, nope I should add something like IsExternalAuthProviderSetViaCLI for breakglasscredential. If the parameter is set, we should skip the interactive mode.

cmd/describe/breakglasscredential/cmd.go
robpblake: @chenz4027 @andreadecorte: @gdbranco @ciaranRoche any myself have had a chat about this. We're against printing out that the user should use `jq` to get details of the `kubeconfig` to login. Reading through the epic, it feels like this is a core workflow, so should be as easy as possible for customers to achieve using only `rosa`.

We've come up with this proposal:

- `rosa create break-glass-credential -c rblake-test -id bar`
  - This would poll until the `kubeconfig` is created and print it as STDOUT at the end of the command
  - Ideally we're targeting the workflow for a customer to be able to do `rosa create break-glass-credential -c rblake-test -id bar > /path/to/kubeconfig`
- `rosa describe break-glass-credential`
   - Supports a `--kubeconfig` flag that replicates the output from the `rosa create break-glass-credential` command
   - Allows a user to do `rosa describe break-glass-credential --kubeconfg > /path/to/kubeconfig` to easily grab the credentials if they forgot them from the `create command`
 - `rosa list break-glass-credential`
 - `rosa delete break-glass-credential`

WDYT?

chenz4027: Thanks @robpblake , an additional --kubeconfig would work and we could remind the user of that parameter when they do `rosa describe breakglasscredential` like what we have here. @andreadecorte WDYT?

andreadecorte: sounds like a solid proposal.
+1 on Maggie's idea to keep an hint for the user in the standard describe to use the --kubeconfig flag if he wants to get the kubeconfig, I think it helps on the discoverability

robpblake: Thanks @chenz4027 @andreadecorte 👍 

I've updated the proposal to capture @chenz4027 suggestion that the `rosa describe break-glass-credential` command includes the suggestion of the `--kubeconfig` flag to just to get the KubeConfig. @chenz4027 Are you happy to re-work this PR to meet that proposal? If yes, I'd suggest documenting it in the JIRA as well so that QE are clear on what needs to be tested.

cheers,

Rob

chenz4027: Thanks @robpblake , yep I'll be working on it. Let me update the Jira and tag you ;)

robpblake (changes requested):
As per discussion, slight rework to this PR to make the core workflows easier to handle.

pkg/breakglasscredential/flags.go
chenz4027: Refactored

cmd/describe/breakglasscredential/cmd.go
chenz4027: @robpblake Added one more parameter for kubeconfig. Would you mind taking another look?

cmd/describe/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,153 @@
...
+	externalAuthService := externalauthprovider.NewExternalAuthService(r.OCMClient)
+	err := externalAuthService.IsExternalAuthProviderSupported(cluster, clusterKey)
+	if err != nil {
+		return err
+	}
+
+	r.Reporter.Debugf("Fetching the break glass credential '%s' for cluster '%s'", breakGlassCredentialId, clusterKey)
+	if !getKubeconfig {
+		r.Reporter.Infof(
+			"To retrieve the full kubeconfig, do: rosa describe break-glass-credential %s -c %s --kubeconfig",
```
robpblake, line 102: @chenz4027 Perhaps : "To retrieve only the kubeconfig for this credential, use: 'rosa describe break-glass-credential %s -c %s --kubeconfig'"?

cmd/create/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,80 @@
...
+		return fmt.Errorf("failed to create a break glass credential for cluster '%s': %s",
+			clusterKey, err)
+	}
+
+	err = breakglasscredential.CreateBreakGlass(cluster, clusterKey, args, r)
+	if err != nil {
+		return err
+	}
+
+	r.Reporter.Infof("Successfully created a break glass credential for cluster '%s'", clusterKey)
```
robpblake, line 77: @chenz4027 We had discussed that this command would poll for the details of the `kubeconfig` and print it out here. At the minute I don't see it doing that?

cmd/create/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,80 @@
...
+	Hidden: true,
+	Args:   cobra.NoArgs,
+}
+
+func init() {
+	flags := Cmd.Flags()
+
+	ocm.AddClusterFlag(Cmd)
+
+	interactive.AddFlag(flags)
```
robpblake, line 35: @chenz4027 You could just pass `Cmd.Flags()` here rather than assigning the `flags` variable. Not a biggie.

cmd/list/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,88 @@
...
+	breakGlassCredentials, err := r.OCMClient.GetBreakGlassCredentials(cluster.ID())
+	if err != nil {
+		return fmt.Errorf("failed to get break glass credentials for cluster '%s': %v", clusterKey, err)
+	}
+
+	if output.HasFlag() {
+		err = output.Print(breakGlassCredentials)
+		if err != nil {
+			r.Reporter.Errorf("%s", err)
+			os.Exit(1)
```
robpblake, line 64: @chenz4027 You don't need the `os.Exit` here. Return the `error` or `nil` 👍

pkg/breakglasscredential/flags.go
```diff
@@ -0,0 +1,153 @@
...
+
+	if err != nil {
+		return fmt.Errorf("failed to create a break glass credential for cluster '%s': %s",
+			clusterKey, err)
+	}
+	return nil
+
+}
+
+func CreateBreakGlassConfig(args *BreakGlassCredentialArgs) (*cmv1.BreakGlassCredential, error) {
```
robpblake, line 113: @chenz4027 It would be good to get some test coverage on these functions 👍

pkg/breakglasscredential/flags.go
```diff
@@ -0,0 +1,153 @@
...
+	usernameFlag   = "username"
+	expirationFlag = "expiration"
+)
+
+type BreakGlassCredentialArgs struct {
+	username           string
+	expirationDuration time.Duration
+}
+
+func AddBreakGlassCredentialFlags(cmd *cobra.Command) *BreakGlassCredentialArgs {
```
robpblake, line 27: @chenz4027 Unit test would be awesome 👍

cmd/create/breakglasscredential/cmd.go
chenz4027: Added the fmt print

chenz4027:
Thanks @robpblake , I'll add more test coverage as suggested.

pkg/breakglasscredential/flags.go
chenz4027: Added the test coverage

cmd/create/breakglasscredential/cmd.go
```diff
@@ -0,0 +1,89 @@
...
+		return fmt.Errorf("failed to create a break glass credential for cluster '%s': %s",
+			clusterKey, err)
+	}
+
+	credentialResponse, err := breakglasscredential.CreateBreakGlass(cluster, clusterKey, args, r)
+	if err != nil {
+		return err
+	}
+
+	credentialId, ok := credentialResponse.GetID()
```
robpblake, line 78: @chenz4027 You can use `credentialResponse.ID()` to avoid the `if` statement. We should always have an ID returned from the API, otherwise we have bigger problems :-)

chenz4027: Thanks Rob, updated 👍

robpblake (approved):
/lgtm

PR #1854, opened 2024-03-15
ROSA CLI V2 Changes

den-rgb:
Final Rebase for the ROSA CLI V2 changes located on the `aws-sdk` [branch](https://github.com/openshift/rosa/tree/aws-sdk/)

PR #1856, opened 2024-03-15
OCM-6753 | fix: hcp private cluster subnet validation

mikeshng:
In the PR https://github.com/openshift/rosa/pull/1854 

The `ValidateHostedClusterSubnets` publicSubnetsCount calculation were changed.
This cause rosa unable to create private hcp clusters in some cases due to validation failure:

```
E: The number of public subnets for a private hosted cluster should be zero
```

This change is to revert back to the previous calculation method. 

Reference: https://issues.redhat.com/browse/OCM-6753

### DIFF
--- a/pkg/ocm/helpers.go
+++ b/pkg/ocm/helpers.go
@@ -734,7 +734,7 @@ func ValidateHostedClusterSubnets(awsClient aws.Client, isPrivate bool, subnetID
 	}
 
 	privateSubnetCount := len(privateSubnets)
-	publicSubnetsCount := len(vpcSubnets) - privateSubnetCount
+	publicSubnetsCount := len(subnets) - privateSubnetCount
 
 	if isPrivate {
 		if publicSubnetsCount > 0 {
--- a/pkg/ocm/helpers_test.go
+++ b/pkg/ocm/helpers_test.go
@@ -21,11 +21,17 @@ import (
 	"net/url"
 	"strings"
 
+	"go.uber.org/mock/gomock"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	ocmCommonValidations "github.com/openshift-online/ocm-common/pkg/ocm/validations"
 	commonUtils "github.com/openshift-online/ocm-common/pkg/utils"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+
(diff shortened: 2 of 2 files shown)


mikeshng:
CC @lucasponce @gbranco as discussed I am simply reverting based on my local testing. The original PR that made this change is a giant rebase PR that's hard to know the full context. Please if you have more knowledge around this area could you double check to make sure by reverting this change won't break other things?

/hold

Maintainers, please unhold when comfortable.

pkg/ocm/helpers.go
```diff
@@ -734,7 +734,7 @@ func ValidateHostedClusterSubnets(awsClient aws.Client, isPrivate bool, subnetID
 	}
 
 	privateSubnetCount := len(privateSubnets)
-	publicSubnetsCount := len(vpcSubnets) - privateSubnetCount
+	publicSubnetsCount := len(subnets) - privateSubnetCount
```
robpblake, line 737: @mikeshng Can we back this up with a unit test in `helpers_test.go`? 

I agree with your changes. I would like us to spend time now to write a test that ensures that we don't inadvertently break this again in the future as our lack of coverage is one of the major reasons why we have this type of regression.

If we add coverage as we go, it makes larger refactors easier in the future as we have a contract that tells us if we're breaking things.

Let me know if you want any help with the test. `Client` interface for AWS is mocked via `gomock` and available in `pkg/aws/mock_client.go`

cheers,

Rob

mikeshng: Hi @robpblake thank you for the review and feedback. I've done as you suggested. Could you PTAL again? Thanks.

pkg/ocm/helpers_test.go
```diff
@@ -396,6 +402,33 @@ var _ = Describe("ValidateSubnetsCount", func() {
...
+		mockClient = mock.NewMockClient(mockCtrl)
+		mockClient.EXPECT().GetVPCSubnets(gomock.Any()).Return(subnets, nil).AnyTimes()
+	})
+	It("should not return an error when only private subnets are present for a private cluster", func() {
+		mockClient.EXPECT().FilterVPCsPrivateSubnets(gomock.Any()).Return([]ec2types.Subnet{subnets[1]}, nil)
+		_, err := ValidateHostedClusterSubnets(mockClient, true, []string{"subnet-private-2"})
+		Expect(err).NotTo(HaveOccurred())
+	})
+	It("should return an error when public subnets are present for a private cluster", func() {
+		mockClient.EXPECT().FilterVPCsPrivateSubnets(gomock.Any()).Return([]ec2types.Subnet{}, nil)
```
den-rgb, line 425: I think this should return a `0` rather then a empty subnet array

robpblake (approved):
/lgtm

PR #1859, opened 2024-03-18
OCM-6768 | fix: Create machine pool with security group - filter options

oriAdler:
For interactive mode, filter out security groups with `red-hat-managed` tag and with name `default`.

Improve the UX, to prevent an error returned from the backend.

### DIFF
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -23,7 +23,6 @@ import (
 	"io"
 	"net/http"
 	"os"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/aws/client.go
```diff
@@ -1279,19 +1278,10 @@ func (c *awsClient) GetSecurityGroupIds(vpcId string) ([]ec2types.SecurityGroup,
 		return []ec2types.SecurityGroup{}, err
 	}
 
-	for _, secGroup := range resp.SecurityGroups {
```
gdbranco, line 1282: Did this not already work?

PR #1867, opened 2024-03-21
OCM-6727 | feat: allow edit to hcp mp aws tags

gdbranco:
[OCM-6727](https://issues.redhat.com//browse/OCM-6727)

### DIFF
--- a/cmd/edit/machinepool/nodepool.go
+++ b/cmd/edit/machinepool/nodepool.go
@@ -27,6 +27,7 @@ func editNodePool(cmd *cobra.Command, nodePoolID string, clusterKey string, clus
 	isAutorepairSet := cmd.Flags().Changed("autorepair")
 	isTuningsConfigSet := cmd.Flags().Changed("tuning-configs")
 	isNodeDrainGracePeriodSet := cmd.Flags().Changed("node-drain-grace-period")
+	isTagsSet := cmd.Flags().Changed("tags")
 
 	// we don't support anymore the version parameter
 	if isVersionSet {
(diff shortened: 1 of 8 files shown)


cmd/edit/machinepool/nodepool.go
```diff
@@ -86,6 +89,14 @@ func editNodePool(cmd *cobra.Command, nodePoolID string, clusterKey string, clus
 	if isTaintsSet || interactive.Enabled() {
 		npBuilder = npBuilder.Taints(taintBuilders...)
 	}
+	var awsNodePoolBuilder *cmv1.AWSNodePoolBuilder
+	if isTagsSet || interactive.Enabled() {
+		awsNodePoolBuilder = cmv1.NewAWSNodePool().Tags(awsTags)
```
robpblake, line 94: @gdbranco Did we confirm with the BU this is what they want? My understanding from the last few conversations on this is that it will cause all nodes in the nodepool to be recreated and we're deliberately blocking that for other operations IIUC.

pkg/helper/machinepools/helpers.go
```diff
@@ -171,6 +172,52 @@ func taintValidator(val interface{}) error {
...
+			os.Exit(1)
+		}
+		if len(tagsInput) > 0 {
+			tags = strings.Split(tagsInput, ",")
+		}
+	}
+	if len(tags) > 0 {
+		if err := aws.UserTagValidator(tags); err != nil {
+			r.Reporter.Errorf("%s", err)
+			os.Exit(1)
```
robpblake, line 211: @gdbranco I think we should start returning errors from functions and not `os.Exit()` as it just makes them impossible to test.

cmd/edit/machinepool/nodepool.go
gdbranco: No, I'll be keeping tags as hidden and remove the interactive though

gdbranco:
Editing won't be done as part of the current feature, there's an issue with how the api handles removing the tag

PR #1870, opened 2024-03-21
OCM-6926 | fix: Support local envs in GetEnv()

tirthct:
https://issues.redhat.com/browse/OCM-6926

**Changed**
- Added complete URL check as a fallback mechanism for GetEnv() in case of specific local or other URLs
- Added test case to confirm the behavior

**Tested**
- Local env should be supported in GetEnv() : tested

### DIFF
--- a/pkg/ocm/config.go
+++ b/pkg/ocm/config.go
@@ -105,6 +105,12 @@ func GetEnv() (string, error) {
 				}
 			}
 		}
+	} else { // Go back to exact URL check (in case of other URLs like local envs, etc.)
+		for env, api := range urlAliases {
+			if api == strings.TrimSuffix(cfg.URL, "/") {
+				return env, nil
+			}
+		}
 	}
 
 	// Special use case for Admin users in the GovCloud environment
--- a/pkg/ocm/config.go
+++ b/pkg/ocm/config.go
@@ -91,13 +91,6 @@ func GetEnv() (string, error) {
 	urlAliases := URLAliases
 	if cfg.FedRAMP {
 		urlAliases = fedramp.URLAliases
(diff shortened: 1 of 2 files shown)


pkg/ocm/config.go
```diff
@@ -105,6 +105,12 @@ func GetEnv() (string, error) {
 				}
 			}
 		}
+	} else { // Go back to exact URL check (in case of other URLs like local envs, etc.)
```
gdbranco, line 108: I'd say we can move [this](https://github.com/openshift/rosa/blob/4d39081f7312cceb0e90e8ac3503d5ab12aa862c/pkg/ocm/config.go#L116) comment up [here](https://github.com/openshift/rosa/blob/4d39081f7312cceb0e90e8ac3503d5ab12aa862c/pkg/ocm/config.go#L93) and just remove the else if [this](https://github.com/openshift/rosa/blob/4d39081f7312cceb0e90e8ac3503d5ab12aa862c/pkg/ocm/config.go#L117) uses `range urlAliases` instead

tirthct: @gdbranco Moved the FedRAMP logic upwards and removed else condition

pkg/ocm/config.go
```diff
@@ -91,6 +91,13 @@ func GetEnv() (string, error) {
 	urlAliases := URLAliases
 	if cfg.FedRAMP {
 		urlAliases = fedramp.URLAliases
+		// Special use case for Admin users in the GovCloud environment
```
gdbranco, line 94: I think this is not required if we are adjusting the urlAliases variable. We can always resort for the fallback as they are the same code block if it doesn't include openshift. What do you think?

gdbranco (approved):

PR #1871, opened 2024-03-25
OCM-6883 | fix: revise the help message for rosa admin commands

marcolan018:
https://issues.redhat.com/browse/OCM-6883
to update the help message for rosa admin commands, to specify that those commands only handle the 'cluster-admin' user, not a common admin user in cluster-admins group

### DIFF
--- a/cmd/create/admin/cmd.go
+++ b/cmd/create/admin/cmd.go
@@ -39,9 +39,9 @@ const MaxPasswordLength = 23
 
 var Cmd = &cobra.Command{
 	Use:   "admin",
-	Short: "Creates an admin user to login to the cluster",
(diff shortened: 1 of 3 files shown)


ciaranRoche:
@marcolan018 can you explain why the describe does not show a cluster admin with a custom name? 

What is the technical aspect to stopping us from describing all types of cluster admin?

marcolan018:
@ciaranRoche 
no blocker things in technical aspect. I just keep the behavior no change. The admin user with custom name is never included in rosa admin commands, regardless the user is created by day-1 operation we start to support recently or by day-2 operation 
do you think we need to change the behavior there?

PR #1874, opened 2024-03-26
OCM-5549 | feat: Disable listing services while using FedRAMP login

hunterkepley:
https://issues.redhat.com/browse/OCM-5549

This MR makes it so when you are logged into FedRAMP envs, you cannot list managed services. Prevents a strange, and confusing error to the user.

### DIFF
--- a/pkg/ocm/managedservice.go
+++ b/pkg/ocm/managedservice.go
@@ -9,6 +9,8 @@ import (
 	"github.com/openshift/rosa/pkg/fedramp"
 )
 
+var fedrampError = fmt.Errorf("managed services are not supported for FedRAMP clusters")
+
 type CreateManagedServiceArgs struct {
 	ServiceType string
 	ClusterName string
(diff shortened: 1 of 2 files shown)


marcolan018:
@hunterkepley 
how about other command such as `rosa describe service`?

hunterkepley:
@ciaranRoche Create is covered from the last MR: https://github.com/openshift/rosa/pull/1851
Will update describe and delete though :+1:

oriAdler:
@hunterkepley should we add the check for the create service command as well?

hunterkepley:
> @hunterkepley should we add the check for the create service command as well?

@oriAdler https://github.com/openshift/rosa/pull/1851/files -- was done in this MR :+1:

pkg/ocm/managedservice.go
```diff
@@ -134,6 +134,10 @@ func (c *Client) CreateManagedService(args CreateManagedServiceArgs) (*msv1.Mana
 }
 
 func (c *Client) ListManagedServices(count int) (*msv1.ManagedServiceList, error) {
+	if fedramp.Enabled() {
+		return nil, fmt.Errorf("managed services are not supported for FedRAMP clusters")
```
robpblake, line 138: @hunterkepley in the spirit of not repeating the message everywhere, how about:

```
var fedrampError = fmt.Errorf("managed services are not supported for FedRAMP clusters")
```
and then methods updated to

```
if fedramp.Enabled() {
		return nil, fedrampError
}
```

hunterkepley: SGTM

hunterkepley:
Not sure why it closed :laughing: but made the change @robpblake

robpblake (approved):
/lgtm

PR #1875, opened 2024-03-26
OCM-6822 | fix: use cluster name in error message

chenz4027:
https://issues.redhat.com/browse/OCM-6822

Error now shows the cluster name instead of cluster id:
```
E: failed to create an external authentication provider for cluster 'magchen-hcp': external_auth.issuer.audiences must contain at least one audience
```

### DIFF
--- a/pkg/externalauthprovider/flags.go
+++ b/pkg/externalauthprovider/flags.go
@@ -79,14 +79,14 @@ func (e *ExternalAuthServiceImpl) CreateExternalAuthProvider(cluster *cmv1.Clust
 	externalAuthConfig, err := CreateExternalAuthConfig(args)
 	if err != nil {
 		return fmt.Errorf("failed to create an external authentication provider for cluster '%s': %s",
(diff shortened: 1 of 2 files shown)


pkg/externalauthprovider/flags.go
```diff
@@ -79,14 +79,14 @@ func (e *ExternalAuthServiceImpl) CreateExternalAuthProvider(cluster *cmv1.Clust
 	externalAuthConfig, err := CreateExternalAuthConfig(args)
 	if err != nil {
 		return fmt.Errorf("failed to create an external authentication provider for cluster '%s': %s",
-			cluster.ID(), err)
+			cluster.Name(), err)
```
gdbranco, line 82: I think we should show the clusterKey provided by the customer instead, this way it will show what the customer uses to identify the cluster. What do you think?

chenz4027: Thanks @gdbranco agreed, let me update it

PR #1878, opened 2024-03-27
OCM-6697 | fix: upgrade role error should print extra info when invoke from upgrade cluster

mikeshng:
In [PR](https://github.com/openshift/rosa/pull/1806/files#diff-0422b55e643b32d83ec9a053112d37486dfe3471aa851274e75e49931c88b013L500-L503), some of the error handlings were removed. I've gone through each one of them and brought back the one that should be present: upgrade role error handling when invoke from upgrade cluster.

[Jira ref](https://issues.redhat.com/browse/OCM-6697)

### DIFF
--- a/cmd/upgrade/roles/cmd.go
+++ b/cmd/upgrade/roles/cmd.go
@@ -122,7 +122,7 @@ func run(cmd *cobra.Command, argv []string) {
 	}
 	args.isInvokedFromClusterUpgrade = isInvokedFromClusterUpgrade
 
-	r.GetClusterKey()
+	clusterKey := r.GetClusterKey()
 	cluster = r.FetchCluster()
 
 	mode, err := aws.GetMode()
@@ -299,6 +299,9 @@ func run(cmd *cobra.Command, argv []string) {
 					LogError(roles.RosaUpgradeAccRolesModeAuto, ocmClient, policyVersion, err, reporter)
 					if args.isInvokedFromClusterUpgrade {
 						reporter.Errorf("%s", err)
+						if reporter.IsTerminal() {
+							reporter.Infof(generateClusterUpgradeInfo(clusterKey, args.clusterUpgradeVersion, mode))
+						}
 						os.Exit(1)
 					}
 					reporter.Errorf("Error upgrading the account role policies: %s", err)
@@ -417,6 +420,9 @@ func run(cmd *cobra.Command, argv []string) {
 		)
 		if err != nil {
 			r.Reporter.Errorf("%s", err)
+			if args.isInvokedFromClusterUpgrade && reporter.IsTerminal() {
+				reporter.Infof(generateClusterUpgradeInfo(clusterKey, args.clusterUpgradeVersion, mode))
+			}
 			os.Exit(1)
 		}
 	}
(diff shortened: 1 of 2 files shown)


cmd/upgrade/roles/cmd.go
```diff
@@ -1173,3 +1180,15 @@ func checkPolicyAndClusterVersionCompatibility(policyVersion, clusterVersion str
...
 	return nil
 }
+
+func printClusterUpgradeInfo(mode string, reporter *rprtr.Object) {
+	if args.isInvokedFromClusterUpgrade && reporter.IsTerminal() {
+		rolesStr := fmt.Sprintf("rosa upgrade roles -c %s --cluster-version=%s --mode=%s",
+			args.clusterID, args.clusterUpgradeVersion, mode)
+		upgradeClusterStr := fmt.Sprintf("rosa upgrade cluster -c %s", args.clusterID)
+		reporter.Infof("Account/Operator Role policies are not valid with upgrade version %s. "+
+			"Run the following command(s) to upgrade the roles and run the upgrade command again:\n\n"+
```
philipwu08, line 1190: Mike, it's clear to me but wonder if this could make it more clear:

"Run the following command(s) to upgrade the roles:\n" +
"\t%s\n\n"+
", then run the upgrade command again:\n" +
"\t%s\n", args.clusterUpgradeVersion, rolesStr, upgradeClusterStr)

mikeshng: Done as suggested. Thanks.

chenz4027:
/lgtm
I kinda wanna propose adding unit test for printClusterUpgradeInfo, but it might be tough for reporter. Up to you Mike.

oriAdler:
LGTM, can you please attach a sample output to the PR? just to get a sense of how the info message looks to the user?

cmd/upgrade/roles/cmd.go
```diff
@@ -1173,3 +1180,16 @@ func checkPolicyAndClusterVersionCompatibility(policyVersion, clusterVersion str
 	}
 	return nil
 }
+
+func printClusterUpgradeInfo(mode string, reporter *rprtr.Object) {
```
robpblake, line 1184: @mikeshng I'd say lets be consistent in our usage of global versus function parameters and avoid globals when we can. So how about:

```
printClusterUpgradeInfo(clusterKey string, clusterUpgradeVersion string, mode string, reporter *rprtr.Object)
```

as a function signature?
```

and remove the `if args.isInvokedFromClusterUpgrade` expression from the function as you're already guarding the function invocation with that anyway in the command.

It makes the function easier to test if we're not reliant on global state.

Comment (robpblake) [reply cmd/upgrade/roles/cmd.go:1184]: You could equally return the command as a `string` from the function, drop the `runtime` parameter and write a unit test for this :-)

So your caller then becomes:

```
	reporter.Infof(printClusterUpgradeInfo(args.clusterId, args.clusterUpgradeVersion, mode)
}
```

Comment (mikeshng) [reply cmd/upgrade/roles/cmd.go]: Done as suggested. Thanks.

**Comment by mikeshng** [comment]
> LGTM, can you please attach a sample output to the PR? just to get a sense of how the info message looks to the user?

```
I: Account/Operator Role policies are not valid with upgrade version 4.15. Run the following command(s) to upgrade the roles:
        rosa upgrade roles -c redacted_cluster_id_string_here --cluster-version=4.15 --mode=auto

, then run the upgrade command again:
	rosa upgrade cluster -c redacted_cluster_id_string_here
```

@oriAdler PTAL thanks

cmd/upgrade/roles/cmd.go
mikeshng: Done as suggested after making some fixes :) Sorry for the email spams.

chenz4027 (approved):
/lgtm

PR #1881, opened 2024-03-27
OCM-6302 | ci: Prow job update to check PR author in commit check

hunterkepley:
[https://issues.redhat.com/browse/OCM-6302](https://issues.redhat.com/browse/OCM-6302)

Updates the hack script, ran by prow, which checks PR commit msgs to also check PR author

### DIFF
--- a/hack/commit-msg-verify.sh
+++ b/hack/commit-msg-verify.sh
@@ -1,11 +1,28 @@
 #!/usr/bin/env bash
 
+# This regex matches the pull number with an author from the job spec. Prow does not have a PULL_AUTHOR job var
+PULL_AUTHOR=$(echo $JOB_SPEC | grep -Po "$PULL_NUMBER,\"author\":\"\K[^\"]*")
+
 # PULL_BASE_SHA and PULL_PULL_SHA are set in prow job
 PULL_BASE_SHA="${PULL_BASE_SHA:-$(git merge-base master HEAD)}"
 PULL_PULL_SHA="${PULL_PULL_SHA:-HEAD}"
 
-pattern="^[A-Z]+-[0-9]+ \| (feat|fix|docs|style|refactor|test|chore|build|ci|perf): .*$"
-
+# Lists of GitHub users for whom the commit validation should be skipped:
+# # This contains the bots used in the ROSA project
+# * openshift-merge-bot[bot] - bot responsible for merigng MRs
+# * openshift-ci[bot] - bot responsible for running tests / approvals
+declare -a skip_pr_authors=(
+  "openshift-ci[bot]"
+  "openshift-merge-bot[bot]"
+)
+echo "The PR Author is \"${PULL_AUTHOR}\""
+for skip_pr_author in "${skip_pr_authors[@]}"
+do
+  if [ "${PULL_AUTHOR}" = "${skip_pr_author}" ]; then
+    echo "The commits created by this PR author (probably bot) should be skipped!!!"
+    exit 0
+  fi
+done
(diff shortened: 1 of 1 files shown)


hack/commit-msg-verify.sh
```diff
@@ -1,21 +1,42 @@
...
+    exit 0
+  fi
+done
+for message in "${commit_messages[@]}"
+do
+  echo "validating commit message:\"$message\""
+  if ! echo "$message" | grep -qE "^[A-Z]+-[0-9]+ \| (feat|fix|docs|style|refactor|test|chore|build|ci|perf): .*$"; then
+    echo "Invalid commit message format. Expected format: JIRA_TICKET | TYPE: MESSAGE"
+    echo "Where:"
+    echo "   JIRA_TICKET is jira ticket ID (for example OCM-xxx)"
```
davidleerh, line 35: Just a question from my end on contributing.
If people outside the RH org wants to contribute to this open source project (with Apache 2.0 license), does that imply that we supply the contributors with the JIRA number via "github issues" prior to allowing them to commit?

hunterkepley: hm. I would assume that's the only way to go about it, but I think we should ask a higher power to be sure; @ciaranRoche

hunterkepley: As discussed in our call, im not too sure, but this logic was already in place, so the same system applies :smiley:

hack/commit-msg-verify.sh
```diff
@@ -1,21 +1,42 @@
...
-    echo "Please check CONTRIBUTING.md for more details"
+# Lists of GitHub users for whom the commit validation should be skipped:
+# # This contains the bots used in the ROSA project
+# * openshift-merge-bot[bot] - bot responsible for merigng MRs
+# * openshift-ci[bot] - bot responsible for running tests / approvals
+# * codecov[bot] - bot responsible for reporting code coverage
+declare -a skip_pr_authors=(
+  "openshift-ci[bot]"
+  "openshift-merge-bot[bot]"
+  "codecov[bot]"
```
robpblake, line 19: @hunterkepley Codecov doesn't/shouldn't have commit permissions on our repo. We should fail if it does, so I'd suggest removing this author from the list.

cheers,

Rob

hack/commit-msg-verify.sh
```diff
@@ -1,21 +1,42 @@
 #!/usr/bin/env bash
 
+# This regex matches the pull number with an author from the job spec. Prow does not have a PULL_AUTHOR job var
+export PULL_AUTHOR=$(echo $JOB_SPEC | grep -Po "$PULL_NUMBER,\"author\":\"\K[^\"]*")
```
robpblake, line 4: @hunterkepley I don't think you need the `export` the `PULL_AUTHOR` variable is local to this script 👍

hack/commit-msg-verify.sh
robpblake: @davidleerh It's a good discussion point.

Our release notes for `rosa` as a product are generated from Jira issues, so I think a practice could be that if/when we had a community contribution, we could create a Jira on their behalf? 

Wanna bring it up at office hours?

robpblake (changes requested):
@hunterkepley Can you remove CodeCov bot as a PR author. It shouldn't have commit permissions on our repo.

hack/commit-msg-verify.sh
hunterkepley: Let's discuss then, I think creating a ticket on their behalf is a great idea especially since the release notes are generated via jira issues

robpblake (approved):
/lgtm

PR #1884, opened 2024-03-28
OCM-6820 | fix: add checks for mandatory parameters

chenz4027:
https://issues.redhat.com/browse/OCM-6820

### DIFF
--- a/pkg/externalauthprovider/flags.go
+++ b/pkg/externalauthprovider/flags.go
@@ -324,34 +324,32 @@ func CreateExternalAuthConfig(args *ExternalAuthProvidersArgs) (*cmv1.ExternalAu
 	externalAuthBuilder := cmv1.NewExternalAuth().ID(args.name)
 	claimValidationRules := args.claimValidationRule
 
-	if args.issuerUrl != "" || args.issuerCaFile != "" || args.issuerAudiences != nil {
-		tokenIssuerBuilder := cmv1.NewTokenIssuer()
-
-		if args.issuerUrl != "" {
-			tokenIssuerBuilder.URL(args.issuerUrl)
-		}
-
-		if args.issuerCaFile != "" {
-			// Get certificate contents
-			ca := ""
-			if args.issuerCaFile != "" {
-				cert, err := os.ReadFile(args.issuerCaFile)
-				if err != nil {
-					return &cmv1.ExternalAuth{}, fmt.Errorf("expected a valid certificate bundle: %s", err)
-				}
-				ca = string(cert)
-			}
-			// Set the CA file, if any
-			if ca != "" {
-				tokenIssuerBuilder.CA(ca)
-			}
-		}
-
-		if args.issuerAudiences != nil {
-			tokenIssuerBuilder.Audiences(args.issuerAudiences...)
-		}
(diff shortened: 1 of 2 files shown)


pkg/externalauthprovider/flags.go
```diff
@@ -324,6 +324,21 @@ func CreateExternalAuthConfig(args *ExternalAuthProvidersArgs) (*cmv1.ExternalAu
...
+	if args.issuerUrl == "" {
+		return &cmv1.ExternalAuth{}, fmt.Errorf(
+			"issuer-url parameter is mandatory for creating an external authentication configuration")
+	}
+	if args.issuerAudiences == nil {
+		return &cmv1.ExternalAuth{}, fmt.Errorf(
+			"issuer-audience parameter is mandatory for creating an external authentication configuration")
+	}
+
 	if args.issuerUrl != "" || args.issuerCaFile != "" || args.issuerAudiences != nil {
```
hunterkepley, line 342: I wonder if we can reduce this now:
```if args.issuerUrl != "" || args.issuerCaFile != "" || args.issuerAudiences != nil {```
Now that we return when `args.issuerUrl == ""` and when `args.issuerAudiences == nil` , it feels redundant to have them in this. Maybe  we can just have `if args.issuerCaFile != "" { ... }`

hunterkepley: WDYT?

hunterkepley, line 342: Deleted my last comment to explain better;

Would just like to bring up something that may could be changed,

`if args.issuerUrl != "" || args.issuerCaFile != "" || args.issuerAudiences != nil {` --> we are checking the opposite conditions from 2 of the if statements above. For example:
`if args.issuerUrl == "" { return ... }`
There really isn't a need to have it in the condition below, because, if we get past this condition, then it is **always** not equal to an empty string.

Same goes for `if args.issuerAudience == nil { return ... }` --> The above if statement returns if it _is_ equal to nil, so there's no reason to check if it's not nil, as it will **always** be not nil after this if statement.

Not sure if it would be fine, given what you are trying to achieve, to only have `if args.issuerCaFile != "" { ... }` on L342, but just bringing this up :smile: LMK what you think of this

hunterkepley: overall, LGTM, just feels like redundant checks in that if statement due to the new if statements introduced

chenz4027: Sounds good, @hunterkepley one question I got is that should I keep the current check parameters logic, or should I move them to condition1 || condition 2 || condition3. The reason I separated them is due to the error messages being different.

hunterkepley: I think it may be best to move inside the `condition1 || condition 2 || condition3` -- this way we run through the least amount of if statements in a best case scenario (no issues), while keeping about the same for worst case
WDYT?

chenz4027: Sounds good to me, will also change that then. Thanks a lot @hunterkepley

hunterkepley:
if tests pass LGTM :slightly_smiling_face:

ciaranRoche (approved):
/lgtm

PR #1890, opened 2024-03-29
OCM-6391 | feat: Deprecate the region flag in commands which do not utilize it

hunterkepley:
https://issues.redhat.com/browse/OCM-6391

We put the region flag in almost every overarching command (`create`, `delete`, `list`, etc) -- yet many of the subcommands do _not_ use region. This can lead to confusion for users. As a solution, I have made a function in the same package following the same naming scheme as the function which adds the region flag. This way, we can disable using the region flag in any command we want with a single line.

### DIFF
--- a/pkg/arguments/arguments.go
+++ b/pkg/arguments/arguments.go
@@ -28,10 +28,12 @@ import (
 	"github.com/openshift/rosa/pkg/aws/profile"
 	"github.com/openshift/rosa/pkg/aws/region"
 	"github.com/openshift/rosa/pkg/debug"
-	"github.com/openshift/rosa/pkg/helper"
 )
 
 const boolType string = "bool"
+
+const regionFlagName = "region"
+const regionDeprecationMessage = "Region flag will be removed from this command in future versions"
 
 var hasUnknownFlags bool
 
@@ -314,19 +316,18 @@ func IsValidMode(modes []string, mode string) bool {
 	return false
 }
 
-func markGlobalFlagsHidden(command *cobra.Command, hidden ...string) {
+func deprecateRegion(command *cobra.Command) {
 	command.PersistentFlags().VisitAll(func(flag *pflag.Flag) {
-		name := flag.Name
-		if helper.Contains(hidden, name) {
-			flag.Hidden = true
+		if flag.Name == regionFlagName {
+			flag.Deprecated = regionDeprecationMessage
 		}
 	})
 }
 
-func MarkRegionHidden(parentCmd *cobra.Command, childrenCmds []*cobra.Command) {
+func MarkRegionDeprecated(parentCmd *cobra.Command, childrenCmds []*cobra.Command) {
+	deprecateRegion(parentCmd)
 	for _, cmd := range childrenCmds {
 		cmd.SetHelpFunc(func(command *cobra.Command, strings []string) {
-			markGlobalFlagsHidden(parentCmd, "region")
 			command.Parent().HelpFunc()(command, strings)
 		})
 	}
--- a/pkg/aws/region/flag.go
+++ b/pkg/aws/region/flag.go
@@ -26,7 +26,7 @@ import (
 	"github.com/openshift/rosa/pkg/helper"
 )
 
-// AddFlag adds the debug flag to the given set of command line flags.
+// AddFlag adds the region flag to the given set of command line flags.
 func AddFlag(flags *pflag.FlagSet) {
 	flags.StringVar(
 		&region,
--- a/cmd/create/cmd.go
+++ b/cmd/create/cmd.go
@@ -75,7 +75,10 @@ func init() {
 	globallyAvailableCommands := []*cobra.Command{
 		accountroles.Cmd, operatorroles.Cmd,
 		userrole.Cmd, ocmrole.Cmd,
-		oidcprovider.Cmd,
+		oidcprovider.Cmd, breakglasscredential.Cmd,
+		admin.Cmd, autoscaler.Cmd, dnsdomains.Cmd,
+		externalauthprovider.Cmd, idp.Cmd, kubeletconfig.Cmd,
+		ocmrole.Cmd, oidcprovider.Cmd, tuningconfigs.Cmd,
 	}
-	arguments.MarkRegionHidden(Cmd, globallyAvailableCommands)
+	arguments.MarkRegionDeprecated(Cmd, globallyAvailableCommands)
 }
--- a/cmd/describe/cmd.go
+++ b/cmd/describe/cmd.go
@@ -58,4 +58,13 @@ func init() {
 	flags := Cmd.PersistentFlags()
 	arguments.AddProfileFlag(flags)
 	arguments.AddRegionFlag(flags)
+
+	globallyAvailableCommands := []*cobra.Command{
+		tuningconfigs.Cmd, cluster.Cmd, service.Cmd,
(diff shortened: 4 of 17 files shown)


marcolan018:
maybe it's better to remove region flag in the overarching commands, and let the sub-command to add it if the flag required.
imho that makes more sense for the dev, to declare a flag when it's required.

hunterkepley:
> maybe it's better to remove region flag in the overarching commands, and let the sub-command to add it if the flag required. imho that makes more sense for the dev, to declare a flag when it's required.

@marcolan018 I wanted to do this originally, but it seems adding it to the overarching command is the 'norm' and it is done everywhere even when all subcommands do not use it. I think maybe there should be a wider discussion with @ciaranRoche and @robpblake about this, I am fine with either solution, just not sure what is 'best' here in terms of design/architecture

hunterkepley:
@marcolan018 Thread on this here: https://redhat-internal.slack.com/archives/C04C3G5MDKN/p1711732133140609

robpblake:
> > maybe it's better to remove region flag in the overarching commands, and let the sub-command to add it if the flag required. imho that makes more sense for the dev, to declare a flag when it's required.
> 
> @marcolan018 I wanted to do this originally, but it seems adding it to the overarching command is the 'norm' and it is done everywhere even when all subcommands do not use it. I think maybe there should be a wider discussion with @ciaranRoche and @robpblake about this, I am fine with either solution, just not sure what is 'best' here in terms of design/architecture

I left a comment on the Slack thread. I agree with @marcolan018 in that I think its preferable to take the time to refactor this in such a way that commands that do support the `--region` flag add it to their flagset and it is completely omitted from the flagsets of commands that do not support it. 

Your solution works, but I don't think its solving our underlying problem which is that we have global flags being added when its not appropriate. I would vote to take slightly more time to solve this now, once, rather than working around the issue but still having the incorrect underlying behaviour.

cheers,

Rob

pkg/aws/region/flag.go
```diff
@@ -24,18 +24,29 @@ import (
...
-		"region",
+		regionFlag,
 		"",
 		"Use a specific AWS region, overriding the AWS_REGION environment variable.",
 	)
 }
 
+// DisableFlag 'disables' the region flag from being used in the given set of command line flags.
+// This command checks if the region flag is used when it shouldn't be, then errors and exits if it is
+func DisableFlag(flags *pflag.FlagSet, reporter *reporter.Object) {
```
robpblake, line 44: @hunterkepley I think it would be useful for this to be shown in the `help` output for a command as well. To achieve that, we need the following for these commands when their defining their flagset:

```
cmd.Flags().MarkDeprecated("region", "Region flag will be removed from this command in future versions")
cmd.Flag("region").Hidden = false
```

So we could wrap that into a function that gets called when creating the flags for a command e.g:

```
func MarkRegionDeprecated(cmd *cobra.Command) {
flag := cmd.Flag("region")
if flag != nil {
  cmd.Flags().MarkDeprecated("cluster", "Region flag will be removed from this command in future versions")
  cmd.Flag("cluster").Hidden = false
  }
}
```

That seems to cause a warning to be printed when the user makes use of the deprecated flag, which would drop the need for the per-command log statement.

pkg/aws/region/flag.go
```diff
@@ -24,18 +24,29 @@ import (
...
 		"",
 		"Use a specific AWS region, overriding the AWS_REGION environment variable.",
 	)
 }
 
+// DeprecateFlag warns the user of region being deprecated in future versions. We should remove this flag
+// and only add region to commands which use it in the future.
+func DeprecateFlag(flags *pflag.FlagSet, reporter *reporter.Object) {
+	if flags.Changed(regionFlag) {
+		reporter.Warnf("Region flag will be deprecated in a future version for this command.")
```
robpblake, line 46: @hunterkepley We should ensure the two deprecation messages are the same. Shall we create a const and refer to that?

pkg/arguments/arguments.go
```diff
@@ -314,19 +314,20 @@ func IsValidMode(modes []string, mode string) bool {
 	return false
 }
 
-func markGlobalFlagsHidden(command *cobra.Command, hidden ...string) {
+func deprecateRegion(command *cobra.Command, hidden ...string) {
```
robpblake, line 317: @hunterkepley The signature of this function feels a bit off. We're passing in a variadic list of fields to be hidden, but have a hard-coded deprecation message. I'd be tempted just to drop the `hidden` parameter and do a hardcoded lookup for the `region` field for now.

pkg/aws/region/flag.go
robpblake: Or perhaps:

```
flag := flags.Lookup("region")
if flag.Changed {
 reporter.Warnf(flag.Deprecated)
}
```

pkg/arguments/arguments.go
```diff
@@ -314,19 +314,20 @@ func IsValidMode(modes []string, mode string) bool {
 	return false
 }
 
-func markGlobalFlagsHidden(command *cobra.Command, hidden ...string) {
+func deprecateRegion(command *cobra.Command) {
 	command.PersistentFlags().VisitAll(func(flag *pflag.Flag) {
 		name := flag.Name
-		if helper.Contains(hidden, name) {
-			flag.Hidden = true
+		if helper.Contains([]string{"region"}, name) {
```
robpblake, line 320: @hunterkepley Can we just simply this to:

```
const regionFlagName = "region"
const regionDeprecationMessage = ""Region flag will be removed from this command in future versions
```

and then

```
if flag.Name == regionFlagName {
    flag.Deprecated = regionDeprecationMessage
}
```

pkg/aws/region/flag.go
```diff
@@ -24,18 +24,29 @@ import (
...
-		"region",
+		regionFlag,
 		"",
 		"Use a specific AWS region, overriding the AWS_REGION environment variable.",
 	)
 }
 
+// DeprecateFlag warns the user of region being deprecated in future versions. We should remove this flag
+// and only add region to commands which use it in the future.
+func DeprecateFlag(flags *pflag.FlagSet, reporter *reporter.Object) {
```
robpblake, line 44: @hunterkepley I don't think is being used any more so we can remove it?

pkg/aws/region/flag.go
```diff
@@ -24,18 +24,29 @@ import (
 	"github.com/spf13/pflag"
 
 	"github.com/openshift/rosa/pkg/helper"
+	"github.com/openshift/rosa/pkg/reporter"
 )
 
-// AddFlag adds the debug flag to the given set of command line flags.
+var regionFlag = "region"
```
robpblake, line 30: @hunterkepley Don't think this is being used either

pkg/arguments/arguments.go
```diff
@@ -314,19 +314,20 @@ func IsValidMode(modes []string, mode string) bool {
...
+			flag.Deprecated =
+				"Region flag will be removed from this command in future versions"
 		}
 	})
 }
 
-func MarkRegionHidden(parentCmd *cobra.Command, childrenCmds []*cobra.Command) {
+func MarkRegionDeprecated(parentCmd *cobra.Command, childrenCmds []*cobra.Command) {
```
robpblake, line 327: @hunterkepley A unit test for this would be awesome if you can 👍

robpblake (approved):
/lgtm

PR #1893, opened 2024-04-01
OCM-6926 | fix: Added back the condition to check for empty env flag

tirthct:
**Changed**
- Added back the condition to check for empty env flag

**Tested**
- FedRAMP login flow should work as expected : tested

### DIFF
--- a/cmd/login/cmd.go
+++ b/cmd/login/cmd.go
@@ -50,6 +50,8 @@ const oauthClientId = "ocm-cli"
 
 var reAttempt bool
 
+var env string
+
 var args struct {
 	tokenURL      string
 	clientID      string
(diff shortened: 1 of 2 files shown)


cmd/login/cmd.go
```diff
@@ -50,6 +50,8 @@ const oauthClientId = "ocm-cli"
 
 var reAttempt bool
 
+var env string
```
gdbranco, line 53: If I'm not mistaken we are slowly trying to move away from "global" variables, would it be possible to rely on args.env or to use exclusively a local variable?

tirthct: We need constantly updated env variable between the run() and CheckAndLogIntoFedramp() methods. We can try and return env from the latter and update in run() but that would just confuse us more. I'd prefer keeping the global variable in this case.

PR #1894, opened 2024-04-02
Display AWS Billing Account on non-HCP

paradigm:
Billing accounts can now be set on ROSA Classic clusters (rather than just HCP clusters) and should be included in `rosa describe cluster` accordingly.

### DIFF
--- a/cmd/describe/cluster/cmd.go
+++ b/cmd/describe/cluster/cmd.go
@@ -236,7 +236,7 @@ func run(cmd *cobra.Command, argv []string) {
 		cluster.Version().ChannelGroup(),
 		clusterDNS,
 		creatorARN.AccountID,
-		BillingAccount(cluster, isHypershift),
+		BillingAccount(cluster),
 		cluster.API().URL(),
 		cluster.Console().URL(),
 		cluster.Region().ID(),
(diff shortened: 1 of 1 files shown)


cristianoveiga:
LGTM

cmd/describe/cluster/cmd.go
```diff
@@ -734,7 +734,7 @@ func formatClusterHypershift(cluster *cmv1.Cluster,
 }
 
 func BillingAccount(cluster *cmv1.Cluster, isHostedControlPlane bool) string {
-	if !isHostedControlPlane || cluster.AWS().BillingAccountID() == "" {
+	if cluster.AWS().BillingAccountID() == "" {
```
marcolan018, line 737: should we remove the `isHostedControlPlane` in the func parameter?

ciaranRoche:
@paradigm can you link a JIRA for this change. 

You will also need to update the commit message for the CI to pass - https://github.com/openshift/rosa/blob/master/CONTRIBUTING.md

paradigm:
Jira ticket is: https://issues.redhat.com/browse/OCM-6145

I've updated the commit message format per `CONTRIBUTING.md` and applied the suggested change to drop the now unnecessary parameter.

PR #1895, opened 2024-04-02
OCM-7110 | fix: GitHub IDP validation

mnecas:
Switch from URI validation to the DNS1123 validation. The openshift does not accept the URI format of domains. So when user specifies some address with prefix https:// it will fail.

### DIFF
--- a/cmd/create/idp/github.go
+++ b/cmd/create/idp/github.go
@@ -190,7 +190,7 @@ func buildGithubIdp(cmd *cobra.Command,
 			Help:     cmd.Flags().Lookup("hostname").Usage,
 			Default:  githubHostname,
 			Validators: []interactive.Validator{
-				interactive.IsURL,
+				interactive.IsValidHostname,
 			},
 		})
 		if err != nil {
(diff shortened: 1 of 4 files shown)


robpblake:
@mnecas You need to update your commit message to: `OCM-7110 | fix: GitHub IDP validation`

cheers,
Rob

cmd/create/idp/github.go
```diff
@@ -19,17 +19,27 @@ package idp
...
 
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/ocm"
 )
 
+// isValidHostname is same validation as in the Open Shift GitHub IDP CRD
+// https://github.com/openshift/kubernetes/blob/91607f5d750ba4002f87d34a12ae1cfd45b45b81/openshift-kube-apiserver/admission/customresourcevalidation/oauth/helpers.go#L13
+//
+//nolint:lll
+func isValidHostname(hostname string) bool {
```
robpblake, line 39: @mnecas A couple of unit tests for valid and invalid hosts would be great!

gdbranco:
Will also require changes to the interactive validation
https://github.com/openshift/rosa/blob/8396138dbd672226e12e8f372247db558cf4037f/cmd/create/idp/github.go#L203

gdbranco (approved):

PR #1897, opened 2024-04-02
OCM-7136 | fix: support create breakglasscredential without optional flags

chenz4027:
https://issues.redhat.com/browse/OCM-7136

### DIFF
--- a/pkg/breakglasscredential/flags.go
+++ b/pkg/breakglasscredential/flags.go
@@ -113,13 +113,15 @@ func CreateBreakGlass(cluster *cmv1.Cluster,
 func CreateBreakGlassConfig(args *BreakGlassCredentialArgs) (*cmv1.BreakGlassCredential, error) {
 	breakGlassBuilder := cmv1.NewBreakGlassCredential()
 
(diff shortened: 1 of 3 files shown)


pkg/breakglasscredential/flags.go
```diff
@@ -113,13 +113,15 @@ func CreateBreakGlass(cluster *cmv1.Cluster,
 func CreateBreakGlassConfig(args *BreakGlassCredentialArgs) (*cmv1.BreakGlassCredential, error) {
```
robpblake, line 113: @chenz4027 Could you add unit tests to cover the new functionality in this function. It would be good to have tests for:

- Passing a `nil` args which should result in no error and an empty `BreakGlassCredential`
- Passing args with just `username` set
- Passing args with just `expirationDuration` set

cheers,

Rob

chenz4027: Will work on it, thanks @robpblake

chenz4027: Updated - @robpblake would you mind taking another look?

PR #1899, opened 2024-04-03
OCM-6840 | fix: Add json output test for external auth

chenz4027:


chenz4027:
@gdbranco hmm not sure why, but it still misses the approved label though

cmd/list/externalauthprovider/cmd.go
```diff
@@ -79,7 +79,7 @@ func runWithRuntime(r *rosa.Runtime, cmd *cobra.Command) error {
 			r.Reporter.Errorf("%s", err)
 			os.Exit(1)
```
robpblake, line 80: @chenz4027 Can we also remove this `os.Exit()` and return the `err` instead?

chenz4027: Sure, @robpblake I'll work on it

andreadecorte:
lgtm

ciaranRoche (commented):
/lgtm

PR #1902, opened 2024-04-05
OCM-6568 | fix: ensure "mode" in interactive mode defaults to users input

ciaranRoche:
JIRA - https://issues.redhat.com/browse/OCM-6568

This change sets the default mode to the value the user passes as a param else defaults to true.


cmd/create/accountroles/cmd.go
```diff
@@ -329,7 +329,7 @@ func run(cmd *cobra.Command, argv []string) {
 		mode, err = interactive.GetOption(interactive.Input{
```
robpblake, line 329: @ciaranRoche I'd be tempted to roll this up into a common function seeing as it's essentially the same thing everywhere apart from the question to the user AFAICS

cmd/create/accountroles/cmd.go
```diff
@@ -329,7 +329,7 @@ func run(cmd *cobra.Command, argv []string) {
...
 			Question: "Role creation mode",
 			Help:     cmd.Flags().Lookup("mode").Usage,
```
robpblake, line 331: @ciaranRoche It's always been there, but should we perform the check for `nil` on the `cmd.Flags().Lookup("mode")` . Not a biggie...

ciaranRoche: Yeah lets get a ticket for this, might be worth baking it into milestone 2 of the improvements, we prob should look into a better more deterministic way to approach flags

pkg/interactive/validation_test.go
```diff
@@ -26,7 +26,7 @@ var _ = Describe("Validation", func() {
 			validator := MinValue(50)
 			err := validator(45)
 			Expect(err).To(HaveOccurred())
-			Expect(err.Error()).To(ContainSubstring("can only validate strings, got int"))
+			Expect(err.Error()).To(ContainSubstring("can only validate strings, got 45"))
```
gdbranco, line 29: Is this expected to be part of this PR?

ciaranRoche: the tests were never run, was missing the suite runner. Just updated to returned value

gdbranco: oh

PR #1909, opened 2024-04-09
OCM-7188: GitHub IDP Add hostname validation

mnecas:
The openshfit validates the GitHub hostname and it can not match the [*.]github.com If the user would put the hostname 'github.com' the hive idp sync would fail as the admission controller would not allow to update the oauth cr

### DIFF
--- a/pkg/interactive/validation.go
+++ b/pkg/interactive/validation.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"regexp"
 	"strconv"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/util/validation"
 	netutils "k8s.io/utils/net"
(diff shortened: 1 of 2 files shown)


pkg/interactive/validation.go
```diff
@@ -63,14 +64,21 @@ func IsURL(val interface{}) error {
 }
 
 func IsValidHostname(val interface{}) error {
-	if !_isValidHostname(val.(string)) {
+	hostname := val.(string)
+	if hostname == "github.com" || strings.HasSuffix(hostname, ".github.com") {
+		return fmt.Errorf(fmt.Sprintf("'%s' hostname cannot be equal [*.]github.com", hostname))
```
robpblake, line 69: @mnecas Missing a `to` in the error message e.g `cannot be equal to [*.]github.com`

pkg/interactive/validation.go
```diff
@@ -63,14 +64,21 @@ func IsURL(val interface{}) error {
...
 }
 
 // _isValidHostname is same validation as in the Open Shift GitHub IDP CRD
+// Validates the hostname DNS1123 format
 // https://github.com/openshift/kubernetes/blob/91607f5d750ba4002f87d34a12ae1cfd45b45b81/openshift-kube-apiserver/admission/customresourcevalidation/oauth/helpers.go#L13
+// and denies the [*.]github.com hostnames
+// https://github.com/openshift/kubernetes/blob/258f1d5fb6491ba65fd8201c827e179432430627/openshift-kube-apiserver/admission/customresourcevalidation/oauth/validate_github.go#L49
 //
 //nolint:lll
 func _isValidHostname(hostname string) bool {
```
robpblake, line 84: @mnecas Personally I'd just collapse `_isValidHostname` into `IsValidHostname` unless there's a specific reason to keep it as a standalone  function?

mnecas: no specific reason I was just following the _isUrl but agree

pkg/interactive/validation.go
mnecas: +1

robpblake (approved):
/lgtm

PR #1913, opened 2024-04-09
OCM-7079 | feat: select security groups interactive - filtering

oriAdler:
Checking if the security group has a cluster tag, and filter accordingly.
Improve UX by preventing an error returned from the backend, for the same validation.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -4062,7 +4062,7 @@ func getSecurityGroups(r *rosa.Runtime, cmd *cobra.Command, isVersionCompatibleC
 			os.Exit(1)
 		}
 		*additionalSgIds = interactiveSgs.
-			GetSecurityGroupIds(r, cmd, vpcId, kind)
+			GetSecurityGroupIds(r, cmd, vpcId, kind, "")
 	}
 	for i, sg := range *additionalSgIds {
 		(*additionalSgIds)[i] = strings.TrimSpace(sg)
--- a/pkg/interactive/securitygroups/security_groups.go
+++ b/pkg/interactive/securitygroups/security_groups.go
@@ -41,7 +41,7 @@ var (
 )
 
 func GetSecurityGroupIds(r *rosa.Runtime, cmd *cobra.Command,
-	targetVpcId string, kind string) []string {
+	targetVpcId string, kind string, id string) []string {
 	possibleSgs, err := r.AWSClient.GetSecurityGroupIds(targetVpcId)
 	if err != nil {
 		r.Reporter.Errorf("There was a problem retrieving security groups for VPC '%s': %v", targetVpcId, err)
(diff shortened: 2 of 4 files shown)


pkg/interactive/securitygroups/security_groups.go
```diff
@@ -77,13 +77,18 @@ func GetSecurityGroupIds(r *rosa.Runtime, cmd *cobra.Command,
...
-func isValidSecurityGroup(sg types.SecurityGroup) bool {
+func isValidSecurityGroup(sg types.SecurityGroup, id string) bool {
 	if aws.Ec2ResourceHasTag(sg.Tags, tags.RedHatManaged, strconv.FormatBool(true)) {
 		return false
 	}
 	if awsSdk.ToString(sg.GroupName) == "default" {
 		return false
 	}
+	if id != "" {
+		if aws.Ec2ResourceHasTag(sg.Tags, fmt.Sprintf("kubernetes.io/cluster/%s", id), "owned") {
```
davidleerh, line 88: Hello~
I'm Just wondering if we can get away with checking substring "kubernetes.io/cluster/" with value "owned" instead of having to pass in the cluster id.

I am assuming this is required for displaying list of additional sec. group options in interactive mode?
As long as the security group is not owned by any cluster, then would it be safe to display for selection?

oriAdler: I agree with this approach.
Currently, we check the tag with the cluster/infra ID in the backend.
If we decide to check the substring, changes should be applied both to CLI and backend, for consistency.

cmd/create/cluster/cmd.go
```diff
@@ -4062,7 +4062,7 @@ func getSecurityGroups(r *rosa.Runtime, cmd *cobra.Command, isVersionCompatibleC
 			os.Exit(1)
 		}
 		*additionalSgIds = interactiveSgs.
-			GetSecurityGroupIds(r, cmd, vpcId, kind)
+			GetSecurityGroupIds(r, cmd, vpcId, kind, "")
```
davidleerh, line 4065: One last question before LGTM from me :slightly_smiling_face: 
During cluster create flow, if the user opts in to use additional sec. groups, I'm wondering if the list of security groups visible should be filtered to the ones that is **not owned** by any cluster.
From what I recall, the resource tag 
kubernetes.io/cluster/{cluster_id}=owned is also used when deleting / cleaning up a cluster.

Please correct me here if I'm incorrect . Trying to better understand the flow!

oriAdler: At cluster creation, we don't have cluster/infra ID to check the owned tag.
If we decide to filter security groups that are owned by any cluster, checking the substring is sufficient, as you suggested.
I agree with your suggestion.
@gdbranco and I had a similar discussion, we can confirm with BU/SRE and open a ticket to track changes in the CLI and backend, to validate if a security group is owned by any cluster.

oriAdler: @davidleerh created this ticket https://issues.redhat.com/browse/OCM-7427 to track filtering upon cluster creation.

PR #1917, opened 2024-04-10
OCM-6729 | feat: Show user tags on describe machine pool for HCP

willkutler:
https://issues.redhat.com/browse/OCM-6729

### DIFF
--- a/cmd/describe/machinepool/cmd_test.go
+++ b/cmd/describe/machinepool/cmd_test.go
@@ -25,6 +25,7 @@ Desired replicas:                      0
 Current replicas:                      
 Instance type:                         m5.xlarge
 Labels:                                
+Tags:                                  
 Taints:                                
 Availability zone:                     us-east-1a
 Subnet:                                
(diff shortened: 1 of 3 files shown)


pkg/ocm/output/machinepools.go
```diff
@@ -40,6 +40,20 @@ func PrintLabels(labels map[string]string) string {
 	return strings.Join(output, ", ")
 }
 
+func PrintUserAwsTags(tags map[string]string) string {
+	if len(tags) == 0 {
+		return ""
+	}
+	output := []string{}
+	for k, v := range tags {
+		if !strings.HasPrefix(k, "aws:") {
```
gdbranco, line 49: Could it have aws prefix if we do not allow it to be supplied? Considering we are not fetching it from AWS I'd say this wouldn't ever show up :thinking:

cmd/describe/machinepool/cmd_test.go
```diff
@@ -25,6 +25,7 @@ Desired replicas:                      0
 Current replicas:                      
 Instance type:                         m5.xlarge
 Labels:                                
+Tags:
```
gdbranco, line 28: Can you also add one test which has tags? So that if the output is changed for some reason it would be picked up

pkg/ocm/output/machinepools.go
willkutler: yeah you're right, just removed

PR #1922, opened 2024-04-11
OCM-7252 | feat: Move describe machinepool funcs to pkg, test, split up, make interface

hunterkepley:
https://issues.redhat.com/browse/OCM-7252

```The following functions need to be moved : 

    describeMachinePool
    describeNodePool
    formatNodePoolOutput

These functions should be decomposed to smaller functions where possible to ensure ease of Unit Testing

An interface for these functions should be created and mocks generated to enable ease of testing for the describe machine pool cmd.

DONE

    Functions have been moved to new machine pool package
    Functions test coverage is at least 70%
```

* Move funcs to new pkg dir (`pkg/machinepool`)
* Improve errors slightly (node pools returned `machine pool` in not found error)
* Split up funcs where possible
* Test as much as possible in pkg
* Make interface + struct using interface
* Prepare mocks


hunterkepley:
Coverage should jump up after next MR/ticket, which will include using the mocks + service interface to test the funcs used directly in the command which print, and do not provide returned output

hunterkepley:
> Coverage should jump up after next MR/ticket, which will include using the mocks + service interface to test the funcs used directly in the command which print, and do not provide returned output

Cannot currently test these as the next MR includes a new runner+mocks, which are needed to run those funcs

willkutler (approved):

cmd/describe/machinepool/cmd_test.go
```diff
@@ -176,7 +176,7 @@ var _ = Describe("Upgrade machine pool", func() {
 				stdout, stderr, err := test.RunWithOutputCaptureAndArgv(runWithRuntime, testRuntime.RosaRuntime,
 					Cmd, &[]string{nodePoolName})
 				Expect(err).ToNot(BeNil())
-				Expect(err.Error()).To(ContainSubstring(fmt.Sprintf("Machine pool '%s' not found", nodePoolName)))
+				Expect(err.Error()).To(ContainSubstring(fmt.Sprintf("Node pool '%s' not found", nodePoolName)))
```
robpblake, line 179: @hunterkepley I don't think we should change to `Node pool`. `Machine pool` is the customer visible term and also what's used to execute this i.e `rosa describe machinepool`.

robpblake:
This is nice work @hunterkepley 🚀

cmd/describe/machinepool/cmd_test.go
hunterkepley: understood

hunterkepley:
@robpblake changed back to `machine pool` in the not found message, PTAL

pkg/machinepool/machinepool.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/openshift/rosa/pkg/output"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+var fetchMessage string = "Fetching %s '%s' for cluster '%s'"
+var notFoundMessage string = "Machine pool '%s' not found"
+
+//go:generate mockgen -source=machinepool.go -package=mocks -destination=machinepool_mock.go
+type MachinePoolService interface {
+	DescribeMachinePool(*rosa.Runtime, *cmv1.Cluster, string, string) error
```
gdbranco, line 19: I believe it is possible to have a single describe as they both take the same arguments, a factory to differentiate based on cluster topology

hunterkepley: Implemented a factory-style solution, and kept it simple. This removed a few lines from the command in the end, which is a nice outcome.

PR #1924, opened 2024-04-11
OCM-7412 | fix: list instance types filter by region requires installer role arn

gdbranco:
[OCM-7412](https://issues.redhat.com//browse/OCM-7412)
## What
Adjusts to allow to supply a role-arn and uses interactive mode to filter existing installer role arns in logged in account


ciaranRoche:
Can we update the ticket to describe what we are changing here, as current linked ticket is closed, want to make sure this goes through QE.

Other than that LGTM

ciaranRoche (approved):
/lgtm

PR #1930, opened 2024-04-15
OCM-7340 | fix: Add info msg for empty case

chenz4027:
https://issues.redhat.com/browse/OCM-7340

### DIFF
--- a/cmd/describe/breakglasscredential/cmd.go
+++ b/cmd/describe/breakglasscredential/cmd.go
@@ -124,6 +124,10 @@ func runWithRuntime(r *rosa.Runtime, cmd *cobra.Command, argv []string) error {
 	}
 
 	if getKubeconfig {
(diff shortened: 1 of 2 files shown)


cmd/describe/breakglasscredential/cmd.go
```diff
@@ -124,6 +124,9 @@ func runWithRuntime(r *rosa.Runtime, cmd *cobra.Command, argv []string) error {
 	}
 
 	if getKubeconfig {
+		if breakGlassCredentialConfig.Kubeconfig() == "" {
+			r.Reporter.Infof("The credential is not ready yet. Please wait a few minutes for it to be fully ready.")
```
andreadecorte, line 128: minor, but we could return directly after, since in any case we are printing an empty string

chenz4027: @andreadecorte Updated

PR #1931, opened 2024-04-15
OCM-6391 | fix: Fixed bug where deprecation shows for parent cmd

hunterkepley:
[OCM-6391](https://issues.redhat.com//browse/OCM-6391) is a previously finished ticket which had a small bug, it is now fixed. The deprecation msg was showing up in parent cmds, when it was only meant to show up in certain subcommands. This MR fixes that

### DIFF
--- a/cmd/create/cmd.go
+++ b/cmd/create/cmd.go
@@ -78,7 +78,7 @@ func init() {
 		oidcprovider.Cmd, breakglasscredential.Cmd,
 		admin.Cmd, autoscaler.Cmd, dnsdomains.Cmd,
 		externalauthprovider.Cmd, idp.Cmd, kubeletconfig.Cmd,
-		ocmrole.Cmd, oidcprovider.Cmd, tuningconfigs.Cmd,
+		oidcprovider.Cmd, tuningconfigs.Cmd,
 	}
 	arguments.MarkRegionDeprecated(Cmd, globallyAvailableCommands)
 }
--- a/pkg/arguments/arguments.go
+++ b/pkg/arguments/arguments.go
@@ -20,6 +20,7 @@ package arguments
 
 import (
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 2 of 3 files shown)


hunterkepley:
did not print deprecation msg on usage of cmds, added manually to each run command in the same func, now prints when running the deprecated cmds

cmd/create/cmd.go
```diff
@@ -78,7 +78,7 @@ func init() {
 		oidcprovider.Cmd, breakglasscredential.Cmd,
 		admin.Cmd, autoscaler.Cmd, dnsdomains.Cmd,
 		externalauthprovider.Cmd, idp.Cmd, kubeletconfig.Cmd,
-		ocmrole.Cmd, oidcprovider.Cmd, tuningconfigs.Cmd,
+		oidcprovider.Cmd, tuningconfigs.Cmd,
```
gdbranco, line 81: Any reason to remove ocmrole.Cmd?

gdbranco: Oh it's already in there

patjlm:
This change will break json output as the message is always printed:
```
$ rosa list clusters -o json
W: Region flag will be removed from this command in future versions
[]
```

pkg/arguments/arguments.go
```diff
@@ -325,10 +326,15 @@ func deprecateRegion(command *cobra.Command) {
...
 func MarkRegionDeprecated(parentCmd *cobra.Command, childrenCmds []*cobra.Command) {
-	deprecateRegion(parentCmd)
 	for _, cmd := range childrenCmds {
 		cmd.SetHelpFunc(func(command *cobra.Command, strings []string) {
+			deprecateRegion(parentCmd)
 			command.Parent().HelpFunc()(command, strings)
 		})
+		currentRun := cmd.Run
+		cmd.Run = func(c *cobra.Command, args []string) {
+			_, _ = fmt.Fprintf(os.Stdout, "%s%s\n", "\u001B[0;33mW:\u001B[m ", regionDeprecationMessage)
```
gdbranco, line 336: We can check if output flag is being requested before printing

PR #1932, opened 2024-04-15
OCM-4806 | fix: align list users with list idps

chenz4027:
https://issues.redhat.com/browse/OCM-4806


mikeshng (approved):
/lgtm

andreadecorte:
/lgtm
this command probably will be refactored but better to do that in a dedicated effort to avoid polluting functional changes

PR #1934, opened 2024-04-16
OCM-6199 | ci: migrate the Dockerfile from the repo release to rosa

yasun1:


images/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.21-openshift-4.16 as builder
```
ciaranRoche, line 1: Can we leave a comment in this file to explain what it is used for to ensure the context of what this dockerfile is known

ciaranRoche (commented):
Small NP inline, other than that looks good

images/Dockerfile
yasun1: How about moving the filename to `Dockerfile.e2e` and leaving `Dockerfile` for regular usage?

ciaranRoche (approved):
/lgtm

PR #1936, opened 2024-04-16
OCM-6391 | fix: only show message when region is changed and no output mode set

gdbranco:


robpblake (approved):
/lgtm

PR #1937, opened 2024-04-16
OCM-6391 | fix: Only print warning when region is used without output flag

hunterkepley:

image

No longer prints warning with -o/--output is used. Only prints warning when region flag is used (of course with commands that have intent to deprecate this flag)

### DIFF
--- a/pkg/arguments/arguments.go
+++ b/pkg/arguments/arguments.go
@@ -34,6 +34,7 @@ import (
 const boolType string = "bool"
 
 const regionFlagName = "region"
+const outputFlagName = "output"
 const regionDeprecationMessage = "Region flag will be removed from this command in future versions"
 
 var hasUnknownFlags bool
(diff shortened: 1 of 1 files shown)


pkg/arguments/arguments.go
```diff
@@ -333,11 +334,13 @@ func MarkRegionDeprecated(parentCmd *cobra.Command, childrenCmds []*cobra.Comman
...
-			hasChangedOutputFlag := outputFlag != nil && outputFlag.Value.String() != outputFlag.DefValue
-			hasChangedRegionFlag := regionFlag != nil && regionFlag.Value.String() != regionFlag.DefValue
-			if hasChangedRegionFlag && !hasChangedOutputFlag {
+			hasRegion := false
+			cmd.Parent().PersistentFlags().VisitAll(func(flag *pflag.Flag) {
+				if flag.Name == regionFlagName {
+					hasRegion = true
+				}
+			})
+			if cmd.Flags().Lookup(outputFlagName).Value.String() == "" && hasRegion {
```
hunterkepley, line 343: note on the value comparison choice - `.Changed` does not seem to function how I would expect with the output flag, it's always true even when not used. This comparison seems to work fine in it's replacement

PR #1938, opened 2024-04-16
OCM-7253 | feat: Refactor describe machine pool cmd to use new default runner

hunterkepley:
[OCM-7253](https://issues.redhat.com//browse/OCM-7253)

Pt 2 to the RCI Milestone 2 for `describe machine-pool`

Refactors the command to use new default runner, and refactors tests

DoD:

    Command uses new default runner
    Command has at least 70% unit test coverage

### DIFF
--- a/cmd/describe/cmd.go
+++ b/cmd/describe/cmd.go
@@ -49,7 +49,8 @@ func init() {
 	Cmd.AddCommand(installation.Cmd)
 	Cmd.AddCommand(upgrade.Cmd)
 	Cmd.AddCommand(tuningconfigs.Cmd)
-	Cmd.AddCommand(machinepool.Cmd)
+	machinePoolCommand := machinepool.NewDescribeMachinePoolCommand()
+	Cmd.AddCommand(machinePoolCommand)
 	Cmd.AddCommand(kubeletconfig.Cmd)
 	Cmd.AddCommand(autoscaler.NewDescribeAutoscalerCommand())
 	Cmd.AddCommand(externalauthprovider.Cmd)
@@ -61,10 +62,10 @@ func init() {
 
 	globallyAvailableCommands := []*cobra.Command{
 		tuningconfigs.Cmd, cluster.Cmd, service.Cmd,
-		machinepool.Cmd, addon.Cmd, upgrade.Cmd,
+		machinePoolCommand, addon.Cmd, upgrade.Cmd,
 		admin.Cmd, breakglasscredential.Cmd,
 		externalauthprovider.Cmd, installation.Cmd,
-		kubeletconfig.Cmd, machinepool.Cmd, upgrade.Cmd,
+		kubeletconfig.Cmd, upgrade.Cmd,
 	}
 	arguments.MarkRegionDeprecated(Cmd, globallyAvailableCommands)
 }
--- a/cmd/describe/machinepool/cmd.go
+++ b/cmd/describe/machinepool/cmd.go
@@ -17,8 +17,8 @@ limitations under the License.
 package machinepool
 
 import (
+	"context"
 	"fmt"
-	"os"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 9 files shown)


cmd/describe/cmd.go
```diff
@@ -49,7 +49,8 @@ func init() {
 	Cmd.AddCommand(installation.Cmd)
 	Cmd.AddCommand(upgrade.Cmd)
 	Cmd.AddCommand(tuningconfigs.Cmd)
-	Cmd.AddCommand(machinepool.Cmd)
+	machinePoolCommand := machinepool.NewDescribeMachinePoolCommand()
+	Cmd.AddCommand(machinePoolCommand)
```
hunterkepley, line 53: the separate definition is temporary due to deprecation of region flag in this command

hunterkepley:
DoD met

cmd/describe/machinepool/cmd.go
```diff
@@ -29,61 +30,68 @@ import (
...
+			}
+		}
+		if args.machinePool == "" {
+			return fmt.Errorf("You need to specify a machine pool name")
+		}
+		clusterKey := runtime.GetClusterKey()
+		cluster := runtime.FetchCluster()
+		if cluster.State() != cmv1.ClusterStateReady {
+			runtime.Reporter.Errorf("Cluster '%s' is not yet ready", clusterKey)
+			os.Exit(1)
```
robpblake, line 89: @hunterkepley We can just return the `error` here and remove the `os.Exit(1)` call 👍

cmd/describe/machinepool/cmd.go
```diff
@@ -29,61 +30,68 @@ import (
...
+const (
+	use     = "machinepool"
+	alias   = "machine-pool"
+	short   = "Show details of a machine pool on a cluster"
+	long    = "Show details of a machine pool on a cluster."
+	example = `  # Show details of a machine pool named "mymachinepool"" on a cluster named "mycluster"
+  rosa describe machinepool --cluster=mycluster --machinepool=mymachinepool`
+)
 
 var args struct {
```
robpblake, line 42: @hunterkepley Take a look at the discussion in this thread: https://github.com/openshift/rosa/pull/1777#discussion_r1512694837

We're trying to find a pattern to avoid the global `args` struct to keep everything encapsulated. That discussion is our first attempt. Give it a try applying it here and let us know if it feels OK for you or if you can see a better way to approach it 👍 

cheers,

Rob

cmd/describe/machinepool/cmd_test.go
```diff
@@ -124,11 +129,34 @@ kind: MachinePool
 `
 )
 
+type stdOutAttributes struct {
```
robpblake, line 132: @hunterkepley This looks generally useful across a lot of testing scenarios. Should we move it into the test helpers or perhaps even introduce it onto the `TestRuntime` as a method that can be easily used? WDYT?

hunterkepley: yeah sounds good, ill add it to the test helpers

cmd/describe/machinepool/cmd.go
```diff
@@ -29,61 +29,68 @@ import (
...
+		// Allow the use also directly the machine pool id as positional parameter
+		if len(argv) == 1 && !cmd.Flag("machinepool").Changed {
+			userOptions.machinepool = argv[0]
+		} else {
+			err := cmd.ParseFlags(argv)
+			if err != nil {
+				return fmt.Errorf("unable to parse flags: %s", err)
+			}
+		}
+		options.Bind(userOptions)
```
robpblake, line 81: @hunterkepley I think it makes sense for the `Bind` method to return an `error` if the user hasn't supplied enough fields or fields that contradict each other. Essentially `Bind` validates the options the user has supplied.

So for all validation of user options it would look like:

```
err  := options.Bind(userOptions)
if err != nil {
 return err
}
```

We can then extend the validation on the options without having to touch the runner.  We can also unit test fairly easily option validation scenarios.

hunterkepley: Sounds good

hunterkepley: Updated tests and added validation + error return to bind

cmd/describe/machinepool/options.go
```diff
@@ -0,0 +1,40 @@
...
+	reporter *reporter.Object
+
+	args DescribeMachinepoolUserOptions
+}
+
+func NewDescribeMachinepoolUserOptions() DescribeMachinepoolUserOptions {
+	return DescribeMachinepoolUserOptions{machinepool: ""}
+}
+
+func NewDescribeMachinepoolOptions() (*DescribeMachinepoolOptions, error) {
```
robpblake, line 23: @hunterkepley Do we need the `error` on the return signature of this function? It only ever returns `nil`

robpblake (approved):
/lgtm

PR #1939, opened 2024-04-16
OCM-6592 | fix: filter out local-zone and wave-length zone associated subnets during 'rosa create machinepool (HCP)'

davidleerh:
New PR
Resolve merge conflicts from: 
https://github.com/openshift/rosa/pull/1834


pkg/aws/client.go
```diff
@@ -613,13 +614,54 @@ func (c *awsClient) FilterVPCsPrivateSubnets(subnets []ec2types.Subnet) ([]ec2ty
...
+	}
+
 	if len(privateSubnets) < 1 {
 		return nil, fmt.Errorf("failed to find private subnets associated with VPC '%s'", *subnets[0].VpcId)
 	}
 
 	return privateSubnets, nil
 }
 
+func (c *awsClient) FilterSubnetsWithStandardAvailabilityZones(subnets []ec2types.Subnet) ([]ec2types.Subnet, error) {
```
ciaranRoche, line 632: Does this func need to be exported?

pkg/aws/client.go
```diff
@@ -613,13 +614,54 @@ func (c *awsClient) FilterVPCsPrivateSubnets(subnets []ec2types.Subnet) ([]ec2ty
...
+	filteredSubnets := []ec2types.Subnet{}
+	subnetAvailabilityZonesList := []string{}
+	standardAvailabilityZonesMap := map[string]bool{}
+
+	// list of availability zones to pass as input
+	for _, subnet := range subnets {
+		subnetAvailabilityZonesList = append(subnetAvailabilityZonesList, *subnet.AvailabilityZone)
```
ciaranRoche, line 639: Can you wrap this in `aws.ToString` just to make the deref a little safer

```
subnetAvailabilityZonesList = append(subnetAvailabilityZonesList, aws.ToString(*subnet.AvailabilityZone))
```

pkg/aws/client.go
```diff
@@ -613,13 +614,54 @@ func (c *awsClient) FilterVPCsPrivateSubnets(subnets []ec2types.Subnet) ([]ec2ty
...
+
+	describeAvailabilityZonesOutput, err := c.ec2Client.DescribeAvailabilityZones(
+		context.Background(),
+		&ec2.DescribeAvailabilityZonesInput{ZoneNames: subnetAvailabilityZonesList})
+	if err != nil {
+		return filteredSubnets, err
+	}
+
+	for _, availabilityZone := range describeAvailabilityZonesOutput.AvailabilityZones {
+		if *availabilityZone.ZoneType == standardZone {
```
ciaranRoche, line 650: Same as above to wrap `*availabilityZone.ZoneType` in the `ToString`

pkg/aws/client.go
```diff
@@ -613,13 +614,54 @@ func (c *awsClient) FilterVPCsPrivateSubnets(subnets []ec2types.Subnet) ([]ec2ty
...
+			standardAvailabilityZonesMap[*availabilityZone.ZoneName] = true
```
ciaranRoche, line 651:
```suggestion
			standardAvailabilityZonesMap[aws.ToString(availabilityZone.ZoneName)] = true
```

pkg/aws/client.go
```diff
@@ -613,13 +614,54 @@ func (c *awsClient) FilterVPCsPrivateSubnets(subnets []ec2types.Subnet) ([]ec2ty
...
+		}
+	}
+
+	for _, subnet := range subnets {
+		_, exists := standardAvailabilityZonesMap[*subnet.AvailabilityZone]
```
ciaranRoche, line 656:
```suggestion
		_, exists := standardAvailabilityZonesMap[aws.ToString(subnet.AvailabilityZone)]
```

pkg/aws/client.go
davidleerh: Nope the function does not need to be exported.
Was following unit test conventions in the client_test file where the methods being tested were called through the "Client" interface, instead of a mock client.

I will go ahead and change this to a non exportable method.

PR #1940, opened 2024-04-16
OCM-5186 | fix: align nodepool actions with machinepools

chenz4027:
https://issues.redhat.com/browse/OCM-5186
https://issues.redhat.com/browse/OCM-4597

### DIFF
--- a/cmd/edit/machinepool/cmd.go
+++ b/cmd/edit/machinepool/cmd.go
@@ -165,9 +165,14 @@ func run(cmd *cobra.Command, argv []string) {
 		}
 	}
 
+	var err error
 	if cluster.Hypershift().Enabled() {
-		editNodePool(cmd, machinePoolID, clusterKey, cluster, r)
+		err = editNodePool(cmd, machinePoolID, clusterKey, cluster, r)
 	} else {
-		editMachinePool(cmd, machinePoolID, clusterKey, cluster, r)
+		err = editMachinePool(cmd, machinePoolID, clusterKey, cluster, r)
+	}
+	if err != nil {
+		r.Reporter.Errorf("%s", err)
+		os.Exit(1)
 	}
 }
--- a/cmd/edit/machinepool/machinepool.go
+++ b/cmd/edit/machinepool/machinepool.go
@@ -1,7 +1,7 @@
 package machinepool
 
 import (
-	"os"
+	"fmt"
 	"regexp"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
@@ -18,11 +18,9 @@ import (
 var machinePoolKeyRE = regexp.MustCompile(`^[a-z]([-a-z0-9]*[a-z0-9])?$`)
 
 func editMachinePool(cmd *cobra.Command, machinePoolID string, clusterKey string, cluster *cmv1.Cluster,
-	r *rosa.Runtime) {
-	var err error
+	r *rosa.Runtime) error {
 	if !machinePoolKeyRE.MatchString(machinePoolID) {
-		r.Reporter.Errorf("Expected a valid identifier for the machine pool")
-		os.Exit(1)
+		return fmt.Errorf("Expected a valid identifier for the machine pool")
 	}
 
 	mpHelpers.HostedClusterOnlyFlag(r, cmd, "version")
@@ -45,8 +43,7 @@ func editMachinePool(cmd *cobra.Command, machinePoolID string, clusterKey string
 	r.Reporter.Debugf("Loading machine pools for cluster '%s'", clusterKey)
 	machinePools, err := r.OCMClient.GetMachinePools(cluster.ID())
 	if err != nil {
-		r.Reporter.Errorf("Failed to get machine pools for cluster '%s': %v", clusterKey, err)
-		os.Exit(1)
+		return fmt.Errorf("Failed to get machine pools for cluster '%s': %v", clusterKey, err)
 	}
 
 	var machinePool *cmv1.MachinePool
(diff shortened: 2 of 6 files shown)


cmd/edit/machinepool/nodepool.go
```diff
@@ -57,7 +57,7 @@ func editNodePool(cmd *cobra.Command, nodePoolID string, clusterKey string, clus
 		os.Exit(1)
```
robpblake, line 57: hey @chenz4027 👋 

I'm wondering if we can do a bit of a tidy up whilst we're here and remove the `os.Exit(1)` calls from this function and instead return the `error`. At the minute we simply cannot test this due to those calls being in the function. Do you have time to do that?

cmd/edit/machinepool/nodepool.go
```diff
@@ -229,13 +231,24 @@ func getNodePoolReplicas(cmd *cobra.Command,
 	autoscaling = args.autoscalingEnabled
 	replicasRequired := existingAutoscaling == nil
 
+	scalingUpdated = isMinReplicasSet || isMaxReplicasSet || isReplicasSet || isAutoscalingSet || interactive.Enabled()
```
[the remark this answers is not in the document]
chenz4027: @robpblake Yep yep

cmd/edit/machinepool/nodepool.go
```diff
@@ -57,7 +57,7 @@ func editNodePool(cmd *cobra.Command, nodePoolID string, clusterKey string, clus
 		os.Exit(1)
 	}
 
-	autoscaling, replicas, minReplicas, maxReplicas := getNodePoolReplicas(cmd, r.Reporter, nodePoolID,
+	autoscaling, replicas, minReplicas, maxReplicas, scalingUpdated, minReplicaUpdated, maxReplicaUpdated := getNodePoolReplicas(cmd, r.Reporter, nodePoolID,
```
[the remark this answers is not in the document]
chenz4027: @robpblake Definitely looks better. One thing though - machinepool was also returning 6 values, which I got the idea from. Would it be okay if I also create a function for machinepool and another one for nodepool?

cmd/edit/machinepool/nodepool.go
chenz4027: @robpblake Of course, will work on it

cmd/edit/machinepool/nodepool.go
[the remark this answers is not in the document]
robpblake: Makes sense to me @chenz4027 👍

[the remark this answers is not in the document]
chenz4027: Sounds good @robpblake , I just updated the code for nodepool, will do the same for machinepool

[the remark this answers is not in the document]
chenz4027: Actually I got a question. the machinepool was checking if those replicas >=0 while nodepool checks >=1, although I thought they should be same though... @robpblake Is it okay to  just check >=0?
```
			if minReplicaUpdated && minReplicas >= 0 {
				asBuilder = asBuilder.MinReplicas(minReplicas)
			}
			if maxReplicaUpdated && maxReplicas >= 0 {
				asBuilder = asBuilder.MaxReplicas(maxReplicas)
			}
```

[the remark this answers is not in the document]
chenz4027: Removed my previous comment as mp and np builders are different

cmd/edit/machinepool/machinepool.go
```diff
@@ -18,11 +18,9 @@ import (
 var machinePoolKeyRE = regexp.MustCompile(`^[a-z]([-a-z0-9]*[a-z0-9])?$`)
 
 func editMachinePool(cmd *cobra.Command, machinePoolID string, clusterKey string, cluster *cmv1.Cluster,
-	r *rosa.Runtime) {
-	var err error
+	r *rosa.Runtime) error {
 	if !machinePoolKeyRE.MatchString(machinePoolID) {
-		r.Reporter.Errorf("Expected a valid identifier for the machine pool")
-		os.Exit(1)
+		return fmt.Errorf("expected a valid identifier for the machine pool")
```
robpblake, line 23: @chenz4027 Please correct me if I'm wrong, but I thought we did this before, only to get feedback from QE that the error messages should be capitalized? I would suggest we don't change the output of messages here and ignore the warning from Go.

cmd/edit/machinepool/cmd.go
```diff
@@ -166,8 +166,16 @@ func run(cmd *cobra.Command, argv []string) {
 	}
 
 	if cluster.Hypershift().Enabled() {
```
robpblake, line 168: @chenz4027 As a suggestion to avoid the repeated `if err` check:

```
var err error

 err = editNodePool(cmd, machinePoolID, clusterKey, cluster, r)
else
 err = editMachinePool(cmd, machinePoolID, clusterKey, cluster, r)
}

if err != nil {
 r.Reporter.Errorf("%s", err)
 os.Exit(1)
}

```

cmd/edit/machinepool/machinepool.go
chenz4027: @robpblake Sure, I'll move them back to capitalized string

cmd/edit/machinepool/cmd.go
chenz4027: Thanks @robpblake , will refactor it

chenz4027: Updated

robpblake (approved):
/lgtm

PR #1941, opened 2024-04-17
OCM-6542 | feat: add cmd to attach policy

marcolan018:
https://issues.redhat.com/browse/OCM-6542


ciaranRoche (commented):
Can you have a look at how we structured the `describe autoscaler` cmd, so we can align this new command to use the new runner method :pray:

cmd/attach/policy/cmd.go
```diff
@@ -0,0 +1,142 @@
+/*
+Copyright (c) 2020 Red Hat, Inc.
```
ciaranRoche, line 2: NP update the year :)
what year is it

marcolan018: fixed

cmd/attach/policy/cmd.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func AttachPolicyRunner() rosa.CommandRunner {
+	return func(_ context.Context, r *rosa.Runtime, cmd *cobra.Command, _ []string) error {
+		mode, err := interactive.GetMode()
+		if err != nil {
+			return err
+		}
+
+		defer r.Cleanup()
```
robpblake, line 85: @marcolan018 You don't need to call this, it's taken care of for you by the `DefaultRunner` 👍

cmd/attach/policy/cmd.go
```diff
@@ -0,0 +1,160 @@
...
+
+		policyArn := args.policyArn
+		_, err = r.AWSClient.IsPolicyExists(policyArn)
```
robpblake, line 88: @marcolan018 Can I make a suggestion of defining an `ArbitraryPolicyService` and having a few simple methods that hide all of this implementation and flow detail from the `Runner` implementation. 

So for example, how about a service interface defined like so:

```
type ArbitraryPolicyService interface {
   AutoAttachArbitraryPolicy(roleName, policyArn) error
   ManualAttachArbitraryPolicy(roleName, policyArn) (string, error)
}
```

You could then create an implementation of it that requires the `AWSClient` and `OCMClient` which you get an instance of via a creator function:

```
arbitraryPolicyService := policies.NewArbitraryPolicyService(r.AWSClient, r.OCMClient)
```

That would then shorten the logic in this `Runner` to _something_ like:

```
mode, err := interactive.GetMode()
if err != nil {
	return err
}

// Determine if interactive mode is needed
if !interactive.Enabled() && !cmd.Flags().Changed("mode") {
	interactive.Enable()
}

if interactive.Enabled() {
	mode, err = interactive.GetOptionMode(cmd, mode, "Attach policy mode")
	if err != nil {
		return fmt.Errorf("Expected a valid attach policy mode: %s", err)
	}
}

arbitraryPolicyService := policies.NewArbitraryPolicyService(r.AWSClient, r.OCMClient)

switch mode {
	case interactive.ModeAuto:
		return arbitraryPolicyService.AutoAttachArbitraryPolicy(roleName, policyArn)
	case interactive.ModeManual:
		cmd, err := arbitraryPolicyService.ManualAttachArbitraryPolicy(roleName, policyArn)
                if err != nil {
                   return err
                }
               r.Reporter.Infof("Run the following command to attach the policy: %s", cmd)
	default:
		return fmt.Errorf("Invalid mode. Allowed values are %s", interactive.Modes)
	}
return nil
```

The main benefit I see of this is a cleaner separation of concerns and ease of testing. The `Runner` is purely focused on doing CLI-stuff e.g getting input from the user and printing output. The actual logic of how this works in encapsulated in the service and if we ever need to change/extend it, the runner doesn't need to be altered.

cmd/attach/policy/cmd.go
```diff
@@ -0,0 +1,160 @@
...
+
+const (
+	use     = "policy"
+	short   = "Attach policy to a role"
+	long    = "Attach existing policy to role in the user’s AWS account"
+	example = `  # Attach policy <policy_arn> to role <role_name>
+  rosa attach policy --policy-arn=<policy_arn> --role-arn=<role_name>`
+)
+
+var args struct {
```
robpblake, line 40: @marcolan018 Could you take a look at this example of how we're trying to remove the need for global `args` struct. It would be good to try and follow the same pattern here: https://github.com/openshift/rosa/pull/1777#discussion_r1512694837

cmd/attach/policy/cmd.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/openshift/rosa/pkg/aws/tags"
+	"github.com/openshift/rosa/pkg/interactive"
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+const (
+	use     = "policy"
+	short   = "Attach policy to a role"
+	long    = "Attach existing policy to role in the user’s AWS account"
```
robpblake, line 35: @marcolan018 "Attach an existing AWS IAM Policy to an AWS IAM Role in the authenticated AWS Account" ?

pkg/test/helpers.go
```diff
@@ -368,3 +373,7 @@ func (t *TestingRuntime) SetCluster(clusterKey string, cluster *v1.Cluster) {
 	t.RosaRuntime.Cluster = cluster
 	t.RosaRuntime.ClusterKey = clusterKey
 }
+
+func GetStringAddress(str string) *string {
```
robpblake, line 377: @marcolan018 Is this used?

cmd/attach/policy/cmd.go
marcolan018: make the code refactoring

cmd/attach/policy/cmd.go
```diff
@@ -0,0 +1,160 @@
...
+
+	"github.com/openshift/rosa/pkg/aws/tags"
+	"github.com/openshift/rosa/pkg/interactive"
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+const (
+	use     = "policy"
+	short   = "Attach policy to a role"
```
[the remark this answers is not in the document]
marcolan018: fixed

cmd/attach/policy/cmd.go
marcolan018: fixed

pkg/test/helpers.go
marcolan018: yes, it's used in cmd_test.go

marcolan018:
@davidleerh @ciaranRoche @robpblake @gdbranco 
 Can you help to review this pr again? I completed changes based on your comments including:
- support multiple policy attach in one command
- check service quota for policy per role before attach
- code refactoring to adopt the latest defaultRunner pattern

cmd/attach/policy/attach_policy_service.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	attachedPolicies, err := c.GetAttachedPolicy(&roleName)
+	if err != nil {
+		return fmt.Errorf("Failed to get attached policies of role %s: %s", roleName, err)
+	}
+	if len(policyArns)+len(attachedPolicies) > int(*quota.Quota.Value) {
+		policySkipped := 0
+		for _, attachedPolicy := range attachedPolicies {
+			if slices.Contains(policyArns, attachedPolicy.PolicyArn) {
+				policySkipped++
```
davidleerh, line 142: Would it be better to output a message back to the user, the list of already attached policy arns?

cmd/attach/policy/attach_policy_service.go
```diff
@@ -0,0 +1,153 @@
...
+			}
+		}
+		if len(policyArns)+len(attachedPolicies)-policySkipped > int(*quota.Quota.Value) {
+			return fmt.Errorf("Exceeds the quota of policies per role: %d."+
```
davidleerh, line 146: I was wondering if a more concise error message here would be easier to read:

"Failed to attach policies due to quota limitations (total limit: x, expected: y)."
x = quota value
y = total of "len(policyArns)+len(attachedPolicies)-policySkipped)"

cmd/attach/policy/attach_policy_service.go
```diff
@@ -0,0 +1,153 @@
...
+	isRedHatManaged := false
+	for _, tag := range role.Tags {
+		if awsutil.ToString(tag.Key) == tags.RedHatManaged &&
+			awsutil.ToString(tag.Value) == "true" {
+			isRedHatManaged = true
+			break
+		}
+	}
+	if !isRedHatManaged {
+		return fmt.Errorf("Rosa only supports attaching policy to RedHat managed role")
```
davidleerh, line 74: Correcting message / for brevity: :) 
"Cannot attach policies to non-ROSA roles."

davidleerh:
aside from few nit picks with error messages, LGTM overall.
Manually tested the flow to see it working as expected.

cmd/attach/policy/attach_policy_service.go
marcolan018: @davidleerh 
seems user is not always interested in such info. The message including all attached policies will make the output redundancy.
Maybe we can just output such info only if the quota is exceeded.
wdyt?

cmd/attach/policy/attach_policy_service.go
marcolan018: message updated

cmd/attach/policy/attach_policy_service.go
marcolan018: message updated

cmd/attach/policy/attach_policy_service.go
davidleerh: IMHO getting the output to show already included policy arns when the quota is exceeded seems redundant as well. I am fine to leave it as it. Self resolving the comment. Thanks for looking.

pkg/policy/policy_service_test.go
```diff
@@ -0,0 +1,117 @@
...
+			awsClient.EXPECT().AttachRolePolicy(roleName, policyArn2).Return(nil)
+			output, err := policySvc.AutoAttachArbitraryPolicy(roleName, policyArns,
+				"sample-account-id", "sample-org-id")
+			fmt.Printf("output is %s", output)
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(output).To(Equal(fmt.Sprintf("Attached policy '%s' to role '%s'\n"+
+				"Attached policy '%s' to role '%s'\n",
+				policyArn1, roleName, policyArn2, roleName)))
+		})
+	})
```
robpblake, line 116: @marcolan018 Could you add a test for the manual attach method on the service as well please?

pkg/test/helpers.go
robpblake: @marcolan018 Can you replace this with `aws.String` instead. We already use that widely throughout the code-base. 

cheers!

Rob

marcolan018: updated, thanks for the remind

PR #1946, opened 2024-04-18
OCM-6391 | fix: Ability to disable region flag deprecation manually

hunterkepley:

When a nested cmd which deprecates region flag is ran (create cluster with describe cluster running from the create's `Run` func), it showed the deprecation msg of the nested cmd. This MR allows a bypass for the deprecation warning

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -3214,7 +3214,12 @@ func run(cmd *cobra.Command, _ []string) {
 				"for more information.")
 	}
 
+	var disableRegionDeprecation bool
+	clusterdescribe.Cmd.LocalFlags().BoolVar(&disableRegionDeprecation, arguments.DisableRegionDeprecationFlagName,
+		true, "Temporarily used for disabling a warning message ran from other commands (no reason to"+
+			" print for cluster describe called inside cluster create, but there is a use for a lone describe.")
 	clusterdescribe.Cmd.Run(clusterdescribe.Cmd, []string{cluster.ID()})
+	disableRegionDeprecation = false // No longer disable
 
 	if isSTS {
 		if mode != "" {
--- a/pkg/arguments/arguments.go
+++ b/pkg/arguments/arguments.go
@@ -37,6 +37,7 @@ const regionFlagName = "region"
 const regionDeprecationMessage = "Region flag will be removed from this command in future versions"
 
 var hasUnknownFlags bool
+var DisableRegionDeprecationFlagName = "disable-region-deprecation"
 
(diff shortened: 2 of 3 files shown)


pkg/arguments/arguments.go
```diff
@@ -335,9 +336,12 @@ func MarkRegionDeprecated(parentCmd *cobra.Command, childrenCmds []*cobra.Comman
 		cmd.Run = func(c *cobra.Command, args []string) {
 			outputFlag := cmd.Flag("output")
 			regionFlag := cmd.Flag("region")
+			disableDeprecationFlag := cmd.Flag(DisableRegionDeprecationFlagName)
 			hasChangedOutputFlag := outputFlag != nil && outputFlag.Value.String() != outputFlag.DefValue
```
davidleerh, line 340: can this be changed to 
`			hasChangedOutputFlag := cmd.Flags().Changed("output")
			hasChangedRegionFlag := cmd.Flags().Changed("region")
			isRegionDeprecationDisabled := cmd.Flags().Changed(DisableRegionDeprecationFlagName)`

instead?

hunterkepley:
@davidleerh
>  we should instead remove the overarching region command in favor of having the region flag specific on each cmd instead

I'm thinking the same thing, planned this ticket https://issues.redhat.com/browse/OCM-7161 which will result in having it specific to each command. I believe 'global' flags in general are not a good idea, as it results in the confusion which brought along the deprecation of this flag in 70ish commands

pkg/arguments/arguments.go
hunterkepley: Wasn't able to get this to work sadly, not sure why :/

hunterkepley: specifically with `region` - i think it has a default value which causes it to be changed even when it's not

pkg/arguments/arguments_test.go
```diff
@@ -47,6 +48,13 @@ var _ = Describe("Client", func() {
 				"",
 				"",
 			)
+			flagSet.BoolVarP(
```
gdbranco, line 51: Should this be hidden or an argv only option? This is just for calling when internal to another command, right?

cmd/create/cluster/cmd.go
```diff
@@ -3214,7 +3214,12 @@ func run(cmd *cobra.Command, _ []string) {
 				"for more information.")
 	}
 
+	var disableRegionDeprecation bool
+	clusterdescribe.Cmd.LocalFlags().BoolVar(&disableRegionDeprecation, arguments.DisableRegionDeprecationFlagName,
```
gdbranco, line 3218: hmm I see you are adding through this :thinking:

PR #1949, opened 2024-04-19
OCM-6419 | feat: fetch OIDC thumbprint from CS

thomasmckay:


cmd/create/cluster/cmd.go
```diff
@@ -3225,7 +3225,16 @@ func run(cmd *cobra.Command, _ []string) {
 			if !output.HasFlag() || r.Reporter.IsTerminal() {
 				r.Reporter.Infof("Preparing to create OIDC Provider.")
 			}
-			oidcprovider.Cmd.Run(oidcprovider.Cmd, []string{clusterName, mode, ""})
+
+			cmd := (*oidcprovider.CreateOidcProvider).NewCommand()
```
robpblake, line 3229: @thomasmckay I would simply call the `run` command on the `cmd` as before as that will take care of creating the `runner` and ensuring the args are bound the flags of the `cmd`.

thomasmckay: This is making it much more difficult to create tests for but I'll keep trying

PR #1950, opened 2024-04-19
OCM-1308 | fix: filter empty subnet id

chenz4027:
https://issues.redhat.com/browse/OCM-1308

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -2058,9 +2058,9 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	// Subnet IDs
-	subnetIDs := args.subnetIDs
+	subnetIDs := helper.FilterEmptyStrings(args.subnetIDs)
 	subnetsProvided := len(subnetIDs) > 0
-	r.Reporter.Debugf("Received the following subnetIDs: %v", args.subnetIDs)
+	r.Reporter.Debugf("Received the following subnetIDs: %v", subnetIDs)
 	// If the user has set the availability zones (allowed for non-BYOVPC clusters), don't prompt the BYOVPC message
 	if !useExistingVPC && !subnetsProvided && !isAvailabilityZonesSet && interactive.Enabled() {
 		existingVPCHelp := "To install into an existing VPC you need to ensure that your VPC is configured " +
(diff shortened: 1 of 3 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -4134,3 +4134,13 @@ func getMachinePoolRootDisk(r *rosa.Runtime, cmd *cobra.Command, version string,
 func clusterHasLongNameWithoutDomainPrefix(clusterName, domainPrefix string) bool {
 	return domainPrefix == "" && len(clusterName) > ocm.MaxClusterDomainPrefixLength
 }
+
+func delete_empty(s []string) []string {
```
mikeshng, line 4138:
```suggestion
func filterEmpty(s []string) []string {
```

Besides the name, I think it’s kind of weird to have a general helper function in here. Is there not a better place to put this?

mikeshng (approved):
/lgtm

pkg/helper/helpers.go
```diff
@@ -258,3 +258,13 @@ func ChunkSlice[T any](slice []T, chunkSize int) [][]T {
 func IsBYOVPC(cluster *cmv1.Cluster) bool {
 	return len(cluster.AWS().SubnetIDs()) > 0
 }
+
+func FilterEmpty(s []string) []string {
```
ciaranRoche, line 262: NP: The naming to be more specific :pray: something like `FilterEmptyStrings` or `RemoveEmptyStrings`

pkg/helper/helpers.go
```diff
@@ -258,3 +258,13 @@ func ChunkSlice[T any](slice []T, chunkSize int) [][]T {
...
+	var r []string
+	for _, str := range s {
+		if str != "" {
+			r = append(r, str)
+		}
+	}
+	return r
+}
```
ciaranRoche, line 270: NP: Can we avoid the single character naming in this func, I know its recommended, but something we tend to avoid across our codebase :pray:

chenz4027: Sure thanks @ciaranRoche , will make the change

chenz4027: Updated FYI @ciaranRoche

PR #1956, opened 2024-04-23
OCM-7307 | fix: remove older attributes for component routes

gdbranco:
[OCM-7307](https://issues.redhat.com//browse/OCM-7307)

Remove older attributes as they are not to be used. Favor `--component-routes` attribute


ciaranRoche (approved):
Given its not in use, we are safe to remove these attributes

PR #1960, opened 2024-04-24
OCM-7573 | fix: ensure version option args are initialised

ciaranRoche:
*WHAT*
When calling BindAndValidate the option args were not initialized, this changes ensures the args are initialized on `NewRosaVersionOptions`

This change also sets the checkVersion to debug in the case of any failures while trying to check the CLI version. Only when we successfully validate the version will we warn the user if its out of date.

JIRA - https://issues.redhat.com/browse/OCM-7573

### DIFF
--- a/cmd/version/options.go
+++ b/cmd/version/options.go
@@ -34,6 +34,7 @@ func NewRosaVersionOptions() (*RosaVersionOptions, error) {
 	return &RosaVersionOptions{
 		verifyRosa: verifyRosa,
 		reporter:   reporter.CreateReporter(),
+		args:       NewRosaVersionUserOptions(),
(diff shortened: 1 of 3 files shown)


cmd/version/options.go
```diff
@@ -34,6 +34,7 @@ func NewRosaVersionOptions() (*RosaVersionOptions, error) {
 	return &RosaVersionOptions{
```
robpblake, line 34: @ciaranRoche Could we add a quick unit test for the `NewRosaVersionOptions` function just so we don't regress this at any point in the future 👍

robpblake (approved):
/lgtm

PR #1961, opened 2024-04-24
OCM-6543 | feat: detach cmd to detach policy

marcolan018:
https://issues.redhat.com/browse/OCM-6543

### DIFF
--- a/pkg/policy/policy_service.go
+++ b/pkg/policy/policy_service.go
@@ -22,6 +22,7 @@ import (
 	"slices"
 
 	awsutil "github.com/aws/aws-sdk-go-v2/aws"
+	awserr "github.com/openshift-online/ocm-common/pkg/aws/errors"
 
 	"github.com/openshift/rosa/pkg/aws"
 	"github.com/openshift/rosa/pkg/aws/tags"
@@ -37,6 +38,9 @@ type PolicyService interface {
 	ValidateAttachOptions(roleName string, policyArns []string) error
 	AutoAttachArbitraryPolicy(roleName string, policyArns []string, accountID, orgID string) (string, error)
 	ManualAttachArbitraryPolicy(roleName string, policyArns []string, accountID, orgID string) string
+	ValidateDetachOptions(roleName string, policyArns []string) error
+	AutoDetachArbitraryPolicy(roleName string, policyArns []string, accountID, orgID string) (string, error)
+	ManualDetachArbitraryPolicy(roleName string, policyArns []string, accountID, orgID string) string
 }
 
 type policyService struct {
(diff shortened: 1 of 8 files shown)


pkg/policy/policy_service.go
```diff
@@ -127,6 +139,48 @@ func (p *policyService) ManualAttachArbitraryPolicy(roleName string, policyArns
...
 }
 
+func (p *policyService) AutoDetachArbitraryPolicy(roleName string, policyArns []string,
+	accountID, orgID string) (string, error) {
+	output := ""
+	for _, policyArn := range policyArns {
+		err := p.AWSClient.DetachRolePolicy(policyArn, roleName)
+		if err != nil {
+			if awserr.IsNoSuchEntityException(err) {
+				output = output + fmt.Sprintf("Currently policy %s is not attached to role %s\n",
```
davidleerh, line 149: I was thinking if it would sound a bit better with message below:
The policy '%s' is currently not attached to role '%s'\n

pkg/policy/policy_service.go
```diff
@@ -127,6 +139,48 @@ func (p *policyService) ManualAttachArbitraryPolicy(roleName string, policyArns
...
+					policyArn, roleName)
+			} else {
+				return output, fmt.Errorf("Failed to detach policy %s from role %s: %s",
```
davidleerh, line 152: "Failed to detach policy '%s' from role '%s': %s",

pkg/policy/policy_service.go
```diff
@@ -56,6 +60,18 @@ func NewPolicyService(OCMClient *ocm.Client, AWSClient aws.Client) PolicyService
...
+		return err
+	}
+	err = validatePolicyQuota(p.AWSClient, roleName, policyArns)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (p *policyService) ValidateDetachOptions(roleName string, policyArns []string) error {
```
davidleerh, line 74: Should this method be renamed so that it is more generic?
something like ValidateRoleAndPolicies?

davidleerh:
added minor np comments but overall lgtm.
Was able to manually validate the command works as expected.

pkg/policy/policy_service.go
marcolan018: updated

pkg/policy/policy_service.go
marcolan018: updated

pkg/policy/policy_service.go
marcolan018: change to a  private func validateRoleAndPolicies which used by ValidateAttachOptions and ValidateDetachOptions

marcolan018:
@davidleerh thanks for the review, can you help to re-review the latest change? Thanks

PR #1964, opened 2024-04-24
OCM-7256 | feat: Move non-cmd funcs to pkg, split, test

hunterkepley:
[OCM-7256](https://issues.redhat.com/browse/OCM-7256) Has the following definition of done:

    Functions have been moved to new machine pool package
    Functions test coverage is at least 70%

The 70% test coverage may not be there yet, as one function requires a runner to be tested. This will be done ALONG WITH the rework in the next ticket: https://issues.redhat.com/browse/OCM-7257

This MR splits the list function up where possible, along with combining machine and node pool funcs into a single function as they have quite a bit of shared code and there's no reason to separate.


pkg/machinepool/machinepool.go
```diff
@@ -29,6 +33,49 @@ func NewMachinePoolService() MachinePoolService {
 	return &machinePool{}
 }
 
+// ListMachinePools lists all machinepools (or, nodepools if hypershift) in a cluster
+func (m *machinePool) ListMachinePools(r *rosa.Runtime, clusterKey string, cluster *cmv1.Cluster, isHypershift bool) {
```
chenz4027, line 37: Is it possible to add some test coverage for ListMachinePools so that the first half of the code is also tested?

hunterkepley: Next ticket in this task will be to move the command to a new runner, which will allow easier testing of functions like this which require a runtime -- will be attempting to get 100% coverage in that MR

WDYT?

chenz4027: @hunterkepley Sounds good, feel free to resolve the thread - I don't have that permission

ciaranRoche, line 37: the `machinePool` target has a mix of pointer, can you align them :pray:

pkg/machinepool/machinepool.go
```diff
@@ -18,6 +21,7 @@ var notFoundMessage string = "Machine pool '%s' not found"
 type MachinePoolService interface {
 	DescribeMachinePool(r *rosa.Runtime, cluster *cmv1.Cluster, clusterKey string, isHypershift bool,
 		machinePoolId string) error
+	ListMachinePools(r *rosa.Runtime, clusterKey string, cluster *cmv1.Cluster, isHypershift bool)
```
ciaranRoche, line 24: Can we update this func to return an error and we leave it up to the cmd to print and exit. Will make this func a lot easier to test too

pkg/machinepool/machinepool.go
```diff
@@ -29,6 +33,49 @@ func NewMachinePoolService() MachinePoolService {
...
+		if !isHypershift {
+			err = output.Print(machinePools)
+		} else {
+			err = output.Print(nodePools)
+		}
+		if err != nil {
+			r.Reporter.Errorf("%s", err)
+			os.Exit(1)
+		}
+		os.Exit(0)
```
ciaranRoche, line 63: Based on the comment above, best we return an error or nil 

```suggestion
		if err != nil {
			return err
		}
		return nil
```

cmd/list/machinepool/cmd.go
```diff
@@ -56,9 +57,6 @@ func run(_ *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
-	if cluster.Hypershift().Enabled() {
-		listNodePools(r, clusterKey, cluster)
-	} else {
-		listMachinePools(r, clusterKey, cluster)
-	}
+	service := machinepool.NewMachinePoolService()
+	service.ListMachinePools(r, clusterKey, cluster, cluster.Hypershift().Enabled())
```
ciaranRoche, line 61: Do we need to pass in `cluster.Hypershift().Enabled()` since we pass in the cluster object we can ref that in the func

pkg/machinepool/machinepool.go
```diff
@@ -29,6 +33,49 @@ func NewMachinePoolService() MachinePoolService {
...
 }
 
+// ListMachinePools lists all machinepools (or, nodepools if hypershift) in a cluster
+func (m *machinePool) ListMachinePools(r *rosa.Runtime, clusterKey string, cluster *cmv1.Cluster, isHypershift bool) {
+	// Load any existing machine pools for this cluster
+	r.Reporter.Debugf("Loading machine pools for cluster '%s'", clusterKey)
+	var err error
+	var machinePools []*cmv1.MachinePool
+	var nodePools []*cmv1.NodePool
+	if !isHypershift {
```
ciaranRoche, line 43: NP : we could remove a lot of if/else if we break up this func. 

something like 
```suggestion
	var err error
	var output string
	if cluster.Hypershift().Enabled() {
		output, err = m.getNodePoolsString(.......)
	}
	output, err = m.getMachinePoolsString(.......)
	if err != nil {
		return err
	}

	// Create the writer that will be used to print the tabulated results:
	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
	fmt.Fprint(writer, output)
	writer.Flush()
	return nil
```

pkg/machinepool/machinepool.go
```diff
@@ -29,10 +32,48 @@ func NewMachinePoolService() MachinePoolService {
...
+	}
+
+	if output.HasFlag() {
+		if !isHypershift {
+			err = output.Print(machinePools)
+		} else {
+			err = output.Print(nodePools)
+		}
+		return err
+	}
```
ciaranRoche, line 58: Sorry one more NP, we can flatten this out, and remove that else statement
```suggestion
	if output.HasFlag() {
		if !isHypershift {
			return output.Print(machinePools)
		}
		return output.Print(nodePools)
	}
```

pkg/machinepool/machinepool.go
```diff
@@ -29,10 +32,46 @@ func NewMachinePoolService() MachinePoolService {
...
+	var err error
+	var machinePools []*cmv1.MachinePool
+	var nodePools []*cmv1.NodePool
+	machinePools, err = r.OCMClient.GetMachinePools(cluster.ID())
+	if isHypershift {
+		nodePools, err = r.OCMClient.GetNodePools(cluster.ID())
+	}
+	if err != nil {
+		return err
+	}
```
ziccardi, line 49: if `isHypershift` is true, the error returned by GetMachinePools is ignored.
```suggestion
	machinePools, err = r.OCMClient.GetMachinePools(cluster.ID())
	if err != nil {
		return err
	}
	if isHypershift {
		nodePools, err = r.OCMClient.GetNodePools(cluster.ID())
	}
	if err != nil {
		return err
	}
```

pkg/machinepool/machinepool.go
```diff
@@ -29,10 +32,46 @@ func NewMachinePoolService() MachinePoolService {
...
+			return output.Print(nodePools)
+		}
+		return output.Print(machinePools)
+	}
+
+	// Create the writer that will be used to print the tabulated results:
+	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
+
+	var finalStringToOutput string
+	finalStringToOutput = getMachinePoolsString(machinePools)
```
ziccardi, line 62:
```suggestion
	finalStringToOutput := getMachinePoolsString(machinePools)
```

pkg/machinepool/machinepool_test.go
```diff
@@ -63,4 +94,33 @@ var _ = Describe("Machinepool and nodepool", func() {
...
+	Context("MachinePools", func() {
+		It("Test printMachinePools", func() {
+			clusterBuilder := cmv1.NewCluster().ID("test").State(cmv1.ClusterStateReady).
+				MachinePools(cmv1.NewMachinePoolList().
+					Items(cmv1.NewMachinePool().ID("np").Replicas(8).Subnets("sn1", "sn2").
+						InstanceType("test instance type").Taints(cmv1.NewTaint().Value("test").
+						Key("taint"))))
+			cluster, err := clusterBuilder.Build()
+			Expect(err).ToNot(HaveOccurred())
+			Expect(err).ToNot(HaveOccurred())
```
ziccardi, line 106:
```suggestion
			Expect(err).ToNot(HaveOccurred())
```

pkg/machinepool/machinepool_test.go
```diff
@@ -23,6 +26,34 @@ var _ = Describe("Machinepool and nodepool", func() {
...
 				NextRun(date)
 		})
+		It("Test printNodePools", func() {
+			clusterBuilder := cmv1.NewCluster().ID("test").State(cmv1.ClusterStateReady).
+				Hypershift(cmv1.NewHypershift().Enabled(true)).NodePools(cmv1.NewNodePoolList().
+				Items(cmv1.NewNodePool().ID("np").Replicas(8).AvailabilityZone("az").
+					Subnet("sn").Version(cmv1.NewVersion().ID("1")).AutoRepair(false)))
+			cluster, err := clusterBuilder.Build()
+			Expect(err).ToNot(HaveOccurred())
+			Expect(err).ToNot(HaveOccurred())
```
ziccardi, line 36:
```suggestion
			Expect(err).ToNot(HaveOccurred())
```

pkg/machinepool/machinepool.go
hunterkepley: Will make this change - the intention was that the error does not matter if hypershift is true, as we don't care about machinepools, so the error should be ignored. But, it isn't good, as it's an extra API call, so will change it

ciaranRoche (approved):
/lgtm

PR #1966, opened 2024-04-25
OCM-7415 | feat: Ensure account roles have expected attached policies

marcolan018:
https://issues.redhat.com/browse/OCM-7415
https://issues.redhat.com/browse/OCM-1752

### DIFF
--- a/cmd/upgrade/roles/cmd.go
+++ b/cmd/upgrade/roles/cmd.go
@@ -34,6 +34,7 @@ import (
 	awscbRoles "github.com/openshift/rosa/pkg/aws/commandbuilder/helper/roles"
 	"github.com/openshift/rosa/pkg/aws/tags"
 	"github.com/openshift/rosa/pkg/helper"
+	"github.com/openshift/rosa/pkg/helper/rolepolicybindings"
 	"github.com/openshift/rosa/pkg/helper/roles"
 	"github.com/openshift/rosa/pkg/interactive"
 	"github.com/openshift/rosa/pkg/interactive/confirm"
@@ -264,6 +265,18 @@ func run(cmd *cobra.Command, argv []string) {
 		spin.Stop()
 	}
 
+	rolePolicyBindings, err := ocmClient.ListRolePolicyBindings(cluster.ID(), true)
+	if err != nil {
+		reporter.Errorf("Failed to get rolePolicyBinding: %s", err)
+		os.Exit(1)
+	}
+	err = rolepolicybindings.CheckRolePolicyBindingStatus(rolePolicyBindings)
+	if err != nil {
+		reporter.Errorf("Error in rolePolicyBinding: %s", err)
+		os.Exit(1)
+	}
+	rolePolicyDetails := rolepolicybindings.TransformToRolePolicyDetails(rolePolicyBindings)
+
 	if !isUpgradeNeedForAccountRolePolicies {
 		reporter.Infof("Account roles/policies for cluster '%s' are already up-to-date.", r.ClusterKey)
 	} else {
@@ -287,6 +300,7 @@ func run(cmd *cobra.Command, argv []string) {
 					accountRolePolicies,
 					policyVersion,
 					isPolicyVersionChosen,
+					rolePolicyDetails,
 				)
 				if err != nil {
 					LogError(roles.RosaUpgradeAccRolesModeAuto, ocmClient, policyVersion, err, reporter)
@@ -323,6 +337,7 @@ func run(cmd *cobra.Command, argv []string) {
 				isUpgradeNeedForAccountRolePolicies,
 				awsClient,
 				policyVersion,
+				rolePolicyDetails,
 			)
 			if err != nil {
 				reporter.Errorf("%s", err)
@@ -410,6 +425,7 @@ func run(cmd *cobra.Command, argv []string) {
 			credRequests,
 			cluster,
 			operatorRolePolicyPrefix,
+			rolePolicyDetails,
 		)
 		if err != nil {
 			r.Reporter.Errorf("%s", err)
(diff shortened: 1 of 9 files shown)


pkg/aws/policies.go
```diff
@@ -1955,6 +1955,38 @@ func (c *awsClient) ListAttachedRolePolicies(roleName string) ([]string, error)
...
+	}
+	policies, err := getAttachedPolicies(c.iamClient, roleName, tagfilter)
+	if err != nil {
+		return "", err
+	}
+	if len(policies) == 0 {
+		return "", nil
+	}
+	if len(policies) > 1 {
+		return "", fmt.Errorf("There are more than one RedHat managed policy attached to the role %s", roleName)
```
davidleerh, line 1985: I think message might be better to indicate that there are more than one "Red Hat managed operator role policy?"

pkg/aws/policies.go
```diff
@@ -1955,6 +1955,38 @@ func (c *awsClient) ListAttachedRolePolicies(roleName string) ([]string, error)
...
+func (c *awsClient) GetAccountRoleDefaultPolicy(roleName string, prefix string) (string, error) {
+	policies, err := getAttachedPolicies(c.iamClient, roleName, getAcctRolePolicyTags(prefix))
+	if err != nil {
+		return "", err
+	}
+	if len(policies) == 0 {
+		return "", nil
+	}
+	if len(policies) > 1 {
+		return "", fmt.Errorf("There are more than one RedHat managed policy attached to the role %s", roleName)
```
davidleerh, line 1967: I think message might be better to indicate that there are more than one "Red Hat managed account role policy?"

davidleerh:
small np in error messages but overall looks good to me.
Thank you for putting great effort and thought into this; considering possible edge case scenarios!

pkg/helper/rolepolicybindings/helpers.go
```diff
@@ -0,0 +1,101 @@
...
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+
+	"github.com/openshift/rosa/pkg/aws"
+)
+
+const (
+	RolePolicyBindingFailedStatus = "failed"
+)
+
+func ValidateRolePolicyBindings(bindings *cmv1.RolePolicyBindingList) error {
```
ciaranRoche, line 34: NP : update the naming of this func to better describe what we are doing here, we are only checking the bindings   status values, we are not doing any other validations IMO

pkg/helper/rolepolicybindings/helpers.go
```diff
@@ -0,0 +1,101 @@
...
+		if binding.Policies() != nil {
+			for _, policy := range binding.Policies() {
+				roleBindings = append(roleBindings, policy.Arn())
+			}
+		}
+		actualBindings[binding.Name()] = roleBindings
+	}
+	missingBindings := map[string][]string{}
+	for _, binding := range desired.Slice() {
+		for _, policy := range binding.Policies() {
```
ciaranRoche, line 58: Can binding be nil?

ciaranRoche (commented):
Small Nitpicks but looks good otherwise

pkg/helper/rolepolicybindings/helpers.go
davidleerh: I believe as long as clusters have roles associated (which is required during cluster create), desired will never be nil.

cmd/upgrade/roles/cmd.go
```diff
@@ -616,11 +612,6 @@ func upgradeAccountRolePoliciesFromCluster(
 		if err != nil {
 			return err
 		}
-		//Delete if present else continue
-		err = awsClient.DeleteInlineRolePolicies(roleName)
```
davidleerh, line 620: Thank you for taking care of this!

pkg/helper/rolepolicybindings/helpers.go
gdbranco: If they delete some bindings from AWS, but not the installer we can still sync right? It might be possible to be nil I think, it is quite cumbersome for a customer to reach such state :thinking:

cmd/upgrade/roles/cmd.go
```diff
@@ -458,6 +474,21 @@ func run(cmd *cobra.Command, argv []string) {
 			)
 		}
 	}
+
+	if isUpgradeNeedForAccountRolePolicies && mode == interactive.ModeAuto || isOperatorPolicyUpgradeNeeded {
+		newRolePolicyBindings, err := ocmClient.ListRolePolicyBindings(cluster.ID(), true)
+		if err != nil {
+			reporter.Warnf("Failed to get rolePolicyBindings after upgrade," +
+				" please check the attached policies on upgraded roles")
```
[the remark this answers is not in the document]
gdbranco: I think we need to be clearer on this, what is the desired state? Is it related only to the extra policies?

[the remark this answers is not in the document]
davidleerh: Since it was unable to retrieve the list of role policy bindings, rosa cannot confirm on behalf of the customer if any of the initial arbitrary policies were dropped. 

The desired state in this case would be what the customer intends to have on each role.

Would a better message be,
"Please ensure that the list of policies attached to roles meet your requirements?"

pkg/helper/rolepolicybindings/helpers.go
davidleerh: the "RolePolicyBindingList" used here is a list of roles which also has a list of policies (if retrievable via AWS).
Since all clusters have roles associated, the list of roles or bindings will always have a value.

To clarify, a binding here is an object that has an array of policies within it. Each object associates to each role of the cluster. A role always exists therefore, a binding object will always exist.

If the permissions of the installer role was inaccessible / not assumable for some reason, the role policy binding list being passed here would still be pointing to a list of bindings (not nil).

Although I personally think the impact is minimal, I'd say it is always safer to have a condition to always check for nils; if binding is nil.
if binding != nil proceed with the loop in 58.
Also in line 49.

pkg/aws/policies.go
marcolan018: fixed

pkg/aws/policies.go
marcolan018: fixed

pkg/helper/rolepolicybindings/helpers.go
marcolan018: change the function name to CheckRolePolicyBindingStatus()

pkg/helper/rolepolicybindings/helpers.go
marcolan018: added the check to avoid possible edge case that binding is nil

cmd/upgrade/roles/cmd.go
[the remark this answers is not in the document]
marcolan018: I changed to "please ensure the required policies attached on upgraded roles", is this ok? @davidleerh @gdbranco

ciaranRoche (approved):
LGTM nice work

cmd/upgrade/roles/cmd.go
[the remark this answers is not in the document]
davidleerh: "Please ensure that the required policies are attached to the upgraded roles."
Thank you :)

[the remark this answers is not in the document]
marcolan018: updated the messages

PR #1967, opened 2024-04-25
OCM-7570 | fix: Small bugfix for rosa describe machinepool

hunterkepley:
OCM-7570 describes a bug for `rosa describe machinepool`. This MR fixes that bug


cmd/describe/machinepool/cmd.go
```diff
@@ -71,6 +71,7 @@ func DescribeMachinePoolRunner(userOptions DescribeMachinepoolUserOptions) rosa.
 			userOptions.machinepool = argv[0]
 		} else {
 			err := cmd.ParseFlags(argv)
+			userOptions.machinepool = cmd.Flag("machinepool").Value.String()
```
robpblake, line 74: @hunterkepley Is `userOptions.machinepool` not set at the entry to this function if you pass the `--machinepool` argument to this command? It should be given the way you initialise the `FlagSet` in the `NewDescribeMachinePool` function

cmd/describe/machinepool/cmd.go
```diff
@@ -71,6 +71,7 @@ func DescribeMachinePoolRunner(userOptions DescribeMachinepoolUserOptions) rosa.
 			userOptions.machinepool = argv[0]
```
robpblake, line 71: @hunterkepley I think you need to replace the `cobra.NoArgs` on the `cmd` definition with `cobra.MaximumNArgs(1)`. That's what is was prior to your refactor and why passing the `workers` name as an arg is being rejected.

cmd/describe/machinepool/cmd.go
```diff
@@ -71,6 +71,7 @@ func DescribeMachinePoolRunner(userOptions DescribeMachinepoolUserOptions) rosa.
...
 		} else {
 			err := cmd.ParseFlags(argv)
```
robpblake, line 73: @hunterkepley I think we can drop this additional `cmd.ParseFlags` call as it should already be done by this point.

cmd/describe/machinepool/cmd.go
hunterkepley: @robpblake fixed

cmd/describe/machinepool/cmd.go
hunterkepley: @robpblake was able to drop that, but not the line mentioned in the first comment you made

hunterkepley: doesn't pass tests after removing

PR #1977, opened 2024-04-30
OCM-7794 | fix: handle nil ptr when rosa version is not retrievable from mirror

davidleerh:


ciaranRoche:
Should be handled via - https://github.com/openshift/rosa/pull/1960

PR #1990, opened 2024-05-01
OCM-7673 | fix: Provide the default value to the external auth

chenz4027:
https://issues.redhat.com/browse/OCM-7673

### DIFF
--- a/pkg/externalauthprovider/flags.go
+++ b/pkg/externalauthprovider/flags.go
@@ -25,6 +25,8 @@ const (
 	claimValidationRuleFlag       = "claim-validation-rule"
 	consoleClientIdFlag           = "console-client-id"
 	consoleClientSecretFlag       = "console-client-secret"
+	defaultClaimMappingUsername   = "email"
+	defaultClaimMappingGroups     = "groups"
 )
 
 type ExternalAuthServiceImpl struct {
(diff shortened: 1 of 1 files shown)


pkg/externalauthprovider/flags.go
```diff
@@ -256,7 +256,7 @@ func GetExternalAuthOptions(
 	if interactive.Enabled() && !cmd.Changed(claimMappingUsernameClaimFlag) {
 		result.claimMappingUsernameClaim, err = interactive.GetString(interactive.Input{
 			Question: "Claim mapping username",
-			Default:  result.claimMappingUsernameClaim,
+			Default:  "email",
```
davidleerh, line 259: NP: Would it be possible to use a const here instead? 
const name like "defaultClaimMappingUsername"?

pkg/externalauthprovider/flags.go
```diff
@@ -268,7 +268,7 @@ func GetExternalAuthOptions(
 	if interactive.Enabled() && !cmd.Changed(claimMappingGroupsClaimFlag) {
 		result.claimMappingGroupsClaim, err = interactive.GetString(interactive.Input{
 			Question: "Claim mapping groups",
-			Default:  result.claimMappingGroupsClaim,
+			Default:  "groups",
```
davidleerh, line 271: NP: Would it be possible to use a const here instead? 
const name like "defaultClaimMappingGroups"?

PR #1992, opened 2024-05-02
OCM-6391 | fix: Add region deprecation disablement to sub-commands

hunterkepley:

is a ticket where we temporarily deprecate the region flag before removing it from commands which do not use it. This is a bug fix for the create cluster command, which housed 4 commands that have this deprecation.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -3214,30 +3214,19 @@ func run(cmd *cobra.Command, _ []string) {
 				"for more information.")
 	}
 
-	disableUsage := "Temporarily used for disabling a warning message ran from other commands (no reason to" +
-		" print for cluster describe called inside cluster create, but there is a use for a lone describe."
-	var disableRegionDeprecation bool
-	clusterdescribe.Cmd.LocalFlags().BoolVar(&disableRegionDeprecation, arguments.DisableRegionDeprecationFlagName,
-		true, disableUsage)
+	arguments.DisableRegionDeprecationWarning = true // disable region deprecation warning
 	clusterdescribe.Cmd.Run(clusterdescribe.Cmd, []string{cluster.ID()})
-	disableRegionDeprecation = false // No longer disable
 
 	if isSTS {
 		if mode != "" {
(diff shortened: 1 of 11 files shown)


pkg/arguments/arguments.go
```diff
@@ -332,15 +335,24 @@ func MarkRegionDeprecated(parentCmd *cobra.Command, childrenCmds []*cobra.Comman
 			deprecateRegion(parentCmd)
 			command.Parent().HelpFunc()(command, strings)
 		})
+
+		disableUsage := "Temporarily used for disabling a warning message ran from other commands (no reason to" +
+			" print for cluster describe called inside cluster create, but there is a use for a lone describe."
+		if cmd.LocalFlags().Lookup(DisableRegionDeprecationFlagName) == nil {
+			cmd.LocalFlags().BoolVar(&DisableRegionDeprecationWarning, DisableRegionDeprecationFlagName,
+				true, disableUsage)
```
gdbranco, line 343: If the default is true doesn't it always disable it though? Considering you are adding the flag to all commands now :thinking:

hunterkepley: i dont believe so, it's based on the pointed-to-bool right?

hunterkepley: oh youre right let me change that just tested

hunterkepley: fixed

cmd/create/cluster/cmd.go
```diff
@@ -3214,30 +3214,19 @@ func run(cmd *cobra.Command, _ []string) {
 				"for more information.")
 	}
 
-	disableUsage := "Temporarily used for disabling a warning message ran from other commands (no reason to" +
-		" print for cluster describe called inside cluster create, but there is a use for a lone describe."
-	var disableRegionDeprecation bool
-	clusterdescribe.Cmd.LocalFlags().BoolVar(&disableRegionDeprecation, arguments.DisableRegionDeprecationFlagName,
-		true, disableUsage)
+	arguments.DisableRegionDeprecationWarning = true // disable region deprecation warning
```
gdbranco, line 3217: Did we check for other commands that have internal calls like this?

PR #2001, opened 2024-05-07
OCM-7881 | fix: skips policy compatibility check when version supplied is empty

gdbranco:
OCM-7881


pkg/aws/policies.go
```diff
@@ -399,6 +399,10 @@ func (c *awsClient) createPolicy(policyArn string, document string, tagList map[
 }
 
 func (c *awsClient) IsPolicyCompatible(policyArn string, version string) (bool, error) {
```
ciaranRoche, line 401: Can we add a basic test for this func before we merge.

gdbranco: Yeah, just to that func we can do. The flow itself would require a lot of additions

PR #2011, opened 2024-05-08
OSD-22376 | refactor: updates fedramp client id configuration

tonytheleg:
Client ID used for keycloak auth is the same for all envs, no need to define it multiple times depending on env

Updates logic to set the appropriate client id when FedRAMP is enabled  and the `--admin` flag is not passed

Changes are mostly for code cleanliness


ciaranRoche (approved):
/lgtm

PR #2019, opened 2024-05-09
OCM-7961 | chore: update the fips flag description

scottd018:
This was added as a way to let users know before attempting a deployment, that FIPS is not available for hosted control plane at this time.  This can be removed at a time when hosted control plane is FIPS validated.  See https://docs.aws.amazon.com/rosa/latest/userguide/rosa-deployment-options.html#rosa-deployment-model-differences for more details.


scottd018:
@ciaranRoche Much appreciated and thanks for staging that up for me!  It should be all fixed now.  Thanks!

PR #2027, opened 2024-05-10
OCM-7263 | feat: Add the ability to list KubeletConfigs for a cluster

robpblake:
This PR adds support for the `rosa list kubeletconfigs` command. This supports KubeletConfigs on ROSA HCP clusters where it is possible to create one or more KubeletConfigs.

This command works for both HCP and Classic clusters. For Classic, there will only ever be one KubeletConfig.

### DIFF
--- a/pkg/ocm/kubeletconfig_test.go
+++ b/pkg/ocm/kubeletconfig_test.go
@@ -2,6 +2,8 @@ package ocm
 
 import (
 	"bytes"
+	"context"
+	"fmt"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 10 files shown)


pkg/ocm/kubeletconfig_test.go
```diff
@@ -177,11 +180,47 @@ var _ = Describe("KubeletConfig", Ordered, func() {
 		Expect(err).To(HaveOccurred())
 	})
 
+	Context("List KubeletConfigs", func() {
+
+		It("Lists all kubeletconfigs", func() {
```
den-rgb, line 185: We should probably have it as `KubeletConfig` just to keep it the same everywhere

robpblake: @den-rgb As per our Slack conversation, I'll fix this in the next PR which is coming shortly rather than going through the whole build for this typo 👍

PR #2028, opened 2024-05-10
OCM-7937 | fix: allow 0 min replicas for classic cluster autoscaling mp

davidleerh:
Able to set autoscaling min replicas to 0 for machine pools (classic) after fix:
image


davidleerh:
FYI the validation to surface errors for max replicas < min replicas exists in clusters service.

PR #2036, opened 2024-05-13
OCM-7265 | feat: Added the ability to edit KubeletConfigs for HCP clusters

robpblake:
This PR adds the ability to edit KubeletConfig for HCP clusters.

### DIFF
--- a/pkg/ocm/kubeletconfig.go
+++ b/pkg/ocm/kubeletconfig.go
@@ -2,6 +2,7 @@ package ocm
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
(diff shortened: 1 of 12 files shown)


pkg/ocm/kubeletconfig.go
```diff
@@ -77,6 +79,33 @@ func (c *Client) UpdateKubeletConfig(clusterID string, args KubeletConfigArgs) (
 	return response.Body(), nil
 }
 
+func (c *Client) FindKubeletConfigByName(
+	ctx context.Context, clusterId string, name string) (*cmv1.KubeletConfig, bool, error) {
+
+	if name == "" {
+		return nil, false, fmt.Errorf("'name' for the KubeletConfig cannot be empty")
```
ciaranRoche, line 86: Is this error enough :thinking: should we also check in the runner prior to finding to ensure a name is passed if a cluster is HCP 
WDYT?

robpblake: Done.

ciaranRoche (approved):
/lgtm

PR #2038, opened 2024-05-13
OCM-7266 | feat: Added support for delete of KubeletConfig for HCP clusters

robpblake:
This PR updates `rosa delete kubeletconfig` with support for HCP clusters.

### DIFF
--- a/cmd/dlt/kubeletconfig/cmd.go
+++ b/cmd/dlt/kubeletconfig/cmd.go
@@ -17,55 +17,80 @@ limitations under the License.
 package kubeletconfig
 
 import (
+	"context"
 	"fmt"
-	"os"
 
 	"github.com/spf13/cobra"
 
 	"github.com/openshift/rosa/pkg/interactive/confirm"
+	. "github.com/openshift/rosa/pkg/kubeletconfig"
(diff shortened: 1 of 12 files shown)


cmd/dlt/kubeletconfig/cmd.go
```diff
@@ -17,55 +17,80 @@ limitations under the License.
...
+		}
+
+		if !cluster.Hypershift().Enabled() {
+			if !PromptUserToAcceptWorkerNodeReboot(OperationDelete, r) {
+				return nil
+			}
+		}
+
+		if cluster.Hypershift().Enabled() {
+			options.Name, err = PromptForName(options.Name)
```
ciaranRoche, line 78: NP : do we only want to prompt for name if name is not passed and or interactive mode is enabled? Sorry I missed this in the other PR, we prompt only if interactive mode is enabled.

ciaranRoche (commented):
Just one question inline, not really a blocker, happy to merge as is and follow up later if we want

cmd/dlt/kubeletconfig/cmd.go
ciaranRoche: following up in another PR

PR #2039, opened 2024-05-13
OCM-7258 | feat: Move delete machinepool/nodepool funcs to pkg

hunterkepley:
[OCM-7258](https://issues.redhat.com//browse/OCM-7258)
---
* Move `cmd` funcs for `delete machinepool` (includes `nodepools`) to `pkg/machinepool` + add to the machine pool service data structure

* Added test for regex

* Will make e2e tests in the 2nd MR (followup to this) where I will be trying to reach 80%+ coverage as I have done for the other `machinepool` commands

* This also combines the machinepool/nodepool func calls to allow a single call for classic/HCP - automatically determined based on the cluster


cmd/dlt/machinepool/cmd.go
```diff
@@ -53,13 +54,16 @@ func run(_ *cobra.Command, argv []string) {
...
+	machinePoolId := argv[0]
 	clusterKey := r.GetClusterKey()
 	cluster := r.FetchCluster()
 
-	if cluster.Hypershift().Enabled() {
-		deleteNodePool(r, machinePoolID, clusterKey, cluster)
-	} else {
-		deleteMachinePool(r, machinePoolID, clusterKey, cluster)
+	service := machinepool.NewMachinePoolService()
+	if !machinepool.MachinePoolKeyRE.MatchString(machinePoolId) {
```
marcolan018, line 62: it‘s better to move the machineId validation right after we got its' value "machinePoolId := argv[0]"

hunterkepley:
Cannot get coverage up until i rewrite the command with the new runner, and add e2e tests. There is an import cycle if I use the test runtime in this package

willkutler (approved):

PR #2040, opened 2024-05-13
OCM-7990 | fix: allow min_replicas 0 with edit machinepools

davidleerh:
FIXED:

A. min and max replicas set at 3
image

B. edit min replicas only to 0
image

C. result
image

### DIFF
--- a/cmd/edit/machinepool/machinepool_test.go
+++ b/cmd/edit/machinepool/machinepool_test.go
@@ -36,16 +36,6 @@ var _ = Describe("Machinepool", func() {
 			asBuilder := cmv1.NewMachinePoolAutoscaling().MaxReplicas(2).MinReplicas(0)
 			Expect(builder).To(Equal(asBuilder))
 		})
-
-		It("editMachinePoolAutoscaling should allow 0 min and 0 max replicas", func() {
-			machinePool, err := cmv1.NewMachinePool().
-				Autoscaling(cmv1.NewMachinePoolAutoscaling().MaxReplicas(2).MinReplicas(1)).
-				Build()
-			Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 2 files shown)


cmd/edit/machinepool/machinepool_test.go
```diff
@@ -36,16 +36,6 @@ var _ = Describe("Machinepool", func() {
 			asBuilder := cmv1.NewMachinePoolAutoscaling().MaxReplicas(2).MinReplicas(0)
 			Expect(builder).To(Equal(asBuilder))
 		})
-
-		It("editMachinePoolAutoscaling should allow 0 min and 0 max replicas", func() {
```
davidleerh, line 40: There is a validation done in CS that prevents users assigning 0 for max mp for autoscaling.

marcolan018: should rosa reports that error before send request to CS side? And then we still need a test case here to verify that

chenz4027: Logic looks good, but I feel you might fail test coverage though as lines were deleted.

davidleerh: Thanks for your review @marcolan018  @chenz4027 
In addition to those, the validations on the interactive mode are missing. I will create a separate card so that we can have a robust number of test cases to make sure we surface input validation errors from the CLI before reaching CS.

Trying to have this land in RC4 as quickly as possible.
If no concerns, would you kindly approve?

marcolan018: @davidleerh I am ok with this

davidleerh: Thank you!
Created OCM-8010 for myself to address this.

PR #2045, opened 2024-05-14
OCM-7268 | feat: Update create machinepool command to support kubeletconfigs

robpblake:
This PR updates the `rosa create machinepool` command with support for the `--kubelet-configs` option, that allows the user to specify a list of KubeletConfig `name` to attach to a specific HCP MachinePool.

### DIFF
--- a/cmd/create/machinepool/nodepool.go
+++ b/cmd/create/machinepool/nodepool.go
@@ -386,6 +386,44 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
 		npBuilder.TuningConfigs(inputTuningConfig...)
 	}
 
+	var inputKubeletConfigs []string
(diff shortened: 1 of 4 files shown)


cmd/create/machinepool/nodepool.go
```diff
@@ -386,6 +386,44 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
 		npBuilder.TuningConfigs(inputTuningConfig...)
 	}
 
+	var inputKubeletConfigs []string
```
robpblake, line 389: We have an active task on-going to refactor the `create machinepool` command to align with the runner model. I'm following the existing implementation at the minute to avoid do a large refactor in this PR.

ciaranRoche (approved):
/lgtm

PR #2046, opened 2024-05-14
OCM-7264 | fix: Ensured that name is optional when creating a KubeletConfig

robpblake:
The `--name` arg is optional when creating a `KubeletConfig`. If omitted, it will be generated on the server-side. We need to ensure we do not send an empty string as the name if the user has not supplied the arg.


ciaranRoche (approved):
/lgtm

PR #2049, opened 2024-05-15
OCM-4458 | test: automate id:71946 validation for create/list/delete idp and admin for exteranl auth cluster

aaraj7:
output log :- https://privatebin.corp.redhat.com/?a82569ade54d98a3#Bebud92EF4E8WmWVFBr77eLdGvBVVYzGG8XRVLAoMHLR

### DIFF
--- a/tests/e2e/hcp_external_auth_test.go
+++ b/tests/e2e/hcp_external_auth_test.go
@@ -110,7 +110,7 @@ var _ = Describe("Rosacli Testing", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 
-		It("Validation for HCP cluster break_glass_credentials create/list/describe/delete can work well via ROSA client - [id:73018]",
+		It("validation for HCP cluster break_glass_credentials create/list/describe/delete can work well via ROSA client - [id:73018]",
 			labels.Medium,
 			func() {
 				By("Create/list/revoke break-glass-credential to non-HCP cluster")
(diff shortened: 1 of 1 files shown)


tests/e2e/hcp_external_auth_test.go
```diff
@@ -193,5 +193,55 @@ var _ = Describe("Rosacli Testing", func() {
...
+			labels.Medium,
+			func() {
+
+				By("Check if cluster is hosted cp cluster and external_auth_providers is enabled")
+				isHostedCP, err := clusterService.IsHostedCPCluster(clusterID)
+				Expect(err).ToNot(HaveOccurred())
+				isExternalAuthEnabled, err := clusterService.IsExternalAuthenticationEnabled(clusterID)
+				Expect(err).ToNot(HaveOccurred())
+
+				if isHostedCP && isExternalAuthEnabled {
```
OAharoni-RedHat, line 207: nit: I think it might be a little cleaner if we used the if to check if it's not an HCP cluster with ExternalAuth and skip running the test in that case rather than having the whole thing inside the if block, e.g.:

```
if !isHostedCP || !isExternalAuthEnabled {
    	Skip("This case is for HCP clusters with External Auth")
}
```

aaraj7: Thanks for it, Updated likewise

tests/e2e/hcp_external_auth_test.go
```diff
@@ -193,5 +193,57 @@ var _ = Describe("Rosacli Testing", func() {
...
+			labels.Medium, labels.NonClassicCluster,
+			func() {
+
+				By("Check if cluster is hosted cp cluster and external_auth_providers is enabled")
+				isHostedCP, err := clusterService.IsHostedCPCluster(clusterID)
+				Expect(err).ToNot(HaveOccurred())
+				isExternalAuthEnabled, err := clusterService.IsExternalAuthenticationEnabled(clusterID)
+				Expect(err).ToNot(HaveOccurred())
+
+				if !isHostedCP || !isExternalAuthEnabled {
```
radtriste, line 207: Do we really need the check on `isHostedCP` if we have the already the `NonClassicCluster` set ?

aaraj7: Just added for double confirmation, I'll remove it if you say?

aaraj7: Done

radtriste (approved):
/lgtm

PR #2050, opened 2024-05-15
OCM-7263 | chore: Added additional aliases to the `rosa list kubeletconfig` command

robpblake:


cmd/list/kubeletconfig/cmd_test.go
```diff
@@ -29,7 +29,7 @@ var _ = Describe("List KubeletConfig Command", func() {
 			Expect(cmd.Use).To(Equal(use))
 			Expect(cmd.Short).To(Equal(short))
 			Expect(cmd.Long).To(Equal(long))
-			Expect(cmd.Aliases).To(ContainElements(alias))
+			Expect(cmd.Aliases).To(ContainElements("kubeletconfig", "kubelet-config", "kubelet-configs"))
```
marcolan018, line 32: can we use ContainElements(aliases) here?

robpblake: @marcolan018 Good point. I've made that change.

PR #2051, opened 2024-05-15
OCM-7488 | test: Automated id:73743 checking rosa version while logged out

jerichokeyne:
output:
```
$ ginkgo -focus 73743 tests/e2e/
Running Suite: e2e tests suite - /home/jericho/work/repos/rosa/tests/e2e
========================================================================
Random Seed: 1715783784

Will run 1 of 64 specs
•SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS

Ran 1 of 64 Specs in 0.469 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 63 Skipped
PASS

Ginkgo ran 1 suite in 1.751934317s
Test Suite Passed
```


tests/e2e/test_rosacli_version.go
```diff
@@ -0,0 +1,54 @@
...
+				By("Make sure to log out")
+				_, logout_err := rosaClient.Runner.Cmd("logout").Run()
+				Expect(logout_err).To(BeNil())
+				By("Get the version output")
+				version_stdout, version_err := rosaClient.Runner.Cmd("version").Run()
+				Expect(version_err).To(BeNil())
+				stdout, e := version_stdout.ReadString('\n')
+				Expect(e).To(BeNil())
+				By("Check the version output")
+				Expect(stdout).NotTo(BeEmpty())
```
OAharoni-RedHat, line 39: Does this line provide any extra coverage that `Expect(stdout).To(ContainSubstring(info.Version))` doesn't already ensure?

tests/e2e/test_rosacli_version.go
```diff
@@ -0,0 +1,54 @@
...
+		var (
+			rosaClient *rosacli.Client
+		)
+
+		BeforeEach(func() {
+			By("Init the client")
+			rosaClient = rosacli.NewClient()
+		})
+
+		It("can get the version of rosa CLI while logged out - [id:73743]",
```
OAharoni-RedHat, line 27: This test doesn't work on older versions of ROSA CLI, right? Should we mark anywhere to skip this test on pre-1.2.39 versions or whatever the number is? I don't know if we have test clusters running on older versions of the software so maybe this is overkill

tests/e2e/test_rosacli_version.go
jerichokeyne: I suppose not. I'll go ahead and remove it

tests/e2e/test_rosacli_version.go
mgahagan73: This did work on older versions, I think we regressed in the .38 timeframe  or so.

jerichokeyne: I'm a bit unclear as to what the behavior is pre-1.2.39. My understanding from the comments on the jira ticket is that it only affects builds from the master branch pre-1.2.39, but 1.2.38 works.

Also, this test assumes that the installed version of `rosa` is the one that matches the codebase as it pulls the version number from the code

OAharoni-RedHat: sounds like from what mike's saying that the code as is should be good to go, so LGTM!

tests/e2e/test_rosacli_version.go
```diff
@@ -0,0 +1,52 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/openshift/rosa/pkg/info"
+	"github.com/openshift/rosa/tests/ci/labels"
+	"github.com/openshift/rosa/tests/utils/exec/rosacli"
+)
+
+var _ = Describe("Get CLI version",
+	labels.Day1,
```
radtriste, line 13: I would have put that one as Day2. We don't have any created resource here

yasun1: We do `rosa login` beyond the test cases. It is complex for us to call `rosa login` back after logout. logout is a destructive operation.

I think validating `rosa version` is more meaningful rather than `logout -> rosa version`.

jerichokeyne: I'm fine with changing the label(s) if needed, but the bug this is related to is specifically about testing that a user can still get the version number without logging in. If logging out is too destructive I could change this to just checking the version and mark the polarion case as manual only if necesary

radtriste: isn't there an other way to set the OCM config env in the command run to a temp file and never login. That would reproduce the "not logged in" behavior to run the `rosa version` command

yasun1: We could run such `logout` cases in an individual step/job. But they need a label(maybe `destructive-logout`) rather than `day1`, `day2`, and so on. These cases are so special, even in destructive scenarios.

yasun1: @jerichokeyne Could you mark the label `destructive-logout` on this case?

jerichokeyne: @yasun1 I added the label `destructive-logout` to this specific case, and I also changed the overall label to `day2` in case we want to add more version tests

radtriste: @yasun1 @jerichokeyne Why not applying the `OCM_CONFIG` env as suggested [here](https://github.com/openshift/rosa/pull/2051#discussion_r1603855548) ?
This would leave the cluster and OCM connection intact and we would not need an extra destructive label on this case
cc @xueli181114

jerichokeyne: @radtriste 
> @yasun1 @jerichokeyne Why not applying the `OCM_CONFIG` env as suggested [here](https://github.com/openshift/rosa/pull/2051#discussion_r1603855548) instead of doing a logout ? This would leave the cluster and OCM connection intact and we would not need an extra destructive label on this case cc @xueli181114

I mocked up a quick way of doing that. I couldn't find an easy way to pass environment variables to the rosa runner, so I just set the environment variable outside of that. But, I'm not sure if that could cause issues with running other tests in parallel

radtriste: @jerichokeyne gonna have a look

radtriste: @jerichokeyne To setup an env to the client, you could:
1. Add a new field to the `runner` struct (like `var envs []string`)
2. Add a new method to the `runner` interface, `AddEnvVar(key,value)`,  which would add the `key=value` to the `envs` field
3. Add a new Method to the `Client` struct, `AddEnvVar(key,value)`, which would call the runner `AddEnvVar` method
4. In the `runner.Run()` method, add this line when setting up the command (line #183): `cmd.Env = append(cmd.Env, r.envs...)`
5. After you create your client in the `BeforeEach`, call the `rosaclient.AddEnvVar("OCM_CONFIG", tmpFile.Name())`

Also, the creation of the temp file for ocm config could be moved to `utils/common` as a `CreateTempOCMConfig` method, wdyt ?

yasun1: Yes, configuring the running environment could solve the issue. `rosa login` generates the credentials under `$HOME/.config/ocm/ocm.json`, which each rosa command will use to connect to the API. If the HOME path is changed, it is in `not login` status.

jerichokeyne: @radtriste, I implemented those changes. Feel free to take another look

tests/ci/labels/runtime.go
```diff
@@ -22,6 +22,7 @@ var Day3 = Label("day3")
 
 // destroy
 var Destroy = Label("destroy")
+var DestructiveLogout = Label("destructive-logout")
```
radtriste, line 25: can be removed

radtriste (commented):
minor change. Rest is all good to me

radtriste (approved):
/lgtm

tests/utils/exec/rosacli/cmd_runner.go
```diff
@@ -35,6 +37,7 @@ func NewRunner() *runner {
 			debug:  false,
 			color:  "auto",
 		},
+		envs: os.Environ(),
```
radtriste, line 40: do we really need that ?

jerichokeyne: In my testing when I would add an environment variable, then the runner would lose access to all the other environment variables which at least on my system would cause issues (e.g. with finding where the `rosa` binary is). It seems like the the behavior is when you set `cmd.Env` to `nil`, then it uses your system environment variables, but once you set `cmd.Env` to a value then it will only set those environment variables and you lose the system ones.

So I think we do need pass in all the other environment variables, but if you think that's better to do elsewhere I can move it

PR #2054, opened 2024-05-15
OCM-7259 | feat: E2E delete machinepool cmd tests; new runner

hunterkepley:
[OCM-7259](https://issues.redhat.com//browse/OCM-7259)
---

* E2E tests for high coverage of command + associated functions
* Every route covered AFAIK
* Testing of new options files + validations including regex validation for both hypershift and classic
* New runner for `delete machinepool` command

### DIFF
--- a/cmd/describe/machinepool/options.go
+++ b/cmd/describe/machinepool/options.go
@@ -13,17 +13,17 @@ type DescribeMachinepoolUserOptions struct {
 type DescribeMachinepoolOptions struct {
 	reporter *reporter.Object
 
-	args DescribeMachinepoolUserOptions
+	args *DescribeMachinepoolUserOptions
 }
 
-func NewDescribeMachinepoolUserOptions() DescribeMachinepoolUserOptions {
-	return DescribeMachinepoolUserOptions{machinepool: ""}
+func NewDescribeMachinepoolUserOptions() *DescribeMachinepoolUserOptions {
+	return &DescribeMachinepoolUserOptions{machinepool: ""}
 }
 
 func NewDescribeMachinepoolOptions() *DescribeMachinepoolOptions {
 	return &DescribeMachinepoolOptions{
 		reporter: reporter.CreateReporter(),
-		args:     DescribeMachinepoolUserOptions{},
+		args:     &DescribeMachinepoolUserOptions{},
 	}
 }
 
@@ -31,10 +31,17 @@ func (m *DescribeMachinepoolOptions) Machinepool() string {
 	return m.args.machinepool
 }
 
-func (m *DescribeMachinepoolOptions) Bind(args DescribeMachinepoolUserOptions) error {
-	if args.machinepool == "" {
+func (m *DescribeMachinepoolOptions) Bind(args *DescribeMachinepoolUserOptions, argv []string) error {
+	m.args = args
+	if m.args.machinepool == "" {
+		if len(argv) > 0 {
+			m.args.machinepool = argv[0]
+		}
+	}
+
+	if m.args.machinepool == "" {
 		return fmt.Errorf("you need to specify a machine pool name")
 	}
-	m.args.machinepool = args.machinepool
+
 	return nil
 }
--- a/cmd/dlt/machinepool/cmd.go
+++ b/cmd/dlt/machinepool/cmd.go
@@ -17,8 +17,8 @@ limitations under the License.
 package machinepool
 
 import (
+	"context"
 	"fmt"
-	"os"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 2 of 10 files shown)


hunterkepley:
ok  	github.com/openshift/rosa/cmd/dlt/machinepool	2.701s	coverage: 97.1% of statements

^ Machinepool command coverage -> 97.1%

cmd/dlt/machinepool/cmd.go
```diff
@@ -27,43 +28,69 @@ import (
...
-	r := rosa.NewRuntime().WithAWS().WithOCM()
-	defer r.Cleanup()
+func DeleteMachinePoolRunner(userOptions DeleteMachinepoolUserOptions) rosa.CommandRunner {
+	return func(_ context.Context, runtime *rosa.Runtime, cmd *cobra.Command, argv []string) error {
+		options := NewDescribeMachinepoolOptions()
+		// Allow the use also directly the machine pool id as positional parameter
+		if len(argv) == 1 && !cmd.Flag("machinepool").Changed {
+			userOptions.machinepool = argv[0]
+		} else {
+			err := cmd.ParseFlags(argv)
```
robpblake, line 74: @hunterkepley This is as a result of not taking in the pointer to the options. If you take in the `pointer`, you can remove the additional `cmd.ParseFlags` call and the additional assignment to the `userOptions.machinepool` field. This will be handled for you by cobra.

cmd/dlt/machinepool/cmd.go
```diff
@@ -27,43 +28,69 @@ import (
...
 }
 
-func run(_ *cobra.Command, argv []string) {
-	r := rosa.NewRuntime().WithAWS().WithOCM()
-	defer r.Cleanup()
+func DeleteMachinePoolRunner(userOptions DeleteMachinepoolUserOptions) rosa.CommandRunner {
+	return func(_ context.Context, runtime *rosa.Runtime, cmd *cobra.Command, argv []string) error {
+		options := NewDescribeMachinepoolOptions()
+		// Allow the use also directly the machine pool id as positional parameter
+		if len(argv) == 1 && !cmd.Flag("machinepool").Changed {
```
robpblake, line 71: @hunterkepley I'd be tempted to move this onto the `options` itself. So for example, `options.BindFromArgs(args []string)` where that function looks like this:

```
func (o *DescribeMachinePoolOptions) BindFromArgs(args []string) {
 if o.Name == "" {
  if len(args) > 0 {
   o.Name = args[0]
  } 
 }
}
```
You could then write a unit test for the options to ensure the `--name` field takes precedence.

cmd/dlt/machinepool/options.go
```diff
@@ -0,0 +1,44 @@
+package machinepool
+
+import (
+	"fmt"
+
+	"github.com/openshift/rosa/pkg/machinepool"
+	"github.com/openshift/rosa/pkg/reporter"
+)
+
+type DeleteMachinepoolUserOptions struct {
```
robpblake, line 10: @hunterkepley Do you mean to call this struct `DeleteMachinePoolUserOptions` ? If it can be shared between `describe` and `delete`, how about something like `NameOnlyMachinePoolOptions` or something like that (naming is hard :-) )

hunterkepley: :+1: may as well reuse, thanks

cmd/dlt/machinepool/cmd.go
hunterkepley: thanks rob :pray: , will do all this when im back from PTO

hunterkepley:
@robpblake Decided to combine the `BindWithArgs` and `Bind` functions into one, as I did not see a reason to call two functions when the second one was pretty much 3 lines. This solution works with both `--machinepool` (flag passed) and a standalone machinepool ID. Also, I have tested every possible combination of args, and it seems to not have changed behavior. New tests added + updated previous tests

thomasmckay:
lgtm

marcolan018:
lgtm

cmd/describe/machinepool/options.go
```diff
@@ -13,25 +13,31 @@ type DescribeMachinepoolUserOptions struct {
...
+		args:     &DescribeMachinepoolUserOptions{},
 	}
 }
 
 func (m *DescribeMachinepoolOptions) Machinepool() string {
 	return m.args.machinepool
 }
 
-func (m *DescribeMachinepoolOptions) Bind(args DescribeMachinepoolUserOptions) error {
+func (m *DescribeMachinepoolOptions) Bind(args *DescribeMachinepoolUserOptions, argv []string) error {
```
robpblake, line 34: @hunterkepley Can we make it a bit easier to read this method ;-)

We have 3 different ways to refer to the same thing I think 🤔 

Does this work:

```
func (m *DescribeMachinepoolOptions) Bind(args *DescribeMachinepoolUserOptions, argv []string) error {
 m.args = args
 if m.args.machinepool == "" {
  if len(argv) > 0 {
   m.args.machinepool = argv[0]
  }
 }

 if m.args.machinepool == "" {
   return fmt.Errorf("you need to specify a machine pool name")
 }

return nil
}
```

PR #2057, opened 2024-05-16
OCM-8058 | fix: Ensure --name option is required for working with KubeletConfigs on HCP clusters

robpblake:
Without the `--name` option, we rely on the generated name from the server-side which can be a little cumbersome. This enforces a better practice in `rosa` where we encourage customers to explicitly name their `KubeletConfig` for better readability.


ciaranRoche (approved):
/lgtm

PR #2061, opened 2024-05-16
OCM-8075 | fix: Ensure users can only supply a single kubeletconfig for HCP MachinePools

robpblake:
This PR tidies up the presentation and validation of the allowed kubelet-configs for a HCP machinepool. Currently whilst the flag is named `--kubelet-configs`, we only support a single entry. This allows us to proceed without having to rename the flag in the future, but makes it much clearer to the user, that at this time, a single kubeletconfig is supported.

This is enforced on the backend, but this provides a better UX for customers using `rosa`.


ciaranRoche (approved):
/lgtm

PR #2062, opened 2024-05-16
OCM-8076 | fix: Ensure the user is prompted if they change the kubeletconfig on their HCP MachinePool

robpblake:
When a user edits the `kubeletconfigs` on their HCP `Machinepool`, we must show them a warning as this will cause all nodes in the nodepool to be recreated to adjust to the new PIDs limit.


ciaranRoche (approved):
/lgtm

PR #2063, opened 2024-05-16
OCM-8055 | fix: Check kubeletconfig exists by name if user specifies it for Classic cluster

robpblake:
For ROSA Classic clusters, the user does not need to pass the `--name` flag to any of the `rosa kubeletconfig` command e.g `rosa create kubeletconfig` because there is only ever a single KubeletConfig for ROSA classic.

However, if the user does decide to pass the `--name` flag, we should verify the KubeletConfig matches so that they get the correct failure and we don't just change things that don't meet their expectations.


ciaranRoche (approved):
/lgtm

PR #2064, opened 2024-05-16
OCM-7855 | fix: return message if there is no breakglass can be revoked

chenz4027:
```
./rosa revoke break-glass-credentials -c magchen-hcp2
I: There are no break glass credentials for cluster 'magchen-hcp2'
```


robpblake (approved):
/lgtm

PR #2065, opened 2024-05-17
OCM-8112 | fix: Force users into interactive mode if omitting a name when creating a KubeletConfig for HCP clusters

robpblake:
If users omit the `--name` flag when creating a KubeletConfig for HCP clusters, this will force them into interactive mode. This aligns us with other commands e.g `create machinepool`.

### DIFF
--- a/cmd/create/kubeletconfig/cmd_test.go
+++ b/cmd/create/kubeletconfig/cmd_test.go
@@ -124,30 +124,6 @@ var _ = Describe("create kubeletconfig", func() {
 				ContainSubstring("Failed getting KubeletConfig for cluster 'cluster'"))
 		})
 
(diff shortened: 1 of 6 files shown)


cmd/create/kubeletconfig/cmd_test.go
```diff
@@ -124,30 +124,6 @@ var _ = Describe("create kubeletconfig", func() {
 				ContainSubstring("Failed getting KubeletConfig for cluster 'cluster'"))
 		})
```
robpblake, line 126: This test had to be deleted as we now force the user into interactive mode and we have no good way of testing interactive mode (yet).

ciaranRoche (approved):
/lgtm

PR #2067, opened 2024-05-17
OCM-8059 | fix: Allow users to use positional args for specifying the name of kubeletconfig

robpblake:
This PR updates the `rosa create kubeletconfig`, `rosa edit kubeletconfig`, `rosa delete kubeletconfig` and `rosa describe kubeletconfig` commands to allow the user to specify the name of the kubeletconfig as a positional arg on the command line.

With this change, the following is now valid as an example:

- `rosa describe kubeletconfig -c mycluster high-pids`

The `--name` flag will always take precedence, so in the following, the name of the kubeletconfig will be `foo`

- `rosa create kubeletconfig -c mycluster --name foo bob`

This aligns us with the UX of other commands.


ciaranRoche (approved):
/lgtm

PR #2079, opened 2024-05-20
OCM-8051 | test: Automated id:66372 create cluster with invalid volume sizes

jerichokeyne:
output:
```
$ TEST_PROFILE=rosa-upgrade-z-stream ginkgo -focus 66372 tests/e2e/
Running Suite: e2e tests suite - /home/jericho/work/repos/rosa/tests/e2e
========================================================================
Random Seed: 1716222234

Will run 1 of 78 specs
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS•SSSSSSSSSSSSSSSSSSSSSSSSSS

Ran 1 of 78 Specs in 124.927 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 77 Skipped
PASS

Ginkgo ran 1 suite in 2m9.195225649s
Test Suite Passed
```

### DIFF
--- a/tests/e2e/e2e_setup_test.go
+++ b/tests/e2e/e2e_setup_test.go
@@ -1,11 +1,14 @@
 package e2e
 
 import (
+	"fmt"
+
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
 	"github.com/openshift/rosa/tests/ci/config"
 	"github.com/openshift/rosa/tests/ci/labels"
+	utilConfig "github.com/openshift/rosa/tests/utils/config"
 	"github.com/openshift/rosa/tests/utils/exec/rosacli"
 	"github.com/openshift/rosa/tests/utils/log"
 	"github.com/openshift/rosa/tests/utils/profilehandler"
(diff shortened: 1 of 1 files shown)


mgahagan73:
if the last size value is supposed to be string, seems like you should use something like "-jkp"  rather than another negative number (we already test for -1)

jerichokeyne:
> if the last size value is supposed to be string, seems like you should use something like "-jkp" rather than another negative number (we already test for -1)

@mgahagan73 I'd like to do that, but `profile.ClusterConfig.VolumeSize` is typed as an integer, so to be able to use the `GenerateClusterCreateFlags` function to get the CLI flags I need it to be an integer at first, and then I can replace the value with a string afterwards.

That being said, if you think it'd be cleaner I could replace that function and just generate the flags myself to avoid that issue

mgahagan73:
ahh got it, if the profile handler is used by anything other than rosacli then I can see why you would have to change it just before running the command. Relaxing the validation on the profile handler (if it wouldn't break anything else) is probably a topic for another PR. As for replacing the value versus replacing the function I don't really have an opinion since both have their upsides and downsides. @radtriste  Does it matter to you either way?

tests/e2e/e2e_setup_test.go
```diff
@@ -21,4 +23,41 @@ var _ = Describe("ROSA CLI Test", func() {
...
+		func() {
+			minSize := 128
+			maxSize := 16384
+
+			// too small, too big, negative, string
+			sizes := []int{minSize / 2, maxSize * 2, -1, -123}
+
+			client := rosacli.NewClient()
+			for _, volSize := range sizes {
+				profile := PH.LoadProfileYamlFileByENV()
```
radtriste, line 39: please load the profile only once before the `for` loop

radtriste:
> ahh got it, if the profile handler is used by anything other than rosacli then I can see why you would have to change it just before running the command. Relaxing the validation on the profile handler (if it wouldn't break anything else) is probably a topic for another PR. As for replacing the value versus replacing the function I don't really have an opinion since both have their upsides and downsides. @radtriste Does it matter to you either way?

Following this discussion, I am just thinking that you could use the `config.RetrieveClusterCreationCommand(...)` and then put the volume size option in the command
see https://github.com/openshift/rosa/blob/master/tests/e2e/test_rosacli_cluster.go#L264 how it is done for another case

jerichokeyne:
@radtriste I believe I've re-implemented it the way you were suggesting. Could you take another look at this PR when you get a chance?

tests/e2e/e2e_setup_test.go
```diff
@@ -30,4 +33,63 @@ var _ = Describe("Cluster preparation", labels.Feature.Cluster, func() {
...
+			var values map[string]string
+			var cmd string
+
+			By("Try a worker disk size that's too small")
+			values = map[string]string{
+				"--worker-disk-size": fmt.Sprintf("%dGiB", minSize-1),
+			}
+			rosaCommand.ReplaceFlagValue(values)
+			cmd = rosaCommand.GetFullCommand()
+			fmt.Println(cmd)
```
radtriste, line 60: please use Logger or remove that line

radtriste (changes requested):

tests/e2e/e2e_setup_test.go
jerichokeyne: Woops, forgot to remove that line after I finished debugging a problem

radtriste (approved):
/lgtm

PR #2080, opened 2024-05-20
OCM-8116 | fix: add validation for username/password of idp creation

marcolan018:
https://issues.redhat.com/browse/OCM-8116


cmd/create/idp/htpasswd.go
```diff
@@ -304,3 +314,19 @@ func parseHtpasswordFile(usersList *map[string]string, filePath string) error {
 
 	return nil
 }
+
+func validateUserPasswd(username, password string) error {
```
davidleerh, line 318: NP can we rename this to something like
validateHtUsernameAndPassword?

marcolan018: changed

PR #2083, opened 2024-05-21
OCM-3701 | feat: Remove version from editing machinepool

chenz4027:
https://issues.redhat.com/browse/OCM-3701


cmd/edit/machinepool/nodepool_test.go
```diff
@@ -121,4 +121,28 @@ var _ = Describe("Nodepool", func() {
 			Expect(invoked).To(BeFalse())
 		})
 	})
+
+	Context("getNodePoolReplicas", func() {
```
chenz4027, line 125: It's not related to the removal of version parameter, just to bump up the test coverage to pass the check.

PR #2084, opened 2024-05-21
OCM-8136 | fix: Delete should exit 1 if fails

chenz4027:
```
./rosa delete machinepool -c magchen-hcp workers -y
E: Error deleting machinepool: Failed to delete machine pool 'workers' on hosted cluster 'magchen-hcp': The last node pool can not be deleted from a cluster.
echo $?
1
```

https://issues.redhat.com/browse/OCM-8136


robpblake (approved):
/lgtm

PR #2090, opened 2024-05-22
OCM-8241 | 'describe addon' convert to runner

thomasmckay:


cmd/describe/addon/cmd_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+var _ = Describe("describe addon", func() {
+	Context("describe addon command", func() {
+		It("returns command", func() {
+			cmd := NewDescribeAddonCommand()
+			Expect(cmd).NotTo(BeNil())
+		})
+	})
+
+	Context("execute describe addon", func() {
```
hunterkepley, line 26: AFAICT, the command isnt executed in these tests, only creation of the command, and the use of the args function from cobra

cmd/describe/addon/cmd_test.go
```diff
@@ -0,0 +1,53 @@
...
+			err := cmd.Args(cmd, []string{})
+			Expect(err).ToNot(BeNil())
+			Expect(err.Error()).To(Equal("Expected exactly " +
+				"one command line argument containing the identifier of the add-on"))
+		})
+
+		It("Succeeds with addon arg", func() {
+			err := cmd.Args(cmd, []string{"name"})
+			Expect(err).To(BeNil())
+		})
```
hunterkepley, line 51: Might be best to include some more tests, maybe including mocking the API for a full end-to-end test of this command. If you want to do full testing of the command, here's an example in [a recent MR](https://github.com/openshift/rosa/pull/2054/files#diff-6d7a1f750109bd069d70a060cbfd0b0e52ec11aecd36e204f3cdb6fc9dd07a62)

cmd/describe/addon/cmd.go
```diff
@@ -28,41 +28,46 @@ import (
...
-	Run: run,
-	Args: func(_ *cobra.Command, argv []string) error {
-		if len(argv) != 1 {
-			return fmt.Errorf(
-				"Expected exactly one command line argument containing the identifier of the add-on")
-		}
-		return nil
-	},
+func NewDescribeAddonCommand() *cobra.Command {
+	cmd := &cobra.Command{
```
hunterkepley, line 32: NP: I think `return &cobra.Command{ ... }` should work here

PR #2102, opened 2024-05-27
OCM-8283 | fix: kubelet-config edit prompt no longer shows up when kubelet-configs are not targeted

den-rgb:
[OCM-8283 ](https://issues.redhat.com/browse/OCM-8283)- Fixing bug to stop kubelet-config warning showing up when kubelet configs are not specified.

### DIFF
--- a/cmd/edit/machinepool/nodepool.go
+++ b/cmd/edit/machinepool/nodepool.go
@@ -176,7 +176,7 @@ func editNodePool(cmd *cobra.Command, nodePoolID string,
 				inputKubeletConfig = nodePool.KubeletConfigs()
 			}
 
-			// Skip if no tuning configs are available
+			// Skip if no kubelet configs are available
 			if len(availableKubeletConfigs) > 0 {
 				inputKubeletConfig, err = interactive.GetMultipleOptions(interactive.Input{
 					Question: "Kubelet config",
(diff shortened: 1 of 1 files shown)


cmd/edit/machinepool/nodepool.go
```diff
@@ -234,7 +234,7 @@ func editNodePool(cmd *cobra.Command, nodePoolID string,
 		return fmt.Errorf("Failed to create machine pool for hosted cluster '%s': %v", clusterKey, err)
 	}
 
-	if !promptForNodePoolNodeRecreate(nodePool, update, PromptToAcceptNodePoolNodeRecreate, r) {
+	if isKubeletConfigSet && !promptForNodePoolNodeRecreate(nodePool, update, PromptToAcceptNodePoolNodeRecreate, r) {
```
ciaranRoche, line 237: I dont think this will work :thinking: as it only checks and will prompt if they pass it as a param. If they set it in interactive mode the prompt wont fire.

IMO I think the issue only happens if interactive mode is not triggered, during this flow we set the update to equal the existing. 

We can either do that always, ensure update = existing, or update the prompt func to take in two `[]string` and move the prompt func up to line 201, inside that if block, should trigger at the correct time, or another option, set `isKubletConfigSet` to true if interactive mode is enabled

PR #2112, opened 2024-05-29
OCM-8330 | fix: Fix delete oidc provider failed

jameszwang:


tests/utils/profilehandler/profile_handler.go
```diff
@@ -850,26 +850,26 @@ func DestroyPreparedUserData(client *rosacli.Client, clusterID string, region st
...
 			log.Logger.Infof("Find prepared oidc config id: %s", ud.OIDCConfigID)
 			_, err := ocmResourceService.DeleteOIDCConfig("--oidc-config-id", ud.OIDCConfigID, "--region", region, "--mode", "auto", "-y")
 			if err != nil {
-				log.Logger.Errorf("Error happened when delete oidc config id: %s", err.Error())
+				log.Logger.Errorf("Error happened when delete oidc config: %s", err.Error())
 				errors = append(errors, err)
 			} else {
 				ud.OIDCConfigID = ""
-				log.Logger.Info("Delete oidc config id successfully ")
+				log.Logger.Info("Delete oidc config successfully ")
```
yasun1, line 862: As all of the resource deletion are using the same format to print out the result, to make the 
```
destroyLog := func(err error, errors []error, resource string) bool {
  if err != nil {
      log.Logger.Errorf("Error happened when delete %s: %s", resource, err.Error())
      errors = append(errors, err)
      return false
  }
  
  log.Logger.Info("Delete %s successfully", resource)
  return true
}

# usage in each block
success := destroyLog(err, errors, "oidc provider")
if success {
  ud.OIDCConfigID = ""
}
```

jameszwang: @yasun1 Fixed according to comments. Can you take a looke? Thank you.

PR #2124, opened 2024-05-31
OCM-8337 | fix: [HCP] always set replicas when updating machinepools

andreadecorte:
We always pass the replicas when editing a machinepool, even if the value has not changed, as this will then be a no-op. Right now we are not setting replicas if they are the same as the current value and this result in an error returned to the user.

Related: [OCM-8337](https://issues.redhat.com//browse/OCM-8337)


andreadecorte:
@chenz4027 PTAL, as linked with https://github.com/openshift/rosa/pull/1940

robpblake (approved):
/lgtm

PR #2131, opened 2024-06-03
OCM-7918 | fix: use space delimiter in AWS tags error msg to avoid confusion

willkutler:
In https://issues.redhat.com/browse/OCM-7918 the code is working as intended, it just parses the tag as having too many delimiters because the value is not separated by a space. Since spaces are not allowed in tag key/values but `:` is, it makes more sense to use a space as a delimiter in the error message to avoid confusion on how the tags are parsed.


gdbranco:
Considering this is a helper would it be possible to include a test?

PR #2140, opened 2024-06-04
OCM-7289 | feat: create/update of maxUnavailable and maxSurge for NodePool

philipwu08:
https://issues.redhat.com/browse/OCM-6654

### DIFF
--- a/cmd/describe/machinepool/cmd_test.go
+++ b/cmd/describe/machinepool/cmd_test.go
@@ -38,6 +38,10 @@ Tuning configs:
 Kubelet configs:                       
 Additional security group IDs:         
 Node drain grace period:               1 minute
+Management upgrade:                    
+ - Type:                               Replace
+ - Max surge:                          1
+ - Max unavailable:                    0
 Message:                               
 `
 	describeStringWithUpgradeOutput = `
@@ -58,6 +62,10 @@ Tuning configs:
 Kubelet configs:                       
 Additional security group IDs:         
 Node drain grace period:               1 minute
+Management upgrade:                    
+ - Type:                               Replace
+ - Max surge:                          1
+ - Max unavailable:                    0
 Message:                               
 Scheduled upgrade:                     scheduled 4.12.25 on 2023-08-07 15:22 UTC
 `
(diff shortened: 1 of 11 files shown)


cmd/describe/machinepool/cmd_test.go
```diff
@@ -38,6 +38,7 @@ Tuning configs:
 Kubelet configs:                       
 Additional security group IDs:         
 Node drain grace period:               1 minute
+Management upgrade:
```
chenz4027, line 41: How come it doesn't have the value for management upgrade?

chenz4027: If there's no management upgrade, should we hide this instead of displaying?

pkg/helper/machinepools/helpers.go
```diff
@@ -341,3 +341,30 @@ func ValidateNodeDrainGracePeriod(val interface{}) error {
 	}
 	return nil
 }
+
+func ValidateUpgradeMaxSurgeUnavailable(val interface{}) error {
+	maxsurgeorunavail := strings.TrimSpace(val.(string))
```
chenz4027, line 346: camal case

pkg/helper/machinepools/helpers.go
```diff
@@ -341,3 +341,30 @@ func ValidateNodeDrainGracePeriod(val interface{}) error {
...
+	if strings.HasSuffix(maxsurgeorunavail, "%") {
+		percent, err := strconv.Atoi(strings.TrimSuffix(maxsurgeorunavail, "%"))
+		if err != nil {
+			return fmt.Errorf("Percentage value '%s' must be an integer", strings.TrimSuffix(maxsurgeorunavail, "%"))
+		}
+		if percent < 0 || percent > 100 {
+			return fmt.Errorf("Percentage value %d must be between 0 and 100", percent)
+		}
+	} else {
+		intMaxsurgeorunavail, err := strconv.Atoi(maxsurgeorunavail)
```
chenz4027, line 360: camal

pkg/machinepool/output.go
```diff
@@ -27,6 +27,7 @@ var nodePoolOutputString string = "\n" +
 	"Kubelet configs:                       %s\n" +
 	"Additional security group IDs:         %s\n" +
 	"Node drain grace period:               %s\n" +
+	"Management upgrade:                    %s\n" +
```
chenz4027, line 30: Comment as above, if it's optional, we should also reflect in describe. Only display `Management upgrade` if we do have max surge and unavailable

philipwu08: It will always have a value, it defaults to maxSurge=1, maxUnavailable=0.

cmd/describe/machinepool/cmd_test.go
philipwu08: It will always have a value. The test nodepool doesn't have manageupgrade upgrade, but that's updated now.

PR #2150, opened 2024-06-07
OCM-6477 | fix: fetch subnets for availability zone

chenz4027:
https://issues.redhat.com/browse/OCM-6477

### DIFF
--- a/cmd/create/machinepool/nodepool.go
+++ b/cmd/create/machinepool/nodepool.go
@@ -6,10 +6,12 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/briandowns/spinner"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 
+	"github.com/openshift/rosa/pkg/aws"
 	"github.com/openshift/rosa/pkg/helper/features"
 	"github.com/openshift/rosa/pkg/helper/machinepools"
 	"github.com/openshift/rosa/pkg/helper/versions"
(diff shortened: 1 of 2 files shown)


cmd/create/machinepool/nodepool.go
```diff
@@ -540,7 +540,11 @@ func getSubnetFromAvailabilityZone(cmd *cobra.Command, r *rosa.Runtime, isAvaila
 		}
 		r.Reporter.Infof("There are several subnets for availability zone '%s'", availabilityZone)
 		interactive.Enable()
-		subnet := getSubnetFromUser(cmd, r, false, cluster)
+		subnet, err := getSubnetFromAvailabilityZone(cmd, r, isAvailabilityZoneSet, cluster)
```
davidleerh, line 543: Hi @chenz4027,
is this recursion expected here? please kindly advise

chenz4027: @davidleerh It is expected. The original error was that even if the user enters the availability zone, the subnets would include all subnets from all zones. Looking through the code, we were getting getSubnetFromUser instead of getSubnetFromAvailabilityZone. However, you are right it's introducing a recursion, let me see if I can revise it to avoid this behaviour

chenz4027: @davidleerh Would you mind taking another look? Refactored

cmd/create/machinepool/nodepool.go
```diff
@@ -540,7 +543,27 @@ func getSubnetFromAvailabilityZone(cmd *cobra.Command, r *rosa.Runtime, isAvaila
 		}
 		r.Reporter.Infof("There are several subnets for availability zone '%s'", availabilityZone)
 		interactive.Enable()
-		subnet := getSubnetFromUser(cmd, r, false, cluster)
+
+		// should only get the subnets from the selected availability zone
+		var subnetOptions []string
+		for _, subnet := range privateSubnets {
```
davidleerh, line 549: Is it required that we re-iterate through the list of privateSubnets to find which private subnets exist in a selected az zone?

Please correct me if I am wrong but wouldn't subnetsMap[availabilityZone] already have the same subnet ids (list of string), which line 551 is already trying to collect again?

Please kindly advise. :)

chenz4027: @davidleerh Yep so previously, the subnetOptions was initiated within getSubnetFromUser. Unfortunately, subnetsMap[availabilityZone] only contains subnetid, which is different from subnetOptions where subnet has a type of types.Subnet. aws.SetSubnetOption(subnet)) needs types.Subnet to construct the string in proper option format. 

In line 550, I was checking if each subnet has the id that is listed from subnetsMap[availabilityZone]

davidleerh: Sorry @chenz4027; 
Adding more eyes to this due to my lack of comprehension :cry: 

Hi @gdbranco would you kindly take a quick look at this? 
Might be my lack of understanding here but I feel like the value stored in subnetOptions == subnetsMap[availabilityZone] and that the iteration is extra compute that is not required?

subnetOptions is a list of strings which should be the same as datatype of subnetsMap[availabilityZone].

gdbranco: I'd say the map can be expanded I suppose, to contain also the subnet type itself. I'm not sure how big this slice ends up being able to be, but I'd say saving some iterations is always worth

chenz4027: Refactored, @gdbranco @davidleerh would you mind taking another look?

davidleerh: lgtm

PR #2152, opened 2024-06-07
OCM-8648 | feat: handle 412 response in upgrade commands

marcolan018:
https://issues.redhat.com/browse/OCM-8648

### DIFF
--- a/cmd/upgrade/roles/cmd.go
+++ b/cmd/upgrade/roles/cmd.go
@@ -66,6 +66,8 @@ const (
 	clusterVersionFlag = "cluster-version"
 	policyVersionFlag  = "policy-version"
 	channelGroupFlag   = "channel-group"
+
+	ArbitraryPolicyNotAvail = "STS arbitrary policies feature is currently not available"
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


cmd/upgrade/roles/cmd.go
```diff
@@ -267,15 +268,24 @@ func run(cmd *cobra.Command, argv []string) {
...
+			os.Exit(1)
+		}
 	}
-	err = rolepolicybindings.CheckRolePolicyBindingStatus(rolePolicyBindings)
-	if err != nil {
-		reporter.Errorf("Error in rolePolicyBinding: %s", err)
-		os.Exit(1)
+
+	rolePolicyDetails := map[string][]aws.PolicyDetail{}
+	if rolePolicyBindings != nil {
```
davidleerh, line 286: instead of checking for nil, should this be checking if the bindings are empty?
example: https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/8086/diffs

marcolan018: the nil is returned by our client: https://github.com/openshift/rosa/blob/03a121c9cb53e0daa7cd39fdf4f39e790034f923/pkg/ocm/rolepolicybinding.go#L31

davidleerh:
In addition to above, if say rolePolicyDetails is empty:
https://github.com/openshift/rosa/pull/2152/files#diff-42e429a79b1a9b4a9fbf33a5dab5ad029f8365750589cd5a1e3402e4ff3ddea1R285

Searching for list of policies on a role that doesn't exist in map
https://github.com/openshift/rosa/pull/2152/files#diff-42e429a79b1a9b4a9fbf33a5dab5ad029f8365750589cd5a1e3402e4ff3ddea1R534

I feel when we pass in the role policies here:
https://github.com/openshift/rosa/pull/2152/files#diff-42e429a79b1a9b4a9fbf33a5dab5ad029f8365750589cd5a1e3402e4ff3ddea1R603

https://github.com/openshift/rosa/pull/2152/files#diff-42e429a79b1a9b4a9fbf33a5dab5ad029f8365750589cd5a1e3402e4ff3ddea1R832

https://github.com/openshift/rosa/pull/2152/files#diff-42e429a79b1a9b4a9fbf33a5dab5ad029f8365750589cd5a1e3402e4ff3ddea1R917

https://github.com/openshift/rosa/pull/2152/files#diff-42e429a79b1a9b4a9fbf33a5dab5ad029f8365750589cd5a1e3402e4ff3ddea1R673

since we try to search a key that doesn't exist in rolePolicyDetails
https://github.com/openshift/rosa/pull/2152/files#diff-42e429a79b1a9b4a9fbf33a5dab5ad029f8365750589cd5a1e3402e4ff3ddea1R285

it might throw seg fault / ptr error? Would you kindly confirm?

marcolan018:
@davidleerh 
in golang, if we tries to get the value of a key which doesn't exist, we should get a nil value but not a ptr exception. So it should be enough to check wether the value is nil in the func handleAccountRolePolicyARN and handleOperatorRolePolicyARN. WDYT?

davidleerh:
CS is unable to output error 412 at this time.
Should we use 503 error code instead?

cmd/upgrade/roles/cmd.go
```diff
@@ -475,7 +485,8 @@ func run(cmd *cobra.Command, argv []string) {
 		}
 	}
 
-	if isUpgradeNeedForAccountRolePolicies && mode == interactive.ModeAuto || isOperatorPolicyUpgradeNeeded {
+	if rolePolicyBindings != nil && isUpgradeNeedForAccountRolePolicies &&
+		mode == interactive.ModeAuto || isOperatorPolicyUpgradeNeeded {
 		newRolePolicyBindings, err := ocmClient.ListRolePolicyBindings(cluster.ID(), true)
 		if err != nil {
 			reporter.Warnf("Failed to get rolePolicyBindings after upgrade." +
```
davidleerh, line 492: When testing locally, I get this output if I force error.
"Failed to get rolePolicyBindings after upgrade. Please ensure that the required policies are attached to the upgraded roles."
If the error is a specific error code, should we also follow this here?
https://github.com/openshift/rosa/pull/2152/files#diff-42e429a79b1a9b4a9fbf33a5dab5ad029f8365750589cd5a1e3402e4ff3ddea1R272

PR #2153, opened 2024-06-09
OCM-8315 | feat: add additional allowed principals config for hcp clusters

mikeshng:
https://issues.redhat.com/browse/OCM-8315

Expose a new param: additional-allowed-principals for hcp create/edit clusters. This param allows users to set the additional allowed principals for their HCP.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -180,9 +180,10 @@ var args struct {
 	tags []string
 
 	// Hypershift options:
-	hostedClusterEnabled bool
-	billingAccount       string
-	noCni                bool
+	hostedClusterEnabled        bool
+	billingAccount              string
+	noCni                       bool
+	additionalAllowedPrincipals []string
 
 	// Cluster Admin
 	createAdminUser      bool
@@ -454,6 +455,15 @@ func initFlags(cmd *cobra.Command) {
 		"",
 		"A file contains a PEM-encoded X.509 certificate bundle that will be "+
 			"added to the nodes' trusted certificate store.")
+
+	flags.StringSliceVar(
+		&args.additionalAllowedPrincipals,
+		"additional-allowed-principals",
+		nil,
+		"A comma-separated list of additional allowed principal ARNs "+
+			"to be added to the Hosted Control Plane's VPC Endpoint Service to enable additional "+
+			"VPC Endpoint connection requests to be automatically accepted.",
+	)
 
 	flags.BoolVar(&args.enableCustomerManagedKey,
 		"enable-customer-managed-key",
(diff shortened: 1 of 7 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -3896,6 +3932,11 @@ func buildCommand(spec ocm.Spec, operatorRolesPrefix string,
 		command += " --no-cni"
 	}
 
+	if len(spec.AdditionalAllowedPrincipals) > 0 {
+		command += fmt.Sprintf(" --additional-allowed-principals %s",
+			strings.Join(spec.AdditionalAllowedPrincipals, ","))
+	}
+
```
chenz4027, line 3939: Got a question for this one. Do we need `--additional-allowed-principals principal1, principal 2` or ``--additional-allowed-principals principal1, --additional-allowed-principals principal 2`

cmd/create/cluster/cmd.go
```diff
@@ -2852,6 +2862,31 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	// Additional Allowed Principals
+	if cmd.Flags().Changed("additional-allowed-principals") && !isHostedCP {
+		r.Reporter.Errorf("Additional Allowed Principals is supported only for Hosted Control Planes")
+		os.Exit(1)
+	}
+	additionalAllowedPrincipals := args.additionalAllowedPrincipals
+	if isHostedCP && interactive.Enabled() {
```
chenz4027, line 2871: Hmm I'm not a big fan of all the if else though. Would it be possible to use validators? One example to compare is with noProxySlice.

I was thinking of something like:
```
aapInputs, err := interactive.GetString(interactive.Input{
			Question: "Additional Allowed Principal ARNs",
			Help:     cmd.Flags().Lookup("additional-allowed-principals").Usage,
			Default:  strings.Join(additionalAllowedPrincipals, ","),
			Validators: [][]interactive.Validator{ValidateAdditionalAllowedPrincipals}
		})
additionalAllowedPrincipals = helper.HandleEmptyStringOnSlice(strings.Split(aapInputs, ","))
```

cmd/edit/cluster/cmd.go
```diff
@@ -500,6 +518,47 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
+	/*******  AdditionalAllowedPrincipals *******/
+	updateAdditionalAllowedPrincipals := false
+	if additionalAllowedPrincipals != nil {
```
chenz4027, line 523: Do we need to consider empty string[] scenario?

cmd/edit/cluster/cmd.go
```diff
@@ -500,6 +518,47 @@ func run(cmd *cobra.Command, _ []string) {
...
+		updateAdditionalAllowedPrincipals = true
+	}
+	if !updateAdditionalAllowedPrincipals && additionalAllowedPrincipals == nil &&
+		interactive.Enabled() {
+		updateAdditionalAllowedPrincipalsValue, err := interactive.GetBool(interactive.Input{
+			Question: "Update additional allowed principals",
+			Default:  updateAdditionalAllowedPrincipals,
```
chenz4027, line 530: Do we need to have an additional question about "Do we want to update?" The below "Additional Allowed Principal ARNs" should be enough though.

cmd/edit/cluster/cmd.go
```diff
@@ -500,6 +518,47 @@ func run(cmd *cobra.Command, _ []string) {
...
+			r.Reporter.Errorf("Expected a valid update-additional-allowed-principals value: %s", err)
+			os.Exit(1)
+		}
+		updateAdditionalAllowedPrincipals = updateAdditionalAllowedPrincipalsValue
+	}
+	if updateAdditionalAllowedPrincipals && interactive.Enabled() {
+		aapInputs, err := interactive.GetString(interactive.Input{
+			Question: "Additional Allowed Principal ARNs",
+			Help:     cmd.Flags().Lookup("additional-allowed-principals").Usage,
+			Default:  cluster.AWS().AdditionalAllowedPrincipals(),
```
chenz4027, line 542: Same as above, validators.

cmd/edit/cluster/cmd.go
mikeshng: Not here because empty string[] is a valid input. It means we will need to update the additional allowed principals by setting it to empty.

cmd/edit/cluster/cmd.go
mikeshng: We do need this because it follows the convention of cluster edit interactive mode by asking "Do you want to update" Y/N before proceeding. The reason is due to this is an optional param so in interactive mode you don't want to ask users for inputs on optional stuffs that they might not want to edit.

cmd/create/cluster/cmd.go
mikeshng: I think the noProxy is validation/validators are redundant.
If you look at the noProxy code they have similar [if slice is populated validation checks ](https://github.com/openshift/rosa/blob/v1.2.40/cmd/create/cluster/cmd.go#L2798-L2811). I think this is because validation needs to be done in param inputs as well.

cmd/edit/cluster/cmd.go
mikeshng: Same reply as my above the edit cluster noProxy code also have similar [if slice is populated validation checks](https://github.com/openshift/rosa/blob/v1.2.40/cmd/edit/cluster/cmd.go#L426-L443).

chenz4027: Sounds good, @mikeshng I don't have permission to resolve the thread though

PR #2165, opened 2024-06-12
OCM-8664 | ci: define override_rosacli_build to support CI running on indicated rosacli binary

xueli181114:


ciaranRoche (approved):
/lgtm

PR #2177, opened 2024-06-18
OCM-8641 | automated case id :68971

AkashKanni:
[OCM-8641](https://issues.redhat.com//browse/OCM-8641) [QE] Automate OCP-68971
https://privatebin.corp.redhat.com/?a0af94b04327d375#BQ2cyp69Sg2bHCjtiUZecCKC7FgSr2Q8rUwXUTGsu7ny

### DIFF
--- a/tests/e2e/test_rosacli_cluster.go
+++ b/tests/e2e/test_rosacli_cluster.go
@@ -10,6 +10,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/openshift-online/ocm-common/pkg/aws/aws_client"
+	"github.com/openshift-online/ocm-common/pkg/test/vpc_client"
 
 	ciConfig "github.com/openshift/rosa/tests/ci/config"
 	"github.com/openshift/rosa/tests/ci/labels"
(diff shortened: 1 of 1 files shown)


tests/e2e/test_rosacli_cluster.go
```diff
@@ -854,6 +855,126 @@ var _ = Describe("Classic cluster creation validation",
 				Expect(err).NotTo(BeNil())
 				Expect(errorOutput.String()).To(ContainSubstring("etcd encryption cannot be disabled on clusters with FIPS mode"))
 			})
+
+		It("Create rosa cluster with additional security groups will work well via rosacli - [id:68971]",
```
xueli181114, line 859: It's a day1-post case. Not day1-supplemental. And this option had been supported in day1

AkashKanni: Thanks @xueli181114, i did update the label can you please take a look again?

xueli181114: @AkashKanni  I think we made things wrong. The title of the case is ```Create rosa cluster with additional security groups will validate well via rosacli```  not ```Create rosa cluster with additional security groups will work well via rosacli``` . It's a validation case of day1-negative.

AkashKanni: @xueli181114 i updated case title and label, could you PTAL again?

tests/e2e/test_rosacli_cluster.go
```diff
@@ -854,6 +855,126 @@ var _ = Describe("Classic cluster creation validation",
...
+					for additionalSecurityGroupFlag, value := range securityGroups {
+						rosalCommand.AddFlags("--dry-run",
+							additionalSecurityGroupFlag, value,
+							"--version", "4.15.11",
+							"--hosted-cp", "-y")
+						output, err := rosaClient.Runner.RunCMD(strings.Split(rosalCommand.GetFullCommand(), " "))
+						Expect(err).To(HaveOccurred())
+						index = strings.Index(additionalSecurityGroupFlag, "a")
+						flagName = additionalSecurityGroupFlag[index:]
+						Expect(output.String()).To(ContainSubstring("Using '301721915996' as billing account"))
```
OAharoni-RedHat, line 903: nit: could we make this billing account a constant?

tests/e2e/test_rosacli_cluster.go
```diff
@@ -854,6 +855,126 @@ var _ = Describe("Classic cluster creation validation",
...
+				profile := profilehandler.LoadProfileYamlFileByENV()
+
+				By("Prepare creation command")
+				var command string
+				var rosalCommand config.Command
+				flags, err := profilehandler.GenerateClusterCreateFlags(profile, rosaClient)
+				Expect(err).To(BeNil())
+
+				command = "rosa create cluster --cluster-name " + profile.ClusterConfig.Name + " " + strings.Join(flags, " ")
+				rosalCommand = config.GenerateCommand(command)
```
OAharoni-RedHat, line 888: forgive my ignorance but what does the l stand for in this variable name? long command?

tests/e2e/test_rosacli_cluster.go
```diff
@@ -854,6 +855,126 @@ var _ = Describe("Classic cluster creation validation",
...
+					}
+
+					By("Try creating cluster with invalid additional security groups")
+					for additionalSecurityGroupFlag, value := range invalidSecurityGroups {
+						rosalCommand.AddFlags("--dry-run", additionalSecurityGroupFlag, "invalid", "--version", ocpVersion4_14, "-y")
+						output, err := rosaClient.Runner.RunCMD(strings.Split(rosalCommand.GetFullCommand(), " "))
+						Expect(err).To(HaveOccurred())
+						Expect(output.String()).To(ContainSubstring("Creating cluster"))
+						Expect(output.String()).To(ContainSubstring(
+							"To view a list of clusters and their status, run 'rosa list clusters'"))
```
OAharoni-RedHat, line 943: out of curiosity, why do we care about this substring existing in this particular output?

tests/e2e/test_rosacli_cluster.go
```diff
@@ -854,6 +855,126 @@ var _ = Describe("Classic cluster creation validation",
...
+						Expect(output.String()).To(ContainSubstring("Security Group ID '%s' doesn't have 'sg-' prefix", value))
+						rosalCommand.DeleteFlag(additionalSecurityGroupFlag, true)
+					}
+
+					By("Try creating cluster with additional security groups with invalid and more than 5 SG ids")
```
OAharoni-RedHat, line 948: I assume we're supposed to fail cluster creation if there are more than 5 Security Groups provided, but I'm not sure that this actually covers that? If we're trying to confirm that we can't input more than 5 security groups, wouldn't it make more sense to use valid IDs here to ensure the error is with the number of SGs rather than the SG itself?

tests/e2e/test_rosacli_cluster.go
xueli181114: This is sensitive info. Don't show billing account in code.
```suggestion
						Expect(output.String()).To(MatchRegex(MatchRegexp(`Using '[0-9]+' as billing account`)))
```

tests/e2e/test_rosacli_cluster.go
AkashKanni: @OAharoni-RedHat yes, 'l' is used being used to indicate its a long rosa command which is triggering cluster creation. several other test cases (for example OCP-38770)  were using the same var name, i thought to use same for consistency, Please let me know if it can be written in some other way to make it more readable. Thanks!

tests/e2e/test_rosacli_cluster.go
AkashKanni: @OAharoni-RedHat i see its a good chance to clarify the doubt in my mind, i was under the impression that if say there are 3 lines in ouput and we want to match the 3rd one, so can we directly match the 3rd without thinking about the earlier two lines?

tests/e2e/test_rosacli_cluster.go
AkashKanni: @OAharoni-RedHat i agree with your thoughts, i guess `invalid` word should be replaced with `valid` here at this test step `Create cluster with invalid  --additional-infra-security-group-ids ,--additional-control-plane-security-group-ids,  --additional-compute-security-group-ids with more than 5 SG ids` @xueli181114 could you please confirm this? Thanks!

tests/e2e/test_rosacli_cluster.go
OAharoni-RedHat: Makes sense to me, I hadn't seen it before is all :). Thanks for the info!

tests/e2e/test_rosacli_cluster.go
OAharoni-RedHat: Hmm I think I may have not been clear. I don't mind that we're checking substrings in general, just that I'm not sure how checking to see if `To view a list of clusters and their status, run 'rosa list clusters'` coming up in the substring helps us reach the goal of this test. It seems to me like we just want to make sure that the 3 line here which checks for `Security Group ID '%s' doesn't have 'sg-' prefix` is the only necessary substring we want to find. I recognize this is a bit of a nitpick, so feel free to ignore it!

tests/e2e/test_rosacli_cluster.go
xueli181114: This is not what we want to verify. Remove line 903 - 905

tests/e2e/test_rosacli_cluster.go
```diff
@@ -854,6 +855,126 @@ var _ = Describe("Classic cluster creation validation",
...
+						"--additional-control-plane-security-group-ids": "invalid",
+						"--additional-compute-security-group-ids":       "invalid",
+					}
+				)
+				profile := profilehandler.LoadProfileYamlFileByENV()
+
+				By("Prepare creation command")
+				var command string
+				var rosalCommand config.Command
+				flags, err := profilehandler.GenerateClusterCreateFlags(profile, rosaClient)
```
xueli181114, line 884: Let's don't use generate flags with profile. Otherwise the created resources will be leak there.

PR #2194, opened 2024-06-25
OCM-2730 | fix: Update error message for linking ocm-role/user-role with not existed entries

Mischulee:
https://issues.redhat.com/browse/OCM-2730


chenz4027:
LGTM

PR #2198, opened 2024-06-26
OCM-9140 | fix: Do not enable interactive for max-surge and max-available

philipwu08:
This makes the CLI tests easier to automate, so that interactive mode does not need to be enabled to update max-surge/max-unavailable.

https://issues.redhat.com/browse/OCM-9140


marcolan018 (approved):
lgtm

PR #2213, opened 2024-07-01
OCM-9240 | test: fixing id:49137,53031

AkashKanni:
[ocm-9240](https://issues.redhat.com//browse/ocm-9240) [CI]Debug and fix: 49137, 53031

$ ginkgo --focus "(53031|49137)" tests/e2e/
  Ginkgo CLI Version:
    2.11.0
    2.17.1 used by e2e

  Ginkgo CLI to match.

  Ginkgo CLI.

  unsupported.
  
Running Suite: ROSA CLI e2e tests suite - /home/akanni/OCP-Repository/rosa/tests/e2e
====================================================================================
Random Seed: 1719825336

Will run 2 of 127 specs
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS••SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS

Ran 2 of 127 Specs in 68.811 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 125 Skipped
PASS

Ginkgo ran 1 suite in 1m9.848237088s
Test Suite Passed

### DIFF
--- a/tests/e2e/test_rosacli_idp.go
+++ b/tests/e2e/test_rosacli_idp.go
@@ -15,6 +15,14 @@ import (
 	ph "github.com/openshift/rosa/tests/utils/profilehandler"
 )
 
+func validateIDPOutput(textData string, clusterID string, idpName string) {
+	Expect(textData).Should(ContainSubstring("Configuring IDP for cluster '%s'", clusterID))
+	Expect(textData).Should(ContainSubstring("Identity Provider '%s' has been created", idpName))
+	Expect(textData).Should(ContainSubstring("To log in to the console, open"))
+	Expect(textData).Should(ContainSubstring("and click on '%s'", idpName))
+
+}
+
 var _ = Describe("Edit IDP",
 	labels.Feature.IDP,
 	func() {
@@ -330,9 +338,7 @@ var _ = Describe("Edit IDP",
 					"-y")
 				Expect(err).To(BeNil())
 				textData = rosaClient.Parser.TextData.Input(output).Parse().Tip()
-				Expect(textData).Should(ContainSubstring("Identity Provider '%s' has been created", idpNames[0]))
-				Expect(textData).Should(ContainSubstring("To log in to the console, open"))
-				Expect(textData).Should(ContainSubstring("and click on '%s'", idpNames[0]))
+				validateIDPOutput(textData, clusterID, idpNames[0])
 
 				By("Create one htpasswd idp with single users")
 				multipleuserPasswd, err = common.GenerateMultipleHtpasswdPairs(2)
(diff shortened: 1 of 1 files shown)


radtriste:
@AkashKanni Ids fixed here and in the issue are not the same. Can you check ?

AkashKanni:
@radtriste i have corrected the issue title on jira card

tests/e2e/test_rosacli_idp.go
```diff
@@ -330,6 +330,7 @@ var _ = Describe("Edit IDP",
 					"-y")
 				Expect(err).To(BeNil())
 				textData = rosaClient.Parser.TextData.Input(output).Parse().Tip()
+				Expect(textData).Should(ContainSubstring("Configuring IDP for cluster '%s'", clusterID))
```
yasun1, line 333: I do not think it is an excellent choice to validate the general message at each checkpoint. We should validate this in one checkpoint, and the others will do the distinguishing validation.

AkashKanni: @yasun1  it is neccessary to validate this step `Expect(textData).Should(ContainSubstring("Configuring IDP for cluster '%s'", clusterID))` for each IDP creation, earlier this was missing thats why the CI job failed  

Do you mean use different validation checkpoints (from the 4 used in this PR) for each idp creation scenario?

•```
 [FAILED] [11.749 seconds]
Edit IDP [It] can create/delete the HTPasswd IDPs - [id:49137] [feature-idp, Critical, day2]
/go/src/github.com/openshift/rosa/tests/e2e/test_rosacli_idp.go:296

  [FAILED] Expected
      <string>: INFO: Configuring IDP for cluster '2c52jne75v1ccbhk444k0ubapp9iqsoe'
      INFO: Identity Provider 'htpasswdn1' has been created.
      
  to contain substring
      <string>: To log in to the console, open
  In [It] at: /go/src/github.com/openshift/rosa/tests/e2e/test_rosacli_idp.go:334 @ 06/26/24 20:40:26.725
```

Comment (yasun1) [reply tests/e2e/test_rosacli_idp.go:333]: I apologize; I did not realize they are for kinds of IDPs; it is truly required for validation.

I realized that almost all of the validation `Expectations are the same; how about collecting them into one function?

File: tests/e2e/test_rosacli_idp.go [hunk 2]
```diff
@@ -15,6 +15,14 @@ import (
 	ph "github.com/openshift/rosa/tests/utils/profilehandler"
 )
 
+func validateOuptput(textData string, clusterID string, idpName string) {
```
yasun1, line 18: As you define the function at the top level. I would suggest using a prominent function name so there is no conflict with the others in the future, e.g., `validateIDPOutput`.

AkashKanni: I have updated the PR and i do like your suggestion for function naming, i will try in incorporate this in future,
could you please have a look again? Thanks!

PR #2216, opened 2024-07-01
OCM-8453 | fix: --multi-availability-zone flag must not be supported for HCP cluster while creating nodepool

Mischulee:
https://issues.redhat.com/browse/OCM-8453

### DIFF
--- a/cmd/create/machinepool/nodepool.go
+++ b/cmd/create/machinepool/nodepool.go
@@ -33,6 +33,12 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
 		os.Exit(1)
 	}
 
+	isMultiAvailabilityZoneSet := cmd.Flags().Changed("multi-availability-zone")
+	if isMultiAvailabilityZoneSet {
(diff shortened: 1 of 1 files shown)


cmd/create/machinepool/nodepool.go
```diff
@@ -33,6 +33,12 @@ func addNodePool(cmd *cobra.Command, clusterKey string, cluster *cmv1.Cluster, r
 		os.Exit(1)
 	}
 
+	isMultiAvailabilityZoneSet := cmd.Flags().Changed("multi-availability-zone")
+	if isMultiAvailabilityZoneSet && cluster.Managed() {
```
lucasponce, line 37: The flow under the nodepool will happen only on Hosted Control Planes.

In the cmd you'll see:

```
		addNodePool(cmd, clusterKey, cluster, r)
	} else {
		addMachinePool(cmd, clusterKey, cluster, r)
	}
```

So, no need to add the "cluster.Managed()".

Perhaps the same validation should be also added in the edit command.

robpblake (approved):
/lgtm

PR #2224, opened 2024-07-03
OCM 7186 - Adding Unit test coverage to Create Machinepool and Nodepool functions

den-rgb:
[OCM-7186](https://issues.redhat.com//browse/OCM-7186) Adding unit test coverage for CreateNodePool and CreateMachinePool functions


cmd/create/machinepool/options.go
```diff
@@ -0,0 +1,40 @@
...
+
+type CreateMachinepoolOptions struct {
+	reporter *reporter.Object
+
+	args *mpOpts.CreateMachinepoolUserOptions
+}
+
+func NewCreateMachinepoolUserOptions() *mpOpts.CreateMachinepoolUserOptions {
+	return &mpOpts.CreateMachinepoolUserOptions{
+		InstanceType:          "m5.xlarge",
```
robpblake, line 16: @den-rgb `m5.xlarge` could be a `const`

pkg/options/machinepool/create.go
```diff
@@ -0,0 +1,259 @@
...
+	)
+
+	flags.StringSliceVar(
+		&options.Tags,
+		"tags",
+		nil,
+		"Apply user defined tags to all resources created by ROSA in AWS. "+
+			"Tags are comma separated, for example: 'key value, foo bar'",
+	)
+
```
robpblake, line 255: @den-rgb I think we're removing the `maxSurge` and `maxUnavailable` args from the CLI again (they were added back in #2259 Can you add them back please.

I'd like to get #2256 merged so we can be sure we're not dropping any other args as part of this refactor.

robpblake:
@den-rgb This needs a rebase from us merging #2259

cmd/create/machinepool/cmd.go
```diff
@@ -17,255 +17,100 @@ limitations under the License.
...
-	flags.StringVar(
-		&args.KubeletConfigs,
-		"kubelet-configs",
-		"",
-		"Name of the kubelet config to be applied to the machine pool. A single kubelet config is allowed. "+
-			"Kubelet config must already exist. "+
-			"This will overwrite any modifications made to node kubelet configs on an ongoing basis.",
-	)
+// Create machine pool based on cluster type
+func (m CreateMachinePool) createMachinePoolBasedOnClusterType(r *rosa.Runtime,
```
robpblake, line 55: @den-rgb I kinda feel like this should be in the `service`. So we have a single method on the `service` which is `createMachinePool` (Machine Pool is the consistent terminology) and internally, the service works out whether it needs to create a Classic MachinePool or a HCP Machine pool.

WDYT?

cmd/create/machinepool/cmd.go
```diff
@@ -17,255 +17,100 @@ limitations under the License.
...
 
-func run(cmd *cobra.Command, _ []string) {
-	r := rosa.NewRuntime().WithOCM()
-	defer r.Cleanup()
-
-	clusterKey := r.GetClusterKey()
-
-	cluster := r.FetchCluster()
+// Validate the cluster's state is ready
+func validateClusterState(cluster *cmv1.Cluster, clusterKey string) error {
```
robpblake, line 101: @den-rgb Same comment here about moving this into the service. I think the service should check to see if the cluster is in `ready` state and return the error if not.

cmd/create/machinepool/cmd.go
```diff
@@ -17,255 +17,100 @@ limitations under the License.
...
-		"",
-		"Root disk size with a suffix like GiB or TiB",
-	)
+// Original function refactored to use the new helper functions
+func CreateMachinepoolRunner(userOptions *mpOpts.CreateMachinepoolUserOptions) rosa.CommandRunner {
+	return func(ctx context.Context, r *rosa.Runtime, cmd *cobra.Command, argv []string) error {
+		var err error
+		options := NewCreateMachinepoolOptions()
+		clusterKey := r.GetClusterKey()
+		options.args = userOptions
```
robpblake, line 70: @den-rgb Should this be calling the `Bind` method on the `options` instance?

cmd/create/machinepool/cmd.go
```diff
@@ -17,255 +17,100 @@ limitations under the License.
...
+		return fmt.Errorf("Cluster '%s' is not yet ready", clusterKey)
 	}
+	return nil
+}
 
-	val, ok := cluster.Properties()[properties.UseLocalCredentials]
-	useLocalCredentials := ok && val == "true"
-
+// Parse labels if the 'labels' flag is set
+func validateLabels(cmd *cobra.Command, args *mpOpts.CreateMachinepoolUserOptions) error {
```
robpblake, line 109: @den-rgb I think we should move this into the `Bind` method on the `CreateMachinePoolOptions`. That method would validate that the user-specified labels are in the correct format, and return an `error` if not.

We can then add a unit test for that as part of your options unit testing.

cmd/create/machinepool/cmd_test.go
```diff
@@ -0,0 +1,115 @@
...
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/rosa/pkg/machinepool"
+	mpOpts "github.com/openshift/rosa/pkg/options/machinepool"
+	"github.com/openshift/rosa/pkg/rosa"
+	"github.com/openshift/rosa/pkg/test"
+)
+
+var _ = Describe("Create machine pool", func() {
```
robpblake, line 17: @den-rgb I love the conciseness and readability of these tests. Nice work!

cmd/create/machinepool/options_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	mpOpts "github.com/openshift/rosa/pkg/options/machinepool"
+	"github.com/openshift/rosa/pkg/reporter"
+)
+
+var _ = Describe("CreateMachinepoolOptions", func() {
```
robpblake, line 11: @den-rgb Same comment here..really nice readability and conciseness on these tests. Very nice!

cmd/create/machinepool/options_test.go
```diff
@@ -0,0 +1,59 @@
...
+		machinepoolOptions *CreateMachinepoolOptions
+		userOptions        *mpOpts.CreateMachinepoolUserOptions
+	)
+
+	BeforeEach(func() {
+		machinepoolOptions = NewCreateMachinepoolOptions()
+		userOptions = NewCreateMachinepoolUserOptions()
+	})
+
+	Describe("NewCreateMachinepoolUserOptions", func() {
```
robpblake, line 22: @den-rgb Minor, but I think we _tend_ to use `Context` under the initial `Describe`. No biggie, and please disregard if there is precedent else where.

robpblake (commented):
Thanks @den-rgb ! Looking good and I really like the tests you've written.

Some minor feedback for you. Have you manually tested this to create a MachinePool?

pkg/options/machinepool/create.go
den-rgb: done thanks for the reminder

cmd/create/machinepool/cmd_test.go
den-rgb: Thank you

cmd/create/machinepool/options_test.go
den-rgb: Thank you :)

cmd/create/machinepool/cmd.go
den-rgb: Yeah good idea I'll move it over

cmd/create/machinepool/cmd.go
den-rgb: I don't know if this would be necessary to move since its only really being used by the runner?

cmd/create/machinepool/cmd.go
den-rgb: The Bind i left in was accidental, the Bind in this instance is redundant because argument order doesnt have to be sequential and the Binds for other methods expect the name of the machinepool to be first, whereas create doesn't

cmd/create/machinepool/cmd.go
den-rgb: I don't know how necessary this is considering if i was to move it to Bind all bind would be checking is the validity of the labels which would be the exact same as having this function in place

den-rgb:
> Thanks @den-rgb ! Looking good and I really like the tests you've written.
> 
> Some minor feedback for you. Have you manually tested this to create a MachinePool?

Yes machinepools can be created successfully, I've also tried the recent regression errors against this pr

cmd/create/machinepool/cmd.go
ciaranRoche: yeah i would plus one what Rob said and move to the service. Keep the runner as dumb as possible, the runner really should not be making decisions. Only parsing/binding args and passing to the service. Leave the service make decisions

cmd/create/machinepool/cmd_test.go
```diff
@@ -0,0 +1,111 @@
...
+	})
+
+	AfterEach(func() {
+		ctrl.Finish()
+	})
+
+	Context("validateClusterState", func() {
+		It("should return nil when the cluster state is ready", func() {
+			err := validateClusterState(mockClassicClusterReady, "test-cluster")
+			Expect(err).To(BeNil())
```
hunterkepley, line 76: In my tests for these I have been doing either
`Expect(err).ToNot(HaveOccurred())` - or - `Expect(...).To(Succeed())`

May be good to have consistency here, not sure the exact difference between these if any, but Massimo has asked for these changes in CS many times before from me so relaying that convention

pkg/machinepool/helper.go
```diff
@@ -262,7 +259,10 @@ func getSubnetFromAvailabilityZone(cmd *cobra.Command, r *rosa.Runtime, isAvaila
 		}
 		r.Reporter.Infof("There are several subnets for availability zone '%s'", availabilityZone)
 		interactive.Enable()
-		subnet := getSubnetFromUser(cmd, r, false, cluster, args)
+		subnet, err := getSubnetFromUser(cmd, r, false, cluster, args)
```
hunterkepley, line 262: good call on this rework, definitely the proper way of handling the errors here

pkg/test/helpers.go
```diff
@@ -260,6 +260,51 @@ func FormatVersionList(versions []*v1.Version) string {
...
+
+	v1.MarshalMachineTypeList(mt, &mtJson)
+
+	return fmt.Sprintf(`
+	{
+		"kind": "MachineTypeList",
+		"page": 1,
+		"size": %d,
+		"total": %d,
+		"items": %s
```
hunterkepley, line 274: Kind of a nitpick, but we can move this into it's own backtick string and re-use:

```
var listString string = ` // Feel free to rename
{
    "kind": "%s",
    "page": 1,
    "size": %d,
    "total": %d,
    "items": %s
}`
```

hunterkepley (commented):
Overall LGTM on first pass, I am interested in the conversation about the validations being in the options with Rob. I have put mine in the options, but I can see both sides being fine

pkg/machinepool/helper.go
den-rgb: thank you :)

cmd/create/machinepool/cmd_test.go
den-rgb: okay I'll apply that to my tests

cmd/create/machinepool/cmd_test.go
```diff
@@ -0,0 +1,111 @@
...
+			mockCmd.Flags().Set("labels", "key1=value1,key2=value2")
+			err := validateLabels(mockCmd, mockArgs)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should return an error when the labels flag is set with invalid labels", func() {
+			mockArgs.Labels = "key1=value1,key2"
+			mockCmd.Flags().Set("labels", "key1=value1,key2")
+			err := validateLabels(mockCmd, mockArgs)
+			Expect(err).NotTo(BeNil())
```
ziccardi, line 108:
```suggestion
			Expect(err).ToNot(HaveOccurred())
```

den-rgb: That change expects an err to not have occured, The test case expects there to be an err

ziccardi: Sorry you are right. The right change should be:
```suggestion
			Expect(err).To(HaveOccurred())
```

ciaranRoche (changes requested):
Reviewed this, looks like autoscaling is not working correctly, in interactive mode. If set to yes it ignores the value.

```
osa create machinepool -c croche2 
I: Enabling interactive mode
? Machine pool name: new1
? Enable autoscaling: Yes
? Replicas: 2
I: Checking available instance types for machine pool 'new1'
? Instance type (default = 'm5.xlarge'): m5.xlarge
? Labels (optional): 
? Taints (optional): 
? Use spot instances: No
? Tags (optional): 
? Root disk size (GiB or TiB): 300 GiB
I: Machine pool 'new1' created successfully on cluster 'croche2'
I: To view the machine pool details, run 'rosa describe machinepool --cluster croche2 --machinepool new1'
I: To view all machine pools, run 'rosa list machinepools --cluster croche2'
╭─ ~/Work/ocm/rosa  on OCM-7186 *2 ························································································· ✔  took 20s  at 09:28:01 
╰─ rosa describe machinepool new1 -c croche2

ID:                                    new1
Cluster ID:                            2cvpp5it9rkr34lr6k3i4cs3tl39h7r4
Autoscaling:                           No
Replicas:                              0
Instance type:                         m5.xlarge
Labels:                                
Taints:                                
Availability zones:                    eu-west-1a
Subnets:                               
Spot instances:                        No
Disk size:                             300 GiB
Additional Security Group IDs:         
Tags:                                  red-hat-managed=true, red-hat-clustertype=rosa
```

Works fine when manually passing the flags.

ciaranRoche:
Also seeing issues when setting replicas, 

```
rosa create machinepool -c croche2                                                                                                    
I: Enabling interactive mode
? Machine pool name: mp2
? Enable autoscaling: No
? Replicas: 2
I: Checking available instance types for machine pool 'mp2'
? Instance type (default = 'm5.xlarge'): m5.xlarge
```
its setting replica to 0
```
rosa describe machinepool mp2 -c croche2

ID:                                    mp2
Cluster ID:                            2cvpp5it9rkr34lr6k3i4cs3tl39h7r4
Autoscaling:                           No
Replicas:                              0
Instance type:                         m5.xlarge
Labels:                                foo=bar, noop=boop
Taints:                                
Availability zones:                    eu-west-1a
Subnets:                               
Spot instances:                        No
Disk size:                             150 GiB
Additional Security Group IDs:         
Tags:                                  foo=bar, noop=boop, red-hat-clustertype=rosa, red-hat-managed=true

```

ciaranRoche:
I have verified locally on Classic, lets get it through now to give time for testing before code freeze. 

Can you ask QE to do full regression on this command @den-rgb

ciaranRoche (approved):
/lgtm

PR #2229, opened 2024-07-05
OCM-9298 | test: Automate OCP-60275 User can edit cluster via rosa-cli

jameszwang:


tests/e2e/test_rosacli_cluster.go
```diff
@@ -332,6 +332,97 @@ var _ = Describe("Edit cluster",
...
+				Expect(err).To(BeNil())
+				if !private {
+					SkipTestOnFeature("private")
+				}
+				isSTS, err := clusterService.IsSTSCluster(clusterID)
+				Expect(err).To(BeNil())
+				isHostedCP, err := clusterService.IsHostedCPCluster(clusterID)
+				Expect(err).To(BeNil())
+
+				defer func() {
```
xueli181114, line 350: Put this defer after the edit command. Otherwise it will fail if the cluster is a private link cluster

tests/e2e/test_rosacli_cluster.go
```diff
@@ -332,6 +332,97 @@ var _ = Describe("Edit cluster",
...
+
+					Expect(CD.Private).To(Equal("No"))
+					Expect(CD.UserWorkloadMonitoring).To(Equal("Disabled"))
+				}
+
+				By("Edit the cluster with '--private' and '--disable-workload-monitoring=false' flags")
+				out, editErr = clusterService.EditCluster(
+					clusterID,
+					"--private",
+					"--disable-workload-monitoring=false",
```
xueli181114, line 409: Can you double check if there is a case already covered `--disabe-workload-monitoring`?

PR #2241, opened 2024-07-10
OCM-9282 | test: automate case id:54869

patriksuba:
ginkgo run -focus 54869 tests/e2e
  Ginkgo CLI Version:
    2.14.0
    2.17.1 used by e2e

  Ginkgo CLI to match.

  Ginkgo CLI.

  unsupported.
  
Running Suite: ROSA CLI e2e tests suite - /home/psuba/workspace/rosa/tests/e2e
==============================================================================
Random Seed: 1720605122

Will run 1 of 137 specs
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS•SSSSSSSSSSSSSSS

Ran 1 of 137 Specs in 7.310 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 136 Skipped
PASS

Ginkgo ran 1 suite in 8.240905127s
Test Suite Passed

### DIFF
--- a/tests/e2e/hcp_cluster_test.go
+++ b/tests/e2e/hcp_cluster_test.go
@@ -8,6 +8,7 @@ import (
 
 	ciConfig "github.com/openshift/rosa/tests/ci/config"
 	"github.com/openshift/rosa/tests/ci/labels"
+	"github.com/openshift/rosa/tests/utils/common/constants"
 	"github.com/openshift/rosa/tests/utils/config"
 	"github.com/openshift/rosa/tests/utils/exec/rosacli"
 	"github.com/openshift/rosa/tests/utils/profilehandler"
@@ -19,11 +20,13 @@ var _ = Describe("HCP cluster testing",
 		defer GinkgoRecover()
 
 		var (
-			clusterID      string
-			rosaClient     *rosacli.Client
-			clusterService rosacli.ClusterService
-			clusterConfig  *config.ClusterConfig
-			profile        *profilehandler.Profile
+			clusterID          string
+			rosaClient         *rosacli.Client
+			clusterService     rosacli.ClusterService
+			clusterConfig      *config.ClusterConfig
+			profile            *profilehandler.Profile
+			machinePoolService rosacli.MachinePoolService
+			ocmResourceService rosacli.OCMResourceService
 		)
 
 		BeforeEach(func() {
@@ -38,6 +41,8 @@ var _ = Describe("HCP cluster testing",
 			var err error
 			clusterConfig, err = config.ParseClusterProfile()
 			Expect(err).ToNot(HaveOccurred())
+			machinePoolService = rosaClient.MachinePool
+			ocmResourceService = rosaClient.OCMResource
 
 			By("Skip testing if the cluster is not a HCP cluster")
 			hostedCluster, err := clusterService.IsHostedCPCluster(clusterID)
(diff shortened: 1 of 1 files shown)


radtriste (approved):
/lgtm

tests/e2e/hcp_cluster_test.go
```diff
@@ -260,4 +260,56 @@ var _ = Describe("HCP cluster testing",
...
+
+				By("Check values of description")
+				Expect(clusterDesc.Name).ToNot(BeEmpty())
+				Expect(clusterDesc.ID).ToNot(BeEmpty())
+				Expect(clusterDesc.ControlPlane).To(Equal("ROSA Service Hosted"))
+				Expect(clusterDesc.OpenshiftVersion).ToNot(BeEmpty())
+				Expect(clusterDesc.DNS).ToNot(BeEmpty())
+				Expect(clusterDesc.APIURL).ToNot(BeEmpty())
+				Expect(clusterDesc.Region).ToNot(BeEmpty())
+				Expect(clusterDesc.Availability[0]).To(HaveKeyWithValue("Control Plane", MatchRegexp("AZ")))
```
xueli181114, line 279: Control Plane must be `MultiAZ`
DataPlane should be calculated from workerpool zones to decide if it is `SingleAZ` or `MultiAZ`

tests/e2e/hcp_cluster_test.go
```diff
@@ -260,4 +260,56 @@ var _ = Describe("HCP cluster testing",
...
+							Expect(value).To(BeNil())
+						}
+					}
+				}
+				Expect(clusterDesc.Network[0]).To(HaveKey("Type"))
+				Expect(clusterDesc.Network[1]).To(HaveKey("Service CIDR"))
+				Expect(clusterDesc.Network[2]).To(HaveKey("Machine CIDR"))
+				Expect(clusterDesc.Network[3]).To(HaveKey("Pod CIDR"))
+				Expect(clusterDesc.Network[4]).To(HaveKey("Host Prefix"))
+				Expect(clusterDesc.Network[5]).To(HaveKeyWithValue("Subnets", MatchRegexp("^subnet-.{17}")))
```
xueli181114, line 299: When it has multi subnets set it will fail.

tests/e2e/hcp_cluster_test.go
```diff
@@ -260,4 +260,56 @@ var _ = Describe("HCP cluster testing",
...
+				Expect(clusterDesc.STSRoleArn).To(MatchRegexp("arn:aws:iam::[0-9]{12}:role/.+-HCP-ROSA-Installer-Role"))
+				Expect(clusterDesc.SupportRoleARN).To(MatchRegexp("arn:aws:iam::[0-9]{12}:role/.+-HCP-ROSA-Support-Role"))
+				Expect(clusterDesc.InstanceIAMRoles[0]).To(HaveKeyWithValue("Worker",
+					MatchRegexp("arn:aws:iam::[0-9]{12}:role/.+-HCP-ROSA-Worker-Role")))
+				Expect(clusterDesc.OperatorIAMRoles).To(HaveEach(MatchRegexp("arn:aws:iam::[0-9]{12}:role/.+")))
+				Expect(clusterDesc.OperatorIAMRoles).To(ContainElements(ContainSubstring("kube-system-kube-controller-manager"),
+					ContainSubstring("kube-system-capa-controller-manager"),
```
xueli181114, line 306: The name is changed with the name prefix, no need to hardcode here. Just verify the description of the operator roles match the result in cluster json body from backend

tests/e2e/hcp_cluster_test.go
patriksuba: The value of subnets is string. Regexp is expecting only one subnet presents which should be present all the time. Even if multiple subnets are in string it will not fail

tests/e2e/hcp_cluster_test.go
patriksuba: Removed hard coded values. I don't get the part "match the result in cluster json body from backend". Could you explain more ?

tests/e2e/hcp_cluster_test.go
patriksuba: Chnaged Control Plane value to be only `MultiAZ`. 
For data plane changed accepted values `SingleAZ` or `MultiAZ`. I am not sure how to calculate from workerpool zones. If only one workerpool is present then Data plane is SingleAZ and if multiple workerpool nodes are present then Data plane is MultipleAZ ?

patriksuba: Added calculation of Data plane value from workerpool number of nodes.

tests/e2e/hcp_cluster_test.go
patriksuba: Described operator roles and matched them with cluster description operator roles.

PR #2256, opened 2024-07-16
OCM-9619 | fix: Test expected ROSA CLI command structure and command Args to prevent regressions

robpblake:
This PR adds tests to try and reduce the number of regressions in ROSA CLI as we refactor the code-base to make it more testable.

In particular this PR focuses on preventing two types of regression:

1. Commands being accidentally removed from the CLI: See #2254 
2. Args being accidentally removed from Commands with the CLI: See #2259 

This PR does two things:

1. First, we introduce a test that verifies the built CLI structure against an expected command tree at build time. The test will fail if commands are missing, or new commands have been added to the CLI without updating the tree. The format for this can be seen [here](https://github.com/openshift/rosa/pull/2256/files#diff-6a987588131d2fcc1757321e47f47af65b4ec285ef5b1e55d47ed7f1262c04bd)
2. Introduce a test that verifies that the list of flags on a command matches the expected list. This test will fail if flags have been removed or flags added without updating the expected list of flags, making removal of a flag a conscious developer decision.

I've updated the PR template to try and draw the attention of contributors to this and also updated the `CONTRIBUTING.MD` to start a style guide for adding new commands.


den-rgb:
lgtm, but is there a reason th etests are only being added to the kubelet config commands? Is it because they will be testing all the commands in that test case anyway so theres no need to duplicate it for everything else?

robpblake:
> lgtm, but is there a reason th etests are only being added to the kubelet config commands? Is it because they will be testing all the commands in that test case anyway so theres no need to duplicate it for everything else?

@den-rgb Good question. I only added it to a few commands to illustrate the approach. If we're happy with it, I can add it to all commands.

ciaranRoche (approved):
/lgtm

ciaranRoche (approved):
/lgtm

PR #2257, opened 2024-07-16
OCM-9626 | fix: Fixed slow tests for break glass credential implementation

robpblake:
This PR updates the break glass credential implementation to allow the caller to specify the poll interval and timeout. This allows us more control in the tests to avoid the long timeouts.


ciaranRoche (commented):
/lgtm

PR #2263, opened 2024-07-17
OCM-9299 | test: Automate ids:57444,75445

jameszwang:


tests/e2e/test_rosacli_upgrade.go
```diff
@@ -259,4 +261,116 @@ var _ = Describe("Cluster Upgrade testing",
 			Expect(output.String()).To(ContainSubstring("are compatible with upgrade"))
 			Expect(output.String()).To(ContainSubstring("Upgrade successfully scheduled for cluster"))
 		})
+
+		It("to upgrade roles/operator-roles and cluster - [id:57444]", labels.Critical, labels.Runtime.Upgrade, func() {
```
xueli181114, line 265: @jameszwang I believe Yu Wang have a case to cover the y stream upgrade. Can you reach to him to see if there is there any necessary to merge the two cases?

jameszwang: @xueli181114  By reading Yu wang's test case(OCP-73731) and script, I think my  two cases are necessary. Yu wang's case is detach/delete policies and re-create them, but my case is upgrading the existed policies. So I don't think his case overlaps with mine.

tests/e2e/test_rosacli_upgrade.go
```diff
@@ -361,6 +364,194 @@ var _ = Describe("Cluster Upgrade testing",
...
+
+							_, err = rosaClient.Runner.RunCMD(cmd2)
+							Expect(err).To(BeNil())
+						} else {
+							cmd := strings.Split(command, " ")
+							if len(cmd) > 0 && cmd[len(cmd)-1] == "" {
+								cmd = cmd[:len(cmd)-1]
+							}
+							_, err := rosaClient.Runner.RunCMD(cmd)
+							Expect(err).To(BeNil())
```
xueli181114, line 550: Didn't check the final result on AWS.

PR #2273, opened 2024-07-22
OCM-9617 | feat: adding extra validation to fix panic introduced with httpTokens

den-rgb:
[OCM-9617](https://issues.redhat.com//browse/OCM-9617) Adding version validation before validating httpToken to avoid panic when version is empty

### DIFF
--- a/pkg/machinepool/machinepool.go
+++ b/pkg/machinepool/machinepool.go
@@ -955,11 +955,9 @@ func (m *machinePool) AddNodePool(cmd *cobra.Command, clusterKey string, cluster
 			return fmt.Errorf("Expected a valid http tokens value : %v", err)
 		}
 	}
+
 	if err = ocm.ValidateHttpTokensValue(httpTokens); err != nil {
 		return fmt.Errorf("Expected a valid http tokens value : %v", err)
-	}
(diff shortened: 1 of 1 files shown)


pkg/machinepool/machinepool.go
```diff
@@ -955,11 +955,13 @@ func (m *machinePool) AddNodePool(cmd *cobra.Command, clusterKey string, cluster
 			return fmt.Errorf("Expected a valid http tokens value : %v", err)
 		}
 	}
-	if err = ocm.ValidateHttpTokensValue(httpTokens); err != nil {
-		return fmt.Errorf("Expected a valid http tokens value : %v", err)
-	}
-	if err := ocm.ValidateHttpTokensVersion(ocm.GetVersionMinor(version), httpTokens); err != nil {
-		return fmt.Errorf(err.Error())
+	if httpTokens != string(cmv1.Ec2MetadataHttpTokensOptional) && version != "" {
```
marcolan018, line 958: the panic error only happened in ValidateHttpTokensVersion func when version is nil or ""?
if that's true, I think the simplest solution is to remove the func here since it's outdated. 
It check cluster version no less than 4.11, which is old and not supported now.

den-rgb: Yeah it panics when the version is "", I'll remove that check in that case

marcolan018:
lgtm

PR #2277, opened 2024-07-23
OCM-7875 | fix: Do not allow creating operator roles with managed policies role arn but no --hosted-cp flag 

Mischulee:
https://issues.redhat.com/browse/OCM-7875


chenz4027:
LGTM

PR #2289, opened 2024-07-26
OCM-9917 | fix: Issue with interactive + replicas edit machinepool

hunterkepley:
Replicas must be  >= 0
Can't set replicas when autoscaling enabled
interactive min/max replicas fix

### DIFF
--- a/cmd/edit/machinepool/cmd_test.go
+++ b/cmd/edit/machinepool/cmd_test.go
@@ -81,8 +81,8 @@ var _ = Describe("Edit Machinepool", func() {
 				cmd := NewEditMachinePoolCommand()
 				Expect(cmd.Flag("cluster").Value.Set(clusterId)).To(Succeed())
 				Expect(cmd.Flags().Set("labels", "test=test")).To(Succeed())
-				Expect(runner(context.Background(), t.RosaRuntime, cmd, []string{"--machinepool",
-					nodePoolId, "--min-replicas", "2", "--enable-autoscaling", "true", "-y"})).To(Succeed())
+				Expect(cmd.Flags().Set("replicas", "1")).To(Succeed())
+				Expect(runner(context.Background(), t.RosaRuntime, cmd, []string{})).To(Succeed())
 			})
 			It("Machinepool ID passed in without flag in random location", func() {
 				// First get
@@ -96,8 +96,11 @@ var _ = Describe("Edit Machinepool", func() {
 				cmd := NewEditMachinePoolCommand()
 				Expect(cmd.Flag("cluster").Value.Set(clusterId)).To(Succeed())
 				Expect(cmd.Flags().Set("labels", "test=test")).To(Succeed())
+				Expect(cmd.Flags().Set("min-replicas", "1")).To(Succeed())
+				Expect(cmd.Flags().Set("max-replicas", "1")).To(Succeed())
+				Expect(cmd.Flags().Set("enable-autoscaling", "true")).To(Succeed())
 				Expect(runner(context.Background(), t.RosaRuntime, cmd,
-					[]string{"--min-replicas", "2", nodePoolId, "--enable-autoscaling", "true", "-y"})).To(Succeed())
+					[]string{})).To(Succeed())
 			})
 		})
 
--- a/pkg/machinepool/machinepool.go
+++ b/pkg/machinepool/machinepool.go
@@ -1399,7 +1399,6 @@ func getMachinePoolReplicas(cmd *cobra.Command,
 	}
 
 	if autoscaling {
-
 		// set default values from previous autoscaling values
 		if !isMinReplicasSet {
 			minReplicas = existingAutoscaling.MinReplicas()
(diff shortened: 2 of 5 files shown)


pkg/machinepool/machinepool.go
```diff
@@ -1449,6 +1456,25 @@ func getMachinePoolReplicas(cmd *cobra.Command,
 			err = fmt.Errorf("Expected a valid number of replicas: %s", err)
 		}
 	}
+	if autoscaling && minReplicas <= 0 {
```
den-rgb, line 1459: Could these checks be put into a function to avoid code repetition and make it cleaner?

pkg/machinepool/machinepool.go
```diff
@@ -2010,6 +2034,25 @@ func getNodePoolReplicas(cmd *cobra.Command,
 			return
 		}
 	}
+	if autoscaling && minReplicas <= 0 {
```
den-rgb, line 2037: Can call the function then here also

pkg/machinepool/machinepool.go
```diff
@@ -1447,8 +1454,13 @@ func getMachinePoolReplicas(cmd *cobra.Command,
 		})
 		if err != nil {
 			err = fmt.Errorf("Expected a valid number of replicas: %s", err)
+			return
```
ciaranRoche, line 1457: NP : Can we update this func to move away from named return values. It leads to errors like this that can be missed. 
Not a blocker for this PR, maybe we can follow up with that change

pkg/machinepool/machinepool.go
```diff
@@ -2010,6 +2020,34 @@ func getNodePoolReplicas(cmd *cobra.Command,
...
 	}
+
+	err = validateEditInput("node", autoscaling, minReplicas, maxReplicas, replicas, isReplicasSet,
+		isAutoscalingSet, nodePoolID)
+
+	return
+}
+
+func validateEditInput(poolType string, autoscaling bool, minReplicas int, maxReplicas int, replicas int,
+	isReplicasSet bool, isAutoscalingSet bool, id string) (err error) {
```
ciaranRoche, line 2031: Lets move away from named return values here. Just to be safe :pray:

pkg/machinepool/machinepool.go
hunterkepley: Yeah I can follow up sometime soon with a refactor, i didn't like how it was set up like this, i definitely had a couple missed during testing haha

cmd/edit/machinepool/cmd_test.go
```diff
@@ -81,8 +81,9 @@ var _ = Describe("Edit Machinepool", func() {
 				cmd := NewEditMachinePoolCommand()
 				Expect(cmd.Flag("cluster").Value.Set(clusterId)).To(Succeed())
 				Expect(cmd.Flags().Set("labels", "test=test")).To(Succeed())
+				Expect(cmd.Flags().Set("replicas", "1")).To(Succeed())
 				Expect(runner(context.Background(), t.RosaRuntime, cmd, []string{"--machinepool",
-					nodePoolId, "--min-replicas", "2", "--enable-autoscaling", "true", "-y"})).To(Succeed())
+					nodePoolId, "--replicas", "2", "--enable-autoscaling", "true", "-y"})).To(Succeed())
```
gdbranco, line 86: Should this succeed? If you enable autoscaling you need to supply min/max instead of replicas, no?

hunterkepley: The args do not matter in testing, but i will remove them anyway (the new runner / cobra works differently when testing)

pkg/machinepool/machinepool.go
```diff
@@ -1519,8 +1531,7 @@ func editMachinePool(cmd *cobra.Command, machinePoolId string,
 		return fmt.Errorf("Failed to get autoscaling or replicas: '%s'", err)
 	}
 
-	if !autoscaling && replicas < 0 ||
-		(autoscaling && isMinReplicasSet && minReplicas < 0) {
+	if !autoscaling && replicas < 0 {
```
marcolan018, line 1534: I guess this has already been checked before in when invoking validateaEditInput()?
why do we need to check this again?

hunterkepley: ill check when im home but i believe interactive needs it here

hunterkepley: moved all to validation afte rereading when i got home

marcolan018: @hunterkepley 
can you check again on this?
I am still confused why we check "if !autoscaling && replicas < 0" here. Before this, we invoke getMachinePoolReplicas() and inside it func validateEditInput() is invoked. There we already checked this condition.
```
	if !autoscaling && replicas <= 0 {
		return fmt.Errorf("Replicas must be a positive number")
	}
```
so IMO, we should remove this check here. Or maybe I miss anything in the whole flow?

marcolan018:
lgtm

PR #2293, opened 2024-07-28
OCM-9958 | ci: make pr check enhancement to match kinds of commit messages

xueli181114:


tests/utils/exec/rosacli/pr_check.go
```diff
@@ -30,22 +32,37 @@ func GetCommitFoucs() (string, error) {
...
-				idStr = strings.Split(idStr, "id:")[1]
-				ids := strings.Split(idStr, ",")
-				tcIDs = append(tcIDs, ids...)
-			}
-		}
+	return output.String(), err
+}
+
+func GetFocusCaseIDs(commitMessage string) (string, error) {
+	reg := regexp.MustCompile(`ids?:([0-9,\s,]*)`)
```
yasun1, line 39: It is not an excellent choice to relax the rule to match the space. We should force the user to follow the format `ids?:1234,1234`.

tests/utils/exec/rosacli/pr_check.go
```diff
@@ -30,22 +32,37 @@ func GetCommitFoucs() (string, error) {
...
-	focus := strings.Join(tcIDs, "|")
-	_, err = common.CreateFileWithContent(config.Test.TestFocusFile, focus)
+	focus = strings.Join(ids, "|")
+	_, err := common.CreateFileWithContent(config.Test.TestFocusFile, focus)
 	return focus, err
 }
+
+func GetFeatureLabelFilter(commitMessage string) (*Labels, error) {
+	featureFilterMap := map[*Labels]regexp.Regexp{
+		&labels.Feature.Machinepool: *regexp.MustCompile(`[Ff]eature\s?:\s?[mM]achine[\s-]?[pP]ools?`),
```
yasun1, line 57: I think we could follow up the same logic on case's, and then do a strings.Contain . The user should get the accurate feature label filter from https://github.com/openshift/rosa/blob/master/tests/ci/labels/features.go

tests/utils/exec/rosacli/pr_check.go
xueli181114: I would respect team members habit.

tests/utils/exec/rosacli/pr_check.go
xueli181114: I would like code resolve all things instead of keeping rules by personality. With this map, whatever the label changed it will  work all the time.

PR #2298, opened 2024-07-29
OCM-3704 | fix: Update error message for create/describe/list/delete machine pool upgrade policy

Mischulee:
https://issues.redhat.com/browse/OCM-3704

### DIFF
--- a/cmd/describe/upgrade/cmd.go
+++ b/cmd/describe/upgrade/cmd.go
@@ -27,6 +27,7 @@ import (
 	"github.com/openshift/rosa/pkg/interactive/confirm"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/rosa"
+	"github.com/openshift/rosa/pkg/upgrade"
 )
 
 var Cmd = &cobra.Command{
@@ -38,7 +39,7 @@ var Cmd = &cobra.Command{
   rosa describe upgrade`,
 	Run:    run,
 	Hidden: false,
-	Args:   cobra.NoArgs,
+	Args:   upgrade.NewUpgradeArgsFunction(true),
 }
 
 var args struct {
--- a/cmd/upgrade/machinepool/cmd.go
+++ b/cmd/upgrade/machinepool/cmd.go
@@ -29,6 +29,7 @@ import (
 	"github.com/openshift/rosa/pkg/interactive/confirm"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/rosa"
+	"github.com/openshift/rosa/pkg/upgrade"
 )
 
 var args struct {
@@ -49,15 +50,8 @@ var Cmd = &cobra.Command{
 
   # Schedule a machinepool upgrade within the hour
   rosa upgrade machinepool np1 -c mycluster --version 4.12.20`,
-	Run: run,
-	Args: func(_ *cobra.Command, argv []string) error {
-		if len(argv) != 1 {
-			return fmt.Errorf(
-				"expected exactly one command line parameter containing the id of the machine pool",
-			)
-		}
-		return nil
-	},
+	Run:  run,
+	Args: upgrade.NewUpgradeArgsFunction(false),
 }
 
 func init() {
--- a/pkg/upgrade/upgrade.go
+++ b/pkg/upgrade/upgrade.go
@@ -1,0 +1,39 @@
+package upgrade
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/rosa/pkg/machinepool"
+)
+
+var (
+	ErrInvalidMachinePoolIdentifier = fmt.Errorf("expected a valid identifier for the machine pool")
+	ErrMissingMachinePoolIdentifier = fmt.Errorf("machine pool identifier is missing")
+	machinepoolFlagName             = "machinepool"
+)
+
+func NewUpgradeArgsFunction(flagSource bool) func(cmd *cobra.Command, argv []string) error {
+	return func(cmd *cobra.Command, argv []string) error {
+		var machinepoolID string
+		var err error
+
+		if flagSource {
+			flags := cmd.Flags()
+			if machinepoolID, err = flags.GetString(machinepoolFlagName); err != nil {
+				return err
+			}
+		} else {
+			if len(argv) != 1 {
+				return ErrMissingMachinePoolIdentifier
+			}
+			machinepoolID = argv[0]
+		}
+
+		if !machinepool.MachinePoolKeyRE.MatchString(machinepoolID) {
+			return ErrInvalidMachinePoolIdentifier
+		}
+		return nil
+	}
+}
--- a/pkg/upgrade/upgrade_test.go
+++ b/pkg/upgrade/upgrade_test.go
@@ -1,0 +1,76 @@
+package upgrade
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/spf13/cobra"
+)
+
+func TestUpgrade(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "upgrade testing")
+}
+
+var _ = Describe("NewUpgradeArgsFunction test", func() {
+	var (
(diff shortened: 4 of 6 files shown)


Mischulee:
Examples:

 ./rosa upgrade machinepool  -c ying-up1
Before:
Error: expected exactly one command line parameter containing the id of the machine pool
Usage:
  rosa upgrade machinepool [flags]
...

Failed to execute root command: expected exactly one command line parameter containing the id of the machine pool



After:
Error: machine pool id is missing
Usage:
  rosa upgrade machinepool [flags]
...

Failed to execute root command: machine pool id is missing


======================================================================================
 ./rosa list upgrade machinepool  -c mparilov
Before:
Error: unknown command "mparilov-hcp2" for "rosa list upgrades"
Usage:
  rosa list upgrades [flags]
...

Failed to execute root command: unknown command "mparilov-hcp2" for "rosa list upgrades"



After:
Error: machine pool id is missing
Usage:
  rosa list upgrades [flags]
 ...

Failed to execute root command: machine pool id is missing

cmd/describe/upgrade/cmd.go
```diff
@@ -38,7 +39,15 @@ var Cmd = &cobra.Command{
   rosa describe upgrade`,
 	Run:    run,
 	Hidden: false,
-	Args:   cobra.NoArgs,
+	Args: func(cmd *cobra.Command, argv []string) error {
```
robpblake, line 42: @Mischulee We are using the same function in multiple locations. Could you please refactor into a method that returns the function to be used. For example:

```
Args: NewUpgradeArgsFunction(),
```

where `NewUpgradeArgsFunction()` is a function that returns the `Args` func e.g:

```
func NewUpgradeArgsFunction() func(cmd *cobra.Command, argv []string) {
 return {
        flags := cmd.Flags()
	machinepoolID, _ := flags.GetString("machinepool")

	if !machinepool.MachinePoolKeyRE.MatchString(machinepoolID) {
		return fmt.Errorf("expected a valid identifier for the machine pool")
	}
	return nil
  }
}
```

Could you also the please add unit tests for the function to make sure it behaves as we expect. I'd suggest creating a new package in `pkg`, perhaps `upgrade` and moving the function and tests there.

There are lots of examples of how we use Ginkgo to write unit tests in the project. One that provides a good example of testing the flags on a `cobra.Command` is `pkg/output/flag_test.go`. Take a look and see if you are able to write your own test. If you have any other questions or need some pointers, please just connect with me on Slack.

cheers,

Rob

cmd/upgrade/machinepool/cmd.go
```diff
@@ -53,7 +53,7 @@ var Cmd = &cobra.Command{
 	Args: func(_ *cobra.Command, argv []string) error {
 		if len(argv) != 1 {
 			return fmt.Errorf(
-				"expected exactly one command line parameter containing the id of the machine pool",
+				"machine pool id is missing",
```
robpblake, line 56: @Mischulee Is there a reason why the message is different and we are not verifying that the machine pool id matches the regex as we do in the other functions?

robpblake (changes requested):
@Mischulee I've left a comment about refactoring to make the code here re-usable and also adding some unit tests.

cmd/upgrade/machinepool/cmd.go
Mischulee: In this case, for example this:.  ./rosa upgrade machinepool -c mparilov
The machinepool is not flag as in other cases, so we can not use the same function. But I can add additional check for regex.

cmd/describe/upgrade/cmd.go
Mischulee: @robpblake Thank you for a helpful review! I refactored the function and added unit test as suggested.

pkg/upgrade/upgrade.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/openshift/rosa/pkg/machinepool"
+)
+
+func NewUpgradeArgsFunction() func(cmd *cobra.Command, argv []string) error {
+	return func(cmd *cobra.Command, argv []string) error {
+		flags := cmd.Flags()
+		machinepoolID, _ := flags.GetString("machinepool")
+
+		if !machinepool.MachinePoolKeyRE.MatchString(machinepoolID) {
+			return fmt.Errorf("expected a valid identifier for the machine pool")
```
robpblake, line 17: @Mischulee We have this string repeated 3 times in this PR, so I'd suggest to make it a `const` in the `upgrade` package.

pkg/upgrade/upgrade.go
```diff
@@ -0,0 +1,21 @@
...
+
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/rosa/pkg/machinepool"
+)
+
+func NewUpgradeArgsFunction() func(cmd *cobra.Command, argv []string) error {
+	return func(cmd *cobra.Command, argv []string) error {
+		flags := cmd.Flags()
+		machinepoolID, _ := flags.GetString("machinepool")
```
robpblake, line 14: @Mischulee We shouldn't be ignoring the `error` here. Could you please add a check for the `err != nil` and return it from the function.

pkg/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/spf13/cobra"
+)
+
+func TestReporter(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "reporter testing")
```
robpblake, line 13: @Mischulee Could you update the text here to correctly identify the test suite. Thanks!

pkg/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,44 @@
+package upgrade
```
robpblake, line 1: @Mischulee Great unit test. Really easy to read!

pkg/upgrade/upgrade.go
robpblake: Alternatively you could make the `error` itself a package-level var and return that e.g:

```
var (
	InvalidMachinePoolIdentifier = fmt.Errorf("expected a valid identifier for the machine pool")
)
```

```
if !machinepool.MachinePoolKeyRE.MatchString(machinepoolID) {
			return InvalidMachinePoolIdentifier
		}
```

cmd/upgrade/machinepool/cmd.go
```diff
@@ -53,9 +54,14 @@ var Cmd = &cobra.Command{
 	Args: func(_ *cobra.Command, argv []string) error {
```
robpblake, line 54: @Mischulee How about you make the source of the `machinepool` name a `bool` on your `NewUpgradeArgsFunction`?  So something like:

```
func NewUpgradeArgsFunction(flagSource bool) func(cmd *cobra.Command, argv []string) error {
	return func(cmd *cobra.Command, argv []string) error {
		var machinepoolID string
		var err error

		if flagSource {
			flags := cmd.Flags()
			if machinepoolID, err = flags.GetString(machinePoolFlagName); err != nil {
				return err
			}
		} else {
			if len(argv) != 1 {
				return InvalidMachinePoolIdentifier
			}
			machinepoolID = argv[0]
		}

		if !machinepool.MachinePoolKeyRE.MatchString(machinepoolID) {
			return InvalidMachinePoolIdentifier
		}
		return nil
	}
}
```

That means we have a single function for all scenarios, and you can add a few more unit tests to verify the flow via the args as well as via the flag.

pkg/upgrade/upgrade.go
Mischulee: You are right, I added the second option.

pkg/upgrade/upgrade.go
Mischulee: Fixed.

pkg/upgrade/upgrade_test.go
Mischulee: Sorry I havent noticed, fixed :sweat_smile:

cmd/upgrade/machinepool/cmd.go
Mischulee: You are right, this looks good and now it will be more reusable. I added more unit tests for this.

robpblake:
Nice work @Mischulee 🚀

PR #2302, opened 2024-07-30
OCM-9984 | fix: Change replica (min, max, and normal) minimum to 0

hunterkepley:
* Changed the minimum for max/min/non-autoscaling replicas to `0`, rather than `-1`, when editing machinepools
* Updated validation unit tests to use `-1` instead of `0`


ciaranRoche (approved):
/lgtm

PR #2305, opened 2024-07-30
OCM-2394 | fix: Delete classic and hosted-cp account roles automatically if -y flag is set

Mischulee:
https://issues.redhat.com/browse/OCM-2394

### DIFF
--- a/cmd/dlt/accountroles/cmd.go
+++ b/cmd/dlt/accountroles/cmd.go
@@ -158,17 +158,6 @@ func run(cmd *cobra.Command, _ []string) {
 
 	if r.Creator.IsGovcloud {
 		deleteHostedCP = false
-	} else if interactive.Enabled() && !cmd.Flags().Changed("hosted-cp") && !cmd.Flags().Changed("classic") {
-		deleteHostedCP, err = interactive.GetBool(interactive.Input{
-			Question: "Delete hosted CP account roles",
-			Help:     cmd.Flags().Lookup("hosted-cp").Usage,
-			Default:  true,
-			Required: false,
-		})
-		if err != nil {
-			r.Reporter.Errorf("Expected a valid value: %s", err)
-			os.Exit(1)
-		}
 	}
 
 	if deleteHostedCP {


cmd/dlt/accountroles/cmd.go
```diff
@@ -158,17 +158,8 @@ func run(cmd *cobra.Command, _ []string) {
...
-		})
-		if err != nil {
-			r.Reporter.Errorf("Expected a valid value: %s", err)
-			os.Exit(1)
-		}
+	} else if !cmd.Flags().Changed("classic") {
+		deleteHostedCP = true
 	}
 
 	if deleteHostedCP {
```
robpblake, line 165: @Mischulee Interestingly, we're not actually taking the value of the `-y` flag into account. If the user omits the flag, it looks like we will proceed to delete the account-roles anyway. That doesn't feel right to me.

Inside the `if deleteHostedCP` and `if deleteClassic` blocks, we should check that the user has passed the `-y` flag. If no, they should be prompted to confirm the delete, if yes, they can be immediately deleted.

Essentially, it should look something like this:

```
if !confirm.Prompt(false, "Are you sure you want to delete all account roles for your ROSA Classic clusters?" ) {
		os.Exit(0)
}
```

With the prompt message changing depending on what's being deleted. WDYT?

Mischulee: @robpblake Hmm I not sure if I really like this. When `-y` flag is not set, it does not automatically delete all the roles, but it is asking user one role by one, to delete them, eg:

./rosa delete account-roles --prefix mparilov-test
? Account role deletion mode: auto
I: Deleting classic account roles
? Delete the account role 'mparilov-test-Installer-Role'? No
? Delete the account role 'mparilov-test-ControlPlane-Role'? No
...

If `-y`  is set, it deletes them automatically.

cmd/dlt/accountroles/cmd.go
```diff
@@ -158,17 +158,8 @@ func run(cmd *cobra.Command, _ []string) {
 
 	if r.Creator.IsGovcloud {
```
robpblake, line 159: @Mischulee I think we can remove the `else if` part of this statement. The `setDeleteRoles` function will return `true` for the assignment to `deletedHostedCP` if either:

- both `--classic` and `--hosted-cp` are omitted
- `--hosted-cp` is explicitly set to true

Otherwise, it should remain `false`.

Do you agree?

cmd/dlt/accountroles/cmd.go
robpblake: @Mischulee I've created https://issues.redhat.com/browse/OCM-10087 to implement the prompting for users when they want to delete account roles. Lets focus this PR on simply getting the `-y` flag to work for HCP clusters.

cmd/dlt/accountroles/cmd.go
Mischulee: @robpblake Yes, this probably makes the most sense. Fixed.

robpblake (approved):
/lgtm

PR #2307, opened 2024-07-31
OCM-9954 | fix: set correct default value for replicas in update mp

marcolan018:


ziccardi:
Although it's not directly related to this PR, the `getMachinePoolReplicas` method is quite lengthy, exceeding 100 lines of code. I recommend splitting the `if autoscaling ... else` sections into two separate helper functions.

pkg/machinepool/machinepool.go
```diff
@@ -1442,19 +1442,21 @@ func getMachinePoolReplicas(cmd *cobra.Command,
...
-			return
+		if interactive.Enabled() || !isReplicasSet && askForScalingParams {
+			replicas, err = interactive.GetInt(interactive.Input{
+				Question: "Replicas",
+				Help:     cmd.Flags().Lookup("replicas").Usage,
+				Default:  replicas,
+				Required: true,
+			})
+			if err != nil {
+				err = fmt.Errorf("Expected a valid number of replicas: %s", err)
```
robpblake, line 1457: @marcolan018 We can just `return` the `fmt.Errorf` directly e.g `return fmt.Errorf("Expected....")`

robpblake: Ignore me. I see we're using named returns from this function.

pkg/machinepool/machinepool.go
```diff
@@ -1442,19 +1442,21 @@ func getMachinePoolReplicas(cmd *cobra.Command,
 				return
 			}
 		}
-	} else if interactive.Enabled() || !isReplicasSet && askForScalingParams {
+	} else {
```
robpblake, line 1445: @marcolan018 I'd really like us to add a unit test for this. This is the type of regression we constantly run into, fix, but don't add a test for to prevent future regressions.

Could you add a unit test for the `getMachinePoolReplicas` function that ensures that if the user has not supplied `--replicas` flag on the CLI, that the returned `replicas` value from this function is what is currently set on the machinepool.

pkg/machinepool/machinepool.go
```diff
@@ -1442,19 +1442,21 @@ func getMachinePoolReplicas(cmd *cobra.Command,
...
+		if interactive.Enabled() || !isReplicasSet && askForScalingParams {
+			replicas, err = interactive.GetInt(interactive.Input{
+				Question: "Replicas",
+				Help:     cmd.Flags().Lookup("replicas").Usage,
+				Default:  replicas,
+				Required: true,
+			})
+			if err != nil {
+				err = fmt.Errorf("Expected a valid number of replicas: %s", err)
+				return
```
ziccardi, line 1458: We use naked return extensively. Naked returns should be used only for very simple functions since they make it hard to understand what we are returning.

Extract from the `go` website:
```
Naked return statements should be used only in short functions, as with the example shown here. They can harm readability in longer functions.
```
I suggest opening another PR to refactor this function or creating an issue to keep track of this.

marcolan018: @ziccardi 
thanks for the comments
I will create an issue to track code refactoring
- split the large method
- don't use naked return
for this pr, let it focus on fixing the blocker issue

pkg/machinepool/machinepool.go
marcolan018: @robpblake added the unit test for `getMachinePoolReplicas` func, please help to re-review

robpblake: Thanks @marcolan018

robpblake (approved):
/lgtm

marcolan018:
@den-rgb @davidleerh this pr needs to be cherry-pick to 1.2.43

PR #2312, opened 2024-07-31
OCM-10009 | feat: Adding taints effect validation

den-rgb:
Addressing issues for this ticket: https://issues.redhat.com/browse/OCM-10009

### DIFF
--- a/pkg/helper/machinepools/helpers.go
+++ b/pkg/helper/machinepools/helpers.go
@@ -3,6 +3,7 @@ package machinepools
 import (
 	"fmt"
 	"os"
+	"slices"
 	"strconv"
 	"strings"
 
@@ -30,6 +31,12 @@ const (
 	MaxNodeDrainTimeInMinutes = 10080
 	MaxNodeDrainTimeInHours   = 168
 )
+
+var allowedTaintEffects = []string{
+	"NoSchedule",
+	"NoExecute",
+	"PreferNoSchedule",
+}
 
 func MinNodePoolReplicaValidator(autoscaling bool) interactive.Validator {
 	return func(val interface{}) error {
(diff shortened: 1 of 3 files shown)


robpblake (changes requested):
@den-rgb Please add a unit test for the function. I've also added a comment about a small refactor to help with the readability of the `if` statement.

cheers,
Rob

pkg/helper/machinepools/helpers.go
```diff
@@ -159,6 +164,24 @@ func ParseTaints(taints string) ([]*cmv1.TaintBuilder, error) {
 	return taintBuilders, nil
 }
 
+func validateMachinePoolTaintEffect(taint string) error {
+	var allowedTaintEffects = []string{
```
robpblake, line 168: @den-rgb This can be package-scoped and not instantiated each time we call the function.

pkg/helper/machinepools/helpers.go
```diff
@@ -159,6 +164,24 @@ func ParseTaints(taints string) ([]*cmv1.TaintBuilder, error) {
...
+		"NoSchedule",
+		"NoExecute",
+		"PreferNoSchedule",
+	}
+	parts := strings.Split(taint, ":")
+	if len(parts) != 2 {
+		return fmt.Errorf("Invalid taint format: '%s'. Expected format is '<key>=<value>:<effect>'", taint)
+	}
+	effect := parts[1]
+	if effect != "" && !slices.Contains(allowedTaintEffects, effect) {
```
robpblake, line 178: @den-rgb Is an empty string a valid effect? This currently allows `""` to be valid. My expectation is that we can just simplify this check here to only `!slices.Contains(allowedTaintEffects, effect)` which would additionally prevent the user from passing an empty string, because it's not in the slice.

robpblake (approved):
/lgtm

PR #2319, opened 2024-08-05
OCM-10143 | test: fix ids:60688,59530,74761

yuwang-RH:
60688: the cluster version was not syncronized to ‘rosa describe cluster’ output, will enhance it to get the cluster version from the cluster config;
And it should skip to recover the role for HCP in AfterEach;
And change the domain-prefix in the dry-run command
59530: It should skip on NON-STS and NON-BYO-OIDC jobs, will fix later
74761: It failed as the region is not the aws default configured one, will fix by add region on `rosa delete oidc-config`

the log is here https://privatebin.corp.redhat.com/?45d42710aee1b787#A8cfxNR8vRxSUNneP4y7onD8vWM1sSvHbqTfpaxxrwXV

### DIFF
--- a/tests/e2e/test_rosacli_operator_roles.go
+++ b/tests/e2e/test_rosacli_operator_roles.go
@@ -30,6 +30,8 @@ var _ = Describe("Edit operator roles", labels.Feature.OperatorRoles, func() {
 		rosaClient             *rosacli.Client
 		ocmResourceService     rosacli.OCMResourceService
 		permissionsBoundaryArn string = "arn:aws:iam::aws:policy/AdministratorAccess"
+		clusterConfig          *config.ClusterConfig
+		err                    error
 	)
 	BeforeEach(func() {
 		By("Init the client")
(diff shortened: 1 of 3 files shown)


tests/e2e/test_rosacli_operator_roles.go
```diff
@@ -105,6 +105,7 @@ var _ = Describe("Edit operator roles", labels.Feature.OperatorRoles, func() {
 				output, err = ocmResourceService.DeleteOIDCConfig(
 					"--oidc-config-id", oidcConfigID,
 					"--mode", "auto",
+					"--region", profile.Region,
```
yasun1, line 108: Although we always provision a cluster using the test profile without overriding its settings, the profile is a template but not an actual configuration for a cluster. In this case, if I specify `Region` from the Prow CI job level, it will cause potential failure.  We should record the cluster configuration in `cluster-config` and use the file as a reference.

yuwang-RH:
@yasun1 Could you help review again? The re-run with the change of retrieving region from cluster config is passed.

yasun1:
@yuwang-RH  Could you check `region` is recorded in `cluster-config` in the running jobs or the previous one?

PR #2322, opened 2024-08-05
OCM-10144 | fix: List upgrade for cluster asking machinepool identifier producing error

Mischulee:
https://issues.redhat.com/browse/OCM-10141

### DIFF
--- a/pkg/upgrade/upgrade.go
+++ b/pkg/upgrade/upgrade.go
@@ -21,6 +21,9 @@ func NewUpgradeArgsFunction(flagSource bool) func(cmd *cobra.Command, argv []str
 
 		if flagSource {
 			flags := cmd.Flags()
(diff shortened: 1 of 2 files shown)


pkg/upgrade/upgrade.go
```diff
@@ -21,6 +21,9 @@ func NewUpgradeArgsFunction(flagSource bool) func(cmd *cobra.Command, argv []str
 
 		if flagSource {
 			flags := cmd.Flags()
+			if !cmd.Flags().Changed(machinepoolFlagName) {
```
robpblake, line 24: @Mischulee Could you add to our existing unit test to verify that if the flag has not been changed, we do not perform the validation.

cheers!

Rob

Mischulee: Added!

robpblake (approved):
/lgtm

PR #2325, opened 2024-08-06
OCM-9805 | test:  Automate id:60278 [Scaling up/down node pool on rosa hcp cluster]

86254860:
Note: 
1. It doesn't check whether node pool instances ready **in default**. And if needed. pls set ENV `export CLUSTER_NODE_POOL_GLOBAL_CHECK=true`, then it will wait and check whether node pool instances ready until timeout.

### DIFF
--- a/tests/e2e/hcp_machine_pool_test.go
+++ b/tests/e2e/hcp_machine_pool_test.go
@@ -1,6 +1,7 @@
 package e2e
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 
@@ -10,17 +11,25 @@ import (
 
 	"github.com/openshift/rosa/tests/ci/labels"
 	"github.com/openshift/rosa/tests/utils/common"
+	"github.com/openshift/rosa/tests/utils/common/constants"
+	"github.com/openshift/rosa/tests/utils/config"
 	"github.com/openshift/rosa/tests/utils/exec/rosacli"
 	ph "github.com/openshift/rosa/tests/utils/profilehandler"
 )
 
-var _ = Describe("Create Machine Pool", labels.Feature.Machinepool, func() {
+var _ = Describe("HCP Machine Pool", labels.Feature.Machinepool, func() {
+	// It doesn't check whether node pool instances ready in default.
+	// If needed for verify hcp node pool's changes, pls set the ENV CLUSTER_NODE_POOL_GLOBAL_CHECK to true,
+	// which will wait for node pool instances ready until timeout.
+	isNodePoolGlobalCheck := config.IsNodePoolGlobalCheck()
 
 	var (
 		rosaClient         *rosacli.Client
 		machinePoolService rosacli.MachinePoolService
 		profile            *ph.Profile
+		isMultiArch        bool
 	)
+
 	BeforeEach(func() {
 		Expect(clusterID).ToNot(BeEmpty(), "Cluster ID is empty, please export the env variable CLUSTER_ID")
 		rosaClient = rosacli.NewClient()
@@ -30,15 +39,19 @@ var _ = Describe("Create Machine Pool", labels.Feature.Machinepool, func() {
 		hostedCluster, err := rosaClient.Cluster.IsHostedCPCluster(clusterID)
 		Expect(err).ToNot(HaveOccurred())
 
+		By("Check whether the cluster is multi arch")
+		isMultiArch, err = rosaClient.Cluster.IsMultiArch(clusterID)
+		Expect(err).ToNot(HaveOccurred())
+
 		profile = ph.LoadProfileYamlFileByENV()
 
 		if !hostedCluster {
 			SkipNotHosted()
 		}
 	})
-	It("to hosted cluster with additional security group IDs will work [id:72195]",
-		labels.Critical, labels.Runtime.Day2,
-		func() {
+
+	Describe("Create/delete/view a machine pool", func() {
+		It("should succeed with additional security group IDs [id:72195]", labels.Critical, labels.Runtime.Day2, func() {
 			By("Load the vpc client of the machinepool")
 			mps, err := rosaClient.MachinePool.ListAndReflectNodePools(clusterID)
 			Expect(err).ToNot(HaveOccurred())
@@ -48,7 +61,9 @@ var _ = Describe("Create Machine Pool", labels.Feature.Machinepool, func() {
 			Expect(err).ToNot(HaveOccurred())
 
 			By("Prepare security groups")
-			sgIDs, err := vpcClient.CreateAdditionalSecurityGroups(3, "72195", "testing for case 72195")
+			// Configure with a random str, which can solve the rerun failure
+			sgPrefix := common.GenerateRandomName("72195", 2)
+			sgIDs, err := vpcClient.CreateAdditionalSecurityGroups(3, sgPrefix, "testing for case 72195")
 			Expect(err).ToNot(HaveOccurred())
 
 			By("Create machinepool with security groups set")
@@ -80,6 +95,7 @@ var _ = Describe("Create Machine Pool", labels.Feature.Machinepool, func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(mpDescription.AdditionalSecurityGroupIDs).To(BeEmpty())
 		})
+	})
 
 	It("machinepool AWS preflight tag validation[id:73638]",
 		labels.Medium, labels.Runtime.Day2,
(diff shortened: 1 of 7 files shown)


tests/utils/common/constants/cluster.go
```diff
@@ -57,4 +57,14 @@ var (
...
 	DefaultInstanceType = "m5.xlarge"
+
+	M5XLarge  = "m5.xlarge"
+	M52XLarge = "m5.2xlarge"
+	M6gXLarge = "m6g.xlarge"
+)
+
+// cpu architecture
+const (
+	AMD = "amd"
```
radtriste, line 68: We should use `amd64` and `arm64` instead

tests/utils/config/cluster.go
```diff
@@ -174,3 +174,13 @@ func GetClusterID() (clusterID string) {
...
 }
+
+// GetNodePoolGlobalCheck Get the nodepool global check flag
+func GetNodePoolGlobalCheck() bool {
+	nodePoolGlobalCheck := os.Getenv("CLUSTER_NODE_POOL_GLOBAL_CHECK")
+	if nodePoolGlobalCheck == "" || nodePoolGlobalCheck == "false" {
+		return false
+	} else {
+		return true
+	}
```
radtriste, line 185: Why not ?

```suggestion
	if nodePoolGlobalCheck == "true" {
		return true
	} 
	return false
```

tests/utils/config/cluster.go
```diff
@@ -174,3 +174,13 @@ func GetClusterID() (clusterID string) {
 func getClusterIDENVExisted() string {
 	return os.Getenv("CLUSTER_ID")
 }
+
+// GetNodePoolGlobalCheck Get the nodepool global check flag
+func GetNodePoolGlobalCheck() bool {
```
radtriste, line 179: Rename to `IsNodePoolGlobalCheck` as it returns a bool

tests/e2e/utils.go
```diff
@@ -19,3 +19,8 @@ func SkipTestOnFeature(feature string) {
 	message := fmt.Sprintf("The test profile is not configured with the feature: %s", feature)
 	Skip(message)
 }
+
+func SkipNotMultiArc() {
```
radtriste, line 23: `SkipNotMultiArch`

tests/e2e/utils.go
```diff
@@ -19,3 +19,8 @@ func SkipTestOnFeature(feature string) {
...
+	message := fmt.Sprintln("The cluster is not multi architecture")
```
radtriste, line 24:
```suggestion
	message := fmt.Sprintln("The cluster can not handle multiple CPU architecture")
```

tests/e2e/hcp_machine_pool_test.go
```diff
@@ -1,23 +1,132 @@
...
 )
 
-var _ = Describe("Create Machine Pool", labels.Feature.Machinepool, func() {
+var _ = Describe("HCP Machine Pool", labels.Feature.Machinepool, func() {
+
+	const (
+		// The timeout may depend on the resource
+		NodePoolCheckTimeout = 30 * time.Minute
+		NodePoolCheckPoll    = 10 * time.Second
+	)
```
radtriste, line 27: Those could be into the `utils/common/constants` package

tests/e2e/hcp_machine_pool_test.go
```diff
@@ -1,23 +1,132 @@
...
 
 	var (
-		rosaClient *rosacli.Client
-		profile    *ph.Profile
+		rosaClient  *rosacli.Client
+		profile     *ph.Profile
+		isMultiArch bool
 	)
+
+	getAutoScaleReplicas := func(mpName string) (error, map[string]int) {
```
radtriste, line 40: usually the error is the last return type

tests/e2e/hcp_machine_pool_test.go
```diff
@@ -1,23 +1,132 @@
...
+		} else {
+			mpDesc, err := rosaClient.MachinePool.DescribeAndReflectNodePool(clusterID, mpName)
+			if err != nil {
+				return false
+			}
+			return mpDesc.CurrentReplicas == fmt.Sprintf("%v", mpDesc.DesiredReplicas)
+		}
+	}
+
+	scaleNodePool := func(mpName string, scaleFlag string, updateReplicas int, isCheck bool) {
```
radtriste, line 82: `isCheck` -> `waitForNPInstancesReady`

tests/e2e/hcp_machine_pool_test.go
```diff
@@ -26,54 +135,241 @@ var _ = Describe("Create Machine Pool", labels.Feature.Machinepool, func() {
 		hostedCluster, err := rosaClient.Cluster.IsHostedCPCluster(clusterID)
 		Expect(err).ToNot(HaveOccurred())
 
+		By("Check the cluster if multi arch")
```
radtriste, line 138:
```suggestion
		By("Check whether the cluster is multi arch")
```

tests/utils/common/constants/cluster.go
86254860: OK

tests/utils/config/cluster.go
86254860: Good point.

tests/e2e/hcp_machine_pool_test.go
```diff
@@ -1,23 +1,126 @@
...
 
 	var (
-		rosaClient *rosacli.Client
-		profile    *ph.Profile
+		rosaClient  *rosacli.Client
+		profile     *ph.Profile
+		isMultiArch bool
 	)
+
+	getAutoScaleReplicas := func(mpName string) (map[string]int, error) {
```
radtriste, line 32: I think that those methods could be part of the machine pool service interface. wdyt ?
Methods should return error and the error would be check in tests.
They could be reused in other tests as well if needed. And that way we don't overload too much the test files

radtriste (commented):
One final comment on code structure

tests/e2e/hcp_machine_pool_test.go
86254860: Yeah. Good catch. I'll adjust `getAutoScaleReplicas` to  the interface `MachinePoolService`.

tests/e2e/hcp_machine_pool_test.go
```diff
@@ -26,54 +103,244 @@ var _ = Describe("Create Machine Pool", labels.Feature.Machinepool, func() {
...
+			Expect(err.Error()).Should(ContainSubstring(expectedErrMsg))
+		},
+
+		Entry("Scale replica too large [id:60278]", constants.M52XLarge, "1000", "exceeds the maximum allowed"),
+		Entry("Scale replica to -1 [id:60278]", constants.M52XLarge, "-1", "Replicas must be a non-negative number"),
+		Entry("Scale replica to a char [id:60278]", constants.M52XLarge, "a", "invalid syntax"),
+	)
+
+	Describe("Scale up/down a machine pool enabling autoscale", func() {
+		When("with the valid min replica and max replica", func() {
```
xueli181114, line 243: May I know why we have `When` here? If not too much benefits we can get from it or not too much required in the code, let's make code simple to read.

tests/e2e/hcp_machine_pool_test.go
```diff
@@ -26,54 +103,244 @@ var _ = Describe("Create Machine Pool", labels.Feature.Machinepool, func() {
...
+					"--instance-type", instanceType,
+					"--enable-autoscaling",
+					"--min-replicas", fmt.Sprintf("%v", minReplica),
+					"--max-replicas", fmt.Sprintf("%v", maxReplica),
+					"-y",
+				)
+				Expect(err).ToNot(HaveOccurred())
+				defer rosaClient.MachinePool.DeleteMachinePool(clusterID, mpName)
+
+				// TODO There's issue here https://issues.redhat.com/browse/OCM-10147
```
xueli181114, line 262: We don't include OCM jira issue in code.

tests/e2e/hcp_machine_pool_test.go
```diff
@@ -26,54 +103,244 @@ var _ = Describe("Create Machine Pool", labels.Feature.Machinepool, func() {
...
+				//By("Scale a machine pool with unchanged parameters")
+				//scaleAutoScaledNodePool(mpName, "unchanged", minReplica, maxReplica, true)
+
+				By("Scale up a machine pool replicas from 1~3 to 2~5")
+				upMinReplica := 2
+				upMaxReplica := 5
+				scaleAutoScaledNodePool(mpName, "up", upMinReplica, upMaxReplica, true)
+
+				// Don't check the current replicas when scale down, because it won't change after reducing min_replica
```
xueli181114, line 271: Is it a bug?

tests/e2e/hcp_machine_pool_test.go
```diff
@@ -1,23 +1,100 @@
...
+			"-y",
+		)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Check the machinepool replicas after scale " + scaleFlag)
+		mpDesc, err := rosaClient.MachinePool.DescribeAndReflectNodePool(clusterID, mpName)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(mpDesc.DesiredReplicas).Should(Equal(updateReplicas))
+
+		if waitForNPInstancesReady && isNodePoolGlobalCheck {
```
xueli181114, line 60: We don't wait for instances ready in rosacli cases. Please put this part to backend. It's not related to rosacli if the instance type can work to the nodepool or not. Just make sure the rosacli is acting well with backend

tests/e2e/hcp_machine_pool_test.go
86254860: Because there are many conditions under the same `Describe("Scale up/down a machine pool enabling autoscale"`, so use `When` to distinguish different sub-scenes.
```
Describe("Scale up/down a machine pool enabling autoscale"...
    When("with the valid min replica and max replica"...
    When("with the invalid min replica and max replica"...

```

tests/e2e/hcp_machine_pool_test.go
86254860: OK, I'll replace the link with jira issue ID, which I can know here still has some issue.

tests/e2e/hcp_machine_pool_test.go
86254860: I confirmed with DEV adecorte. He said it's as expected.

tests/e2e/hcp_machine_pool_test.go
86254860: Hmm, I indeed added it for `rosacli cases`. For an extreme example, if the nodes are ready from CMS, but it may  show wrong `Current Replicas` or `Desired Replicas` due to `rosacli issue`, so I added this logic, but it won't check in default. It only checks when setting the ENV `export CLUSTER_NODE_POOL_GLOBAL_CHECK=true` if needed.

xueli181114: If that is the checkpoint let's just make sure `Desired Replicas= <nodes in backend>` and at the same time `Current Replicas=<desired nodes in the response body>`

tests/e2e/hcp_machine_pool_test.go
xueli181114: Why not include the case ID in the jira card of the bug? So when you verify the bug you can update the script accordingly to the fixing.

tests/e2e/hcp_machine_pool_test.go
xueli181114: If it is expected, no need to clarify here. Let's clarify in the case in Polarion.

xueli181114: `It` spec already indicated the testing purpose. Too many layers will make the code hard to maintain.

PR #2327, opened 2024-08-06
OCM-9974 | fix: Change options validations to allow 0 min replicas as well

hunterkepley:
In the fix for [OCM-9974](https://issues.redhat.com//browse/OCM-9974), I had forgotten to also fix the options validations. This change is included here + updated tests

### DIFF
--- a/cmd/edit/machinepool/options.go
+++ b/cmd/edit/machinepool/options.go
@@ -64,11 +64,11 @@ func (m *EditMachinepoolOptions) Bind(args *EditMachinepoolUserOptions, argv []s
 	}
 
 	if m.args.autoscalingEnabled {
-		if m.args.minReplicas <= 0 {
-			return fmt.Errorf("Min replicas must be greater than zero when autoscaling is enabled")
+		if m.args.minReplicas < 0 {
+			return fmt.Errorf("Min replicas must be a number that is 0 or greater when autoscaling is enabled")
 		}
-		if m.args.maxReplicas <= 0 {
-			return fmt.Errorf("Max replicas must be greater than zero when autoscaling is enabled")
+		if m.args.maxReplicas < 0 {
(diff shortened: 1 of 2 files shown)


cmd/edit/machinepool/options.go
```diff
@@ -64,11 +64,11 @@ func (m *EditMachinepoolOptions) Bind(args *EditMachinepoolUserOptions, argv []s
 	}
 
 	if m.args.autoscalingEnabled {
-		if m.args.minReplicas <= 0 {
-			return fmt.Errorf("Min replicas must be greater than zero when autoscaling is enabled")
+		if m.args.minReplicas < 0 {
+			return fmt.Errorf("Min replicas must be a positive number when autoscaling is enabled")
```
davidleerh, line 68: should it be worded, "must be a number that is 0 or greater?"

lucasponce:
Is this fix for both ROSA Classic and HCP ? or only for Classic?

den-rgb:
> Is this fix for both ROSA Classic and HCP ? or only for Classic?

It's for both

robpblake (approved):
/approve

lucasponce:
But ROSA HCP doesn't support min-replicas to 0, it will fail from backend, which it could be ok

hunterkepley:
@lucasponce Is there a list of differences for HCP and classic machinepool replica count somewhere? It seems QE and engineering have a disconnect on this, which has caused back and forth changes which could have been done in one change

lucasponce:
@hunterkepley the available documentation is typically high-level, and some of the details are part of the API itself.

On this case, this is a limitation/feature from the Hypershift API one, in the case of Cluster Service, this is available in the API tests.

https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/cmd/clusters-service/servecmd/apiserver/node_pool_handlers_test.go?ref_type=heads#L485

PR #2335, opened 2024-08-08
OCM-9776 | test: Automate OCP-55883 Upgrade STS rosa cluster across Y stream via ROSA cli

jameszwang:


tests/e2e/test_rosacli_upgrade.go
```diff
@@ -449,6 +504,25 @@ var _ = Describe("Describe/List rosa upgrade",
 			})
 	})
 
+func FindUpgradeVersion(v rosacli.VersionService, channelGroup string, hostedCP bool, clusterVersion string) (string,
```
yasun1, line 507: IMO, we should do the Y-stream upgrading in the process y->y+1->y+2. So we should find the latest y+1 version.

tests/e2e/test_rosacli_upgrade.go
```diff
@@ -314,6 +309,66 @@ var _ = Describe("Cluster Upgrade testing",
...
+					"-y",
+				)
+				Expect(err).To(BeNil())
+				Expect(output.String()).To(ContainSubstring("Upgrade successfully scheduled for cluster"))
+			}
+
+			By("Check upgrade state")
+			err = WaitForUpgradeToState(upgradeService, clusterID, con.Scheduled, 4)
+			Expect(err).To(BeNil())
+			err = WaitForUpgradeToState(upgradeService, clusterID, con.Started, 70)
```
yasun1, line 369: Do we not wait for the cluster upgrading to be finished?

tests/e2e/test_rosacli_upgrade.go
jameszwang: Update PR according to our discussion, sorted upgrade path versions.

tests/e2e/test_rosacli_upgrade.go
jameszwang: According to our discussion, we verify the upgrade state currently and will enhance this part in the future.

tests/e2e/test_rosacli_upgrade.go
```diff
@@ -449,6 +507,81 @@ var _ = Describe("Describe/List rosa upgrade",
...
+
+func FindUpperYStreamVersion(v rosacli.VersionService, channelGroup string, hostedCP bool,
+	clusterVersion string) (string, string, error) {
+	clusterVersionList, err := v.ListAndReflectVersions(channelGroup, hostedCP)
+	if err != nil {
+		return "", "", err
+	}
+	versions, err := clusterVersionList.FindYStreamUpgradeVersions(clusterVersion)
+	// Sorted version from high to low
+	sortedVersions := SortVersions(versions)
```
yasun1, line 571: This line should be put after 574.

tests/e2e/test_rosacli_upgrade.go
```diff
@@ -449,6 +507,81 @@ var _ = Describe("Describe/List rosa upgrade",
...
+	for i, v := range versions {
+		parsedVersions[i] = ParseVersion(v)
+	}
+	sort.Sort(sort.Reverse(VersionSlice(parsedVersions)))
+	sortedVersions := make([]string, len(versions))
+	for i, v := range parsedVersions {
+		sortedVersions[i] = v.Original
+	}
+	return sortedVersions
+}
```
yasun1, line 561: We have had `sort` functions https://github.com/openshift/rosa/blob/master/tests/utils/exec/rosacli/version_service.go#L221 for this. You just need to use it directly in `FindUpperYStreamVersion`.

PR #2340, opened 2024-08-09
OCM-9704 | feat: persist refreshed tokens

tylercreller:
Modifies the behavior for token management in ROSA
* **Before**: When the initial access token is expired, new access tokens would not be persisted to the config, ROSA would use the refresh token to get a new access token on every request
* **Now**: New tokens are always persisted to the OCM configuration. This will allow the SDK to use valid tokens more often and prevent additional round-trips to SSO
* **On log polling**: On every callback request from getting updated logs during a watch, tokens will now be refreshed. This is required for environments with short-lived refresh tokens such as FedRAMP.

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -85,7 +85,7 @@ var _ = Describe("Config", Ordered, func() {
 
 		BeforeAll(func() {
 			tmpdir, err = os.MkdirTemp("/tmp", ".ocm-config-*")
-			Expect(err).To(BeNil())
+			Expect(err).NotTo(HaveOccurred())
 			os.Setenv("OCM_CONFIG", tmpdir+"/ocm_config.json")
 		})
 
@@ -101,7 +101,7 @@ var _ = Describe("Config", Ordered, func() {
 			Save(cfg)
 
 			myconf, err := Load()
-			Expect(err).To(BeNil())
+			Expect(err).NotTo(HaveOccurred())
 			Expect(myconf.URL).To(Equal(url))
 		})
 	})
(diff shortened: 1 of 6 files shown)


pkg/config/config_test.go
```diff
@@ -121,6 +121,41 @@ var _ = Describe("Config", Ordered, func() {
...
 		})
 	})
+
+	When("Persisting tokens", Ordered, func() {
+		var tmpdir string
+		var err error
+
+		BeforeAll(func() {
+			tmpdir, err = os.MkdirTemp("/tmp", ".ocm-config-*")
+			Expect(err).To(BeNil())
```
hunterkepley, line 131: NP: we've been moving towards `Expect(err).ToNot(HaveOccurred())`

hunterkepley (commented):
Overall LGTM- one nitpick, and i would suggest some negative test cases but the errors that can occur seem to come from other funcs already made, or, from nil checks. 

For sanity reasons, may be good to provide nil values and make sure things are fine

pkg/config/config_test.go
tylercreller: updated, thanks

tylercreller:
> Overall LGTM- one nitpick, and i would suggest some negative test cases but the errors that can occur seem to come from other funcs already made, or, from nil checks.
> 
> For sanity reasons, may be good to provide nil values and make sure things are fine

Fair enough! Added some additional coverage around invalid sdk connection and failure to talk to SSO.

ciaranRoche (approved):
/lgtm

PR #2342, opened 2024-08-09
OCM-6419 | feat: Move fetching thumbprint from ROSA to backend

hunterkepley:
https://issues.redhat.com/browse/OCM-6419

Replaces ROSA based fetching for thumbprint with backend (CS) API calls to circumvent inability to fetch thumbprints when users use a proxy

### DIFF
--- a/cmd/create/oidcprovider/cmd.go
+++ b/cmd/create/oidcprovider/cmd.go
@@ -21,7 +21,6 @@ import (
 	"os"
 	"strings"
 
-	"github.com/openshift-online/ocm-common/pkg/rosa/oidcconfigs"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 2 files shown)


cmd/create/oidcprovider/cmd.go
```diff
@@ -190,7 +190,7 @@ func run(cmd *cobra.Command, argv []string) {
 		if !confirm.Prompt(true, confirmPromptMessage) {
 			os.Exit(0)
 		}
-		err = createProvider(r, oidcEndpointURL, clusterId)
+		err = createProvider(r, oidcEndpointURL, args.oidcConfigId, clusterId)
```
marcolan018, line 193: is it possible args.oidcConfigId equals to ""?

hunterkepley: If it does, then it requires clusterId, if both are `""` - the api returns a proper bad request error

cmd/create/oidcprovider/cmd.go
```diff
@@ -226,14 +226,18 @@ func run(cmd *cobra.Command, argv []string) {
 	}
 }
 
-func createProvider(r *rosa.Runtime, oidcEndpointUrl string, clusterId string) error {
-	thumbprint, err := oidcconfigs.FetchThumbprint(oidcEndpointUrl)
```
marcolan018, line 230: I found this func() is still used I manual mode, should we also change this path?

hunterkepley: Thanks, fixed

cmd/create/oidcprovider/cmd.go
hunterkepley: Looking at other places where an oidc config id is needed in this command, all use `args`

PR #2346, opened 2024-08-12
OCM-10375 | fix: update edit machinepool error message for missing identifier

Mischulee:
https://issues.redhat.com/browse/OCM-10375

### DIFF
--- a/cmd/edit/machinepool/cmd.go
+++ b/cmd/edit/machinepool/cmd.go
@@ -52,7 +52,7 @@ func NewEditMachinePoolCommand() *cobra.Command {
 		Long:    long,
 		Aliases: aliases,
 		Example: example,
-		Args:    cobra.MinimumNArgs(1),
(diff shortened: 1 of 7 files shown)


cmd/edit/machinepool/cmd.go
```diff
@@ -52,7 +53,7 @@ func NewEditMachinePoolCommand() *cobra.Command {
 		Long:    long,
 		Aliases: aliases,
 		Example: example,
-		Args:    cobra.MinimumNArgs(1),
+		Args:    upgrade.NewUpgradeArgsFunction(false),
```
andreadecorte, line 56: as it is used also in a non upgrade scenario, to avoid confusion I would suggest to rename it to `NewMachinepoolArgsFunction`and move it somewhere under pkg/machinepool

Mischulee: Good point. I renamed the function and moved it under `pkg/machinepool`.

andreadecorte:
lgtm

PR #2350, opened 2024-08-13
OCM-10374 | chore: Bump ocm-sdk-go to v0.1.436

mnecas:

PR #2351, opened 2024-08-13
OCM-10374 | fix: Add Limited Support Reason Override support

mnecas:
Right now we have new override parameters in the LSR so we can keep the LSR visible to the SRE but it can be overriden when there is support exception. The cluster status shows only active LSR, without the override.

Depends on: https://github.com/openshift/rosa/pull/2350

### DIFF
--- a/pkg/ocm/limited_support.go
+++ b/pkg/ocm/limited_support.go
@@ -12,9 +12,11 @@ func (c *Client) GetLimitedSupportReasons(clusterID string) (
 		LimitedSupportReasons()
 	page := 1
 	size := 100
+	search := "override_enabled='f'"
 	for {
 		response, err := collection.List().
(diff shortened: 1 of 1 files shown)


mnecas:
Moving to draft as it is depends on: https://github.com/openshift/rosa/pull/2350

gdbranco:
If the reason is overridden should we expose to customer at all? Would he care if it shouldn't affect the cluster?

mnecas:
I have updated the MR to only show the active limited support reasons

tzvatot:
LGTM

mnecas:
As this change does not require the SDK bump, removing the draft.

mnecas:
@gdbranco could you ptal once more/ merge it if possible?

PR #2353, opened 2024-08-14
OCM-9773 | test: Automate id:45745

jameszwang:


tests/e2e/test_rosacli_operator_roles.go
```diff
@@ -1080,3 +1080,144 @@ var _ = Describe("Create cluster with oprator roles which are attaching managed
...
+				Version:      "y-1",
+				ChannelGroup: "candidate",
+				Region:       "us-west-2",
+			}
+			customProfile.NamePrefix = constants.DefaultNamePrefix
+		})
+
+		AfterEach(func() {
+			if clusterID != "" {
+				By("Delete cluster")
```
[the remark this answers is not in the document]
jameszwang: Thanks for the comment.The **By** here is just a description of test step; the cluster deletion operation is already existed in the clusterService with name **DeleteCluster**.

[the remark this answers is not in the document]
86254860: Hmm, what I mean is that can we abstract the logic from Ln 1119~1140 to a common function, and invoke it through `defer <function>` in each case not in "AfterEach". Then other cases, e.g. I have a scenario which need to create and delete cluster, then I won't need to rewrite the same function in my AfterEach. If it can be a function of ClusterService, then I can invoke it in my case.

tests/e2e/test_rosacli_operator_roles.go
```diff
@@ -1080,3 +1080,144 @@ var _ = Describe("Create cluster with oprator roles which are attaching managed
...
+				},
+				Version:      "y-1",
+				ChannelGroup: "candidate",
+				Region:       "us-west-2",
+			}
+			customProfile.NamePrefix = constants.DefaultNamePrefix
+		})
+
+		AfterEach(func() {
+			if clusterID != "" {
```
86254860, line 1119: Hmm, what I mean is that can we abstract the logic from Ln 1119~1140 to a common function, and invoke it through defer <function> in each case not in "AfterEach". Then other cases, e.g. I have a scenario which need to create and delete cluster, then I won't need to rewrite the same function in my AfterEach. If it can be a function of ClusterService, then I can invoke it in my case.

jameszwang: As we discussed earlier, use **DestroyResourceByProfile** function to destroy prepared user data.

tests/e2e/test_rosacli_operator_roles.go
```diff
@@ -1080,3 +1080,132 @@ var _ = Describe("Create cluster with oprator roles which are attaching managed
...
+				err = clusterService.WaitClusterDeleted(clusterID, 3, 30)
+				Expect(err).To(BeNil())
+			}
+
+			By("Clean resource")
+			errs := profilehandler.DestroyResourceByProfile(customProfile, rosaClient)
+			Expect(len(errs)).To(Equal(0))
+		})
+
+		It("to create and upgrade operator roles in auto mode - [id:45745]",
```
yuwang-RH, line 1135: OCP-45745 on Polarion is described as the scenario that update account roles and operator-roles then upgrade cluster.
In you automation bellow, it seems missing the step to upgrade cluster after the roles upgrade.
And this scenarios we need to the account-roles and operator-roles versions are Y-1 before the cluster upgrade, that's the real customers scenario.

jameszwang: updated according to comments.

PR #2355, opened 2024-08-14
OCM-10221 | fix: forbid to create hcp cluster without billing account

Mischulee:
https://issues.redhat.com/browse/OCM-10221


cmd/create/cluster/cmd.go
```diff
@@ -1128,6 +1128,7 @@ func run(cmd *cobra.Command, _ []string) {
 					)
 				} else {
 					r.Reporter.Errorf("A billing account is required for Hosted Control Plane clusters. %s", listBillingAccountMessage)
+					os.Exit(1)
```
gdbranco, line 1131: Could you refactor to capture this validation within a helper function and add a small unit test around just the function?

Mischulee: Sure, I created helper function for providing billing account and added unit tests for it.

PR #2356, opened 2024-08-15
OCM-10455 | feat: Add a configuration for a custom user agent

devguyio:
This is needed by arbitrary services which use the client package to communicate with ocm / rosa.

### DIFF
--- a/pkg/ocm/client.go
+++ b/pkg/ocm/client.go
@@ -120,10 +120,14 @@ func (b *ClientBuilder) Build() (result *Client, err error) {
 	builder.Logger(logger)
 
 	userAgent := info.DefaultUserAgent
+	version := info.DefaultVersion
 	if b.cfg.UserAgent != "" {
 		userAgent = b.cfg.UserAgent
 	}
-	builder.Agent(userAgent + "/" + info.Version + " " + sdk.DefaultAgent)
+	if b.cfg.Version != "" {
+		version = b.cfg.Version
+	}
+	builder.Agent(userAgent + "/" + version + " " + sdk.DefaultAgent)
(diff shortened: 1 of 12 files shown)


pkg/ocm/client.go
```diff
@@ -118,7 +118,12 @@ func (b *ClientBuilder) Build() (result *Client, err error) {
 	// values in the configuration, so that default values won't be overridden:
 	builder := sdk.NewConnectionBuilder()
 	builder.Logger(logger)
-	builder.Agent(info.UserAgent + "/" + info.Version + " " + sdk.DefaultAgent)
+
+	userAgent := info.DefaultUserAgent
+	if b.cfg.UserAgent != "" {
```
den-rgb, line 123: shouldn't this be checking if it is empty?

devguyio: I'm setting the `userAgent` variable to the default value, then in line 123 I'm checking if the config has a non-empty UseAgent, I set the `userAgent` value to the provided one instead of the default one.

pkg/ocm/client.go
```diff
@@ -118,7 +118,12 @@ func (b *ClientBuilder) Build() (result *Client, err error) {
...
+		userAgent = b.cfg.UserAgent
+	}
+	builder.Agent(userAgent + "/" + info.Version + " " + sdk.DefaultAgent)
```
gdbranco, line 126: In your custom agent will you also be supplying capa version/release?

devguyio: Initially this wasn't part of the requirements, but your comment made think more of it and I think it makes sense to be added as well. Pushed a new commit. Thanks!

PR #2357, opened 2024-08-16
OCM-10438 | ci: Enhance account-roles list due to a known issue

yuwang-RH:
[OCM-10438](https://issues.redhat.com//browse/OCM-10438) | ci: Enhance account-roles list due to a known issue

The cluster preparation step failed sometime as it fails `rosa list account-roles` as some irrelevant roles to the testing is deleted during the query. To make sure the cluster preparation can pass, add 3 try times when that issue happens

The testing with profile which needs to create account-roles is passed on local with this change.

### DIFF
--- a/tests/utils/profilehandler/data_preparation.go
+++ b/tests/utils/profilehandler/data_preparation.go
@@ -254,14 +254,22 @@ func PrepareAccountRoles(client *rosacli.Client,
 	if err != nil {
 		return
 	}
-	accRoleList, output, err := client.OCMResource.ListAccountRole()
-	if err != nil {
-		err = fmt.Errorf("error happens when list account-roles, %s", output.String())
-		return
+	var accoutRoles *rosacli.AccountRolesUnit
+	maxRetries := 3
+	for i := 0; i < maxRetries; i++ {
+		accRoleList, output, err := client.OCMResource.ListAccountRole()
+		if err != nil && strings.Contains(err.Error(), "cannot be found") {
+			log.Logger.Infof("Some IAM role cannot be found, Retrying... (%d/%d)\n", i+1, maxRetries)
+			continue
+		} else if err != nil && !strings.Contains(err.Error(), "cannot be found") {
+			err = fmt.Errorf("error happens when list account-roles, %s", output.String())
+			return &rosacli.AccountRolesUnit{}, err
+		} else {
+			accoutRoles = accRoleList.DigAccountRoles(namePrefix, hcp)
+			break
+		}
 	}
-	roleDig := accRoleList.DigAccountRoles(namePrefix, hcp)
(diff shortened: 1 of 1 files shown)


tests/utils/profilehandler/data_preparation.go
```diff
@@ -254,14 +254,25 @@ func PrepareAccountRoles(client *rosacli.Client,
...
-	accRoleList, output, err := client.OCMResource.ListAccountRole()
-	if err != nil {
-		err = fmt.Errorf("error happens when list account-roles, %s", output.String())
-		return
+	var accoutRoles *rosacli.AccountRolesUnit
+	var interErr error
+	maxRetries := 3
+	for i := 0; i < maxRetries; i++ {
+		accRoleList, output, err := client.OCMResource.ListAccountRole()
+		if err != nil {
```
xueli181114, line 262:
```suggestion
		if err != nil &&strings.Contains(err.Error(), "cannot be found"){
```
This can reduce the `if`/`else` in code

tests/utils/profilehandler/data_preparation.go
```diff
@@ -254,14 +254,25 @@ func PrepareAccountRoles(client *rosacli.Client,
 	if err != nil {
 		return
 	}
-	accRoleList, output, err := client.OCMResource.ListAccountRole()
-	if err != nil {
-		err = fmt.Errorf("error happens when list account-roles, %s", output.String())
-		return
+	var accoutRoles *rosacli.AccountRolesUnit
+	var interErr error
```
xueli181114, line 258: `interErr` is not used. Remove it

tests/utils/profilehandler/data_preparation.go
```diff
@@ -254,14 +254,25 @@ func PrepareAccountRoles(client *rosacli.Client,
...
+			}
+		} else {
+			accoutRoles = accRoleList.DigAccountRoles(namePrefix, hcp)
+			break
+		}
 	}
-	roleDig := accRoleList.DigAccountRoles(namePrefix, hcp)
-
-	return roleDig, nil
+	return accoutRoles, interErr
```
xueli181114, line 275:
```suggestion
	return accoutRoles, nil
```

tests/utils/profilehandler/data_preparation.go
yuwang-RH: @xueli181114 There are three condition here,
- err  != nil and strings.Contains(err.Error(), "cannot be found")
- err != nil and !strings.Contains(err.Error(), "cannot be found")
- err = nil

The way as you suggest also needs two if-else as the current one, from my side, it is more clear with current implement. Anyway, I updated as you suggest, please take a look again.

PR #2362, opened 2024-08-19
OCM-10406 | test: Automated id:75535 fix test case

86254860:


tests/e2e/test_rosacli_cluster_post.go
```diff
@@ -397,45 +397,36 @@ var _ = Describe("Healthy check",
 
 		It("with multiAZ will work - [id:75535]", labels.Runtime.Day1Post, labels.Critical,
 			func() {
+				if !isHosted {
+					SkipNotHosted()
```
yasun1, line 401: Do we have the checkpoint to validate `multiAZ` is meaningless to HCP? @xueli181114 @yingzhanredhat

PR #2363, opened 2024-08-19
OCM-10489 | test: fix ids:72195,73753,73765 due to version throttle

xueli181114:


tests/e2e/hcp_machine_pool_test.go
```diff
@@ -68,12 +76,21 @@ var _ = Describe("HCP Machine Pool", labels.Feature.Machinepool, func() {
...
+			output, err := rosaClient.MachinePool.CreateMachinePool(clusterID, mpName,
 				"--additional-security-group-ids", strings.Join(sgIDs, ","),
 				"--replicas", "1",
 				"-y",
 			)
+			if clusterVersion.LessThan(throttleVersion) {
+				// Low version cannot support security group set
+				Expect(err).To(HaveOccurred())
+				Expect(output.String()).Should(ContainSubstring(
+					"Additional security groups are not supported for '%s'",
```
OAharoni-RedHat, line 88: Do we expect this test to fail like this? In other words, are we setting up clusters with version mismatches on purpose that would propagate this error? To me it kind of looks like we're running this test on clusters we're not supposed to if we can't set up additional security groups if that makes any sense

xueli181114: It is expected error. And we do had jobs creating clusters in low versions which is not supported for this. And it's not deserved to have another. We don't want too many labels it will make things more complicated. With OCP releases and versions EOL the error expectation part will be removed.

OAharoni-RedHat: I see in some of the other changes in the PR we use the skip functionality, would it not make sense to use that here as well?

xueli181114: That will cause the low version validation cannot be covered. Skip must make sure there is a case cover the skipped scenario. For this case, it's not deserved to have another case since it will be deprecated with cluster version EOL.

OAharoni-RedHat: understood, thank you!

PR #2367, opened 2024-08-20
OCM-10479 | ci: fix id:64187,67348

yingzhanredhat:
https://privatebin.corp.redhat.com/?9533cf78ea378a6a#AufRcorrpwkAej4icdkTacT3AQMmamwFAoyBTodCrQZ4


tests/e2e/rosa_autoscaler_test.go
```diff
@@ -471,7 +471,7 @@ var _ = Describe("Autoscaler", labels.Feature.Autoscaler, func() {
 							"--ignore-daemonsets-utilization",
 							"--max-node-provision-time", "10m",
 							"--balancing-ignored-labels", "aaa",
-							"--max-nodes-total", "1000",
+							"--max-nodes-total", "10",
```
xueli181114, line 474: Can we set a higher value here? Once the autosccaler removing failed, it will cause other case failure

PR #2373, opened 2024-08-21
OCM-10543 | test: fix ids: 75210,73731,57408,38770,74661,60688 some failures

yuwang-RH:
[OCM-10543](https://issues.redhat.com//browse/OCM-10543) | test: fix ids: 75210,73731,57408,38770,74661,60688 some failures on rosacli ci

75210: update ExtractOIDCProviderFromOidcUrl to support the not-classic oidc config
73731: update expected output for hosted-cp cluster
57408: add one minute to wait time
38770: add condition for not autoscaling enabled cluster
74661: update the role matcher to adapt the long cluster prefix
60688: skip on share vpc clusters as the base domain cannot be reused

### DIFF
--- a/tests/e2e/test_rosacli_account_roles.go
+++ b/tests/e2e/test_rosacli_account_roles.go
@@ -1,9 +1,13 @@
 package e2e
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"strings"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -545,6 +549,23 @@ var _ = Describe("Edit account roles", labels.Feature.AccountRoles, func() {
 				Expect(err).To(BeNil())
 				policiesArn = append(policiesArn, pArn)
 			}
+			err = wait.PollUntilContextTimeout(
+				context.Background(),
+				20*time.Second,
+				200*time.Second,
+				false,
+				func(context.Context) (bool, error) {
+					hasPoliciWaitSync := false
+					for _, arn := range policiesArn {
+						policy, err := awsClient.GetIAMPolicy(arn)
+						if err != nil || policy == nil {
+							hasPoliciWaitSync = true
+							break
+						}
+					}
+					return !hasPoliciWaitSync, err
+				})
+			common.AssertWaitPollNoErr(err, "Network verification result are not ready after 200")
 
 			By("Prepare verson for testing")
 			versionService := rosaClient.Version
--- a/tests/e2e/test_rosacli_cluster.go
+++ b/tests/e2e/test_rosacli_cluster.go
@@ -584,36 +584,39 @@ var _ = Describe("Classic cluster creation validation",
 					Expect(stdout.String()).To(ContainSubstring(v))
 				}
 				if rosalCommand.CheckFlagExist("--multi-az") {
-					invalidReplicasErrorMapMultiAZ := map[string]string{
-						"2":  "Multi AZ cluster requires at least 3 compute nodes",
-						"0":  "Multi AZ cluster requires at least 3 compute nodes",
-						"-3": "must be non-negative",
-						"5":  "Multi AZ clusters require that the number of compute nodes be a multiple of 3",
-					}
-					for k, v := range invalidReplicasErrorMapMultiAZ {
-						rosalCommand.ReplaceFlagValue(map[string]string{
-							"--replicas": k,
-						})
-						stdout, err := rosaClient.Runner.RunCMD(strings.Split(rosalCommand.GetFullCommand(), " "))
-						Expect(err).NotTo(BeNil())
-						Expect(stdout.String()).To(ContainSubstring(v))
+					if !profile.ClusterConfig.AutoscalerEnabled {
+						invalidReplicasErrorMapMultiAZ := map[string]string{
+							"2":  "Multi AZ cluster requires at least 3 compute nodes",
+							"0":  "Multi AZ cluster requires at least 3 compute nodes",
+							"-3": "must be non-negative",
+							"5":  "Multi AZ clusters require that the number of compute nodes be a multiple of 3",
+						}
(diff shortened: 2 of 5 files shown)


tests/utils/common/oidc.go
```diff
@@ -109,15 +109,18 @@ func ParseIssuerURLFromCommand(command string) string {
...
 	path := strings.TrimPrefix(parsedURL.Path, "/")
-	fmt.Printf("The past host and path is %s and %s\n", host, path)
-	oidcProvider := fmt.Sprintf("%s/%s", host, path)
-	oidcProvider = strings.Split(oidcProvider, " ")[0]
-
+	if path != "" {
+		oidcProvider = fmt.Sprintf("%s/%s", host, path)
+	} else {
+		oidcProvider = host
+	}
```
yasun1, line 124: I'd like to write it like this:
```
oidcProvider = host
if path != "" {
  oidcProvider = fmt.Sprintf("%s/%s", host, path)
}

File: tests/e2e/test_rosacli_account_roles.go [hunk 2]
```diff
@@ -545,6 +546,7 @@ var _ = Describe("Edit account roles", labels.Feature.AccountRoles, func() {
 				Expect(err).To(BeNil())
 				policiesArn = append(policiesArn, pArn)
 			}
+			time.Sleep(1 * time.Minute)
```
yasun1, line 549: Let us use a function to wait for the result but not a fuzzy Waiting.

tests/e2e/test_rosacli_upgrade.go
```diff
@@ -269,9 +277,13 @@ var _ = Describe("Cluster Upgrade testing",
 				)
 			}
 			Expect(err).To(BeNil())
-			Expect(output.String()).To(ContainSubstring("are already up-to-date"))
 			Expect(output.String()).To(ContainSubstring("are compatible with upgrade"))
 			Expect(output.String()).To(ContainSubstring("Upgrade successfully scheduled for cluster"))
+			if profile.ClusterConfig.HCP {
```
yasun1, line 282: What is the difference between `isHosted` and `profile.ClusterConfig.HCP`. I do not recommend using `profile.xxx`; as I said before, it is an attribute from the template,  but not the real one from the cluster.

tests/e2e/test_rosacli_cluster.go
```diff
@@ -584,36 +584,39 @@ var _ = Describe("Classic cluster creation validation",
...
-						"5":  "Multi AZ clusters require that the number of compute nodes be a multiple of 3",
-					}
-					for k, v := range invalidReplicasErrorMapMultiAZ {
-						rosalCommand.ReplaceFlagValue(map[string]string{
-							"--replicas": k,
-						})
-						stdout, err := rosaClient.Runner.RunCMD(strings.Split(rosalCommand.GetFullCommand(), " "))
-						Expect(err).NotTo(BeNil())
-						Expect(stdout.String()).To(ContainSubstring(v))
+					if !profile.ClusterConfig.AutoscalerEnabled {
```
86254860, line 587: How about separating these testing logic about `if rosalCommand.CheckFlagExist("--multi-az") and else part` from this case, then the logic may be more clear? 

It may be divided into two cases when `profile.ClusterConfig.AutoscalerEnabled` is `true` or `not`, may use `skip`(like ln 2158~2160)  instead of nested if.

Just a suggestion.

tests/utils/common/oidc.go
yuwang-RH: I don't know what difference with the if-else way, please let me know why you think that's better. 
However there is no impact, I will update even you have no reason why you suggest this?

tests/e2e/test_rosacli_upgrade.go
yuwang-RH: Thanks. No difference from my side. I will update like you suggest.

tests/e2e/test_rosacli_cluster.go
yuwang-RH: @86254860 Using nested if because we want to use one, or as least as possible, test case to cover on more cluster profile, at least keep a possibility to rich with more condition.

tests/e2e/test_rosacli_account_roles.go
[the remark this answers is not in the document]
yuwang-RH: @yasun1 @86254860 Let's discuss in slack channel. 
This time sleep is just for the occurrent failure when AWS is not synchronized the new created resources. In my cases, it creates some policy and roles, a simple time.Sleep can avoid this kind of failure. we uses wait.PollUntilContextTimeout at some place to wait some resources ready, but it is unnecessary in my case to check all the roles and policies sync to AWS. That will increase the unnecessary requests to AWS. 
If we do as you suggested, I cannot see any benefit while increasing the complexity and also increase the chance of failure when query resource from AWS.

yuwang-RH:
Attach log for 57408 which update to use PollUntilContextTimeout replacing time.Sleep
`
yuwan1-mac:rosa yuwan$ ginkgo --focus id:57408 tests/e2e/
Running Suite: ROSA CLI e2e tests suite - /Users/yuwan/workplace/rosacli_automation/rosa/tests/e2e
==================================================================================================
Random Seed: 1724230830

Will run 1 of 184 specs
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS•SS

Ran 1 of 184 Specs in 578.325 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 183 Skipped
PASS

`

tests/utils/common/oidc.go
yasun1: The way I recommend could improve the readability.

PR #2374, opened 2024-08-21
OCM-10545 | test: fix id: 72162

yuwang-RH:
[OCM-10545](https://issues.redhat.com//browse/OCM-10545) | test: fix id: 72162


yuwang-RH:
Log is here

PR #2377, opened 2024-08-21
OCM-10450 | fix: Pass in OIDC Config ID when creating oidcprovider from create/oidcconfig

hunterkepley:
Fixes issue where oidc provider was not automatically created when creating an oidc config due to neither an oidc config id, nor a cluster id being passed in

### DIFF
--- a/cmd/create/oidcconfig/cmd.go
+++ b/cmd/create/oidcconfig/cmd.go
@@ -294,23 +294,31 @@ func run(cmd *cobra.Command, _ []string) {
 		r.Reporter.Errorf("%s", err)
 		os.Exit(1)
 	}
-	oidcConfigStrategy.execute(r)
+	oidcConfigId := oidcConfigStrategy.execute(r)
 	if !args.rawFiles {
 		arguments.DisableRegionDeprecationWarning = true // disable region deprecation warning
-		oidcprovider.Cmd.Run(oidcprovider.Cmd, []string{"", mode, oidcConfigInput.IssuerUrl})
+		providerArgs := []string{"", mode, oidcConfigInput.IssuerUrl}
+		if oidcConfigId != "" {
+			providerArgs = append(providerArgs, "--oidc-config-id", oidcConfigId)
+		} else {
+			r.Reporter.Errorf("Unable to attempt creation of OIDC provider; oidc config ID"+
+				" not found / not created successfully: %s", err)
+			os.Exit(1)
+		}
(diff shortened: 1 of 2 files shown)


cmd/create/oidcconfig/cmd.go
```diff
@@ -294,23 +294,29 @@ func run(cmd *cobra.Command, _ []string) {
...
+	oidcConfigId := oidcConfigStrategy.execute(r)
 	if !args.rawFiles {
 		arguments.DisableRegionDeprecationWarning = true // disable region deprecation warning
-		oidcprovider.Cmd.Run(oidcprovider.Cmd, []string{"", mode, oidcConfigInput.IssuerUrl})
+		providerArgs := []string{"", mode, oidcConfigInput.IssuerUrl}
+		if oidcConfigId != "" {
+			providerArgs = append(providerArgs, "--oidc-config-id", oidcConfigId)
+		} else if r.Cluster != nil && r.Cluster.ID() != "" {
+			providerArgs = append(providerArgs, "--cluster", r.Cluster.ID())
+		}
```
robpblake, line 305: @hunterkepley Should we provide a clear error here if neither is available? Can that happen?

hunterkepley: It seems after talking with @gdbranco - that we will always have an OIDC config in this case. I will provide an error though, and remove the cluster ID stuff

robpblake (approved):
/approve

cmd/create/oidcconfig/cmd.go
```diff
@@ -294,23 +294,31 @@ func run(cmd *cobra.Command, _ []string) {
...
-	oidcConfigStrategy.execute(r)
+	oidcConfigId := oidcConfigStrategy.execute(r)
 	if !args.rawFiles {
 		arguments.DisableRegionDeprecationWarning = true // disable region deprecation warning
-		oidcprovider.Cmd.Run(oidcprovider.Cmd, []string{"", mode, oidcConfigInput.IssuerUrl})
+		providerArgs := []string{"", mode, oidcConfigInput.IssuerUrl}
+		if oidcConfigId != "" {
+			providerArgs = append(providerArgs, "--oidc-config-id", oidcConfigId)
+		} else {
+			r.Reporter.Errorf("Unable to attempt creation of OIDC provider; oidc config ID"+
```
gdbranco, line 304: I think you need to identify the strategy that the user requested, if it is manual then you can't say it won't be able :thinking: even though the id will still be emtpy

PR #2378, opened 2024-08-21
OCM-9784 | test: automate case id:72538

patriksuba:
Successfully tested for all HCP profiles. For profile with autoscaling enabled test will not work as specific number of subnets have to be used.

### DIFF
--- a/tests/e2e/test_rosacli_cluster.go
+++ b/tests/e2e/test_rosacli_cluster.go
@@ -18,6 +18,7 @@ import (
 	"github.com/openshift/rosa/tests/utils/common/constants"
 	"github.com/openshift/rosa/tests/utils/config"
 	"github.com/openshift/rosa/tests/utils/exec/rosacli"
+	"github.com/openshift/rosa/tests/utils/log"
 	"github.com/openshift/rosa/tests/utils/profilehandler"
 )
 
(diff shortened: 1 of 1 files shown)


yasun1:
I agree with the automation, but I'd like to leave it to the ROSA HCP QE members to double-check the checkpoints.

tests/e2e/test_rosacli_cluster.go
```diff
@@ -1642,6 +1643,144 @@ var _ = Describe("HCP cluster creation negative testing",
 						ContainSubstring(
 							"ERR: Additional Allowed Principals is supported only for Hosted Control Planes"))
 			})
+
+		It("HCP cluster creation subnets validation - [id:72538]",
+			labels.High, labels.Runtime.Day1Negative,
+			func() {
+				if profile.ClusterConfig.Private {
```
radtriste, line 1650: AFAIK, this test is HCP only

tests/e2e/test_rosacli_cluster.go
```diff
@@ -1642,6 +1643,144 @@ var _ = Describe("HCP cluster creation negative testing",
...
+					rosalCommand.DeleteFlag("--private", false)
+					rosalCommand.DeleteFlag("--private-link", false)
+				}
+				if profile.ClusterConfig.Autoscale {
+					Skip("Test do not work with autoscaling enabled")
+				}
+				log.Logger.Info(profile.Name)
+				log.Logger.Info(strings.Split(rosalCommand.GetFullCommand(), " "))
```
radtriste, line 1658: Can you put these as Debug ?

tests/e2e/test_rosacli_cluster.go
```diff
@@ -1642,6 +1643,144 @@ var _ = Describe("HCP cluster creation negative testing",
...
+
+				By("Prepare a vpc for the testing")
+				vpc, err := vpc_client.PrepareVPC(vpcName, region, "", false, "")
+				Expect(err).ToNot(HaveOccurred())
+				defer vpc.DeleteVPCChain()
+
+				subnetMap, err := profilehandler.PrepareSubnets(vpc, region, []string{}, true)
+				Expect(err).ToNot(HaveOccurred())
+
+				additionalPrivateSubnet, err := vpc.CreatePrivateSubnet(availabilityZone, true)
```
radtriste, line 1683: I would initialize `availabilityZone` with the AZ from  the first subnet in `subnetMap`. That way you make sure you have a duplicated AZ

tests/e2e/test_rosacli_cluster.go
patriksuba: Yes it is. Testing group is only for HCP negative.

tests/e2e/test_rosacli_cluster.go
```diff
@@ -1645,6 +1646,146 @@ var _ = Describe("HCP cluster creation negative testing",
 						ContainSubstring(
 							"ERR: Additional Allowed Principals is supported only for Hosted Control Planes"))
 			})
+
+		It("HCP cluster creation subnets validation - [id:72538]",
+			labels.High, labels.Runtime.Day1Negative,
```
xueli181114, line 1651: Day1 negative cases has no profile setting.

patriksuba: https://github.com/openshift/rosa/blob/master/tests/e2e/test_rosacli_cluster.go#L1431

radtriste: should be fine then

tests/e2e/test_rosacli_cluster.go
radtriste: should be fine then

radtriste (approved):
/lgtm

PR #2380, opened 2024-08-22
OCM-9995 | feat: Adding rosa quickstart POC

den-rgb:
to run the command -> 
```
rosa bootstrap rosa-quickstart-default-vpc --param Name=rosa-quickstart-default --param VpcCidr=10.0.0.0/16 --param Region=us-west-2 --param AvailabilityZonesCount=2 --param Tags=key1=value1,key2=value2
```

*Whats New*

- Added region
- added params
- different templates will be called depending on the first arg passed in eg:`rosa-quickstart-default-vpc`
- Modified Tags handling + added default tags
- Account + Org validation
- vpc template no longer requires s3 bucket hosting
- no longer creates nested stacks


cmd/create/quick-start-cluster/vpc.yaml
```diff
@@ -0,0 +1,111 @@
...
+  S3VPCEndpoint:
+    Type: AWS::EC2::VPCEndpoint
+    Properties:
+      VpcId: !Ref VPC
+      ServiceName: !Sub "com.amazonaws.${AWS::Region}.s3"
+
+  PublicSubnets:
+    Type: AWS::CloudFormation::Stack
+    Properties:
+      TemplateURL: "https://quickstart-creation-bucket.s3.us-west-2.amazonaws.com/templates/public_subnets.yaml"
```
ciaranRoche, line 46: Is there any reason for making these templates instead of a single file? 

I am just wondering if there is an easy way to bubble up the outputs into a single stack?

cmd/create/quick-start-cluster/cmd.go
```diff
@@ -0,0 +1,222 @@
...
+    defer r.Cleanup()
+
+    err := createVPCStack("cmd/create/quick-start-cluster/vpc.yaml", args.namePrefix, args.vpcCidr, args.availabilityZonesCount, args.tags)
+    if err != nil {
+        r.Reporter.Errorf(err.Error())
+        os.Exit(1)
+    }
+
+    r.Reporter.Infof("VPC created")
+    r.Reporter.Infof("To view all created resource stacks, run `aws cloudformation list-stacks --stack-status-filter CREATE_COMPLETE`")
```
ciaranRoche, line 87: it would be good if you can get the outputs of the required resources needed to create a cluster with

cmd/bootstrap/quick-start-cluster/cmd.go
```diff
@@ -0,0 +1,217 @@
...
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/cloudformation"
+	cfTypes "github.com/aws/aws-sdk-go-v2/service/cloudformation/types"
+	"github.com/openshift/rosa/pkg/output"
+    "github.com/sirupsen/logrus"
+)
+
+var Cmd = makeCmd()
+
+func makeCmd() *cobra.Command {
```
robpblake, line 22: @den-rgb Rather than a command-per-quickstart, I was thinking we could use a folder structure as the quick start name and then have a generic command that works with a known folder structure. 

So for example, to support the `rosa create quickstart hcp-vpc-networking`, we'd have a folder structure like:

- quickstart/hcp-vpc-networking
- quickstart/hcp-vpc-networking/cloudformation-template.json
- quickstart/hcp-vpc-networking/metadata.json <-- Not sure if this is needed, but maybe. It could be used to keep larger descriptions of the quick start that we can't store in the CloudFormation template

So given the above structure, when the user runs `rosa create quickstart hcp-vpc-networking`, the generic command would:

- Verify that the folder exists
- Verify that the CloudFormation template exists

For parameters, the command could parse the CloudFormation template and look at the list of known/supported parameters and we allow a generic way of passing these e.g 

`rosa create quickstart hcp-vpc-network -pname=foo -ptags=foo=bar,bob=baz ...`

Very much aligned with the `oc process -f $template -p ...`

With that model, creating new quick starts is simply a case of providing new cloudformation templates (and the metadata.json) if we think we need it.

cmd/bootstrap/cmd.go
```diff
@@ -0,0 +1,289 @@
...
+	for k, v := range userTags {
+		combinedTags[k] = v
+	}
+	return combinedTags
+}
+
+// selectTemplate selects the appropriate template file based on the parameters
+func selectTemplate(command string) string {
+	templates := map[string]string{
+		"rosa-quickstart-default-vpc":     "cmd/bootstrap/templates/vpc.yaml",
```
robpblake, line 138: @den-rgb I'd suggest to make the template directory structure match the requested "scenario". So if I'm requesting `rosa-quickstart-default-vpc`, then the CloudFormation template must exist at `cmd/bootstrap/templates/rosa-quickstart-default-vpc/cloudformation.yaml`

Directory-per-"scenario" means this function can be generic e.g check that the resource exists in the specified location, and means that we can embed additional metadata in the directory in the future if we wish e.g we could have `cmd/boostrap/templates/rosa-quickstart-default-vpc/metadata.json` in the future that provides lots of readable information for a `rosa describe bootstrap` command.

cmd/bootstrap/cmd.go
```diff
@@ -0,0 +1,289 @@
...
+	// Create a CloudFormation client
+	logger.Info("Creating CloudFormation client")
+	cfClient := cloudformation.NewFromConfig(cfg)
+
+	// Create the stack
+	logger.Info("Creating CloudFormation stack")
+	_, err = cfClient.CreateStack(context.TODO(), &cloudformation.CreateStackInput{
+		StackName:    aws.String(params["Name"]),
+		TemplateBody: aws.String(string(templateBody)),
+		Parameters: []cfTypes.Parameter{
```
robpblake, line 179: @den-rgb Again I'd expect this part to be generic in that we simply convert your `map[string]string` into a `[]cfTypes.Parameter`. So something like:

```
var cfParams []cfTypes.Parameter

for k,v := range params {
  cfParams = append(cfParams, {"ParameterKey": aws.String(k), "ParameterValue": aws.String(v)}
}
```

cmd/bootstrap/cmd.go
```diff
@@ -0,0 +1,289 @@
...
+	r.Reporter.Infof("VPC created")
+}
+
+// parseParams converts the list of parameter strings into a map and sets default values
+func parseParams(params []string) (map[string]string, map[string]string) {
+	result := map[string]string{
+		"Name":                   "rosa-quickstart-default",
+		"VpcCidr":                "10.0.0.0/16",
+		"Region":                 "us-west-2",
+		"AvailabilityZonesCount": "2",
```
robpblake, line 93: @den-rgb Out of interest, why do we need multiple Availability zones?

cmd/bootstrap/cmd.go
```diff
@@ -0,0 +1,275 @@
...
+		r.Reporter.Errorf(err.Error())
+		os.Exit(1)
+	}
+
+	r.Reporter.Infof("VPC created")
+}
+
+// parseParams converts the list of parameter strings into a map and sets default values
+func parseParams(params []string) (map[string]string, map[string]string) {
+	result := map[string]string{
```
robpblake, line 89: @den-rgb These defaults should probably be stored in the CloudFormation template. Take a look here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html

Another move to keep this command generic and have the CloudFormation Template control most things.

cmd/bootstrap/cmd.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+	for k, v := range userTags {
+		combinedTags[k] = v
+	}
+	return combinedTags
+}
+
+// selectTemplate selects the appropriate template file based on the parameters
+func selectTemplate(command string) string {
+	return fmt.Sprintf("cmd/bootstrap/templates/%s/vpc.yaml", command)
```
robpblake, line 137: @den-rgb I know this is a POC, but I'd suggest calling the template `cloudformation.yaml`. It keeps it generic as the to the stack of resources that are created there.

cmd/bootstrap/cmd.go
den-rgb: just had two as a random number for testing, no reason, i have reverted it back to 1

PR #2381, opened 2024-08-22
OCM-10452 | fix: Fix cluster creation with oidc provider

hunterkepley:
passes in oidc config id to oidc provider creation cmd.

### DIFF
--- a/cmd/create/oidcconfig/cmd.go
+++ b/cmd/create/oidcconfig/cmd.go
@@ -300,6 +300,11 @@ func run(cmd *cobra.Command, _ []string) {
 		providerArgs := []string{"", mode, oidcConfigInput.IssuerUrl}
 		if oidcConfigId != "" {
 			providerArgs = append(providerArgs, "--oidc-config-id", oidcConfigId)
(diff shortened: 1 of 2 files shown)


cmd/create/oidcconfig/cmd.go
```diff
@@ -300,6 +300,12 @@ func run(cmd *cobra.Command, _ []string) {
 		providerArgs := []string{"", mode, oidcConfigInput.IssuerUrl}
 		if oidcConfigId != "" {
 			providerArgs = append(providerArgs, "--oidc-config-id", oidcConfigId)
+			err = oidcprovider.Cmd.Flags().Set("oidc-config-id", oidcConfigId)
+			if err != nil {
+				r.Reporter.Errorf("Unable to attempt creation of OIDC provider; oidc config ID"+
+					" not found / not created successfully: %s", err)
+			}
+			oidcprovider.Cmd.Flags().Lookup("oidc-config-id").Changed = true
```
marcolan018, line 308: I think we don't need this line if already invoke Flags().Set in line 303

PR #2388, opened 2024-08-26
OCM-10467 | fix: don't allow cluster name longer than 15 characters when installed into shared vpc

marcolan018:
https://issues.redhat.com/browse/OCM-10467
with the change in CS side https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/8491
hosted zone name is required to be as same as  `cluster_domain_prefix.based_domain`. If radom string as the domin prefix, there is no way for users to create hosted zone name same as it. To avoid this, if cluster installed into shared vpc, ROSA CLI will validate the cluster name and domain prefix, to valide either name length is no longer than 15 or domain prefix set, that can guarantee no random string generated in cs side as the domain prefix


cmd/create/cluster/cmd.go
```diff
@@ -2256,6 +2256,13 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	if isSubnetBelongToSharedVpc(r, awsCreator.AccountID, subnetIDs, mapSubnetIDToSubnet) {
+
+		if clusterHasLongNameWithoutDomainPrefix(clusterName, domainPrefix) {
+			r.Reporter.Errorf("Installing a cluster into shared VPC is only supported for cluster " +
+				"which has a name no longer than 15 characters or with domain prefix")
+			os.Exit(1)
+		}
```
machi1990, line 2264: Let's also involve the docs team to modify the part around cluster DNS in following doc: https://docs.openshift.com/rosa/rosa_install_access_delete_clusters/rosa-shared-vpc-config.html

instead of;

```
Create a private hosted zone in the [Route 53 section of the AWS console](https://us-east-1.console.aws.amazon.com/route53/v2/). In the hosted zone configuration, the domain name is <cluster_name>.<reserved_dns_domain>. The private hosted zone must be associated with the created VPC.
``` 

it should be
```
Create a private hosted zone in the [Route 53 section of the AWS console](https://us-east-1.console.aws.amazon.com/route53/v2/). In the hosted zone configuration, the domain name is <cluster_domain_prefix>.<reserved_dns_domain>. The private hosted zone must be associated with the created VPC.
```
Also instead of 
```
After you create the Operator roles, share the full domain name, which is created with <intended_cluster_name>.<reserved_dns_domain>, your Ingress Operator Cloud Credentials role’s ARN, and your Installer role’s ARN with the VPC Owner to continue configuration.
```

it should be 
```
After you create the Operator roles, share the full domain name, which is created with <intended_cluster_domain_prefix>.<reserved_dns_domain>, your Ingress Operator Cloud Credentials role’s ARN, and your Installer role’s ARN with the VPC Owner to continue configuration.
```

cmd/create/cluster/cmd.go
```diff
@@ -2256,6 +2256,13 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	if isSubnetBelongToSharedVpc(r, awsCreator.AccountID, subnetIDs, mapSubnetIDToSubnet) {
+
+		if clusterHasLongNameWithoutDomainPrefix(clusterName, domainPrefix) {
+			r.Reporter.Errorf("Installing a cluster into shared VPC is only supported for cluster " +
+				"which has a name no longer than 15 characters or with domain prefix")
```
machi1990, line 2262:
```suggestion
			r.Reporter.Errorf("Installing a cluster into shared VPC is only supported for cluster " +
				"which has a name no longer than %s characters or with a cluster domain prefix", ocm.MaxClusterDomainPrefixLength)
```

cmd/create/cluster/cmd.go
marcolan018: thanks @machi1990 
I will create a doc issue to make the change

machi1990:
lgtm

@gdbranco can you give this a look as well? Thanks

PR #2392, opened 2024-08-27
OCM-10676 | fix: Remove manual cluster ID flag update for create/oidcprovider

hunterkepley:
Manually setting the cluster ID caused the command to not be 'programmatically called', removed 2 lines to fix this


robpblake (approved):
/lgtm

PR #2395, opened 2024-08-28
OCM-9876 | feat: enable custom worker disk sizes on node pools

willkutler:
- for day 2 node pool creation (rosa create machinepool cmd) this adds support for the --disk-size flag that is enabled for classic machine pools
- for day 1 node pool creation ROSA already has support as the --worker-disk-size flag is automatically passed to CS regardless of cluster type

### DIFF
--- a/tests/e2e/test_rosacli_cluster_post.go
+++ b/tests/e2e/test_rosacli_cluster_post.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 	nets "net/http"
-	"net/url"
 	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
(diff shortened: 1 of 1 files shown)


pkg/machinepool/machinepool.go
```diff
@@ -808,7 +808,57 @@ func (m *machinePool) CreateNodePools(r *rosa.Runtime, cmd *cobra.Command, clust
...
+			return err
+		}
+	}
+
+	npBuilder.AWSNodePool(createAwsNodePoolBuilder(
+		instanceType,
+		securityGroupIds,
+		httpTokens,
+		awsTags,
+		238,
```
robpblake, line 860: @willkutler The hard-coded disk size isn't correct.

pkg/interactive/validation.go
```diff
@@ -272,3 +272,22 @@ func MachinePoolRootDiskSizeValidator(version string) Validator {
 		return diskValidator.ValidateMachinePoolRootDiskSize(version, size)
 	}
 }
+
+func NodePoolRootDiskSizeValidator() Validator {
```
robpblake, line 276: @willkutler Could you add a unit test for this please.

PR #2396, opened 2024-08-28
OCM-9780 | test: automated id:65900,76481 install with use-local-credentials will work

jerichokeyne:
Automated id:65900 which tests that a cluster installed with the `--use-local-credentials` flag will install correctly

output:
```
$ TEST_PROFILE=rosa-non-sts-local-creds ginkgo run --focus 65900 tests/e2e/
Running Suite: ROSA CLI e2e tests suite - /home/jericho/work/repos/rosa/tests/e2e
=================================================================================
Random Seed: 1724876476

Will run 1 of 156 specs
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS•SSSSSSSSSSSSS

Ran 1 of 156 Specs in 0.436 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 155 Skipped
PASS

Ginkgo ran 1 suite in 1.722011s
Test Suite Passed
```


OAharoni-RedHat:
test LGTM, could you explain to me what the flag does just for my own sake?

jerichokeyne:
@OAharoni-RedHat I believe it just makes it so that when you install a non-sts cluster it will use your AWS credentials from your machine

jerichokeyne:
> > @OAharoni-RedHat I believe it just makes it so that when you install a non-sts cluster it will use your AWS credentials from your machine
> 
> @jerichokeyne maybe i'm going to out myself as clueless here but isn't that what it already does? Since we need to set up an `AWS_PROFILE` for it to grab info from already right? Or am I misunderstanding

I'll be honest I'm not completely sure what the difference is. I think the difference is that both will use your local credentials for installing, but this flag makes the cluster use your credentials instead of the osdCcsAdmin credentials. It's apparently something we don't fully support even though it's something we test for :shrug:

jfrazierRH:
It looks like the step for testing vs an STS cluster is missing. It might make sense to split that into a separate negative test case.

Also, does the cluster itself describe that it used local credentials in any way? Or is it only evident in the json structure?

jerichokeyne:
> It looks like the step for testing vs an STS cluster is missing. It might make sense to split that into a separate negative test case.
Ah, looks like I forgot that part. I'll have to do some more research into how to do that since it's an error at install time
> Also, does the cluster itself describe that it used local credentials in any way? Or is it only evident in the json structure?
As far as I know it's only evident in the json

jerichokeyne:
@jfrazierRH I added that negative test case

tests/e2e/test_rosacli_cluster.go
```diff
@@ -1018,6 +1018,19 @@ var _ = Describe("Classic cluster creation validation",
 				Expect(errorOutput.String()).To(ContainSubstring("etcd encryption cannot be disabled on clusters with FIPS mode"))
 			})
 
+		It("validate use-local-credentials won't work with sts - [id:65900]",
```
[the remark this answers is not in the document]
jerichokeyne: Sorry it took a while to get to your feedback. But, I believe this test doesn't use the profiles at all, so since I specify the `--sts` flag it should always be an STS cluster

tests/ci/data/profiles/rosa-classic.yaml
```diff
@@ -214,4 +214,31 @@ profiles:
     oidc_config: ""
     shared_vpc: false
     imdsv2: "optional"
-    admin_enabled: false
\ No newline at end of file
+    admin_enabled: false
+- as: rosa-non-sts-local-creds
```
radtriste, line 218: Is that meant to be launched from a job ?
If yes, I don't think we need a new profile. You could have here this profile property be configurable also via an environment variable `USE_LOCAL_CREDENTIALS` which would set the profile property accordingly and configure your job to use the `rosa-non-sts-advanced` profile with the env var `USE_LOCAL_CREDENTIALS: true`

jerichokeyne: I added a env var option now. I set it up currently so that when set to true it will enable the profile option, but if it's set to anything else it'll leave the profile option alone

radtriste: so you don't need to keep that profile here. And you would need to create a new job in `openshift/release` for this specific case

tests/utils/profilehandler/profile_handler.go
```diff
@@ -66,11 +66,16 @@ func LoadProfileYamlFileByENV() *Profile {
...
 	if config.Test.GlobalENV.ComputeMachineType != "" {
-		log.Logger.Infof("Got global env settings for INSTANCE_TYPE, overwritten the profile setting with value %s",
+		log.Logger.Infof("Got global env settings for COMPUTE_MACHINE_TYPE, overwritten the profile setting with value %s",
 			config.Test.GlobalENV.ComputeMachineType)
 		profile.ClusterConfig.InstanceType = config.Test.GlobalENV.ComputeMachineType
 	}
 
+	if config.Test.GlobalENV.UseLocalCredentials {
+		log.Logger.Info("Got global env setting for USE_LOCAL_CREDENTIALS, overwritten the profile setting to true")
+		profile.ClusterConfig.UseLocalCredentials = true
```
radtriste, line 76: please check that `config.Test.GlobalENV.UseLocalCredentials == "true"`

yuwang-RH:
@jerichokeyne For the scenario of creating non-sts clustet with setting '--use-local-credentials', it is using the cluster ENV to control as you implemented in this PR. But we still have no such job to test it.

It is fine to use ENV to decide if the cluster creation uses 'use-local-credentials' or not. Considering rosacli testing is a separated part from backend, and to avoid too many jobs exsiting, we can use a supplemental case to cover OCP-65900. 
- Create a non-sts cluster with setting '--use-local-credentials'
- Wait it to be installing status
- Check the properties of the cluster spec has '"use_local_credentials": "true"'.
- Then delete the cluster
In future, we can enable one job for the cluster using 'use-local-credentials' if it is needed.WDYT?

jerichokeyne:
@yuwang-RH I added a test based on your suggestion. I left the old test that relies on a new profile/job for testing properly for now. My thoughts being that it also then tests the negative test case (cluster with `--use-local-credential` won't have the property set). WDYT? I'm willing to remove it, or just add a negative test to the other test section that I wrote.

PR #2401, opened 2024-08-30
OCM-10812 | ci: Generate common functions of rosacli CI steps to prow_ci.sh

xueli181114:


tests/prow_ci.sh
```diff
@@ -30,3 +30,147 @@ override_rosacli_build () {
...
+
+# setup aws credentials for the CI job run
+configure_aws () {
+  # configure aws region
+  if [[ -z "$REGION" ]]; then
+    REGION=${LEASED_RESOURCE}
+  fi
+  echo "[CI] Configured AWS region to ${REGION}"
+  # configure aws credetials
+  AWSCRED="${CLUSTER_PROFILE_DIR}/.awscred"
```
yasun1, line 42: The method with the hardcode variables depended too much on the Prow CI framework. When we call `configure_aws` in the build process, the better way should be `configure_aws $AWSCRED $REGION`

So are the below methods.

tests/prow_ci.sh
```diff
@@ -30,3 +30,147 @@ override_rosacli_build () {
...
+  if [ -z $JUNIT_XML ]; then
+    echo "JUNIT_XML is empty, please define it and call generate_junit to generate the value"
+    exit 1
+  fi [ -z $TEST_TIMEOUT ]; then
+    # set a default value in case it is empty
+    TEST_TIMEOUT="4h"
+  fi
+  cmd="rosatest --ginkgo.v --ginkgo.no-color \
+  --ginkgo.timeout ${TEST_TIMEOUT} \
+  --ginkgo.junit-report $JUNIT_XML \
```
yasun1, line 143: Are you sure to run all of the test cases if there is no `FOCUS`?

tests/prow_ci.sh
```diff
@@ -30,3 +30,147 @@ override_rosacli_build () {
...
+  if [[ -f "${SHARED_DIR}/junit.tar.gz" ]]; then
+      echo "[CI] extract existing junit file package"
+      tar -xvf "${SHARED_DIR}/junit.tar.gz" -C $JUNIT_TEMP_DIR
+  fi
+}
+
+# generate_running_cmd is used to generate rosatest running command
+# TEST_TIMEOUT is step env variable used to define the timeout of the test run
+# JUNIT_XML should be generated by function generate_junit
+generate_running_cmd () {
```
yasun1, line 133: Enveloping the final command on the back end is not a good choice. The `.sh` scripts are our build process scripts. Putting the running command in the apparent location could help us understand what we are doing. It is inconvenient to get this from another repository.

So is `generate_label_filter_switch`.

tests/prow_ci.sh
```diff
@@ -30,3 +30,147 @@ override_rosacli_build () {
...
+  fi
+
+  JUNIT_XML="${JUNIT_TEMP_DIR}/${junit_file_name}.xml"
+
+  echo "[CI] the junit temp dir is $JUNIT_TEMP_DIR and junit file will be $JUNIT_XML" 
+}
+
+# extract_existing_junit is used to extract existing junit file from previous steps
+# JUNIT_TEMP_DIR should be clarified in step and generated by generate_junit
+extract_existing_junit () {
```
yasun1, line 119: In the future, I plan to combine the runtime steps into one and use a variable instead of multiple steps to improve the flexibility of the test chain. If that is true, we do not need to extract/package the `junit.tar.gz` in each step. 

Besides, for the current status, I think it is better to generate a different same-prefix `*.tar.gz` and let the reported step do one-time extract/package.

tests/prow_ci.sh
xueli181114: There is  ${LABEL_FILTER_SWITCH} controlled the the cases need to run.

tests/prow_ci.sh
xueli181114: There will be log output. And also, there is nothing careful from other teams about testing command running.

tests/prow_ci.sh
xueli181114: Let's implement as step by step. This is the final expected implementation way. But it's long term resolution.

tests/prow_ci.sh
```diff
@@ -30,3 +30,163 @@ override_rosacli_build () {
...
+# configure_aws will setup aws credentials for the CI job run
+# It requires two parameters 
+# the first one for aws credential file path which is required
+# the second one for aws region which is required
+# the third one for second aws cred info for shared vpc scenario which is optional
+# usage: configure_aws <aws file path> <region> <shared vpc credential file>
+configure_aws () {
+  # configure aws region
+  if [[ -z "$1"]] || [[ -z "$2" ]]; then
+    echo "ERROR: aws credential file path and region is required. Please call command like $ configure_aws <aws path> <region>"
```
yasun1, line 43: `<aws path>` -> `credential path`

tests/prow_ci.sh
```diff
@@ -30,3 +30,163 @@ override_rosacli_build () {
...
   cd -
 }
+
+# configure_aws will setup aws credentials for the CI job run
+# It requires two parameters 
+# the first one for aws credential file path which is required
+# the second one for aws region which is required
+# the third one for second aws cred info for shared vpc scenario which is optional
+# usage: configure_aws <aws file path> <region> <shared vpc credential file>
+configure_aws () {
```
yasun1, line 40: maybe `configure_aws` - > `aws_config` could be better.

tests/prow_ci.sh
```diff
@@ -30,3 +30,163 @@ override_rosacli_build () {
...
+  if [[ -f "${awscred}" ]]; then
+    export AWS_SHARED_CREDENTIALS_FILE="${awscred}"
+    export AWS_DEFAULT_REGION="$2"
+  else
+    echo "ERROR: aws credential file $1 doesn't exist"
+    exit 1
+  fi
+
+  # configure shared vpc credentials
+  if [[ ! -z "$3" ]]; then
```
yasun1, line 58: Another method `aws_config_shared_vpc` is better.

tests/prow_ci.sh
xueli181114: `configure_aws` is an action about the function. I prefer to keep it.

yasun1: Based on your preferable.

tests/prow_ci.sh
xueli181114: Let's use `configure_aws` since it's an action

xueli181114: Moved it out.

tests/prow_ci.sh
xueli181114: Updated

tests/prow_ci.sh
xueli181114: Updated

PR #2403, opened 2024-08-30
OCM-10779 | fix: Error when setting multi-az true for HCP

den-rgb:
Fix for regression issue https://issues.redhat.com/browse/OCM-10779


ciaranRoche (approved):
/lgtm

PR #2407, opened 2024-08-30
OCM-10676 | fix: Assign cluster ID var in create/oidc_provider

hunterkepley:
Assigns cluster ID flag to the right var to be used when creating an OIDC provider


ciaranRoche (approved):
/lgtm

PR #2415, opened 2024-09-03
OCM-10856 | test: fix id:55883 fixes ci job "e2e-rosa-sts-upgrade-y-stream-prod-f3"

AkashKanni:
[OCM-10856](https://issues.redhat.com//browse/OCM-10856) | test: fix id:55883

$ echo $TEST_PROFILE
rosa-upgrade-y-stream

$ ginkgo run --label-filter day1 tests/e2e --timeout 2h
  Ginkgo CLI Version:
    2.11.0
    2.17.1 used by e2e

  Ginkgo CLI to match.

  Ginkgo CLI.

  unsupported.
  
Running Suite: ROSA CLI e2e tests suite - /home/akanni/OCP-Repository/rosa/tests/e2e
====================================================================================
Random Seed: 1725340859

Will run 1 of 198 specs
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS•SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS

Ran 1 of 198 Specs in 2540.315 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 197 Skipped
PASS

Ginkgo ran 1 suite in 42m26.837645809s
Test Suite Passed

### DIFF
--- a/tests/e2e/test_rosacli_upgrade.go
+++ b/tests/e2e/test_rosacli_upgrade.go
@@ -53,6 +53,14 @@ var _ = Describe("Cluster Upgrade testing",
 		})
 
 		AfterEach(func() {
+			if profile.Version == con.YStreamPreviousVersion {
+				By("Delete cluster upgrade")
+				output, err := upgradeService.DeleteUpgrade("-c", clusterID, "-y")
+				Expect(err).ToNot(HaveOccurred())
+				Expect(output.String()).To(ContainSubstring(
+					"Successfully canceled scheduled upgrade on cluster '%s'", clusterID))
+			}
+
 			By("Clean remaining resources")
 			err := rosaClient.CleanResources(clusterID)
 			Expect(err).ToNot(HaveOccurred())


tests/e2e/test_rosacli_upgrade.go
```diff
@@ -53,6 +53,14 @@ var _ = Describe("Cluster Upgrade testing",
 		})
 
 		AfterEach(func() {
+			if profile.Version == con.YStreamPreviousVersion {
+				By("Delete cluster upgrade")
+				output, err := upgradeService.DeleteUpgrade("-c", clusterID, "-y")
+				Expect(err).ToNot(HaveOccurred())
+				Expect(output.String()).To(ContainSubstring(
+					"Successfully canceled scheduled upgrade on cluster '%s'", clusterID))
```
yuwang-RH, line 61: May I know more about the background of this fix? 
If this is a fix on some specific failed case, why cannot it delete the upgrade policy in the case block?

AkashKanni: @yuwang-RH this fix is for a CI job `rosa-sts-upgrade-y-stream` not for a specific test case,  here in test case `ocp-73731` an upgrade is scheduled but not deleted after running the case in AfterEach block, and in case `ocp-55883` it again tries to upgrade the same cluster and then due to the cluster has already an upgrade scheduled it fails to schedule another upgrade on that cluster.

more information on slack thread :- https://redhat-internal.slack.com/archives/C0519LZH1RA/p1724177182281089

AkashKanni: for the CI job `rosa-sts-upgrade-y-stream` more than one test case is doing upgrade task, so instead of having a `delete upgrade`  block for every case i guess better to keep it at AfterEach, please let me know if any better ideas we can use here. Thanks!!

radtriste: Why not trying to remove the upgrade all the time and not specific to specific Version (and forget about checking the error if no upgrade is there as it would fail in case upgrade is already deleted or ?)

PR #2416, opened 2024-09-03
OCM-10628 | ci: Make NAME_PREFIX required for cluster creation

xueli181114:


radtriste (approved):
/lgtm

tests/README.md
```diff
@@ -102,11 +102,13 @@ For the test cases, we need `$ make install` to make the rosa command line insta
 
 2. Export the profile name as an environment variable
   * `$ export TEST_PROFILE=<PROFILE NAME>`
+3. Export the name prefix of the cluster and resources created, no longer than 15 chars.
+  * `$ export NAME_PREFIX=<your alias>`
```
yasun1, line 106: Prow CI provide a global environment boolean variable `CI`. How about
```
if !CI {
  NAME_PREFIX="local"
}

radtriste: Name prefix in local should be set by QE/dev so that we can easily identify owner of a cluster
For example, mine are starting with `tr-`

yasun1: I mean, we force adding a prefix to avoid the user giving a similar name to the CI.

PR #2419, opened 2024-09-04
OCM-10868 | test: Automate id:56224

jameszwang:


tests/e2e/test_rosacli_cluster.go
```diff
@@ -3067,3 +3063,105 @@ var _ = Describe("Sts cluster creation supplemental testing",
...
+				AccountRoleConfig: &profilehandler.AccountRoleConfig{
+					Path:               "/aa/bb/",
+					PermissionBoundary: "",
+				},
+				Version:      "latest",
+				ChannelGroup: "candidate",
+				Region:       "us-west-2",
+			}
+			customProfile.NamePrefix = constants.DefaultNamePrefix
+
```
xueli181114, line 3108: Set Name prefix in case level to make it related to case id. Otherwise leak resource cannot be traced back.

tests/e2e/test_rosacli_cluster.go
```diff
@@ -3067,3 +3063,105 @@ var _ = Describe("Sts cluster creation supplemental testing",
 				Expect(clusterID).ToNot(BeNil())
 			})
 	})
+
+var _ = Describe("User can set availability zones to create rosa multi-az STS cluster supplemental testing",
```
xueli181114, line 3067: Why a new Describe?  Can this be put to other Describe? I believe there are other day1 supplemental cases existing

tests/e2e/test_rosacli_cluster.go
```diff
@@ -3067,3 +3063,105 @@ var _ = Describe("Sts cluster creation supplemental testing",
...
+
+		It("User can set availability zones to create rosa multi-az STS cluster - [id:56224]",
+			labels.Critical, labels.Runtime.Day1Supplemental,
+			func() {
+				By("Create classic sts cluster in auto mode")
+				testingClusterName = "rosa56224"
+				flags, err := profilehandler.GenerateClusterCreateFlags(customProfile, rosaClient)
+				Expect(err).ToNot(HaveOccurred())
+
+				command := "rosa create cluster --cluster-name " + testingClusterName + " " + strings.Join(flags, " ")
```
xueli181114, line 3142: I don't understand the implementation here. Why not user rosaclient.ClusterService to create the cluster. This kind of the command cannot output the log. It will be very hard to debug in future with failure

tests/e2e/test_rosacli_cluster.go
```diff
@@ -3067,3 +3063,105 @@ var _ = Describe("Sts cluster creation supplemental testing",
...
+			By("Delete operator-roles")
+			_, err = ocmResourceService.DeleteOperatorRoles(
+				"-c", clusterID,
+				"--mode", "auto",
+				"-y",
+			)
+			Expect(err).To(BeNil())
+
+			By("Clean resource")
+			errs := profilehandler.DestroyResourceByProfile(customProfile, rosaClient)
```
xueli181114, line 3130: With this implementation, the pre-submit jobs won't pass. It will destroy the cluster resources instead of the resources created by this case.

tests/e2e/test_rosacli_cluster.go
jameszwang: Updated using the existed day1 supplemental Describe.

tests/e2e/test_rosacli_cluster.go
jameszwang: Updated using clusterService.Create() function to create cluster.

tests/e2e/test_rosacli_cluster.go
jameszwang: Updated using case id as name prefix.

tests/e2e/test_rosacli_cluster.go
jameszwang: Updated using ocmResourceService delete account roles and operator roles.

PR #2429, opened 2024-09-05
OCM-10943 | feat: show disk size in output for list node pools

willkutler:
https://issues.redhat.com/browse/OCM-10943

- displays disk size in list node pool cmd
- defaults to "default" which is the same as machinepool
- unit testing

### DIFF
--- a/pkg/ocm/output/nodepools.go
+++ b/pkg/ocm/output/nodepools.go
@@ -22,6 +22,7 @@ import (
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 
+	"github.com/openshift/rosa/pkg/helper"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/output"
 )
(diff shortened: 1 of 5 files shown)


pkg/ocm/output/nodepools.go
```diff
@@ -134,3 +135,15 @@ func PrintNodePoolManagementUpgrade(upgrade *cmv1.NodePoolManagementUpgrade) str
 
 	return ""
 }
+
+func PrintNodePoolDiskSize(aws *cmv1.AWSNodePool) string {
+	diskSizeStr := "default"
```
marcolan018, line 140: why not use the number of default value, but use "default" string here?

gdbranco: I suppose in case default changes internally. Even if the default value is returned in the API it will change it to whatever the number is. It is just in case some record does not have the value set yet I suppose

willkutler: yeah I think @gdbranco is right, I also just want the default to be the same as it already is for classic machinepools. the "default" string should never appear anyway once the API changes are merged since that value will always be set

PR #2430, opened 2024-09-05
OCM-7315 | feat: create and edit client support for configurable registries

chenz4027:
https://issues.redhat.com/browse/OCM-7315

### DIFF
--- a/cmd/edit/cluster/cmd.go
+++ b/cmd/edit/cluster/cmd.go
@@ -27,6 +27,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/openshift/rosa/pkg/aws"
+	"github.com/openshift/rosa/pkg/clusterregistryconfig"
 	"github.com/openshift/rosa/pkg/helper"
 	"github.com/openshift/rosa/pkg/helper/roles"
 	"github.com/openshift/rosa/pkg/input"
@@ -57,7 +58,17 @@ var args struct {
 
 	// Other options
 	additionalAllowedPrincipals []string
+
+	// Cluster Registry
+	allowedRegistries          []string
+	blockedRegistries          []string
+	insecureRegistries         []string
+	allowedRegistriesForImport string
+	platformAllowlist          string
+	additionalTrustedCa        string
 }
+
+var clusterRegistryConfigArgs *clusterregistryconfig.ClusterRegistryConfigArgs
 
 var Cmd = &cobra.Command{
 	Use:   "cluster",
@@ -163,6 +174,8 @@ func init() {
 			"to be added to the Hosted Control Plane's VPC Endpoint Service to enable additional "+
 			"VPC Endpoint connection requests to be automatically accepted.",
 	)
+
+	clusterRegistryConfigArgs = clusterregistryconfig.AddClusterRegistryConfigFlags(Cmd)
 }
 
 func run(cmd *cobra.Command, _ []string) {
@@ -176,9 +189,13 @@ func run(cmd *cobra.Command, _ []string) {
 		changedFlags := false
 		for _, flag := range []string{"expiration-time", "expiration", "private",
 			"disable-workload-monitoring", "http-proxy", "https-proxy", "no-proxy",
-			"additional-trust-bundle-file", "additional-allowed-principals", "audit-log-arn"} {
+			"additional-trust-bundle-file", "additional-allowed-principals", "audit-log-arn",
+			"registry-config-allowed-registries", "registry-config-blocked-registries",
+			"registry-config-insecure-registries", "allowed-registries-for-import",
+			"registry-config-platform-allowlist", "registry-config-additional-trusted-ca"} {
 			if cmd.Flags().Changed(flag) {
 				changedFlags = true
+				break
 			}
 		}
 		if !changedFlags {
@@ -615,6 +632,30 @@ func run(cmd *cobra.Command, _ []string) {
 		clusterConfig.AdditionalAllowedPrincipals = additionalAllowedPrincipals
 	}
 
+	clusterRegistryConfigArgs, err = clusterregistryconfig.GetClusterRegistryConfigOptions(
+		cmd.Flags(), clusterRegistryConfigArgs, aws.IsHostedCP(cluster), cluster)
+	if err != nil {
+		r.Reporter.Errorf("%s", err)
+		os.Exit(1)
+	}
+	if clusterRegistryConfigArgs != nil {
+		allowedRegistries, blockedRegistries, insecureRegistries,
+			additionalTrustedCa, allowedRegistriesForImport := clusterregistryconfig.GetClusterRegistryConfigArgs(
+			clusterRegistryConfigArgs)
+		clusterConfig.AllowedRegistries = allowedRegistries
+		clusterConfig.BlockedRegistries = blockedRegistries
+		clusterConfig.InsecureRegistries = insecureRegistries
+		if additionalTrustedCa != "" {
+			ca, err := clusterregistryconfig.BuildAdditionalTrustedCAFromInputFile(additionalTrustedCa)
+			if err != nil {
+				r.Reporter.Errorf("%s", err)
+				os.Exit(1)
+			}
+			clusterConfig.AdditionalTrustedCa = ca
+			clusterConfig.AdditionalTrustedCaFile = additionalTrustedCa
+		}
+		clusterConfig.AllowedRegistriesForImport = allowedRegistriesForImport
+	}
 	if auditLogRole != nil {
 		clusterConfig.AuditLogRoleARN = new(string)
 		*clusterConfig.AuditLogRoleARN = *auditLogRole
--- a/pkg/clusterregistryconfig/flags.go
+++ b/pkg/clusterregistryconfig/flags.go
@@ -1,0 +1,287 @@
+package clusterregistryconfig
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"github.com/openshift/rosa/pkg/helper"
+	"github.com/openshift/rosa/pkg/input"
+	"github.com/openshift/rosa/pkg/interactive"
+	"github.com/openshift/rosa/pkg/ocm"
+)
+
+const (
+	allowedRegistriesFlag          = "registry-config-allowed-registries"
+	insecureRegistriesFlag         = "registry-config-insecure-registries"
+	blockedRegistriesFlag          = "registry-config-blocked-registries"
+	platformAllowlistFlag          = "registry-config-platform-allowlist"
+	additionalTrustedCaPathFlag    = "registry-config-additional-trusted-ca"
+	allowedRegistriesForImportFlag = "registry-config-allowed-registries-for-import"
+)
+
+type ClusterRegistryConfigArgs struct {
+	allowedRegistries          []string
+	blockedRegistries          []string
+	insecureRegistries         []string
+	allowedRegistriesForImport string
+	platformAllowlist          string
+	additionalTrustedCa        string
+}
+
+func AddClusterRegistryConfigFlags(cmd *cobra.Command) *ClusterRegistryConfigArgs {
+	args := &ClusterRegistryConfigArgs{}
+
+	cmd.Flags().StringSliceVar(
+		&args.allowedRegistries,
+		allowedRegistriesFlag,
+		nil,
+		"A comma-separated list of registries for which image pull and push actions are allowed.",
+	)
+
+	cmd.Flags().StringSliceVar(
+		&args.insecureRegistries,
+		insecureRegistriesFlag,
+		nil,
+		"A comma-separated list of registries which do not have a valid TLS certificate or only support HTTP connections.",
+	)
+
+	cmd.Flags().StringSliceVar(
+		&args.blockedRegistries,
+		blockedRegistriesFlag,
+		nil,
+		"A comma-separated list of registries for which image pull and push actions are denied.",
+	)
+
+	cmd.Flags().StringVar(
+		&args.allowedRegistriesForImport,
+		allowedRegistriesForImportFlag,
+		"",
+		"Limits the container image registries from which normal users can import images. "+
+			"The format should be a comma-separated list of 'domainName:insecure'. "+
+			"'domainName' specifies a domain name for the registry. "+
+			"'insecure' indicates whether the registry is secure or insecure.",
+	)
+
+	cmd.Flags().StringVar(
+		&args.platformAllowlist,
+		platformAllowlistFlag,
+		"",
+		"A reference to the id of the list of registries that needs to be whitelisted for the platform to work. "+
+			"It can be omitted at creation and updating and its lifecycle can be managed separately if needed.",
+	)
+	cmd.Flags().MarkHidden("platform-allowlist")
+
+	cmd.Flags().StringVar(
+		&args.additionalTrustedCa,
+		additionalTrustedCaPathFlag,
+		"",
+		"A json file containing the registry hostname as the key,"+
+			" and the PEM-encoded certificate as the value, for each additional registry CA to trust.")
+
+	return args
+}
+
+func GetClusterRegistryConfigArgs(args *ClusterRegistryConfigArgs) (
+	[]string, []string, []string, string, string) {
+	return args.allowedRegistries, args.blockedRegistries,
+		args.insecureRegistries, args.additionalTrustedCa, args.allowedRegistriesForImport
+}
+
+func GetClusterRegistryConfigOptions(cmd *pflag.FlagSet,
+	args *ClusterRegistryConfigArgs, isHostedCP bool, cluster *cmv1.Cluster) (
+	*ClusterRegistryConfigArgs, error) {
+
+	if args != nil && !isHostedCP {
+		return nil, fmt.Errorf("Setting the registry config is only supported for hosted clusters")
+	}
+
+	result := &ClusterRegistryConfigArgs{}
+
+	result.allowedRegistries = args.allowedRegistries
+	result.insecureRegistries = args.insecureRegistries
+	result.blockedRegistries = args.blockedRegistries
+	result.additionalTrustedCa = args.additionalTrustedCa
+	result.allowedRegistriesForImport = args.allowedRegistriesForImport
+	result.platformAllowlist = args.platformAllowlist
+
+	if !IsClusterRegistryConfigSetViaCLI(cmd) && !interactive.Enabled() {
+		return nil, nil
+	}
+
+	defaultAllowedRegistries := args.allowedRegistries
+	defaultBlockedRegistries := args.blockedRegistries
+	defaultInsecureRegistries := args.insecureRegistries
+	defaultAllowedRegistriesForImport := args.allowedRegistriesForImport
+
+	if defaultAllowedRegistries == nil {
+		defaultAllowedRegistries = cluster.RegistryConfig().RegistrySources().AllowedRegistries()
+	}
+	if defaultBlockedRegistries == nil {
+		defaultBlockedRegistries = cluster.RegistryConfig().RegistrySources().BlockedRegistries()
+	}
+	if defaultInsecureRegistries == nil {
+		defaultInsecureRegistries = cluster.RegistryConfig().RegistrySources().InsecureRegistries()
+	}
+	if defaultAllowedRegistriesForImport == "" {
+		var list []string
+		for _, location := range cluster.RegistryConfig().AllowedRegistriesForImport() {
+			list = append(list, fmt.Sprintf("%s:%s", location.DomainName(), strconv.FormatBool(location.Insecure())))
+		}
+		defaultAllowedRegistriesForImport = strings.Join(list, ",")
+	}
+
+	enableRegistriesConfig := IsClusterRegistryConfigSetViaCLI(cmd)
+
+	if !enableRegistriesConfig && interactive.Enabled() && isHostedCP {
+		updateRegistriesConfigValue, err := interactive.GetBool(interactive.Input{
+			Question: "Enable registries config",
+			Default:  enableRegistriesConfig,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("Expected a valid value: %s", err)
+		}
+		enableRegistriesConfig = updateRegistriesConfigValue
+	}
+
+	if enableRegistriesConfig && interactive.Enabled() {
+		allowedRegistriesInputs, err := interactive.GetString(interactive.Input{
+			Question: "Allowed Registries",
+			Help:     cmd.Lookup(allowedRegistriesFlag).Usage,
+			Default:  strings.Join(defaultAllowedRegistries, ","),
+		})
+		if err != nil {
+			return nil, fmt.Errorf("Expected a valid value for allowed registries: %s", err)
+		}
+		result.allowedRegistries = helper.HandleEmptyStringOnSlice(strings.Split(allowedRegistriesInputs, ","))
+
+		blockedRegistriesInputs, err := interactive.GetString(interactive.Input{
+			Question: "Blocked Registries",
+			Help:     cmd.Lookup(blockedRegistriesFlag).Usage,
+			Default:  strings.Join(defaultBlockedRegistries, ","),
+		})
+		if err != nil {
+			return nil, fmt.Errorf("Expected a valid value for blocked registries: %s", err)
+		}
+		result.blockedRegistries = helper.HandleEmptyStringOnSlice(strings.Split(blockedRegistriesInputs, ","))
+
+		insecureRegistriesInputs, err := interactive.GetString(interactive.Input{
+			Question: "Insecure Registries",
+			Help:     cmd.Lookup(insecureRegistriesFlag).Usage,
+			Default:  strings.Join(defaultInsecureRegistries, ","),
+		})
+		if err != nil {
+			return nil, fmt.Errorf("Expected a valid value for insecure registries: %s", err)
+		}
+		result.insecureRegistries = helper.HandleEmptyStringOnSlice(strings.Split(insecureRegistriesInputs, ","))
+
+		result.allowedRegistriesForImport, err = interactive.GetString(interactive.Input{
+			Question: "Allowed Registries For Import",
+			Help:     cmd.Lookup(allowedRegistriesForImportFlag).Usage,
+			Default:  defaultAllowedRegistriesForImport,
+			Validators: []interactive.Validator{
+				ocm.ValidateAllowedRegistriesForImport,
+			},
+		})
+		if err != nil {
+			return nil, fmt.Errorf("Expected a valid value for allowed registries for import: %s", err)
+		}
+
+		result.additionalTrustedCa, err = interactive.GetString(interactive.Input{
+			Question: "Registry Additional Trusted CA",
+			Help:     cmd.Lookup(additionalTrustedCaPathFlag).Usage,
+			Default:  args.additionalTrustedCa,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("Expected a valid certificate: %s", err)
+		}
+	}
+	if err := ocm.ValidateAllowedRegistriesForImport(result.allowedRegistriesForImport); err != nil {
+		return nil, fmt.Errorf("Expected valid allowed registries for import values: %v", err)
+	}
+
+	return result, nil
+}
+
+func IsClusterRegistryConfigSetViaCLI(cmd *pflag.FlagSet) bool {
+	for _, parameter := range []string{allowedRegistriesFlag,
+		insecureRegistriesFlag, blockedRegistriesFlag, platformAllowlistFlag,
+		allowedRegistriesForImportFlag, additionalTrustedCaPathFlag} {
+
+		if cmd.Changed(parameter) {
+			return true
+		}
+	}
+
+	return false
+}
+
+func BuildRegistryConfigOptions(spec ocm.Spec) string {
+	command := ""
+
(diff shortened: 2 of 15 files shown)


cmd/edit/cluster/cmd.go
```diff
@@ -463,6 +541,91 @@ func run(cmd *cobra.Command, _ []string) {
...
+			Help:     cmd.Flags().Lookup("platform-allowlist").Usage,
+			Default:  platformAllowlist,
+		})
+		if err != nil {
+			r.Reporter.Errorf("Expected a valid value for platform allowlist: %s", err)
+			os.Exit(1)
+		}
+
+		additionalTrustedCa, err = interactive.GetString(interactive.Input{
+			Question: "Additional Trusted CA file path",
```
mikeshng, line 619: Should we say registry here?

pkg/ocm/clusters_test.go
```diff
@@ -27,6 +27,47 @@ var _ = Describe("New Operator Iam Role From Cmv1", func() {
...
+		It("Returns nil", func() {
+			config := Spec{}
+			output := BuildRegistryConfig(config)
+			Expect(output).To(BeNil())
+		})
+		It("Returns expected output", func() {
+			config := Spec{
+				AllowedRegistries:          []string{"example.com", "quay.io"},
+				InsecureRegistries:         []string{"insecure.com"},
+				AllowedRegistriesForImport: []string{"quay.io"},
```
mikeshng, line 41: It's missing some fields here. I counted 6 and this is only testing 3.

cmd/edit/cluster/clusterregistryconfig_test.go
```diff
@@ -0,0 +1,41 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(ca["userRegistry.io"]).To(
+				Equal("-----BEGIN CERTIFICATE-----\n/abc\n-----END CERTIFICATE-----"))
+
+		})
+		It("KO: fail if the spec file is invalid", func() {
+			_, err := BuildAdditionalTrustedCAFromInputFile("not-exist")
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(
+				Equal("Expected a valid additional trusted certificate spec file: open not-exist: no such file or directory"))
```
ziccardi, line 38:
```suggestion
			Expect(err).To(MatchError("Expected a valid additional trusted certificate spec file: open not-exist: no such file or directory"))
```

cmd/edit/cluster/cmd.go
```diff
@@ -463,6 +541,91 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
+	/*******  Registries Config *******/
+	updateRegistriesConfig := false
+	if insecureRegistries != nil || blockedRegistries != nil || allowedRegistriesForImport != nil ||
+		allowedRegistries != nil || platformAllowlist != "" || additionalTrustedCa != "" {
+		updateRegistriesConfig = true
+	}
```
ziccardi, line 549: This could be simplified
```suggestion
	updateRegistriesConfig := insecureRegistries != nil || blockedRegistries != nil || allowedRegistriesForImport != nil ||	allowedRegistries != nil || platformAllowlist != "" || additionalTrustedCa != ""
```

cmd/edit/cluster/cmd.go
ziccardi, line 619: It seems to me that a path to a file containing a single certificate is required here, not a certificate chain or a keystore, correct?

In both cases, I think we should specify whether we expect the input file to be DER or PEM encoded.

chenz4027:
Thanks @mikeshng @ziccardi for your thorough reviews, I'll look into the threads today and address those 
👍

cmd/edit/cluster/spec.json
andreadecorte, line 1: nit but probably worth having a more specific name like specRegistryAddiitionalTrustedCa

cmd/edit/cluster/cmd.go
chenz4027: It actually expects it in a json in {registry.io: cert content} format, let me add a validation for it.

cmd/edit/cluster/clusterregistryconfig_test.go
chenz4027: Updated to use match error.

cmd/edit/cluster/clusterregistryconfig.go
```diff
@@ -0,0 +1,44 @@
...
+
+func BuildAdditionalTrustedCAFromInputFile(specPath string) (map[string]string, error) {
+	specJson, err := input.UnmarshalInputFile(specPath)
+	if err != nil {
+		return nil, fmt.Errorf("Expected a valid additional trusted certificate spec file: %v", err)
+	}
+	form := make(map[string]string)
+	for k, v := range specJson {
+		form[k] = v.(string)
+	}
```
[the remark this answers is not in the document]
chenz4027: Added the check for the spec

cmd/edit/cluster/cmd.go
chenz4027: Refined the updateRegistriesConfig

cmd/edit/cluster/cmd.go
chenz4027: Changed to "Registry Additional Trusted CA", would it be better?

cmd/edit/cluster/spec.json
chenz4027: Updated the name

cmd/edit/cluster/cmd.go
```diff
@@ -163,6 +171,49 @@ func init() {
...
+		"blocked-registries",
+		nil,
+		"A comma-separated list of registries for which image pull and push actions are denied.",
+	)
+
+	flags.StringSliceVar(
+		&args.allowedRegistriesForImport,
+		"allowed-registries-for-import",
+		nil,
+		"A comma-separated list of registries that limits the container image registries that normal users may import",
```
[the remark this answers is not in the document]
chenz4027: I changed it to take registryName:bool format, would that be okay?

cmd/edit/cluster/cmd.go
```diff
@@ -163,6 +171,49 @@ func init() {
...
+		"",
+		"A comma-separated list of internal registries that needs to be whitelisted for the platform to work. "+
+			"It can be omitted at creation and updating and its lifecycle can be managed separately if needed.",
+	)
+
+	flags.StringVar(
+		&args.additionalTrustedCa,
+		"additional-trusted-ca",
+		"",
+		"A map containing the registry hostname as the key,"+
```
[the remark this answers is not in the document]
chenz4027: Changed to "registry-additional-trusted-ca", WDYT?

cmd/edit/cluster/cmd.go
chenz4027: Added the validator

chenz4027:
Still need to work on adding more tests, moving to draft

cmd/edit/cluster/clusterregistryconfig_test.go
```diff
@@ -0,0 +1,41 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(ca["userRegistry.io"]).To(
+				Equal("-----BEGIN CERTIFICATE-----\n/abc\n-----END CERTIFICATE-----"))
+
+		})
+		It("KO: fail if the spec file is invalid", func() {
+			_, err := BuildAdditionalTrustedCAFromInputFile("not-exist")
+			Expect(err).To(HaveOccurred())
+			Expect(err).To(MatchError(
+				"Expected a valid additional trusted certificate spec file: open not-exist: no such file or directory"))
```
ziccardi, line 38: When using `MatchError`, `Expect(err).To(HaveOccurred())` is not needed: `MatchError` will fail if the `err` is nil
```suggestion
			Expect(err).To(MatchError(
				"Expected a valid additional trusted certificate spec file: open not-exist: no such file or directory"))
```

cmd/edit/cluster/clusterregistryconfig.go
```diff
@@ -0,0 +1,48 @@
...
+	specJson, err := input.UnmarshalInputFile(specPath)
+	if err != nil {
+		return nil, fmt.Errorf("Expected a valid additional trusted certificate spec file: %v", err)
+	}
+	form := make(map[string]string)
+	var ok bool
+	for k, v := range specJson {
+		form[k], ok = v.(string)
+		if !ok {
+			return nil, fmt.Errorf("expect value to be string, got %T", v)
```
ziccardi, line 37: Perhaps we could add a bit more detail here. Since we already have the key, it would be more convenient for the user if we included that as well.

I recommend printing the received value instead of just the type (or both, if possible).

All the other errors start with an uppercase letter. Is this because they are printed directly to the user?
If that's the case, this one should also start with an uppercase letter.

Otherwise, Go's best practice is to always return errors in lowercase, allowing them to be easily concatenated to log messages like:
```go
fmt.Println("An error has occurred: ", err)
```

cmd/edit/cluster/clusterregistryconfig_test.go
chenz4027: Thanks @ziccardi 👍

cmd/edit/cluster/clusterregistryconfig.go
chenz4027: Thanks @ziccardi , will let me add more info to the error log & look into upper case

pkg/ocm/clusters_test.go
```diff
@@ -27,6 +27,58 @@ var _ = Describe("New Operator Iam Role From Cmv1", func() {
...
+					Insecure(!true).DomainName("quay.io")).
+				RegistrySources(cmv1.NewRegistrySources().InsecureRegistries("insecure.com").
+					BlockedRegistries("insecure.com").
+					AllowedRegistries([]string{"example.com", "quay.io"}...))
+			Expect(output).To(Equal(expectedOutput))
+		})
+	})
+})
+
+var _ = Context("Generate a query", func() {
```
chenz4027, line 64: Not related to the PR, just add some coverage.

cmd/edit/cluster/cmd.go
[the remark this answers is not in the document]
mikeshng: I am ok with the updated description. Feel free to resolve this thread.

cmd/edit/cluster/cmd.go
```diff
@@ -463,6 +545,98 @@ func run(cmd *cobra.Command, _ []string) {
...
+			Question: "Allowed Registries For Import",
+			Help:     cmd.Flags().Lookup("allowed-registries-for-import").Usage,
+			Default:  allowedRegistriesForImport,
+			Validators: []interactive.Validator{
+				ocm.ValidateAllowedRegistriesForImport,
+			},
+		})
+		if err != nil {
+			r.Reporter.Errorf("expected a valid value for allowed registries for import: %s", err)
+			os.Exit(1)
```
ziccardi, line 612: Upon reviewing the reporter code, I noticed that the Reporter object outputs the error to the console. In my opinion, r.Reporter.Errorf should follow the convention of using an uppercase letter, though I'll defer that decision to someone with more familiarity with the project.

The Go best practice regarding lowercase error messages applies specifically to returned errors from functions, allowing them to be easily concatenated with other strings, as you're doing with the err variable in this case.

ziccardi (changes requested):
I've noticed some inconsistency in the messages passed to Reporter.Errorf, where sometimes they are lowercase and other times uppercase.

Since Errorf outputs the message directly to stderr, I believe it would be better for these messages to start with an uppercase letter. However, I defer the final decision to those with a deeper understanding of the project.

Regardless of the choice we make, I think it's important to maintain consistency—either lowercase or uppercase across the board.

To clarify a previous point: when I mentioned Go's best practice for returning errors in lowercase, I was referring specifically to errors returned by functions, not those written to the console.

cmd/edit/cluster/cmd.go
```diff
@@ -163,6 +172,52 @@ func init() {
...
+		nil,
+		"A comma-separated list of registries for which image pull and push actions are denied.",
+	)
+
+	flags.StringVar(
+		&args.allowedRegistriesForImport,
+		"allowed-registries-for-import",
+		"",
+		"Limits the container image registries from which normal users can import images. "+
+			"The format should be a comma-separated list of 'domanName:insecure'. "+
```
andreadecorte, line 202:
```suggestion
			"The format should be a comma-separated list of 'domainName:insecure'. "+
```

cmd/edit/cluster/cmd.go
```diff
@@ -163,6 +172,52 @@ func init() {
...
+			"'domainName' specifies a domain name for the registry. "+
+			"'insecure' indicates whether the registry is secure or insecure.",
+	)
+
+	flags.StringVar(
+		&args.platformAllowlist,
+		"platform-allowlist",
+		"",
+		"A comma-separated list of internal registries that needs to be whitelisted for the platform to work. "+
```
andreadecorte, line 211: this is a reference to a RegistryAllowlistId, not a list. 
I think we should change that comment + mark this as hidden for the initial phase, as we don't expect changes at the beginning, so no need to expose it to the user

cmd/edit/cluster/cmd.go
```diff
@@ -463,6 +545,98 @@ func run(cmd *cobra.Command, _ []string) {
...
+			Validators: []interactive.Validator{
+				ocm.ValidateAllowedRegistriesForImport,
+			},
+		})
+		if err != nil {
+			r.Reporter.Errorf("expected a valid value for allowed registries for import: %s", err)
+			os.Exit(1)
+		}
+
+		platformAllowlist, err = interactive.GetString(interactive.Input{
```
andreadecorte, line 615: I think we should remove this from interactive flow, let's just leave it as hidden flag.
Reason is that this is a specific flow in case platform registries change, otherwise user doesn't need to care about it.

pkg/ocm/helpers.go
```diff
@@ -151,6 +151,45 @@ func ValidateHTTPProxy(val interface{}) error {
...
+		for _, registry := range registries {
+			if !idRE.MatchString(registry) {
+				return fmt.Errorf("invalid identifier '%s' for 'allowed registries for import.' "+
+					"Should be in a <registry>:<boolean> format. "+
+					"The boolean indicates whether the registry is secure or not.", registry)
+			}
+		}
+	case reflect.Slice:
+		if reflect.TypeOf(input).Elem().Kind() != reflect.String {
+			return fmt.Errorf("unable to verify allowed , incompatible type, expected slice of string got: '%s'",
```
andreadecorte, line 184: missing something after allowed

pkg/ocm/helpers_test.go
```diff
@@ -473,6 +473,40 @@ var _ = Describe("ClusterNameValidator()", func() {
...
+var _ = Describe("ValidateRegistryAdditionalCa", func() {
+	DescribeTable("ValidateRegistryAdditionalCa() test cases", func(input map[string]string, shouldErr bool) {
+		err := ValidateRegistryAdditionalCa(input)
+		if shouldErr {
+			Expect(err).To(HaveOccurred())
+		} else {
+			Expect(err).NotTo(HaveOccurred())
+		}
+	},
+		Entry("should error when a they key is not a PEM certificate", map[string]string{
```
andreadecorte, line 485:
```suggestion
		Entry("should error when the value is not a PEM certificate", map[string]string{
```

pkg/ocm/helpers_test.go
```diff
@@ -473,6 +473,40 @@ var _ = Describe("ClusterNameValidator()", func() {
...
+			"registry.io": "abc",
+		}, true),
+		Entry("should not error when a they key is a PEM certificate", map[string]string{
```
andreadecorte, line 488:
```suggestion
		Entry("should not error when a the value is a PEM certificate", map[string]string{
```

cmd/edit/cluster/cmd.go
```diff
@@ -615,6 +789,53 @@ func run(cmd *cobra.Command, _ []string) {
...
+		clusterConfig.BlockedRegistries = blockedRegistries
+	}
+	if insecureRegistries != nil {
+		clusterConfig.InsecureRegistries = insecureRegistries
+	}
+	if allowedRegistriesForImport != "" {
+		obj := map[string]bool{}
+		list := helper.HandleEmptyStringOnSlice(strings.Split(allowedRegistriesForImport, ","))
+		for _, registry := range list {
+			registryObj := helper.HandleEmptyStringOnSlice(strings.Split(registry, ":"))
```
andreadecorte, line 805: we can have a registry with a port and it will have a ":" inside, do you have a test on this?
e.g. "registry.io:80:insecure" works correctly?

cmd/edit/cluster/cmd.go
chenz4027: Thanks @ziccardi 👍 Will review the output. If it's directly to console, will start with upper case

chenz4027:
Updated the code, @andreadecorte @ziccardi Could you help take another review?
I'll have a separate pr for create/edit.

cmd/edit/cluster/clusterregistryconfig.go
```diff
@@ -0,0 +1,73 @@
...
+		registryObj := helper.HandleEmptyStringOnSlice(strings.Split(registry, ":"))
+		objLen := len(registryObj)
+		if objLen >= 2 {
+			// insecure will always be the last item
+			insecure := registryObj[objLen-1]
+			// the registryName will remain the same
+			registryName := strings.Join(registryObj[:objLen-1], ":")
+			boolValue, err := strconv.ParseBool(insecure)
+			if err != nil {
+				return obj, fmt.Errorf("failed to convert the value '%s' to bool: %s", insecure, err)
```
ziccardi, line 67:
```suggestion
				return nil, fmt.Errorf("failed to convert the value '%s' to bool: %s", insecure, err)
```

cmd/edit/cluster/cmd.go
```diff
@@ -176,7 +231,9 @@ func run(cmd *cobra.Command, _ []string) {
 		changedFlags := false
 		for _, flag := range []string{"expiration-time", "expiration", "private",
 			"disable-workload-monitoring", "http-proxy", "https-proxy", "no-proxy",
-			"additional-trust-bundle-file", "additional-allowed-principals", "audit-log-arn"} {
+			"additional-trust-bundle-file", "additional-allowed-principals", "audit-log-arn",
+			"allowed-registries", "blocked-registries", "insecure-registries", "allowed-registries-for-import",
+			"platform-allowlist", "registry-additional-trusted-ca"} {
 			if cmd.Flags().Changed(flag) {
 				changedFlags = true
 			}
```
ziccardi, line 239: As soon as we set `changeFlags` to `true` there is no reason to keep iterating
```suggestion
			if cmd.Flags().Changed(flag) {
				changedFlags = true
				break
			}
```

ziccardi (commented):
Added minor, optional comments.
Looks good to me overall.

pkg/clusterregistryconfig/flags.go
```diff
@@ -0,0 +1,252 @@
...
+
+		if cmd.Changed(parameter) {
+			return true
+		}
+	}
+
+	return false
+}
+
+func BuildRegistryConfig(args *ClusterRegistryConfigArgs) (*cmv1.ClusterRegistryConfigBuilder, error) {
```
chenz4027, line 151: introduced the panic issue `panic: runtime error: invalid memory address or nil pointer dereference`
Looking...

cmd/edit/cluster/clusterregistryconfig.go
chenz4027: Updated to nil

pkg/ocm/helpers.go
chenz4027: Added text "element"

pkg/ocm/helpers_test.go
chenz4027: Updated the text

cmd/edit/cluster/cmd.go
chenz4027: Removed from interactive mode

cmd/edit/cluster/cmd.go
chenz4027: Thanks for the catch, changed the text

cmd/edit/cluster/cmd.go
chenz4027: Hide the args and changed the text to "The id of the internal registry that needs to be whitelisted for the platform to work. It can be omitted at creation and updating and its lifecycle can be managed separately if needed."

chenz4027:
@mikeshng @ziccardi @andreadecorte Sorry for the re-write, the scope extends from edit to both create and edit, thus the refactor. Would you mind taking some additional looks? Thanks in advance :)

pkg/clusterregistryconfig/flags.go
```diff
@@ -0,0 +1,350 @@
...
+			"The format should be a comma-separated list of 'domainName:insecure'. "+
+			"'domainName' specifies a domain name for the registry. "+
+			"'insecure' indicates whether the registry is secure or insecure.",
+	)
+
+	cmd.Flags().StringVar(
+		&args.platformAllowlist,
+		platformAllowlistFlag,
+		"",
+		"The id of the internal registry that needs to be whitelisted for the platform to work. "+
```
andreadecorte, line 74:
```suggestion
		"A reference to the id of the list of registries that needs to be whitelisted for the platform to work. "+
```

chenz4027: Updated the description

pkg/clusterregistryconfig/flags.go
```diff
@@ -0,0 +1,350 @@
...
+func GetClusterRegistryConfigArgs(args *ClusterRegistryConfigArgs) (
+	[]string, []string, []string, string, string) {
+	return args.allowedRegistries, args.blockedRegistries,
+		args.insecureRegistries, args.additionalTrustedCa, args.allowedRegistriesForImport
+}
+
+func GetClusterRegistryConfigOptions(cmd *pflag.FlagSet,
+	args *ClusterRegistryConfigArgs, isHostedCP bool) (
+	*ClusterRegistryConfigArgs, error) {
+	var err error
```
ziccardi, line 98: No need to declare an error variable here

```suggestion
	var err error
```

pkg/clusterregistryconfig/flags.go
```diff
@@ -0,0 +1,350 @@
...
+		result.additionalTrustedCa, err = interactive.GetString(interactive.Input{
+			Question: "Registry Additional Trusted CA",
+			Help:     cmd.Lookup("registry-additional-trusted-ca").Usage,
+			Default:  args.additionalTrustedCa,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("Expected a valid certificate: %s", err)
+		}
+	}
+	if err = ocm.ValidateAllowedRegistriesForImport(result.allowedRegistriesForImport); err != nil {
```
ziccardi, line 184:
```suggestion
	if err := ocm.ValidateAllowedRegistriesForImport(result.allowedRegistriesForImport); err != nil {
```

pkg/clusterregistryconfig/flags.go
```diff
@@ -0,0 +1,350 @@
...
+		return nil, fmt.Errorf("Expected valid allowed registries for import values: %v", err)
+	}
+
+	return result, err
```
ziccardi, line 188:
```suggestion
	return result, nil
```

ziccardi (changes requested):
Reusing an `err` variable throughout an entire function diminishes the function's readability. This is particularly true when the function concludes with `return something, err`, forcing the reader to trace through the entire code to grasp which value is ultimately being returned.

To enhance clarity, errors should be handled and returned at the earliest possible point, and the scope of the `err` variable should be kept as confined as feasible. This approach limits the cognitive overhead and ensures that the code remains intuitive to follow.

pkg/clusterregistryconfig/flags.go
chenz4027: Thanks @ziccardi , updated with the new change :)

pkg/clusterregistryconfig/flags.go
```diff
@@ -0,0 +1,349 @@
...
+				cmv1.NewRegistryAllowlist().ID(args.platformAllowlist))
+		}
+		if args.allowedRegistriesForImport != "" {
+			obj, err := BuildAllowedRegistriesForImport(args.allowedRegistriesForImport)
+			if err != nil {
+				return nil, fmt.Errorf("Failed to build allowed registries for import, received error: %s", err)
+			}
+
+			//construct the location list
+			locationList := []*cmv1.RegistryLocationBuilder{}
```
ziccardi, line 249: No need for this instantiation: all slice functions (`append`, `len`) are nil safe.
```suggestion
			var locationList []*cmv1.RegistryLocationBuilder
```

chenz4027: Updated the var

chenz4027:
Updated with recommendations, @andreadecorte @ziccardi Could you help take a final look? 👀

pkg/clusterregistryconfig/flags.go
```diff
@@ -0,0 +1,349 @@
...
+	result.blockedRegistries = args.blockedRegistries
+	result.additionalTrustedCa = args.additionalTrustedCa
+	result.allowedRegistriesForImport = args.allowedRegistriesForImport
+	result.platformAllowlist = args.platformAllowlist
+
+	if !IsClusterRegistryConfigSetViaCLI(cmd) {
+		if !interactive.Enabled() {
+			return nil, nil
+		}
+	}
```
ziccardi, line 116: This could be simplified
```suggestion
	if !IsClusterRegistryConfigSetViaCLI(cmd) &&  !interactive.Enabled() {
			return nil, nil
	}
```

ziccardi (approved):
Added just a very small and optional comment.

Looks good to me.

pkg/clusterregistryconfig/flags.go
chenz4027: Updated :)

andreadecorte (approved):
tested locally too, it looks good for creation/editing, thanks

pkg/clusterregistryconfig/flags.go
```diff
@@ -0,0 +1,369 @@
...
+	"github.com/spf13/pflag"
+
+	"github.com/openshift/rosa/pkg/helper"
+	"github.com/openshift/rosa/pkg/input"
+	"github.com/openshift/rosa/pkg/interactive"
+	"github.com/openshift/rosa/pkg/ocm"
+)
+
+const (
+	allowedRegistriesFlag          = "allowed-registries"
```
gdbranco, line 19: Can we include a prefix of registry-config on these? It helps when looking at the help output or understanding what the flags are for

pkg/clusterregistryconfig/flags.go
```diff
@@ -0,0 +1,369 @@
...
+			Default:  defaultAllowedRegistriesForImport,
+			Validators: []interactive.Validator{
+				ocm.ValidateAllowedRegistriesForImport,
+			},
+		})
+		if err != nil {
+			return nil, fmt.Errorf("Expected a valid value for allowed registries for import: %s", err)
+		}
+
+		result.additionalTrustedCa, err = interactive.GetString(interactive.Input{
```
gdbranco, line 194: For certificates we usually supply the file and pull the information from it if I'm not mistaken
[Example](https://github.com/openshift/rosa/blob/965cea4de7c13531d682a6e650a111a76a6f611e/cmd/create/cluster/cmd.go#L2837) from additional trusted ca for proxy config

pkg/clusterregistryconfig/flags.go
```diff
@@ -0,0 +1,369 @@
...
+	"github.com/openshift/rosa/pkg/interactive"
+	"github.com/openshift/rosa/pkg/ocm"
+)
+
+const (
+	allowedRegistriesFlag          = "allowed-registries"
+	insecureRegistriesFlag         = "insecure-registries"
+	blockedRegistriesFlag          = "blocked-registries"
+	platformAllowlistFlag          = "platform-allowlist"
+	additionalTrustedCaFlag        = "registry-additional-trusted-ca"
```
gdbranco, line 23: Please mention this should be a file path on the flag

pkg/clusterregistryconfig/flags.go
gdbranco: hmm, I see this would not work as it is a json file with possibly many certificates, nvm

pkg/ocm/clusters.go
```diff
@@ -174,6 +174,15 @@ type Spec struct {
...
 	AdditionalControlPlaneSecurityGroupIds []string
+
+	// Registry Config
+	AllowedRegistries          []string
+	BlockedRegistries          []string
+	InsecureRegistries         []string
+	AllowedRegistriesForImport string
+	PlatformAllowlist          string
+	AdditionalTrustedCa        string
+	ClusterRegistryConfig      *cmv1.ClusterRegistryConfigBuilder
```
gdbranco, line 185: Does it make sense to include the builder in this spec?
I feel like either the builder or the attributes should be enough :thinking:

pkg/clusterregistryconfig/flags.go
chenz4027: Thanks so much @gdbranco , what kind of prefix are you referring to?

gdbranco: Like instead of 'allowed-registries' something like 'registry-config-allowed-registries' just so it is clear for what the flag is for

Example, '--default-ingress-excluded-namespaces' this makes it clear it is a setting for the default ingress

chenz4027: Ahhh got it! Will change the name

pkg/clusterregistryconfig/flags.go
chenz4027: Added file path

pkg/ocm/clusters.go
chenz4027: Removed the builder

PR #2451, opened 2024-09-12
OCM-11096 | ci: Fix the result reporting issue

xueli181114:


tests/utils/exec/reportportal/report_portal.go
```diff
@@ -133,7 +134,7 @@ func GenerateReportXMLFile() (int, int, map[string][]Testcase, map[string][]Test
 	passedNum := 0
 	issuedNum := 0
 
-	xmlFileList := ListFiles(config.Test.ArtifactDir, ".xml")
+	xmlFileList := ListFiles(config.Test.OutputDir, ".xml")
```
yasun1, line 137: I remembered why I got the junit.xml from `ArtifactDir` but not `Shared_dir`.  Look at the old code in release:
```
tar -xvf ${SHARED_DIR}/junit.tar.gz -C ${ARTIFACT_DIR}

log "INFO: Generate report portal report ..."
rosatest --ginkgo.v --ginkgo.no-color --ginkgo.timeout "10m" --ginkgo.label-filter "e2e-report"
log "\nTest results:"
cat "$ARTIFACT_DIR/e2e-test-results.json"
# Remove the old junit.xml file
rm -rf ${ARTIFACT_DIR}/*.xml
```

We should not extract the XML files into Shared_dir; otherwise, as you know, the secret can not be enveloped. So we should extract the files into the `tmp` directory or `ArtifactDir`(could be better).

xueli181114: Let's remove the xml after report

PR #2452, opened 2024-09-12
OCM-11097 | test: automated id:76093

yuwang-RH:
[OCM-11097](https://issues.redhat.com//browse/OCM-11097) | test: automated id:76093
log: https://privatebin.corp.redhat.com/?0450c771f234d708#CcZhmPiUStFvXPwt5k4Mm7wNqpARXJSp9hkZopFWUjxZ

### DIFF
--- a/tests/e2e/test_rosacli_cluster.go
+++ b/tests/e2e/test_rosacli_cluster.go
@@ -3510,3 +3510,117 @@ var _ = Describe("Sts cluster creation supplemental testing",
 				Expect(zones).To(Equal("[us-east-2a us-east-2b us-east-2c]"))
 			})
 	})
+
+var _ = Describe("Sts cluster with BYO oidc flow creation supplemental testing",
+	labels.Feature.Cluster,
+	func() {
+		defer GinkgoRecover()
+
+		var (
+			rosaClient         *rosacli.Client
+			clusterService     rosacli.ClusterService
+			customProfile      *profilehandler.Profile
+			clusterID          string
+			ocmResourceService rosacli.OCMResourceService
+			testingClusterName string
+		)
+		BeforeEach(func() {
+			By("Init the client")
+			rosaClient = rosacli.NewClient()
+			clusterService = rosaClient.Cluster
+			ocmResourceService = rosaClient.OCMResource
+
+			By("Prepare custom profile")
+			customProfile = &profilehandler.Profile{
+				ClusterConfig: &profilehandler.ClusterConfig{
+					HCP:           false,
+					MultiAZ:       false,
+					STS:           true,
+					OIDCConfig:    "managed",
+					NetworkingSet: false,
+					BYOVPC:        false,
+				},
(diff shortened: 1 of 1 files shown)


tests/e2e/test_rosacli_cluster.go
```diff
@@ -3510,3 +3510,117 @@ var _ = Describe("Sts cluster creation supplemental testing",
...
+				err = clusterService.WaitClusterDeleted(clusterID, 3, 35)
+				Expect(err).To(BeNil())
+			}
+			By("Clean resource")
+			errs := profilehandler.DestroyResourceByProfile(customProfile, rosaClient)
+			Expect(len(errs)).To(Equal(0))
+		})
+
+		It("Create STS cluster with oidc config id but no oidc provider via rosacli in auto mode - [id:76093]",
+			labels.Critical, labels.Runtime.Day1Supplemental,
```
OAharoni-RedHat, line 3571: I noticed that the polarion test case for this says it's a Medium importance test, whichever one isn't accurate should be updated

tests/e2e/test_rosacli_cluster.go
```diff
@@ -3510,3 +3510,117 @@ var _ = Describe("Sts cluster creation supplemental testing",
...
+			ocmResourceService rosacli.OCMResourceService
+			testingClusterName string
+		)
+		BeforeEach(func() {
+			By("Init the client")
+			rosaClient = rosacli.NewClient()
+			clusterService = rosaClient.Cluster
+			ocmResourceService = rosaClient.OCMResource
+
+			By("Prepare custom profile")
```
OAharoni-RedHat, line 3533: Is there any rule of thumb on when we choose to create a custom profile like here and when we make a full profile in the yaml file?

tests/e2e/test_rosacli_cluster.go
yuwang-RH: It is a case for cluster creation, I will update the importance in Polarion to Critical.

tests/e2e/test_rosacli_cluster.go
yuwang-RH: A full profile would be created for day-1 cluster creation ,also using for day-2 test case, for each profile should have one CI job using it.
For the test case labeled "labels.Runtime.Day1Supplemental", which will be executed by the supplemental CI job, means no prepared cluster for this kind of cases, it needs to create a cluster on demand, using custom profile is one way to prepare the cluster for it.

yuwang-RH: By the way, the cases labeled "labels.Runtime.Day1Supplemental" itself should have the full cycle of the created cluster. While the cases with the label like day-2, don't need the cleanup step which will be done by the job step.

OAharoni-RedHat: Interesting, thank you for the explanation!

PR #2462, opened 2024-09-16
OCM-10679 | fix: use default user agent

devguyio:
After #2356 there were a few places where the default user agent was hardcoded and needed to be fixed


ciaranRoche (approved):
/lgtm

PR #2471, opened 2024-09-18
OCM-11241 | fix: update description for 36293

msmigiel-rh:
For test 36293, the description said that it was doing a delete, but from looking at the code, no delete was happening, so I updated the description to only mention that it was only creating and listing the machine pools.


jfrazierRH:
/lgtm

Did you change the TC description as well?

msmigiel-rh:
Yes, I updated the description of the test case.

PR #2476, opened 2024-09-19
OCM-11000 | ci: Automate ids:67149,76200

radtriste:
https://issues.redhat.com/browse/OCM-11000

### DIFF
--- a/tests/e2e/hcp_machine_pool_test.go
+++ b/tests/e2e/hcp_machine_pool_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/openshift/rosa/tests/utils/constants"
 	"github.com/openshift/rosa/tests/utils/exec/rosacli"
 	"github.com/openshift/rosa/tests/utils/helper"
+	. "github.com/openshift/rosa/tests/utils/log"
 	ph "github.com/openshift/rosa/tests/utils/profilehandler"
 )
 
@@ -51,11 +52,11 @@ var _ = Describe("HCP Machine Pool", labels.Feature.Machinepool, func() {
 		isMultiArch, err = rosaClient.Cluster.IsMultiArch(clusterID)
 		Expect(err).ToNot(HaveOccurred())
 
-		profile = ph.LoadProfileYamlFileByENV()
-
 		if !hostedCluster {
 			SkipNotHosted()
 		}
+
+		profile = ph.LoadProfileYamlFileByENV()
 	})
 
 	Describe("Create/delete/view a machine pool", func() {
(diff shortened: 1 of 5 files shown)


radtriste:
@OAharoni-RedHat @jfrazierRH could you check that one ?

tests/e2e/hcp_machine_pool_test.go
```diff
@@ -575,5 +577,170 @@ var _ = Describe("HCP Machine Pool", labels.Feature.Machinepool, func() {
...
+					clusterConfig.Version.RawID,
+					"",
+					"")
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring(fmt.Sprintf("Cluster '%s' is not yet ready", notReadyClusterID)))
+			} else {
+				Logger.Info("No cluster found in state not-ready. Skipping this step")
+			}
+
+			By("Find a version has available upgrade versions")
```
jfrazierRH, line 638: For each of these blocks, would it make sense to have a single execution statement and that loops through passed in flags? 

Something similar to this: https://github.com/openshift/rosa/blob/3a9ac1fa489b6d6595f43338c2674af2d463a5ea/tests/e2e/test_rosacli_node_pool.go#L1150

radtriste: We are in this case testing different cli options and we use either `CreateAutomaticUpgrade` or `CreateManualUpgrade` method.
I would see it as not readable to have a huge map containing all cases.

tests/e2e/test_rosacli_machine_pool.go
```diff
@@ -1426,3 +1426,74 @@ var _ = Describe("Edit machinepool",
...
+			}
+
+		})
+
+		AfterEach(func() {
+			By("Clean remaining resources")
+			rosaClient.CleanResources(clusterID)
+		})
+
+		It("will fail - [id:76200]",
```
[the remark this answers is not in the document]
radtriste: updated

tests/e2e/hcp_machine_pool_test.go
```diff
@@ -575,5 +577,170 @@ var _ = Describe("HCP Machine Pool", labels.Feature.Machinepool, func() {
...
+				_, err = machinePoolUpgradeService.CreateManualUpgrade(
+					notReadyClusterID,
+					"anything",
+					clusterConfig.Version.RawID,
+					"",
+					"")
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring(fmt.Sprintf("Cluster '%s' is not yet ready", notReadyClusterID)))
+			} else {
+				Logger.Info("No cluster found in state not-ready. Skipping this step")
```
OAharoni-RedHat, line 635: How often do we expect a cluster to be available in a non-ready step?

radtriste: non ready can be error/installing/uninstalling... so there should be quite many

radtriste:
test failures are not related to current change

PR #2478, opened 2024-09-19
OCM-6053 | feat: Day1 SG support for HCP

hunterkepley:
Supports adding SGs to HCP clusters in day 1

Currently, CS has a disabled feature toggle, so this change allows the feature to work, but it will not work until the toggle is enabled

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -883,9 +883,21 @@ func run(cmd *cobra.Command, _ []string) {
 
 	// validate flags for cluster admin
 	isHostedCP := args.hostedClusterEnabled
-	if isHostedCP && fedramp.Enabled() {
-		r.Reporter.Errorf("Fedramp does not currently support Hosted Control Plane clusters. Please use classic")
-		os.Exit(1)
+	if isHostedCP {
+		if fedramp.Enabled() {
+			r.Reporter.Errorf("Fedramp does not currently support Hosted Control Plane clusters. Please use classic")
+			os.Exit(1)
+		}
+		if cmd.Flag(securitygroups.InfraSecurityGroupFlag).Changed {
+			r.Reporter.Errorf("Cannot use '%s' flag with Hosted Control Plane clusters, only '%s' is "+
+				"supported", securitygroups.InfraSecurityGroupFlag, securitygroups.ComputeSecurityGroupFlag)
+			os.Exit(1)
+		}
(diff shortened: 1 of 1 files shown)


hunterkepley:
Do not merge until https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/8611 is merged- i believe it will break HCP cluster creation otherwise

cmd/create/cluster/cmd.go
```diff
@@ -2571,20 +2571,23 @@ func run(cmd *cobra.Command, _ []string) {
...
-	getSecurityGroups(r, cmd, isVersionCompatibleComputeSgIds,
-		securitygroups.InfraKind, useExistingVPC, isHostedCP, subnets,
-		subnetIDs, &additionalInfraSecurityGroupIds)
-
 	additionalControlPlaneSecurityGroupIds := args.additionalControlPlaneSecurityGroupIds
-	getSecurityGroups(r, cmd, isVersionCompatibleComputeSgIds,
-		securitygroups.ControlPlaneKind, useExistingVPC, isHostedCP, subnets,
-		subnetIDs, &additionalControlPlaneSecurityGroupIds)
+	if !isHostedCP {
+		getSecurityGroups(r, cmd, isVersionCompatibleComputeSgIds,
```
gdbranco, line 2583: I think you may now need to check if the flag changed for these two sets and return an error to the customer if it is hostedCp

hunterkepley: good catch, will add that along with any other requested changes soon

cmd/describe/cluster/cmd.go
```diff
@@ -743,10 +743,17 @@ Nodes:
...
 	}
+	hasSgsCompute := len(cluster.AWS().AdditionalComputeSecurityGroupIds()) > 0
 	hasSgsControlPlane := len(cluster.AWS().AdditionalControlPlaneSecurityGroupIds()) > 0
 	hasSgsInfra := len(cluster.AWS().AdditionalInfraSecurityGroupIds()) > 0
-	if hasSgsControlPlane || hasSgsInfra {
+	if hasSgsControlPlane || hasSgsInfra || hasSgsCompute {
 		nodeConfig += " - Additional Security Group IDs:\n"
+		if hasSgsCompute {
+			nodeConfig += fmt.Sprintf(
+				"   - Compute:		%s\n",
```
gdbranco, line 753: Oh this is intentionally left out because the customer is supposed to check it from describe machine pool :)

PR #2484, opened 2024-09-23
OCM-7315 | fix: when classic topology return nil

gdbranco:
[OCM-7315](https://issues.redhat.com//browse/OCM-7315)


chenz4027:
Thanks @gdbranco for the quick fix 👍

PR #2490, opened 2024-09-23
OCM-11364 | fix: Regression, HCP nodepool creation validation used for classic

hunterkepley:
Fix for issue customer faced after 1.2.45 released

Logic which checked # of min replicas for newly created machinepools was applied to both HCP and classic- though there is a difference between what is supported (when autoscaling is enabled)

for HCP, >0 min replicas
for classic, >=0 min replicas

This fix adds different logic paths in the validation + fixes unit tests

### DIFF
--- a/pkg/machinepool/machinepool.go
+++ b/pkg/machinepool/machinepool.go
@@ -228,7 +228,8 @@ func (m *machinePool) CreateMachinePool(r *rosa.Runtime, cmd *cobra.Command, clu
 		}
 	}
 
-	minReplicas, maxReplicas, replicas, autoscaling, err := manageReplicas(cmd, args, multiAZMachinePool)
+	minReplicas, maxReplicas, replicas, autoscaling, err := manageReplicas(cmd, args, multiAZMachinePool,
+		cluster.Hypershift().Enabled())
 	if err != nil {
 		return err
 	}
@@ -588,7 +589,8 @@ func (m *machinePool) CreateNodePools(r *rosa.Runtime, cmd *cobra.Command, clust
 		}
 	}
 
-	minReplicas, maxReplicas, replicas, autoscaling, err := manageReplicas(cmd, args, false)
+	minReplicas, maxReplicas, replicas, autoscaling, err := manageReplicas(cmd, args, false,
+		cluster.Hypershift().Enabled())
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


pkg/machinepool/machinepool.go
```diff
@@ -228,7 +228,7 @@ func (m *machinePool) CreateMachinePool(r *rosa.Runtime, cmd *cobra.Command, clu
 		}
 	}
 
-	minReplicas, maxReplicas, replicas, autoscaling, err := manageReplicas(cmd, args, multiAZMachinePool)
+	minReplicas, maxReplicas, replicas, autoscaling, err := manageReplicas(cmd, args, multiAZMachinePool, false)
```
davidleerh, line 231: NP instead of false, would it be better to supply the boolean value from the cluster instead?
cluster.Hypershift().Enabled()

pkg/machinepool/machinepool.go
```diff
@@ -588,7 +588,7 @@ func (m *machinePool) CreateNodePools(r *rosa.Runtime, cmd *cobra.Command, clust
 		}
 	}
 
-	minReplicas, maxReplicas, replicas, autoscaling, err := manageReplicas(cmd, args, false)
+	minReplicas, maxReplicas, replicas, autoscaling, err := manageReplicas(cmd, args, false, true)
```
davidleerh, line 591: NP instead of false, would it be better to supply the boolean value from the cluster instead?
cluster.Hypershift().Enabled()

pkg/machinepool/machinepool.go
```diff
@@ -1323,7 +1325,7 @@ func getMachinePoolReplicas(cmd *cobra.Command,
 				Default:  minReplicas,
 				Required: replicasRequired,
 				Validators: []interactive.Validator{
-					minReplicaValidator(multiAZ, false),
+					minReplicaValidator(multiAZ, false, false),
```
gdbranco, line 1328: Shouldn't this be `cluster.Hypershift().Enabled()`? same for the ones below it

gdbranco: hmm I see, it is within a classic only function. ok

hunterkepley: @gdbranco There isn't a cluster for this func, would you rather me pass that into the args here? This func is used for machinepools only, so it's safe to say it's not hypershift (`Hypershift().Enabled()` will always be false)

PR #2499, opened 2024-09-24
OCM-7780 | feat: support approve/deny access request in ROSA CLI

marcolan018:
https://issues.redhat.com/browse/OCM-7780
https://issues.redhat.com/browse/OCM-7781

add `rosa create decision` command to support approve/deny access request

### DIFF
--- a/cmd/create/decision/cmd.go
+++ b/cmd/create/decision/cmd.go
@@ -1,0 +1,107 @@
+package decision
+
+import (
+	"context"
+	"strings"
+
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
+
+	v1 "github.com/openshift-online/ocm-sdk-go/accesstransparency/v1"
+	"github.com/spf13/cobra"
+	errors "github.com/zgalor/weberr"
+
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+const (
+	use     = "decision"
+	short   = "Create a decision for an access request"
+	long    = short
+	example = `  # Create a decision for an access request to approve it
+  rosa create decision --access-request <access_request_id> --decision Approved
+  `
+)
+
+type Options struct {
+	accessRequest string
+	decision      string
+	justification string
+}
+
+func NewDecisionOptions() *Options {
+	return &Options{}
+}
+
+func NewCreateDecisionCommand() *cobra.Command {
+
+	options := NewDecisionOptions()
+	cmd := &cobra.Command{
+		Use:     use,
+		Aliases: []string{"access-request"},
+		Short:   short,
+		Long:    long,
+		Example: example,
+		Run:     rosa.DefaultRunner(rosa.RuntimeWithOCM(), CreateDecisionRunner(options)),
+		Args:    cobra.NoArgs,
+	}
+
+	flags := cmd.Flags()
+	flags.StringVarP(
+		&options.accessRequest,
+		"access-request",
+		"a",
+		"",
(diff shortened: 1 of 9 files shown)


cmd/create/decision/cmd.go
```diff
@@ -0,0 +1,107 @@
...
+		"a",
+		"",
+		"ID of the Access Request to add decision (required).",
+	)
+	flags.StringVarP(
+		&options.decision,
+		"decision",
+		"d",
+		"",
+		"Decision created for the access request, valid value are 'Approved' or 'Denied' (required).",
```
hunterkepley, line 62: NP: `value` -> `values`

cmd/create/decision/cmd_test.go
```diff
@@ -0,0 +1,97 @@
...
+			options = &DecisionOptions{
+				accessRequest: "fake-id",
+			}
+			output.SetOutput("")
+			interactive.SetEnabled(false)
+		})
+
+		AfterEach(func() {
+			output.SetOutput("")
+			interactive.SetEnabled(false)
```
hunterkepley, line 57: I think we could get away with setting `interactive.SetEnabled(false)` only in the BeforeEach, but if im wrong no big deal

cmd/create/decision/cmd_test.go
```diff
@@ -0,0 +1,97 @@
...
+			decision, err := v1.NewDecision().ID("fake-decision-id").Decision(v1.DecisionDecisionDenied).Build()
+			Expect(err).NotTo(HaveOccurred())
+			t.ApiServer.AppendHandlers(
+				testing.RespondWithJSON(
+					http.StatusOK, FormatResource(decision)))
+			runner := CreateDecisionRunner(options)
+			t.StdOutReader.Record()
+			err = runner(nil, t.RosaRuntime, nil, nil)
+			Expect(err).NotTo(HaveOccurred())
+			stdOut, _ := t.StdOutReader.Read()
```
hunterkepley, line 92: happy to see my stdout reader used :smile:

hunterkepley (commented):
I wonder if we should implement the `options` file since there is some specific logic surrounding the `justification` flag required when a decision is denied. `edit/machinepool` has an example of this options file

Otherwise, looks good and thanks for using the new runner

cmd/create/decision/cmd.go
```diff
@@ -0,0 +1,107 @@
...
+const (
+	use     = "decision"
+	short   = "Create a decision for an access request"
+	long    = short
+	example = `  # Create a decision for an access request to approve it
+  rosa create decision --access-request <access_request_id> --decision Approved
+  `
+)
+
+type DecisionOptions struct {
```
AlexVulaj, line 26: NP - can we name this `Options` instead since the package name is "decision"? This avoids stuttering when using the type in code, i.e. `decision.DecisionOptions` vs. `decision.Options`.

This is also called out in "Naming package contents" here: https://go.dev/blog/package-names

cmd/create/decision/cmd.go
marcolan018: fixed

cmd/create/decision/cmd_test.go
marcolan018: I removed all interactive related code since interactive mode is not supported for this command

cmd/create/decision/cmd.go
marcolan018: +1
updated the name, thanks for the suggestion

PR #2515, opened 2024-09-27
OCM-11482 | fix: loosen registry regex check

chenz4027:
https://issues.redhat.com/browse/OCM-11484


andreadecorte (approved):
I see that you already have tests on it, so I think we're good on that

chenz4027:
> I see that you already have tests on it, so I think we're good on that

@andreadecorte Yeah before the regex was more strict to only allow ".". Since we don't really restrict the name validation in the backend, cli will align.

PR #2517, opened 2024-09-28
OCM-7783 | feat: add rosa describe accessrequest command

marcolan018:
https://issues.redhat.com/browse/OCM-7783

samples:
```
bash-5.2$ ./rosa describe accessrequest --id 2msh3sGFmkhzAXf3j56faQvESMO

ID:                                2msh3sGFmkhzAXf3j56faQvESMO
Subscription ID:                   2msfSLdXe59mjR8VcMVl3KoCGU3
Cluster ID:                        2e5edtubdsalag01sg2mp9lnrt58fef7
Support Case ID:                   
Requested By:                      llan.openshift
Created At:                        Wed Oct  2 12:09:54 UTC 2024
Respond By:                        Sat Oct  5 12:09:54 UTC 2024
Request Duration:                  8h
Justification:                     test
Status:                            Denied
Decisions:                           
  - Decision:                      Denied
    Decided By:                    llan.openshift
    Created At:                    Wed Oct  2 12:10:48 UTC 2024
    Justification:                 test2
  - Decision:                      Approved
    Decided By:                    llan.openshift
    Created At:                    Wed Oct  2 12:10:25 UTC 2024
    Justification:                 test2
```

### DIFF
--- a/cmd/create/decision/cmd.go
+++ b/cmd/create/decision/cmd.go
@@ -38,7 +38,6 @@ func NewCreateDecisionCommand() *cobra.Command {
 	options := NewDecisionOptions()
 	cmd := &cobra.Command{
 		Use:     use,
-		Aliases: []string{"access-request"},
 		Short:   short,
 		Long:    long,
 		Example: example,
--- a/pkg/ocm/accessrequests.go
+++ b/pkg/ocm/accessrequests.go
@@ -1,6 +1,10 @@
 package ocm
 
-import v1 "github.com/openshift-online/ocm-sdk-go/accesstransparency/v1"
+import (
+	"net/http"
+
+	v1 "github.com/openshift-online/ocm-sdk-go/accesstransparency/v1"
+)
 
 func (c *Client) CreateDecision(accessRequest string, decision string, justification string) error {
 	decisionSpec, err := v1.NewDecision().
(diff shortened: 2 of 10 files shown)


gdbranco:
Could you check what the UI also exposes? I'm not sure if we want to show the internal support case or not

Also when showing the decisions, what do you think of ordering by timestamp descending? That way the latest decision will be on top

cmd/create/decision/cmd.go
```diff
@@ -38,7 +38,6 @@ func NewCreateDecisionCommand() *cobra.Command {
 	options := NewDecisionOptions()
 	cmd := &cobra.Command{
 		Use:     use,
-		Aliases: []string{"access-request"},
```
AlexVulaj, line 41: Out of curiosity - why are we removing this alias?

pkg/ocm/accessrequests.go
```diff
@@ -20,3 +24,15 @@ func (c *Client) CreateDecision(accessRequest string, decision string, justifica
 	}
 	return nil
 }
+
+func (c *Client) GetAccessRequest(id string) (*v1.AccessRequest, bool, error) {
```
AlexVulaj, line 28: Is there ever a situation where this function could return "true" for the second param and an error in the 3rd param? I ask because it's unclear to me what the second parameter's purpose is.

cmd/create/decision/cmd.go
marcolan018: it's 'decision', not 'access-request'

pkg/ocm/accessrequests.go
marcolan018: it's impossible. Our code guarantees the error is always nil when 2nd return is true.
The purpose of the 2nd return is to clarify notfound error and other errors. So that we can give more user-friendly message in the output when cannot find access request using the specified id

AlexVulaj: I'm wondering if `GetAccessRequest` should just return an error for a 404 and let it bubble up to be handled further up, or maybe even just a nil `*v1.AccessRequest`. That way we can avoid having to check multiple return params from call sites.

PR #2518, opened 2024-09-30
OCM-11538 | feat: Adding rosa create network command

den-rgb:
[OCM-11538](https://issues.redhat.com/browse/OCM-11538)
to run the command -> 
```
rosa create network rosa-quickstart-default-vpc --param Name=rosa-quickstart-default --param VpcCidr=10.0.0.0/16 --param Region=us-west-2 --param AvailabilityZonesCount=2 --param Tags=key1=value1,key2=value2
```

*Features*

- Added region
- added params
- different templates will be called depending on the first arg passed in eg:`rosa-quickstart-default-vpc`
- Modified Tags handling + added default tags
- Account + Org validation
- vpc template no longer requires s3 bucket hosting
- no longer creates nested stacks
- Added template parameter reading for `--help` functionality
- Added `--mode=manual`
- Added zero egress support to template


hunterkepley (commented):
Looking really good, a few comments

hunterkepley:
rebasing with master should fix the pipeline failure, it was fixed after you signed off today :smile:

cmd/create/network/cmd.go
```diff
@@ -0,0 +1,154 @@
...
+	interactive.AddModeFlag(cmd)
+
+	cmd.SetHelpFunc(func(cmd *cobra.Command, args []string) {
+		templateDir := "cmd/create/network/templates"
+		err := filepath.WalkDir(templateDir, func(path string, d fs.DirEntry, err error) error {
+			if err != nil {
+				return err
+			}
+			if !d.IsDir() && strings.HasSuffix(d.Name(), ".yaml") {
+				templateBody, err := os.ReadFile(path)
```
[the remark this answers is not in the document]
hunterkepley: well i notice now it's using a dirWalk, so maybe putting it in the same exact folder as the other template is a bad idea. Not sure what's best here, but i like the idea of consolidating the templates in the 'master folder' we have in the rosa root dir

cmd/create/network/cmd.go
```diff
@@ -0,0 +1,154 @@
...
+			return r.Reporter.Errorf(err.Error())
+		}
+
+		parsedParams, parsedTags := network.ParseParams(userOptions.Params)
+
+		if parsedParams["Name"] == "" {
+			r.Logger.Debugf("Name not provided, using default name %s", r.Creator.AccountID)
+			parsedParams["Name"] = "rosa-network-stack-" + r.Creator.AccountID
+		}
+		if parsedParams["Region"] == "" {
```
hunterkepley, line 104: if these parsedParams don't exist at all, will these if's error? IIRC go will panic if you try to grab a value from a key which does not exist

hunterkepley: Only asking because i didnt see a negative test for the validation with the func which parses the params, might be wrong about this either way :+1:

cmd/create/network/cmd_test.go
```diff
@@ -0,0 +1,64 @@
...
+		ctrl = gomock.NewController(GinkgoT())
+		mockArgs = &bsOpts.NetworkUserOptions{}
+	})
+
+	AfterEach(func() {
+		ctrl.Finish()
+	})
+
+	Context("parseParams", func() {
+		It("should correctly parse the tags and parameters", func() {
```
hunterkepley, line 49: i think a negative test would be good here, related to above comment (if need be)

pkg/network/helper.go
```diff
@@ -0,0 +1,129 @@
...
+	cfTypes "github.com/aws/aws-sdk-go-v2/service/cloudformation/types"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	ColorReset  = "\033[0m"
+	ColorRed    = "\033[31m"
+	ColorGreen  = "\033[32m"
+	ColorYellow = "\033[33m"
+	ColorBlue   = "\033[34m"
```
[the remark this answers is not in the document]
den-rgb: I believe it would be appropriate to move them, but at the moment other parts of ROSA dont really use it. It could be done as part of a separate ticket, to apply the colours to more logs and to move them to a separate file.

pkg/options/network/create.go
```diff
@@ -0,0 +1,58 @@
...
+	return m.args
+}
+
+func BuildNetworkCommandWithOptions() (*cobra.Command, *NetworkUserOptions) {
+	options := NewNetworkUserOptions()
+	cmd := &cobra.Command{
+		Use:     use,
+		Short:   short,
+		Long:    long,
+		Aliases: []string{"networks", "boot-strap", "boot-straps"},
```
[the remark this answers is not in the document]
den-rgb: The bootstrap was left over from the previous implementation whoops, i dont think its necessary to keep it in

cmd/create/network/cmd.go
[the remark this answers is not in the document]
den-rgb: The template wont be held here permanently, I only have it in here for now for review purposes, The templates will be a separate optional zip file that you will be able to download along with the ROSA CLI that you can move to the correct directory in this case `rosa/cmd/create/network` . This is just the implementation agreed upon before switching it to use the ocm api.

cmd/create/network/cmd.go
den-rgb: I mean considering we will probably be writing these cf templates, the standard for these templates should be that they will at least have a region and a name parameter, otherwise we might get resources scattered across different regions with an unidentifiable resource name. But I'll test this out just to be sure.

pkg/network/helper.go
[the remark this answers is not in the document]
hunterkepley: sounds good to me, id like to make a UX ticket at some point on the new team doing this, I think it would be great for listing clusters for example

cmd/create/network/cmd.go
hunterkepley: My main concern is getting a panic rather than a readable error here- no worries if it errors, just to more quickly identify what is missing

den-rgb:
```
E: failed to create stack, operation error CloudFormation: CreateStack, https response error StatusCode: 400, RequestID: 9ae9fb2e-d8b2-407a-b11c-56dc1e0ad886, api error ValidationError: Parameters: [Region] do not exist in the template
```
This is the error you get if the parameter doesnt exist in the template, it should be fine as is

cmd/create/network/cmd.go
```diff
@@ -0,0 +1,157 @@
...
+
+	"github.com/openshift/rosa/pkg/interactive"
+	"github.com/openshift/rosa/pkg/network"
+	helper "github.com/openshift/rosa/pkg/network"
+	"github.com/openshift/rosa/pkg/ocm"
+	bsOpts "github.com/openshift/rosa/pkg/options/network"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+type NetworkSpec struct {
```
AlexVulaj, line 23: I have a preference for not repeating the package name when naming types. For example, if this were accessed externally it would look like `network.NetworkSpec`, when `network.Spec` also conveys the same information.

This is also suggested by the Go blog if you search https://go.dev/blog/package-names

cmd/create/network/cmd.go
```diff
@@ -0,0 +1,157 @@
...
+	Service network.NetworkService
+}
+
+type networkStruct struct {
```
AlexVulaj, line 27: Is this struct needed? It looks like it's boxing the "service" without any additional methods or functionality attached - could we just use the service directly?

cmd/create/network/cmd.go
```diff
@@ -0,0 +1,157 @@
...
+			fmt.Println(err)
+		}
+
+		fmt.Println("\n" + cmd.UsageString())
+	})
+
+	return cmd
+}
+
+func NetworkRunner(userOptions *bsOpts.NetworkUserOptions) rosa.CommandRunner {
```
AlexVulaj, line 85: Same comment here about the repetition

cmd/create/network/cmd.go
AlexVulaj: There are a few other instances of functions/structs/etc in this PR that the same comment applies to as well. I won't tag all of them to keep the review a bit cleaner, but I'll leave it up to your discretion.

pkg/network/helper.go
```diff
@@ -0,0 +1,140 @@
...
+	case cfTypes.ResourceStatusCreateComplete, cfTypes.ResourceStatusUpdateComplete:
+		return ColorGreen
+	case cfTypes.ResourceStatusCreateFailed, cfTypes.ResourceStatusDeleteFailed, cfTypes.ResourceStatusUpdateFailed:
+		return ColorRed
+	default:
+		return ColorYellow
+	}
+}
+
+// parseParams converts the list of parameter strings into a map and sets default values
```
AlexVulaj, line 73:
```suggestion
// ParseParams converts the list of parameter strings into a map and sets default values
```

pkg/network/helper.go
```diff
@@ -0,0 +1,140 @@
...
+				return nil, nil, errors.New("duplicate parameter key")
+			}
+			result[parts[0]] = parts[1]
+		}
+	}
+
+	return result, userTags, nil
+}
+
+// selectTemplate selects the appropriate template file based on the template name
```
AlexVulaj, line 107:
```suggestion
// SelectTemplate selects the appropriate template file based on the template name
```

pkg/network/network.go
```diff
@@ -0,0 +1,129 @@
...
+type network struct {
+}
+
+var _ NetworkService = &network{}
+
+func NewNetworkService() NetworkService {
+	return &network{}
+}
+
+// createStack creates a CloudFormation stack
```
AlexVulaj, line 30:
```suggestion
// CreateStack creates a CloudFormation stack
```

cmd/create/network/cmd.go
```diff
@@ -0,0 +1,157 @@
...
+
+type NetworkSpec struct {
+	Service network.NetworkService
+}
+
+type networkStruct struct {
+	service network.NetworkService
+}
+
+func NewNetwork(spec NetworkSpec) networkStruct {
```
AlexVulaj, line 31: If I've missed something in my above comment and we keep `networkStruct`, can we export it? I've run into issues with testing and visibility when returning non-exported types as part of exported functions.

cmd/create/network/cmd.go
```diff
@@ -0,0 +1,141 @@
...
+
+				parameters, ok := templateMap["Parameters"].(map[string]interface{})
+				if !ok {
+					fmt.Printf("No parameters found in the CloudFormation template %s\n", d.Name())
+					return nil
+				}
+
+				fmt.Printf("Available parameters in %s/%s:\n", filepath.Base(filepath.Dir(path)), d.Name())
+				for paramName := range parameters {
+					fmt.Printf("  %s\n", paramName)
```
hunterkepley, line 56: Just noticed youre using print in some spots, is this intended? I think we normally use the logger

den-rgb: the use of fmt is intentional otherwise the help command output would look strange if we were to use the logger

PR #2523, opened 2024-10-01
OCM-11556 | fix: refactor multiple ca tests from getClusterRegistryConfig

chenz4027:
https://issues.redhat.com/browse/OCM-11556

### DIFF
--- a/cmd/describe/cluster/cmd_test.go
+++ b/cmd/describe/cluster/cmd_test.go
@@ -140,6 +140,7 @@ var _ = Describe("getClusterRegistryConfig", func() {
 	It("Should return expected output", func() {
 		mockCa := make(map[string]string)
 		mockCa["registry.io"] = "-----BEGIN CERTIFICATE-----\nlalala\n-----END CERTIFICATE-----\n"
+		mockCa["registry.io2"] = "-----BEGIN CERTIFICATE-----\nlalala\n-----END CERTIFICATE-----\n"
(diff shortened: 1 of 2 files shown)


andreadecorte (approved):

cmd/describe/cluster/cmd_test.go
```diff
@@ -149,6 +151,12 @@ var _ = Describe("getClusterRegistryConfig", func() {
 				DomainName("quay.io").Insecure(true)).
 			PlatformAllowlist(cmv1.NewRegistryAllowlist().ID("test-id"))).Build()
 		Expect(err).NotTo(HaveOccurred())
+		mockCaOutput := ""
+		for registry := range mockCa {
+			mockCaOutput = fmt.Sprintf("%s"+
+				"    - %s: REDACTED\n", mockCaOutput, registry)
+
+		}
```
gdbranco, line 159: hmm, I meant to use sprintf directly on the map, but you can't do it because of redacted is that it?
I suggest to pull the map keys, sort it then use that list instead of the map to go through the keys and write the redacted alongside it then :thinking: 
Getting the keys via a loop on the map directly results in it being able to be out of order again

chenz4027: I see, thanks @gdbranco , let me try again

chenz4027: Updated, @gdbranco mind taking another look?

PR #2528, opened 2024-10-04
OCM-11611 | ci: Fix id:75525

radtriste:
https://issues.redhat.com/browse/OCM-11611

### DIFF
--- a/tests/e2e/test_rosacli_cluster_post.go
+++ b/tests/e2e/test_rosacli_cluster_post.go
@@ -392,10 +392,10 @@ var _ = Describe("Healthy check",
 
 		It("with policy path will work - [id:75525]", labels.Runtime.Day1Post, labels.High,
 			func() {
-				path := profile.AccountRoleConfig.Path
-				if path == "" {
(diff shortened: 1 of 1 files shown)


tests/e2e/test_rosacli_cluster_post.go
```diff
@@ -392,10 +392,10 @@ var _ = Describe("Healthy check",
 
 		It("with policy path will work - [id:75525]", labels.Runtime.Day1Post, labels.High,
 			func() {
-				path := profile.AccountRoleConfig.Path
-				if path == "" {
+				if profile.AccountRoleConfig == nil || profile.AccountRoleConfig.Path == "" {
```
AkashKanni, line 395: @radtriste why not using `path` variable by storing value `profile.AccountRoleConfig.Path` in it?
is there something which changes in the behavior when we store it into a variable and use?

AkashKanni: okay, got it Thanks!

PR #2532, opened 2024-10-04
OCM-11496 | fix: Change language when updating cluster reg config

patrickjennings:
https://issues.redhat.com/browse/OCM-11496

### DIFF
--- a/pkg/clusterregistryconfig/flags.go
+++ b/pkg/clusterregistryconfig/flags.go
@@ -147,13 +147,15 @@ func GetClusterRegistryConfigOptions(cmd *pflag.FlagSet,
 		enableRegistriesConfig = true
 	}
 
-	if interactive.Enabled() && isHostedCP {
+	if interactive.Enabled() {
+		regConfigQuestion := GetClusterRegistryConfigQuestion(cluster)
+
 		updateRegistriesConfigValue, err := interactive.GetBool(interactive.Input{
-			Question: "Enable registries config",
+			Question: regConfigQuestion,
 			Default:  enableRegistriesConfig,
 		})
 		if err != nil {
-			return nil, fmt.Errorf("Expected a valid value: %s", err)
+			return nil, fmt.Errorf("Expected a valid registries config value: %s", err)
 		}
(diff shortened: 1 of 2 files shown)


pkg/clusterregistryconfig/flags.go
```diff
@@ -147,12 +147,17 @@ func GetClusterRegistryConfigOptions(cmd *pflag.FlagSet,
 	}
 
 	if interactive.Enabled() && isHostedCP {
+		regConfigQuestion := "Enable registries config"
```
philipwu08, line 150: Maybe extract the next few lines to a separate function so that we can add unit tests for it.

patrickjennings: Let me know if the changes I pushed are what you are expecting.

pkg/clusterregistryconfig/flags_test.go
```diff
@@ -37,12 +38,12 @@ var _ = Describe("Cluster Registry Config tests", func() {
 				Equal("-----BEGIN CERTIFICATE-----\n/abc\n-----END CERTIFICATE-----"))
 
 		})
-		It("KO: fail if the spec file is invalid", func() {
+		It("OK: fail if the spec file is invalid", func() {
```
chenz4027, line 41: Oh @patrickjennings it's not a typo. KO is for tests with expected failures while OK is for best case scenario. No need to change those.

patrickjennings: Sorry about that. "KO", nice convention! Reverted those changes.

chenz4027: Np! Also it's conventional to have only one commit per each change. CS side has the option to squash the commits but I don't think rosa has it.
You can do 
```
git rebase -i HEAD~2
```
comment the 2nd commit
And then force push the PR.

patrickjennings: Gotcha. Rebased and squashed the commits.

chenz4027: Perfect, thanks @patrickjennings

PR #2539, opened 2024-10-09
OCM-11711 | fix: Adding a default template if no template is specified

den-rgb:
Adding a default template if no template is specified in the `rosa create network` command
[OCM-11711](https://issues.redhat.com/browse/OCM-11711)


cmd/create/network/cmd.go
```diff
@@ -103,6 +103,10 @@ func NetworkRunner(userOptions *opts.NetworkUserOptions) rosa.CommandRunner {
 			}
 		}
 
+		if templateCommand == "" {
```
marcolan018, line 106: is it better to assign an initial value when define the var `templateCommand`?

hunterkepley (approved):
LGTM

PR #2540, opened 2024-10-09
OCM-7782 | feat: support list access requests

marcolan018:
https://issues.redhat.com/browse/OCM-7782

sample output:
```
bash-5.2$ ./rosa list accessrequests
STATE     ID                           CLUSTER ID                        UPDATED AT
Pending   2mwPzSn8hbXydav8WfNA3RZzAHF  2e6adt54d7tbkemhmog6jtcvq0hvd5a5  Thu Oct  3 19:48:47 UTC 2024
Pending   2mwBGYxLOAzhPgBiJqrQMIrP4Fv  2e68g3s0jd6jn383kq3stms6ar702tm3  Thu Oct  3 17:47:42 UTC 2024
Approved  2mwA4CtaEl7AMgA8afJNaCPTb96  2e5edtubdsalag01sg2mp9lnrt58fef7  Thu Oct  3 17:38:10 UTC 2024
I: Run the following command to approve or deny the access request:

   rosa create decision --access-request <ID> --decision Approved
   rosa create decision --access-request <ID> --decision Denied
```
sample output for specified cluster:
```
bash-5.2$ ./rosa list accessrequests -c marcohp2
STATE    ID                           UPDATED AT
Pending  2mwPzSn8hbXydav8WfNA3RZzAHF  Thu Oct  3 19:48:47 UTC 2024
I: Run the following command to approve or deny the access request:

   rosa create decision --access-request 2mwPzSn8hbXydav8WfNA3RZzAHF --decision Approved
   rosa create decision --access-request 2mwPzSn8hbXydav8WfNA3RZzAHF --decision Denied
```


cmd/list/accessrequests/cmd.go
```diff
@@ -0,0 +1,117 @@
...
+					"   rosa create decision --access-request %s --decision Approved\n"+
+					"   rosa create decision --access-request %s --decision Denied\n",
+					pendingId, pendingId)
+			}
+		}
+		return nil
+	}
+}
+
+func getAccessRequestsOutput(clusterId string, accessRequests []*v1.AccessRequest) (string, bool, string) {
```
hunterkepley, line 88: I think we can move this func to the `pkg/ocm/accessrequests.go` file, WDYT?

cmd/list/accessrequests/cmd.go
```diff
@@ -0,0 +1,117 @@
...
+			writer := tabwriter.NewWriter(os.Stdout, 0, 4, 2, ' ', 0)
+			output, hasPending, pendingId := getAccessRequestsOutput(clusterId, accessRequests)
+			fmt.Fprint(writer, output)
+			err = writer.Flush()
+			if err != nil {
+				return err
+			}
+			if hasPending {
+				r.Reporter.Infof("Run the following command to approve or deny the access request:\n\n"+
+					"   rosa create decision --access-request %s --decision Approved\n"+
```
hunterkepley, line 79: Since the flag for creating a decision using a `-` between `access` and `request` - I think it makes sense (along with other naming convention in ROSA) to change the default for this command to `access-request` rather than `accessrequest`
WDYT?

hunterkepley (commented):
Overall looks great, some restructuring ideas + I think we may want to name these files `access_requests` to keep naming convention up with the rest of ROSA

cmd/list/accessrequests/cmd.go
```diff
@@ -0,0 +1,117 @@
...
+				} else {
+					r.Reporter.Infof("There are no Access Requests for cluster '%s'.", r.ClusterKey)
+				}
+				return nil
+			}
+			writer := tabwriter.NewWriter(os.Stdout, 0, 4, 2, ' ', 0)
+			output, hasPending, pendingId := getAccessRequestsOutput(clusterId, accessRequests)
+			fmt.Fprint(writer, output)
+			err = writer.Flush()
+			if err != nil {
```
hunterkepley, line 74: NP:
`if writer.Flush() != nil {`

cmd/list/accessrequests/cmd.go
marcolan018: if needed, need to move to place like pkg/accessrequests/output.og. Since this func is not ocm related, pkg/ocm/accessrequests.go is not a good place for it.
Because this func is only for list command and not resuable in other commands, also there are no other utility functions, so I don't create a new folder `accessrequest` under pkg

marcolan018:
@hunterkepley thanks for the comments. Can you take a look again on the latest change? thanks

cmd/list/accessrequests/cmd.go
hunterkepley: understood- was mainly following convention on the refactors Rob wanted to accomplish, keeping the commands simple and all related funcs in `pkg` (i agree `ocm` is not the best subfolder for this)

But I plan to do a major refactoring of ROSA once we switch teams anyway :smile:

hunterkepley:
LGTM- I think we should rename the files to `access_request...` (snakecase) to keep naming convention up. I am unsure how i feel about making access request a single word file-wise

cmd/list/accessrequests/cmd.go
```diff
@@ -0,0 +1,116 @@
...
+					r.Reporter.Infof("There are no Access Requests in Pending or Approved status.")
+				} else {
+					r.Reporter.Infof("There are no Access Requests for cluster '%s'.", r.ClusterKey)
+				}
+				return nil
+			}
+			writer := tabwriter.NewWriter(os.Stdout, 0, 4, 2, ' ', 0)
+			output, hasPending, pendingId := getAccessRequestsOutput(clusterId, accessRequests)
+			fmt.Fprint(writer, output)
+			if writer.Flush() != nil {
```
den-rgb, line 73: Could possibly handle the error like this for better readability
```
if err := writer.Flush(); err != nil {
				return err
			}
```

cmd/list/accessrequests/cmd.go
```diff
@@ -0,0 +1,116 @@
...
+			return err
+		}
+		if output.HasFlag() {
+			output.Print(accessRequests)
+		} else {
+			if len(accessRequests) == 0 {
+				if clusterId == "" {
+					r.Reporter.Infof("There are no Access Requests in Pending or Approved status.")
+				} else {
+					r.Reporter.Infof("There are no Access Requests for cluster '%s'.", r.ClusterKey)
```
den-rgb, line 66: will r.ClusterKey always have a valid value, its not explicitly being set anywhere?

marcolan018: it's set in when invoke r.GetClusterKey() when detect `cluster` flag changed

cmd/list/accessrequests/cmd.go
marcolan018: thanks for pointing it out. That's a bug in the code, fixed

cmd/list/accessrequests/cmd.go
```diff
@@ -0,0 +1,116 @@
...
+					pendingId, pendingId)
+			}
+		}
+		return nil
+	}
+}
+
+func getAccessRequestsOutput(clusterId string, accessRequests []*v1.AccessRequest) (string, bool, string) {
+	output := "STATE\tID\tCLUSTER ID\tUPDATED AT\n"
+	if clusterId != "" {
```
robpblake, line 89: @marcolan018 I'm curious...is `clusterId` ever an empty string? IIRC, when we create the AccessRequest we lookup the `Subscription` and fetch the `clusterId` from there 🤔

cmd/list/accessrequests/cmd.go
```diff
@@ -0,0 +1,116 @@
...
+				return nil
+			}
+			writer := tabwriter.NewWriter(os.Stdout, 0, 4, 2, ' ', 0)
+			output, hasPending, pendingId := getAccessRequestsOutput(clusterId, accessRequests)
+			fmt.Fprint(writer, output)
+			if err := writer.Flush(); err != nil {
+				return err
+			}
+			if hasPending {
+				r.Reporter.Infof("Run the following command to approve or deny the access request:\n\n"+
```
robpblake, line 77: @marcolan018 "access request" should be "Access Request"

cmd/list/accessrequests/cmd.go
robpblake: @marcolan018 Ah I see the flow here. You only print out the `clusterId` if the user has passed it as a CLI parameter.

_Personally_, I would always print out the `clusterId` when showing details of the Access Request as it's a key part of the information required by customers.

cmd/list/accessrequests/cmd.go
```diff
@@ -0,0 +1,116 @@
...
+	output := "STATE\tID\tCLUSTER ID\tUPDATED AT\n"
+	if clusterId != "" {
+		output = "STATE\tID\tUPDATED AT\n"
+	}
+	hasPending := false
+	id := "<ID>"
+	for _, accessRequest := range accessRequests {
+		if accessRequest.Status().State() == v1.AccessRequestStatePending {
+			hasPending = true
+			if clusterId != "" {
```
robpblake, line 97: @marcolan018 If I follow this correctly, if there are `Pending` Access Requests, but the user only executes `rosa list access-requests`, then id is returned as `<ID>` which is then shown to the user in the log statement on how to approve/deny the access requests?

cmd/list/accessrequests/cmd.go
marcolan018: @robpblake 
will it be redundant to display the `clusterId` if user already specified the `--cluster` flag? Also in the whole list the value will be the same.

cmd/list/accessrequests/cmd.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/output"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+const (
+	use   = "access-request"
+	short = "List access requests"
```
robpblake, line 20: @marcolan018 "Access Requests" should be used through out

cmd/list/accessrequests/cmd.go
```diff
@@ -0,0 +1,116 @@
...
+				}
+				return nil
+			}
+			writer := tabwriter.NewWriter(os.Stdout, 0, 4, 2, ' ', 0)
+			output, hasPending, pendingId := getAccessRequestsOutput(clusterId, accessRequests)
+			fmt.Fprint(writer, output)
+			if err := writer.Flush(); err != nil {
+				return err
+			}
+			if hasPending {
```
robpblake, line 76: @marcolan018 Are you OK with this message if the customer has multiple, pending Access Requests?

cmd/list/accessrequests/cmd.go
marcolan018: hmmm, since user maybe set cluster name as the identity, I changed to always include clusterId in the output

cmd/list/accessrequests/cmd.go
marcolan018: fixed

cmd/list/accessrequests/cmd.go
marcolan018: @robpblake 
that's current implementation, if there is Pending Access Request
if user specify `--cluster`, then log out something like 
`rosa create decision --access-request 2nFVxq5ki0ktsRtmhrSyGf98TVB  --decision Approved`
if user does not specify `--cluster` flag, log out
`rosa create decision --access-request <ID>  --decision Approved`

do you have concern on this? And any recommendation about the output?

PR #2544, opened 2024-10-11
OCM-11810 | feat: Improved help message for rosa create network

den-rgb:
[OCM-11810](https://issues.redhat.com/browse/OCM-11810)


pkg/options/network/create.go
```diff
@@ -11,7 +11,12 @@ const (
 	short   = "Network aws cloudformation stack"
 	long    = "Network aws cloudformation stack using predefined yaml templates. "
 	example = `  # Create a aws cloudformation stack
-  rosa create network <template-name> --param Param1=Value1 --param Param2=Value2 `
+  rosa create network <template-name> --param Param1=Value1 --param Param2=Value2 ` +
+		"\n\n" + `  # Rosa quick start HCP VPC example` +
+		"\n" + `  rosa create network rosa-quickstart-default-vpc --param Region=us-west-2` +
+		` --param Name=quickstart-stack --param AvailabilityZoneCount=1 --param VpcCidr=10.0.0.0/16` +
+		"\n\n" + `  # To delete the aws cloudformation stack` +
```
hunterkepley, line 18: NP: `AWS Cloudformation` - we use capitals with acronyms for `HPC VPC` - just for consistency

hunterkepley: Since we're saying `the ...` rather than referencing the command here

pkg/options/network/create.go
```diff
@@ -11,7 +11,12 @@ const (
 	short   = "Network aws cloudformation stack"
 	long    = "Network aws cloudformation stack using predefined yaml templates. "
 	example = `  # Create a aws cloudformation stack
-  rosa create network <template-name> --param Param1=Value1 --param Param2=Value2 `
+  rosa create network <template-name> --param Param1=Value1 --param Param2=Value2 ` +
+		"\n\n" + `  # Rosa quick start HCP VPC example` +
```
robpblake, line 15: @den-rgb `ROSA` rather than `Rosa`

pkg/options/network/create.go
```diff
@@ -11,7 +11,12 @@ const (
 	short   = "Network aws cloudformation stack"
```
robpblake, line 11: @den-rgb Personally, these feel a bit too literal. From a customer perspective, the role of this command is to quickly create your network configuration for your HCP clusters. As a suggestion (unless this has already been given the +1 elsewhere)

- short: Create required network configuration for ROSA HCP clusters
- long: Quickly create the required network configuration in AWS to support deployment of ROSA HCP clusters
- example: Create the required network configuration for a publicly accessible, ROSA HCP cluster

I also find the inclusion of the `to delete the AWS Cloudformation stack` here a bit confusing as it is an implementation detail and feels more appropriate to a log statement after the stack has been created.

den-rgb: It was QE that asked for a more literal example as well as the generic one, we also have to remember this functionality is aimed at new users
As discussed with the team previously we decided we didnt want the delete command after cluster creation and it should be included in the help instead.

den-rgb: Also this command isn't only going to be used for HCP related resources it can be used for anything, so it wouldn't make sense to have the long and the short as above

PR #2565, opened 2024-10-25
OCM-12100 | fix: Only attempt provider creation with auto mode

hunterkepley:


cmd/create/oidcconfig/cmd.go
```diff
@@ -295,25 +295,23 @@ func run(cmd *cobra.Command, _ []string) {
...
+	if !args.rawFiles {
+		arguments.DisableRegionDeprecationWarning = true // disable region deprecation warning
+		providerArgs := []string{"", mode, oidcConfigInput.IssuerUrl}
+		if oidcConfigId != "" {
+			providerArgs = append(providerArgs, "--oidc-config-id", oidcConfigId)
+			err = oidcprovider.Cmd.Flags().Set("oidc-config-id", oidcConfigId)
+			if err != nil {
 				r.Reporter.Errorf("Unable to attempt creation of OIDC provider; oidc config ID"+
 					" not found / not created successfully: %s", err)
-				os.Exit(1)
```
davidleerh, line 312: can we keep the os.Exit(1) here for early exit?

PR #2606, opened 2024-11-08
OCM-11043 | test: automated case id:76543

jfrazierRH:
ginkgo run -focus 76543 tests/e2e
Running Suite: ROSA CLI e2e tests suite - /Users/jfrazier/workspace/rosa/tests/e2e
==================================================================================
Random Seed: 1731037536

Will run 1 of 252 specs
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS•SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS

Ran 1 of 252 Specs in 13.325 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 251 Skipped
PASS

Ginkgo ran 1 suite in 20.007141041s
Test Suite Passed


tests/ci/data/profiles/rosa-hcp.yaml
```diff
@@ -62,6 +62,34 @@ profiles:
   account-role:
     path: "/test/hcp/"
     permission_boundary: "arn:aws:iam::aws:policy/AdministratorAccess"
+- as: rosa-hcp-ze
```
yingzhanredhat, line 65: what does this name:rosa-hcp-ze  mean?

tests/ci/data/profiles/rosa-hcp.yaml
```diff
@@ -62,6 +62,34 @@ profiles:
...
+  version: 4.14.38
```
yingzhanredhat, line 66: How about other versions? we use latest version usually

tests/ci/data/profiles/rosa-hcp.yaml
yingzhanredhat: Why not add zero_egress to exist profile?

tests/e2e/hcp_cluster_test.go
```diff
@@ -657,4 +658,34 @@ var _ = Describe("HCP cluster testing",
 					}
 				}
 			})
+
+		It("create zero-egress HCP cluster - [id:76543]",
+			labels.High, labels.Runtime.Day1,
```
yingzhanredhat, line 663: I think it can move the step:CreateMachinePool in day2 case

tests/ci/data/profiles/rosa-hcp.yaml
jfrazierRH: rosa-hcp-(zero egress)
I added it as a separate profile since there are restrictions about the version itself.

tests/ci/data/profiles/rosa-hcp.yaml
jfrazierRH: Currently, zero_egress is only supported for 4.14.y versions. Once the GA is released, I can change this to latest

tests/e2e/hcp_cluster_test.go
jfrazierRH: Even though its part of the test case? I know that having a machinepool created on a ZE cluster was an area we wanted tested.

yingzhanredhat: According to the rule, we have day1\day2\destoy ... cases. We will filter the case by orde.So if the case is for day1, it would be better to not contain day2 step. We can move it in day2 case other than ignore the step.

jfrazierRH: No problem, I'll remove the mp portion.

PR #2607, opened 2024-11-08
OCM-12420 | feat: Adding default cf template for binary builds

den-rgb:
[OCM-12420](https://issues.redhat.com/browse/OCM-12420)


hunterkepley (approved):
/lgtm

PR #2609, opened 2024-11-11
OCM-12438 | fix: changing default template dir in help message

den-rgb:
[OCM-12438](https://issues.redhat.com/browse/OCM-12438)


hunterkepley (approved):
/lgtm

PR #2610, opened 2024-11-11
OCM-12435 | fix: Updated info message for when no template is specified

den-rgb:
[OCM-12435](https://issues.redhat.com/browse/OCM-12435)


hunterkepley (approved):
/lgtm

PR #2620, opened 2024-11-12
OCM-12482 | fix: Fixed tempalte dir env var to work

den-rgb:
[OCM-12482](https://issues.redhat.com/browse/OCM-12482)

### DIFF
--- a/pkg/options/network/create.go
+++ b/pkg/options/network/create.go
@@ -85,6 +85,10 @@ func BuildNetworkCommandWithOptions() (*cobra.Command, *NetworkUserOptions) {
 		Args:    cobra.MaximumNArgs(1),
 		Hidden:  false,
 	}
+	var exportedTempDir string
+	if options.TemplateDir != "" {
+		exportedTempDir = options.TemplateDir
+	}
 
 	flags := cmd.Flags()
 	flags.StringVar(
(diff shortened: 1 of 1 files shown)


pkg/options/network/create.go
```diff
@@ -87,6 +91,9 @@ func BuildNetworkCommandWithOptions() (*cobra.Command, *NetworkUserOptions) {
 		defaultTemplateDir,
 		"Use a specific template directory, overriding the OCM_TEMPLATE_DIR environment variable.",
 	)
+	if exportedTempDir != "" {
+		options.TemplateDir = exportedTempDir
```
hunterkepley, line 95: Not sure I understand this change,

I see this is supposed to set user-inputted `template-dir` to it's original value after the flag is created, but I feel like it would be the same outcome if we changes L91's default value to be `options.templateDir`

Where does the env var feed in? should we be using that as the default value here on L91, and maybe disabling auto mode input later for `template-dir`?

den-rgb: Personally I'd be all for replacing L91 , but i do see why it might be confusing to see the default being the path set by the OCM_TEMPLATE_DIR env var

hunterkepley: unless the env var is set to the `options.templateDir` var prior to this, then it makes sense

hunterkepley: nm yeah I see that, LGTM

den-rgb:
cherrypick release_1.2.48

PR #2628, opened 2024-11-13
OCM-12362 | feat: Add HCP shared-vpc support to accountroles

hunterkepley:
Allows creation of account-roles for shared-vpc use. This attaches a policy which allows assuming of 2 roles the user provides:

1) the `route53 role arn`
2) the `vpc endpoint role arn`

These are important as we do not currently have managed policies set up for shared-vpc, and without these, the user will be unable to create an HCP sharedVPC cluster

Output and results:

Output is the same as if you were to create accountroles prior to this change, results (once providing the required flags containing the role arns + creating hosted-cp roles) is as follows:

image

Attaching a new policy, containing an `sts:AssumeRole` action for the 2 resources provided by the user


cmd/create/accountroles/creators.go
```diff
@@ -434,3 +444,31 @@ func buildAttachRolePolicyCommand(accRoleName string, policyARN string) string {
 		AddParam(awscb.PolicyArn, policyARN).
 		Build()
 }
+
+func attachHcpSharedVpcPolicy(r *rosa.Runtime, route53RoleArn string, vpcEndpointRoleArn string, roleName string,
+	path string, defaultPolicyVersion string) error {
+	policyDetails := "{\n  \"Version\": \"2012-10-17\",\n  \"Statement\": {\n    \"Effect\": \"Allow\",\n    " +
```
gdbranco, line 450: this could be a const outside of the func

cmd/create/accountroles/cmd.go
```diff
@@ -133,6 +140,22 @@ func init() {
...
+		"Role ARN associated with the private hosted zone used for Hosted Control Plane cluster shared VPC, this "+
+			"role contains policies to be used with route53",
+	)
+
+	flags.StringVar(
+		&args.vpcEndpointRoleArn,
+		vpcEndpointRoleArnFlag,
+		"",
+		"Role ARN associated with the shared VPC used for Hosted Control Plane clusters, this role contains "+
+			"policies to be used with the vpc endpoint",
```
robpblake, line 156: @hunterkepley Consistency of `VPC` in this sentence. We switch between upper and lower case

cmd/create/accountroles/cmd.go
```diff
@@ -133,6 +140,22 @@ func init() {
 		"Create only classic Rosa account roles",
 	)
 
+	flags.StringVar(
+		&args.route53RoleArn,
+		route53RoleArnFlag,
+		"",
+		"Role ARN associated with the private hosted zone used for Hosted Control Plane cluster shared VPC, this "+
+			"role contains policies to be used with route53",
```
robpblake, line 148: @hunterkepley I think it's probably best to go for `Route 53`. WDYT?

cmd/create/accountroles/cmd.go
```diff
@@ -133,6 +140,22 @@ func init() {
 		"Create only classic Rosa account roles",
 	)
 
+	flags.StringVar(
```
robpblake, line 143: @gdbranco / @hunterkepley What's the expected flag names? Looking at https://issues.redhat.com/browse/OCM-12291, i see `–cross-account-hosted-zone-role-arn` and `–cross-account-vpc-endpoint-service` which is different to what we have here.

cmd/create/accountroles/cmd.go
```diff
@@ -32,6 +32,11 @@ import (
 	"github.com/openshift/rosa/pkg/rosa"
 )
 
+const (
+	route53RoleArnFlag     = "route53-role-arn"
```
robpblake, line 36: @hunterkepley / @gdbranco For my understanding: Is it expected that the user creates these roles somewhere else to get the ARNs or are we waiting on Managed Policies from AWS?

cmd/create/accountroles/validation.go
```diff
@@ -0,0 +1,36 @@
+package accountroles
+
+import errors "github.com/zgalor/weberr"
+
+func validateSharedVpcInputs(hostedCp bool, vpcEndpointRoleArn string,
+	route53RoleArn string) (bool, error) {
+
+	if !hostedCp {
+		if vpcEndpointRoleArn != "" {
+			return false, errors.UserErrorf("Can only use '%s' flag for Hosted Control Plane account roles",
```
robpblake, line 10: @hunterkepley I'd consider extracting these error messages to a `const` to avoid the duplication e.g:

```
return false, errors.UserErrorf(hostedCpOnlyErrorMessage, vpcEndpointRoleArnFlag)
```

where `hostedCpOnlyErrorMessage` is the string const.

robpblake:
@hunterkepley Is interactive mode required to get these role ARNs from the user?

cmd/create/accountroles/validation.go
```diff
@@ -0,0 +1,36 @@
+package accountroles
+
+import errors "github.com/zgalor/weberr"
+
+func validateSharedVpcInputs(hostedCp bool, vpcEndpointRoleArn string,
+	route53RoleArn string) (bool, error) {
+
+	if !hostedCp {
+		if vpcEndpointRoleArn != "" {
```
robpblake, line 9: @hunterkepley Take a look at the `machinepools.HostedClusterOnlyFlag` function. Is there anything we can do to at least standardise the messaging we give to a user when they're trying to user a HCP-only flag with a classic cluster?

cmd/create/accountroles/cmd.go
gdbranco: They are expected to create it in adhoc manner for the moment

cmd/create/accountroles/cmd.go
```diff
@@ -383,7 +412,7 @@ func run(cmd *cobra.Command, argv []string) {
 	}
 
 	input := buildRolesCreationInput(prefix, permissionsBoundary, r.Creator.AccountID, env, policies,
-		policyVersion, path)
+		policyVersion, path, isHcpSharedVpc)
```
gdbranco, line 415: Would rather not include a bool param if possible

cmd/create/accountroles/cmd.go
hunterkepley: Yup- this will eventually be replaced, but for release, we are expecting adhoc roles

hunterkepley:
> @hunterkepley Is interactive mode required to get these role ARNs from the user?

Nope; as with the operator roles manual mode/deletion MR: https://github.com/openshift/rosa/pull/2630
There will be one for account roles following. It would double the line count for review though sadly

cmd/create/accountroles/cmd.go
robpblake: Thanks. 

@hunterkepley I'd suggest updating the description of https://issues.redhat.com/browse/OCM-12291 just to make things clearer when we get to QE etc.

cmd/create/accountroles/cmd.go
robpblake: Thanks for the explanation.

hunterkepley:
Failure related to test which requires CS update- if this gets approved I have gotten the OK to remove the failing test to unblock

cmd/create/accountroles/cmd.go
gdbranco: If there is too much refactoring required we can move ahead

hunterkepley: I think it would be best to move ahead-
Im not sure I see a benefit from making one function which returns a struct containing `hostedCp: false` and another func returning the same struct with `hostedCp: true` - would like to discuss further for the future at another point 🙂

gdbranco:
sgtm

PR #2630, opened 2024-11-13
OCM-12538 | feat: Manual mode for create&delete operator-roles with hcp sharedvpc

hunterkepley:
Allows `--mode manual` for creating and deleting operator roles with the created&attached HCP SharedVPC policies.

### DIFF
--- a/cmd/create/operatorroles/common_utils.go
+++ b/cmd/create/operatorroles/common_utils.go
@@ -4,13 +4,27 @@ import (
 	"fmt"
 
 	awsCommonUtils "github.com/openshift-online/ocm-common/pkg/aws/utils"
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	errors "github.com/zgalor/weberr"
 
 	"github.com/openshift/rosa/pkg/aws"
+	awscb "github.com/openshift/rosa/pkg/aws/commandbuilder"
+	"github.com/openshift/rosa/pkg/aws/tags"
+	"github.com/openshift/rosa/pkg/helper"
 	"github.com/openshift/rosa/pkg/rosa"
 )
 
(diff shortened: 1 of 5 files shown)


cmd/create/operatorroles/common_utils.go
```diff
@@ -60,3 +64,49 @@ func validateIngressOperatorPolicyOverride(r *rosa.Runtime, policyArn string, sh
...
+		map[string]string{tags.RedHatManaged: helper.True}, path)
+	if err != nil {
+		return policyArn, err
+	}
+	return policyArn, nil
+}
+
+func getHcpSharedVpcPolicyDetails(r *rosa.Runtime, roleArn string, roleName string, iamTags map[string]string,
+	path string) (string, string) {
+	policyDetails := ` \
```
robpblake, line 89: @hunterkepley Can we extract this to a `const`? It looks the same as the definition of the `policyDetails` variable on line 70?

cmd/create/operatorroles/common_utils.go
```diff
@@ -60,3 +64,49 @@ func validateIngressOperatorPolicyOverride(r *rosa.Runtime, policyArn string, sh
 
 	return nil
 }
+
+func getHcpSharedVpcPolicy(r *rosa.Runtime, roleArn string, roleName string,
```
robpblake, line 68: @hunterkepley Shall we add unit tests for these functions?

cmd/create/operatorroles/common_utils.go
hunterkepley: I can extract to a const, but sadly it is not the same as the var on `L70`. Using the var on `L70` does not work with the AWS CLI's `--policy-document` body, and will not create the policy. It must be a byte string with no `\n`'s

Technically I could create a file like managed policies, but it is much cleaner and simpler for the user like this

PR #2635, opened 2024-11-19
OCM-12660 | feat: Refactor op + acc roles sharedvpc policies

hunterkepley:
Adds route53 assume role policy to control plane + gets rid of needless func for EnsurePolicyWithName + refactors a couple of things (fixing a bug)

### DIFF
--- a/cmd/create/accountroles/cmd.go
+++ b/cmd/create/accountroles/cmd.go
@@ -171,10 +171,6 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
-	if !cmd.Flag("hosted-cp").Changed {
-		rosa.HostedClusterOnlyFlag(r, cmd, route53RoleArnFlag)
-		rosa.HostedClusterOnlyFlag(r, cmd, vpcEndpointRoleArnFlag)
-	}
 	isHcpSharedVpc, err := validateSharedVpcInputs(args.hostedCP, args.vpcEndpointRoleArn, args.route53RoleArn)
 	if err != nil {
 		r.Reporter.Errorf("%s", err)
@@ -407,6 +403,11 @@ func run(cmd *cobra.Command, argv []string) {
 			os.Exit(1)
 		}
 		isHostedCPValueSet = true
+	}
+
+	if !createHostedCP {
+		rosa.HostedClusterOnlyFlag(r, cmd, route53RoleArnFlag)
+		rosa.HostedClusterOnlyFlag(r, cmd, vpcEndpointRoleArnFlag)
 	}
 
 	rolesCreator, createRoles := initCreator(r, managedPolicies, createClassic, createHostedCP,
--- a/cmd/create/accountroles/creators.go
+++ b/cmd/create/accountroles/creators.go
@@ -201,7 +201,7 @@ func (up *unmanagedPoliciesCreator) printCommands(r *rosa.Runtime, input *accoun
 
 		createPolicy := buildCreatePolicyCommand(policyName, policyDocument, iamTags, input.path)
 
-		policyARN := aws.GetPolicyARN(r.Creator.Partition, input.accountID, accRoleName, input.path)
+		policyARN := aws.GetPolicyArnWithSuffix(r.Creator.Partition, input.accountID, accRoleName, input.path)
 
 		attachRolePolicy := buildAttachRolePolicyCommand(accRoleName, policyARN)
 
@@ -279,7 +279,7 @@ func createRoleUnmanagedPolicy(r *rosa.Runtime, input *accountRolesCreationInput
 
 	policyPermissionDetail := aws.GetPolicyDetails(input.policies, filename)
 
-	policyARN := aws.GetPolicyARN(r.Creator.Partition, r.Creator.AccountID, accRoleName, input.path)
+	policyARN := aws.GetPolicyArnWithSuffix(r.Creator.Partition, r.Creator.AccountID, accRoleName, input.path)
 
 	r.Reporter.Debugf("Creating permission policy '%s'", policyARN)
 	if args.forcePolicyCreation {
@@ -467,9 +467,9 @@ func attachHcpSharedVpcPolicy(r *rosa.Runtime, sharedVpcRoleArn string, roleName
 		return err
 	}
 	policyName := fmt.Sprintf(aws.AssumeRolePolicyPrefix, userProvidedRoleName)
-	policyArn := aws.GetPolicyARN(r.Creator.Partition, r.Creator.AccountID, roleName, path)
-	policyArn, err = r.AWSClient.EnsurePolicyWithName(policyArn, policyDetails,
-		defaultPolicyVersion, policyTags, path, policyName)
+	policyArn := aws.GetPolicyArn(r.Creator.Partition, r.Creator.AccountID, policyName, path)
+	policyArn, err = r.AWSClient.EnsurePolicy(policyArn, policyDetails,
+		defaultPolicyVersion, policyTags, path)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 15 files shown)


cmd/create/operatorroles/by_prefix.go
```diff
@@ -304,6 +304,7 @@ func createRolesByPrefix(r *rosa.Runtime, prefix string, permissionsBoundary str
 		}
 
 		var policyArn string
+		var policyArns []string
```
hunterkepley, line 307: This change is needed with multiple policy attachments. Originally, the normal managed policy was being eaten

cmd/create/accountroles/creators.go
```diff
@@ -467,9 +467,10 @@ func attachHcpSharedVpcPolicy(r *rosa.Runtime, sharedVpcRoleArn string, roleName
 		return err
 	}
 	policyName := fmt.Sprintf(aws.AssumeRolePolicyPrefix, userProvidedRoleName)
-	policyArn := aws.GetPolicyARN(r.Creator.Partition, r.Creator.AccountID, roleName, path)
-	policyArn, err = r.AWSClient.EnsurePolicyWithName(policyArn, policyDetails,
-		defaultPolicyVersion, policyTags, path, policyName)
+	fmt.Println(path)
```
gdbranco, line 470: Missed to remove

cmd/create/operatorroles/by_prefix.go
```diff
@@ -398,6 +406,16 @@ func createRolesByPrefix(r *rosa.Runtime, prefix string, permissionsBoundary str
 		if err != nil {
 			return err
 		}
+
+		if len(policyArns) > 0 {
```
gdbranco, line 410: Is it required to check the length? If there are none the loop will just be skipped anyway

cmd/create/operatorroles/by_prefix.go
```diff
@@ -398,6 +406,16 @@ func createRolesByPrefix(r *rosa.Runtime, prefix string, permissionsBoundary str
...
+			for _, arn := range policyArns {
+				r.Reporter.Debugf("Attaching HCP shared VPC permission policy '%s' to role '%s'", arn, roleName)
```
gdbranco, line 412: missed to remove

cmd/create/operatorroles/by_prefix.go
```diff
@@ -304,6 +304,7 @@ func createRolesByPrefix(r *rosa.Runtime, prefix string, permissionsBoundary str
 		}
 
 		var policyArn string
```
gdbranco, line 306: Considering this only exists now to do [this](https://github.com/openshift/rosa/blob/71c1af575bf4fbd6d88610a8201cf2402df132bd/cmd/create/operatorroles/by_prefix.go#L405), can't it also be included in the new slice of policy arns?

pkg/aws/helpers.go
```diff
@@ -455,6 +455,10 @@ func GetPolicyARN(partition string, accountID string, name string, path string)
 	return getPolicyARN(partition, accountID, GetPolicyName(name), path)
 }
 
+func GetNormalPolicyARN(partition string, accountID string, name string, path string) string {
```
gdbranco, line 458: Can we rename the other one to GetOcmPolicyArn and this to GetPolicyArn instead? :thinking: Not sure if it would make it clearer or not to be truthful

cmd/create/operatorroles/common_utils.go
```diff
@@ -75,17 +74,19 @@ func validateIngressOperatorPolicyOverride(r *rosa.Runtime, policyArn string, sh
 	return nil
 }
 
-func getHcpSharedVpcPolicy(r *rosa.Runtime, roleArn string, roleName string,
-	operator *cmv1.STSOperator, path string, defaultPolicyVersion string) (string, error) {
+func getHcpSharedVpcPolicy(r *rosa.Runtime, roleArn string, path string, defaultPolicyVersion string) (string, error) {
 	policyDetails := "{\n  \"Version\": \"2012-10-17\",\n  \"Statement\": {\n    \"Effect\": \"Allow\",\n    " +
```
gdbranco, line 78: could be a const

cmd/create/operatorroles/cmd.go
```diff
@@ -162,7 +162,9 @@ func run(cmd *cobra.Command, argv []string) {
 	r := rosa.NewRuntime().WithAWS().WithOCM()
 	defer r.Cleanup()
 
-	rosa.HostedClusterOnlyFlag(r, cmd, vpcEndpointRoleArnFlag)
+	if !cmd.Flag("hosted-cp").Changed {
```
gdbranco, line 165: Not entirely sure if I follow, also considering the cmd is suplied to the function this could be done internally instead :thinking:

cmd/create/operatorroles/by_prefix.go
marcolan018: +1, the check is unecessary

pkg/aws/helpers.go
hunterkepley: Yeah this is a tough one to name. I think ill rename the other to `GetPolicyArnWithSuffix` and this one as `GetPolicyArn`

cmd/create/operatorroles/cmd.go
hunterkepley: Updated func

cmd/create/accountroles/cmd.go
```diff
@@ -409,6 +405,11 @@ func run(cmd *cobra.Command, argv []string) {
 		isHostedCPValueSet = true
 	}
 
+	if !createHostedCP {
+		rosa.HostedClusterOnlyFlag(r, cmd, route53RoleArnFlag)
+		rosa.HostedClusterOnlyFlag(r, cmd, vpcEndpointRoleArnFlag)
+	}
+
```
hunterkepley, line 412: Moved this here so it will only check if not also creating hostedCP roles (if creating both)

cmd/create/operatorroles/by_prefix.go
```diff
@@ -312,17 +313,24 @@ func createRolesByPrefix(r *rosa.Runtime, prefix string, permissionsBoundary str
...
+					sharedVpcPolicyArn, err := getHcpSharedVpcPolicy(r, sharedVpcRoleArn, roleName, defaultPolicyVersion)
 					if err != nil {
 						return err
 					}
+					policyArns = append(policyArns, sharedVpcPolicyArn)
 				} else if credrequest == aws.ControlPlaneCloudCredentialsRoleType {
-					policyArn, err = getHcpSharedVpcPolicy(r, sharedVpcEndpointRoleArn, roleName,
-						operator, path, defaultPolicyVersion)
+
+					sharedVpcPolicyArn, err := getHcpSharedVpcPolicy(r, sharedVpcEndpointRoleArn, path,
```
gdbranco, line 323: Could be a loop over []string{sharedVpcEndpointRoleArn, sharedVpcRoleArn} instead of repeating the pattern twice

pkg/rosa/helpers.go
```diff
@@ -6,10 +6,14 @@ import (
...
 )
 
+const hostedCpFlagName = "hosted-cp"
+
 func HostedClusterOnlyFlag(r *Runtime, cmd *cobra.Command, flagName string) {
-	isFlagSet := cmd.Flags().Changed(flagName)
-	if isFlagSet {
-		r.Reporter.Errorf("Setting the `%s` flag is only supported for hosted clusters", flagName)
-		os.Exit(1)
+	if cmd.Flag(hostedCpFlagName) == nil && (cmd.Flag(hostedCpFlagName) != nil && !cmd.Flag(hostedCpFlagName).Changed) {
```
gdbranco, line 12: How is it going to be both nil and not nil :thinking: ? Or maybe I'm just reading this wrong

hunterkepley: Ugh, meant `||`

PR #2645, opened 2024-11-25
OCM-12543 | feat: Manual mode for account roles with shared-vpc for hcp

hunterkepley:
Also has changes made to validation of sharedvpc role arn flags, making sure to account for when a user creates both classic and HCP accountroles


cmd/create/accountroles/cmd.go
```diff
@@ -437,6 +452,7 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 
+	fmt.Println(isHcpSharedVpc)
```
gdbranco, line 455: missed to remove

cmd/create/accountroles/creators.go
```diff
@@ -384,6 +384,38 @@ func (hcp *hcpManagedPoliciesCreator) printCommands(r *rosa.Runtime, input *acco
 				return err
 			}
 
+			isHcpInstallerRole := role.Name == aws.HCPAccountRoles[aws.HCPInstallerRole].Name
+
+			if isHcpInstallerRole && input.isSharedVpc { // HCP shared VPC
+				// Route53 Policy
+				userProvidedRoleName, err := aws.GetResourceIdFromARN(args.route53RoleArn)
```
gdbranco, line 391: I find that a loop may help in readability considering it does the same steps for both

cmd/create/accountroles/validation.go
```diff
@@ -6,25 +6,26 @@ import (
...
-	route53RoleArn string) (bool, error) {
-	if hostedCp {
-		if vpcEndpointRoleArn != "" && route53RoleArn == "" {
-			return false, errors.UserErrorf(
-				arguments.MustUseBothFlagsErrorMessage,
-				route53RoleArnFlag,
-				vpcEndpointRoleArnFlag,
-			)
-		}
+func validateSharedVpcInputs(vpcEndpointRoleArn string, route53RoleArn string) (bool, error) {
```
gdbranco, line 9: we should probably have a separate package like roles/validation because the behavior seems to be shared

PR #2647, opened 2024-11-25
OCM-12805 | fix: Refactor manual mode for create/oproles

hunterkepley:
Adds both policies to control plane op role (originally the ask was for 1 of the new policies to be added to this role, but that has changed)

### DIFF
--- a/cmd/create/operatorroles/by_prefix.go
+++ b/cmd/create/operatorroles/by_prefix.go
@@ -522,29 +522,64 @@ func buildCommandsFromPrefix(r *rosa.Runtime, env string,
 			Build()
 
 		var attachSharedVpcRolePolicy string
-		var createPolicyCommand string
+		var policyCommands []string
+
+		type manualSharedVpcPolicyDetails struct {
+			command string
+			name    string
+		}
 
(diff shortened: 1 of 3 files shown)


cmd/create/operatorroles/by_prefix.go
```diff
@@ -522,29 +522,51 @@ func buildCommandsFromPrefix(r *rosa.Runtime, env string,
...
-					iamTags, path)
+
+			// Precreate HCP shared VPC policies for less memory usage + time to execute
+			var policyDetails = make(map[string][]string)
+			createPolicyCommand, policyName, err := getHcpSharedVpcPolicyDetails(r, sharedVpcRoleArn,
+				iamTags)
+			if err != nil {
+				return "", err
+			}
+			policyDetails[aws.IngressOperatorCloudCredentialsRoleType] = []string{createPolicyCommand, policyName}
```
gdbranco, line 536: nit: really don't like using arbitrary structures like that. Would you mind creating some transition struct instead?
Like 
```
type manualPolicyData struct {
command string
name string
}
```

PR #2656, opened 2024-11-27
OCM-12806 | feat: Delete operator roles auto mode (and changes to account roles)

hunterkepley:
There is a map which we return from the function which merges with maps from the last call (deleteOperatorRole is called once per role so we must merge these, and only print out warnings that we could not delete the policies ONCE PER POLICY). The issue is that, without this, you will get prints 1) if they are deleted in the end and 2) multiple times for the route53 policy

With these changes, it only prints the warning when it is unable to delete the policy one time. One role may be unable to delete a policy (controlplane) if the other still has the role attached (ingress). We cannot do this before or after function calls because it is designed to ask the user per every oprole if they want to delete the role.

There is an extra change to account roles which makes sure you are using hcpsharedvpc roles before asking if you want to attempt deleting the hcpsharedvpc roles. This way it doesn't ask if you are using non-sharedvpc hcp account or op roles

In the end, this results in good UX, where the user is asked once if they want to delete these policies, and will tell you **once** if the policy was deleted, or unable to be deleted due to attachments to other resources

### DIFF
--- a/pkg/roles/utils.go
+++ b/pkg/roles/utils.go
@@ -58,3 +58,29 @@ func GetHcpSharedVpcPolicyDetails(r *rosa.Runtime, roleArn string) (bool, string
 
 	return existsQuery != nil, createPolicy, policyName, nil
 }
+
+func CheckIfRolesAreHcpSharedVpc(r *rosa.Runtime, roles []string) bool {
+	isHcpSharedVpc := false
+	for _, roleName := range roles {
+		ptrRoleName := roleName
+		attachedPolicies, err := r.AWSClient.GetPolicyDetailsFromRole(&ptrRoleName)
+		if err != nil {
+			r.Reporter.Errorf("Failed to get policy details for role '%s': %v", roleName, err)
+		}
+		for _, attachedPolicy := range attachedPolicies {
+			rhManaged := false
+			hcpSharedVpc := false
+			for _, tag := range attachedPolicy.Policy.Tags {
+				if *tag.Key == tags.RedHatManaged {
+					rhManaged = true
+				} else if *tag.Key == tags.HcpSharedVpc {
+					hcpSharedVpc = true
+				}
+			}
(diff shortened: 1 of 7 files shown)


pkg/roles/utils.go
```diff
@@ -58,3 +58,29 @@ func GetHcpSharedVpcPolicyDetails(r *rosa.Runtime, roleArn string) (bool, string
...
+func CheckIfRolesAreHcpSharedVpc(r *rosa.Runtime, roles []string) bool {
+	isHcpSharedVpc := false
+	for _, roleName := range roles {
+		ptrRoleName := roleName
+		attachedPolicies, err := r.AWSClient.GetPolicyDetailsFromRole(&ptrRoleName)
+		if err != nil {
+			r.Reporter.Errorf("Failed to get policy details for role '%s': %v", roleName, err)
+		}
+		for _, attachedPolicy := range attachedPolicies {
+			rhManaged := false
```
den-rgb, line 71: Might be worth making lines 71 - 79 in to a helper func since its being duplicated in the follow up draft pr

pkg/roles/utils.go
```diff
@@ -58,3 +58,29 @@ func GetHcpSharedVpcPolicyDetails(r *rosa.Runtime, roleArn string) (bool, string
 
 	return existsQuery != nil, createPolicy, policyName, nil
 }
+
+func CheckIfRolesAreHcpSharedVpc(r *rosa.Runtime, roles []string) bool {
+	isHcpSharedVpc := false
+	for _, roleName := range roles {
+		ptrRoleName := roleName
```
den-rgb, line 65: can we not just use `&roleName`

hunterkepley: Using a direct pointer of a for loop object (`for _, OBJECT := range ...`) results in a failure to build in golfing, since it is a temporary object and the memory address is freed every iteration of the for loop. I think it's called `implicit memory aliasing`

PR #2660, opened 2024-11-28
OCM-12828 | fix: Do not print create policy commands more than once op roles

hunterkepley:


cmd/create/operatorroles/by_prefix.go
```diff
@@ -529,52 +531,58 @@ func buildCommandsFromPrefix(r *rosa.Runtime, env string,
...
-			exists, createPolicyCommand, policyName, err := roles.GetHcpSharedVpcPolicyDetails(r, sharedVpcRoleArn)
-			if err != nil {
-				return "", err
-			}
-			policyDetails[aws.IngressOperatorCloudCredentialsRoleType] = roles.ManualSharedVpcPolicyDetails{
-				Command:       createPolicyCommand,
-				Name:          policyName,
-				AlreadyExists: exists,
+			if _, ok := policyDetails[aws.IngressOperatorCloudCredentialsRoleType]; !ok {
+				exists, createPolicyCommand, policyName, err := roles.GetHcpSharedVpcPolicyDetails(r, sharedVpcRoleArn)
```
den-rgb, line 535: Could this be modified to use a common function since L534 -543 and L546 - 557 are pretty much the same

cmd/create/operatorroles/by_prefix.go
```diff
@@ -529,52 +531,58 @@ func buildCommandsFromPrefix(r *rosa.Runtime, env string,
...
 						policyDetails[credrequest] = details
 					}
 				}
 			case aws.ControlPlaneCloudCredentialsRoleType:
 				for i, details := range policyDetails {
 					policies = append(policies, details.Name)
-					if details.AlreadyExists { // Skip creation if already exists
-						continue
+					if !details.AlreadyExists {
+						if details.AlreadyExists { // Skip creation if already exists
```
den-rgb, line 578: This is unecessary

cmd/create/operatorroles/by_prefix.go
hunterkepley: Set up a function for this and realized it may not be as useful as I expected. It's duplicated but it's a function call with an error check, and then adding the output to a map, when put into a function it takes 4/5 parameters, set up something else to reduce duplication in a less-confusing-to-future-contributors way

hunterkepley: I think in this case, after moving to another function; readability is impacted negatively somewhat, and I believe the function made is too specific to be used in another place in the codebase. The function ends up being something like `attachHcpSharedVpcPolicyDetailsByOperatorRole` with 4 parameters, for essentially 2 lines of code (ignoring the error check since it will be returned and checks again in the function shown here anyway)

The code is a function call, then passing the output into a map, even though maps are a reference type and this can be done by just passing a map in and not returning the values, it feels too specific to add an extra function call / jump around point for future contributors to look at

cmd/create/operatorroles/by_prefix.go
hunterkepley: Done

PR #2662, opened 2024-12-02
OCM-12851| feat: default values for CAS Max Nodes Total and validation added for count of mp worker nodes

davidleerh:


pkg/clusterautoscaler/flags.go
```diff
@@ -945,3 +953,50 @@ func PrefillAutoscalerArgs(cmd *cobra.Command, autoscalerArgs *AutoscalerArgs,
...
+	return workerNodeCount + infraNodeCount + masterNodeCount
+}
+
+// classic clusters only
+func validateAutoscalerMaxNodesTotal(autoscalerValidationArgs *AutoscalerValidationArgs) interactive.Validator {
+	return func(val interface{}) error {
+		maxNodesTotalDefaultValue := getAutoscalerMaxNodesTotalDefaultValue(0, autoscalerValidationArgs)
+		number, err := strconv.Atoi(fmt.Sprintf("%v", val))
+		if err != nil {
+			return fmt.Errorf("should provide an integer number between 0 to 2147483647")
```
hunterkepley, line 990: I think go has a built-in value for the max int value allowed (`math.MaxInt`) - I see this number is used a couple times in this MR is why I bring it up :) 

Hope I'm not being too nitpicky with some of my comments

hunterkepley: Unless there's a specific max we need here, like if we are converting to an int16 or int8 or something in CS (not sure if there's a constant value for a max int8 🤔 )

pkg/machinepool/machinepool_test.go
```diff
@@ -550,9 +550,16 @@ var _ = Describe("Utility Functions", func() {
...
 
 		BeforeEach(func() {
-			validator = minReplicaValidator(true, false, false) // or false for non-multiAZ
+			replicaSizeValidationArgs = &ReplicaSizeValidationArgs{
+				ClusterVersion: "openshift-v4.14.14",
+				MultiAz:        true,
+				IsHostedCp:     false,
+				Autoscaling:    false,
+			}
+			validator = minReplicaValidator(replicaSizeValidationArgs) // or false for non-multiAZ
```
hunterkepley, line 562: I very much welcome this change haha, `minReplicaValidator(true, false, false)` is not a pretty function call..

hunterkepley (commented):
looks great overall, couple nitpicks, couple discussion points/safety requests 🙂

cmd/create/cluster/cmd.go
```diff
@@ -3610,35 +3626,53 @@ func isValidCidrRange(
...
+			replicaSizeValidationArgs.IsHostedCp,
+			replicaSizeValidationArgs.PrivateSubnetsCount,
+		)
 	}
 }
 
-func maxReplicaValidator(multiAZ bool, minReplicas int, isHostedCP bool,
-	privateSubnetsCount int) interactive.Validator {
+func maxReplicaValidator(
+	minReplicas int, replicaSizeValidationArgs *machinepool.ReplicaSizeValidationArgs) interactive.Validator {
```
gdbranco, line 3657: Any reason not to include the min replicas to the validation args?

gdbranco: Or maybe have the max replicas validator function be part of the ReplicaSizeValidationArgs struct instead so you don't need to supply it but just apply based on existing instance

pkg/machinepool/helper.go
```diff
@@ -198,33 +208,35 @@ func maxReplicaValidator(minReplicas int, multiAZMachinePool bool) interactive.V
...
 			return fmt.Errorf("Multi AZ clusters require that the replicas be a multiple of 3")
 		}
-		return nil
+		return ValidateClusterVersionWithMaxNodesLimit(
+			replicaSizeValidationArgs.ClusterVersion, maxReplicas, replicaSizeValidationArgs.IsHostedCp)
 	}
 }
 
-func minReplicaValidator(multiAZMachinePool bool, autoscaling bool, isHypershift bool) interactive.Validator {
+func minReplicaValidator(replicaSizeValidationArgs *ReplicaSizeValidationArgs) interactive.Validator {
```
gdbranco, line 219: Could be part of the ReplicaSizeValidationArgs struct instead

pkg/machinepool/helper.go
```diff
@@ -189,7 +199,7 @@ func getMachinePoolAvailabilityZones(r *rosa.Runtime, cluster *cmv1.Cluster, mul
 	return cluster.Nodes().AvailabilityZones(), nil
 }
 
-func maxReplicaValidator(minReplicas int, multiAZMachinePool bool) interactive.Validator {
+func maxReplicaValidator(minReplicas int, replicaSizeValidationArgs *ReplicaSizeValidationArgs) interactive.Validator {
```
gdbranco, line 202: Could be part of the ReplicaSizeValidationArgs struct instead

pkg/clusterautoscaler/flags.go
```diff
@@ -153,7 +160,7 @@ func AddClusterAutoscalerFlags(cmd *cobra.Command, prefix string) *AutoscalerArg
 	cmd.Flags().IntVar(
 		&args.ResourceLimits.MaxNodesTotal,
 		fmt.Sprintf("%s%s", prefix, maxNodesTotalFlag),
-		249,
+		0,
```
gdbranco, line 163: Should default be 0? Maybe nil and use pointer to int instead?

cmd/create/cluster/cmd.go
davidleerh: I can go ahead and add "MinReplicas" an attribute for the "ReplicaSizeValidationArgs" struct.
However, I am personally a bit hesitant with setting "maxReplicaValidator" or "minReplicaValidator" as a method for the struct because, it seems we have two functions, (from different pkg) sharing the same function name but with the same signature. 

https://github.com/openshift/rosa/blob/cas_validations_rosacli/cmd/create/cluster/cmd.go#L3629-L3678
VS.
https://github.com/openshift/rosa/blob/cas_validations_rosacli/pkg/machinepool/helper.go#L202-L241

The two functions have slightly different validation conditions, which I fear may cause some regression issues if I abruptly merge them together. 

A spike should be created as a separate effort to see if min max replica size validations should differ between day 1 and day 2 events. WDYT?

pkg/machinepool/helper.go
davidleerh: https://github.com/openshift/rosa/pull/2662#discussion_r1870186738

pkg/machinepool/helper.go
davidleerh: https://github.com/openshift/rosa/pull/2662#discussion_r1870186738

pkg/clusterautoscaler/flags.go
davidleerh: Setting 0 seems to omit the "default" suggestion in the "-help" option screen in the CLI.
Since this value tends to change depending on multiple variable conditions, i've felt that 0 might be the best substitution to hide the default suggestion, as well as prevent any "nil" ptr exceptions.

gdbranco: Oh I see

cmd/create/cluster/cmd.go
gdbranco: I don't think you need to merge it together but just have two methods with different names :thinking: , considering they check different things anyways

PR #2674, opened 2024-12-03
OCM-12871 | feat: Add manual mode for deleting hcp sharedvpc policies [oproles]

hunterkepley:
Also fixes bug with multiple printouts of manual mode for deleting the hcp sharedvpc policies for accountroles

### DIFF
--- a/cmd/dlt/accountroles/cmd.go
+++ b/cmd/dlt/accountroles/cmd.go
@@ -237,9 +237,10 @@ func deleteAccountRoles(r *rosa.Runtime, env string, prefix string, clusters []*
 		if roles.CheckIfRolesAreHcpSharedVpc(r, finalRoleList) &&
 			confirm.Prompt(true, "Create commands to delete Hosted CP shared VPC policies?") {
 			for _, role := range finalRoleList {
-				policiesOutput, err = r.AWSClient.GetPolicyDetailsFromRole(awssdk.String(role))
+				policies, err := r.AWSClient.GetPolicyDetailsFromRole(awssdk.String(role))
+				policiesOutput = append(policiesOutput, policies...)
 				if err != nil {
-					r.Reporter.Infof("There was an error getting details of policies attached to role '%s': %v",
+					r.Reporter.Warnf("There was an error getting details of policies attached to role '%s': %v",
 						role, err)
 				}
 			}
@@ -319,6 +320,8 @@ func buildCommand(roleNames []string, policyMap map[string][]aws.PolicyDetail,
 	arbitraryPolicyMap map[string][]aws.PolicyDetail, managedPolicies bool,
 	hcpSharedVpcPoliciesOutput []*iam.GetPolicyOutput) string {
 	commands := []string{}
+	hcpSharedVpcPolicyCommands := make(map[string]string) // Ensures no duplicate delete policy cmds for hcp sharedvpc
+
 	for _, roleName := range roleNames {
 		policyDetails := policyMap[roleName]
 		excludedPolicyDetails := arbitraryPolicyMap[roleName]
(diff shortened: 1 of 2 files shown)


cmd/dlt/operatorrole/cmd.go
```diff
@@ -270,7 +273,22 @@ func run(cmd *cobra.Command, _ []string) {
...
+
+		// Get HCP shared vpc policy details if the user is deleting roles related to HCP shared vpc
+		policiesOutput := make([]*iam.GetPolicyOutput, 0)
+		if roles.CheckIfRolesAreHcpSharedVpc(r, foundOperatorRoles) &&
+			confirm.Prompt(true, "Create commands to delete Hosted CP shared VPC policies?") {
+			for _, role := range foundOperatorRoles {
+				policies, err := r.AWSClient.GetPolicyDetailsFromRole(awssdk.String(role))
+				policiesOutput = append(policiesOutput, policies...)
+				if err != nil {
+					r.Reporter.Infof("There was an error getting details of policies attached to role '%s': %v",
```
marcolan018, line 285: why not use r.Reporter.Errorf() here?

cmd/dlt/operatorrole/cmd.go
```diff
@@ -341,6 +362,30 @@ func buildCommand(r *rosa.Runtime, roleNames []string, policyMap map[string][]st
 			AddParam(awscb.RoleName, roleName).
 			Build()
 		commands = append(commands, deleteRole)
+
+		if len(hcpSharedVpcPoliciesOutput) > 0 { // Delete HCP shared VPC policies
```
marcolan018, line 366: seems this line is not necessary

cmd/dlt/operatorrole/cmd.go
```diff
@@ -341,6 +362,30 @@ func buildCommand(r *rosa.Runtime, roleNames []string, policyMap map[string][]st
...
+				hasRhManagedTag := false
+				hasHcpSharedVpcTag := false
+				for _, tag := range hcpSharedVpcPolicy.Policy.Tags {
+					if *tag.Key == tags.RedHatManaged {
+						hasRhManagedTag = true
+					} else if *tag.Key == tags.HcpSharedVpc {
+						hasHcpSharedVpcTag = true
+					}
+				}
+				if hasHcpSharedVpcTag && hasRhManagedTag {
```
marcolan018, line 377: is it possible that a policy only has tag tag.HcpSharedVpc, but doesn't have tags.RedHatManaged?

cmd/dlt/operatorrole/cmd.go
hunterkepley: It is better to not error here as customers are not expected to have set up the right permissions for our aws client to work with fetching policies, so we want to provide the normal role deletion commands even if the client is unable to fetch the policies

cmd/dlt/operatorrole/cmd.go
hunterkepley: hm, I guess not no, maybe its not needed, but the policies created will always have both tags, WDYT?

hunterkepley: maybe it's not needed to check for both* but it's not possible to create these policies without both

cmd/dlt/accountroles/cmd.go
```diff
@@ -364,26 +367,29 @@ func buildCommand(roleNames []string, policyMap map[string][]aws.PolicyDetail,
...
+				deletePolicy := awscb.NewIAMCommandBuilder().
+					SetCommand(awscb.DeletePolicy).
+					AddParam(awscb.PolicyName, *hcpSharedVpcPolicy.Policy.PolicyName).
+					Build()
+				hcpSharedVpcPolicyCommands[*hcpSharedVpcPolicy.Policy.PolicyName] = deletePolicy
+			}
 		}
 	}
+
+	for _, command := range hcpSharedVpcPolicyCommands {
```
gdbranco, line 391: Does the order of the commands matter? If it does it shouldn't be a map

cmd/dlt/operatorrole/cmd.go
gdbranco: Should it be a warn instead then?

cmd/dlt/operatorrole/cmd.go
```diff
@@ -341,6 +362,30 @@ func buildCommand(r *rosa.Runtime, roleNames []string, policyMap map[string][]st
...
+				deletePolicy := awscb.NewIAMCommandBuilder().
+					SetCommand(awscb.DeletePolicy).
+					AddParam(awscb.PolicyName, *hcpSharedVpcPolicy.Policy.PolicyName).
+					Build()
+				hcpSharedVpcPolicyCommands[*hcpSharedVpcPolicy.Policy.PolicyName] = deletePolicy
+			}
+		}
+	}
+
+	for _, command := range hcpSharedVpcPolicyCommands {
```
gdbranco, line 387: Does the order of the commands matter? If it does it shouldn't be a map

cmd/dlt/accountroles/cmd.go
hunterkepley: Order should not matter, deleting the policies can be done in any order I would think since they can only be deleted if they are not attached to anything, and the commands are printed last. the policies have no connection to each other

PR #2690, opened 2024-12-05
OCM-12960 | fix: Don't use path for hcp shared vpc policies

hunterkepley:


gdbranco:
Closing as we need path, required to consider other solutions

PR #2692, opened 2024-12-09
OCM-12964 | feat: Allow use of flag for deleting hcpsharedvpc policies

hunterkepley:
Allows use of script with manual mode, and skipping of prompt when using auto/interactive when deleting hcp shared vpc policies

### DIFF
--- a/cmd/dlt/accountroles/cmd.go
+++ b/cmd/dlt/accountroles/cmd.go
@@ -37,10 +37,15 @@ import (
 	"github.com/openshift/rosa/pkg/rosa"
 )
 
+const (
+	deleteHcpSharedVpcPoliciesFlagName = "delete-hcp-shared-vpc-policies"
+)
+
 var args struct {
-	prefix   string
-	hostedCP bool
-	classic  bool
+	prefix                     string
+	hostedCP                   bool
+	classic                    bool
+	deleteHcpSharedVpcPolicies bool
 }
 
 var Cmd = &cobra.Command{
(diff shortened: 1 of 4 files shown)


cmd/dlt/accountroles/cmd.go
```diff
@@ -182,8 +194,8 @@ func setDeleteRoles(isClassicFlagSet bool, isHostedCPFlagSet bool) (bool, bool)
 	return isClassicFlagSet, isHostedCPFlagSet
 }
 
-func deleteAccountRoles(r *rosa.Runtime, env string, prefix string, clusters []*cmv1.Cluster, mode string,
-	hostedCP bool) error {
+func deleteAccountRoles(r *rosa.Runtime, cmd *cobra.Command, env string, prefix string, clusters []*cmv1.Cluster,
+	mode string, hostedCP bool) error {
```
gdbranco, line 198: If you are going to supply the cmd I suppose you could remove the mode param :thinking:

cmd/dlt/accountroles/cmd.go
```diff
@@ -203,28 +215,32 @@ func deleteAccountRoles(r *rosa.Runtime, env string, prefix string, clusters []*
...
+					continue
+				}
+				r.Reporter.Infof("Deleting account role '%s'", role)
+				err := r.AWSClient.DeleteAccountRole(role, prefix, managedPolicies, deleteHcpSharedVpcPolicies)
+				if err != nil {
+					r.Reporter.Warnf("There was an error deleting the account roles or policies: %s", err)
+					continue
+				}
 			}
+			r.Reporter.Infof(fmt.Sprintf("Successfully deleted the %saccount roles", roleTypeString))
```
den-rgb, line 242: Just formatting fix needed here

cmd/dlt/accountroles/cmd.go
hunterkepley: Ah actually after looking at it, I don't think we should, `mode` is a var scoped to the `run` function which can be changed after the flag itself

hunterkepley: I could include the `cmd.Flag(...).Changed` value though instead

PR #2696, opened 2024-12-09
OCM-12775 | feat: by_cluster_key operator role creation hcpsharedvpc

hunterkepley:


cmd/create/cluster/cmd.go
```diff
@@ -3461,6 +3461,13 @@ func run(cmd *cobra.Command, _ []string) {
 				if permissionsBoundary != "" {
 					rolesCMD = fmt.Sprintf("%s --permissions-boundary %s", rolesCMD, permissionsBoundary)
 				}
+				// HCP Shared VPC
+				if route53RoleArn != "" {
+					rolesCMD = fmt.Sprintf("%s --%s %s --hosted-cp", rolesCMD, route53RoleArnFlag, route53RoleArn)
```
marcolan018, line 3466: is "--hosted-cp" desired here?

hunterkepley: Yup 🙂  after testing it was required when making operator roles. This removes a manual addition of this flag from the user

PR #2700, opened 2024-12-10
OCM-13019 | feat: Describe Cluster changes for hcpsharedvpc

hunterkepley:


cmd/describe/cluster/cmd.go
```diff
@@ -447,13 +447,19 @@ func run(cmd *cobra.Command, argv []string) {
 			cluster.AWS().STS().OIDCEndpointURL(), managementType)
 	}
 	if cluster.AWS().PrivateHostedZoneID() != "" {
-		str = fmt.Sprintf("%s"+"Private Hosted Zone:\n", str)
+		str = fmt.Sprintf("%s"+"Shared VPC Config:\n", str)
 		str = fmt.Sprintf("%s"+
-			" - ID:                      %s\n", str,
+			" - Ingress Private HZ ID:   %s\n", str,
 			cluster.AWS().PrivateHostedZoneID())
 		str = fmt.Sprintf("%s"+
```
gdbranco, line 454: should check if it is hypershift I think

gdbranco: But we still need to provide the old params for non hypershift :0

hunterkepley: oh sorry, got confused from the requirements (**replace** X with Y) - will put back for non HS

PR #2702, opened 2024-12-10
OCM-13016 | fix: Include path in create acc/op roles manual mode

hunterkepley:


pkg/roles/utils.go
```diff
@@ -36,10 +36,11 @@ func GetHcpSharedVpcPolicyDetails(r *rosa.Runtime, roleArn string) (bool, string
 	if err != nil {
 		return false, "", "", err
 	}
+	path, err := aws.GetPathFromARN(roleArn)
```
marcolan018, line 39: the "err" is ignored?

hunterkepley: oops thanks

PR #2719, opened 2024-12-12
OCM-13068 | fix: Add required flags to oprole creation for cluster manual mode

hunterkepley:


cmd/create/cluster/cmd.go
```diff
@@ -3465,6 +3465,13 @@ func run(cmd *cobra.Command, _ []string) {
 			if !output.HasFlag() || r.Reporter.IsTerminal() {
 				r.Reporter.Infof("Preparing to create operator roles.")
 			}
+			// Set flags for HCP shared VPC if needed
+			if vpcEndpointRoleArn != "" && route53RoleArn != "" && isHostedCP {
+				operatorroles.Cmd.Flags().Set(route53RoleArnFlag, route53RoleArn)
+				operatorroles.Cmd.Flags().Set(vpcEndpointRoleArnFlag, vpcEndpointRoleArn)
+				operatorroles.Cmd.Flags().Set(operatorroles.HostedCpFlag, helper.True)
```
gdbranco, line 3472:
```suggestion
				operatorroles.Cmd.Flags().Set(operatorroles.HostedCpFlag, isHostedCP)
```
Wouldn't that work?

hunterkepley: should be true no matter what here but would definitely be cleaner

hunterkepley: oh, `.Set` requires a string

davidleerh: how about using "strconv.FormatBool(isHostedCP)"?

PR #2727, opened 2024-12-16
OCM-13124 | fix: Make route53 role arn usable for classic oproles

hunterkepley:
the route53 role arn was mistakenly put as HCP only for `create/operatorroles` - removed this validation

### DIFF
--- a/cmd/create/operatorroles/cmd.go
+++ b/cmd/create/operatorroles/cmd.go
@@ -181,7 +181,6 @@ func run(cmd *cobra.Command, argv []string) {
 	var isHcpSharedVpc bool
 	var err error
 	if !args.hostedCp {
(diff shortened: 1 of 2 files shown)


cmd/create/operatorroles/cmd.go
```diff
@@ -181,7 +181,6 @@ func run(cmd *cobra.Command, argv []string) {
 	var isHcpSharedVpc bool
 	var err error
 	if !args.hostedCp {
-		rosa.HostedClusterOnlyFlag(r, cmd, hostedZoneRoleArnFlag)
```
gdbranco, line 184: It should include the other hosted zone though, right? The hcp internal comms

hunterkepley: ahh my bad, thanks, Monday brain scrambles

PR #2739, opened 2024-12-23
OCM-13218 | test: fix the failure of cluster provision and resource clean up

xueli181114:


tests/utils/handler/interface.go
```diff
@@ -65,20 +65,26 @@ type ClusterConfig struct {
...
-	Region               string `json:"region,omitempty"`
-	ResourceShareArn     string `json:"resource_share,omitempty"`
-	SharedVPCRole        string `json:"shared_vpc_role,omitempty"`
-	VpcID                string `json:"vpc_id,omitempty"`
+	AccountRolesPrefix   string                `json:"account_roles_prefix,omitempty"`
+	AdditionalPrincipals string                `json:"additional_principals,omitempty"`
+	AuditLogArn          string                `json:"audit_log,omitempty"`
+	DNSDomain            string                `json:"dns_domain,omitempty"`
+	EtcdKMSKey           string                `json:"etcd_kms_key,omitempty"`
+	FromSharedAWSAccount *FromSharedAWSAccount `json:"from_shard_aws_account,omitempty"`
```
yasun1, line 73: typo error `from_shard_aws_account`?

tests/utils/handler/interface.go
```diff
@@ -65,20 +65,26 @@ type ClusterConfig struct {
...
+	HostedZoneID         string                `json:"hosted_zone_id,omitempty"`
+	KMSKey               string                `json:"kms_key,omitempty"`
+	OIDCConfigID         string                `json:"oidc_config_id,omitempty"`
+	OIDCProviderID       string                `json:"oidc_provider_id,omitempty"`
+	OperatorRolesPrefix  string                `json:"operator_roles_prefix,omitempty"`
+	Region               string                `json:"region,omitempty"`
+	ResourceShareArn     string                `json:"resource_share,omitempty"`
+	SharedVPCRole        string                `json:"shared_vpc_role,omitempty"`
+	VpcID                string                `json:"vpc_id,omitempty"`
+}
```
yasun1, line 83: Which resources are created by the regular AWS account, and which resources are created by the Shared AWS account, it is better to separate the resources created by the Shared AWS account in an individual block.

xueli181114: That is a better resolution needs more research to see if any other place relied on the resources.json. How about make this merged to resolve the problem of resources leak which is affecting all CI jobs?

PR #2743, opened 2025-01-08
OCM-13330 | ci: fix additional principle issue for rosa classic clust…

yuwang-RH:
One new involved issue when creating cluster with profiles of rosa classic clusters in ci job is caused by the empty '--additional-allowed-principals' flag


tests/utils/handler/cluster_handler.go
```diff
@@ -397,7 +397,7 @@ func (ch *clusterHandler) GenerateClusterCreateFlags() ([]string, error) {
 		if sharedVPCAdditionalPrincipalsForHostedCP != "" {
 			flags = append(flags, "--additional-allowed-principals",
 				fmt.Sprintf("%s,%s", sharedVPCAdditionalPrincipalsForHostedCP, additionalPrincipalRoleArn))
-		} else {
+		} else if additionalPrincipalRoleArn != "" {
 			flags = append(flags, "--additional-allowed-principals", additionalPrincipalRoleArn)
```
yingzhanredhat, line 401: I think --additional-allowed-principals is only supported for HCP cluster.

PR #2744, opened 2025-01-10
OCM-13406 | ci: fix one function issue leading the destroy and post-destroy cases failure

yuwang-RH:
OCM-13406 | ci: fix one function issue leading the destroy and post-destroy cases failure

In the jobs which additional principle enabled, the case 74556 creates an additional principle role using a temple resourceHandller and override the user data, this will lead the destroy step and post-destroy cases failures. See the logs of the failed cases in https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-rosa-master-e2e-rosa-hcp-private-link-critical-high-f3/1877074610999005184 as an example. 

The log with this fix is here, https://privatebin.corp.redhat.com/?4f51d2b672ed5c47#EV4QiJCuRX2xNwmj3eZsyhpD7YL7429HLUGnp4Bxdj86


tests/utils/handler/cluster_handler.go
```diff
@@ -394,11 +394,13 @@ func (ch *clusterHandler) GenerateClusterCreateFlags() ([]string, error) {
...
 		}
-		if sharedVPCAdditionalPrincipalsForHostedCP != "" {
+		if sharedVPCAdditionalPrincipalsForHostedCP != "" && additionalPrincipalRoleArn != "" {
 			flags = append(flags, "--additional-allowed-principals",
 				fmt.Sprintf("%s,%s", sharedVPCAdditionalPrincipalsForHostedCP, additionalPrincipalRoleArn))
-		} else if additionalPrincipalRoleArn != "" {
+		} else if sharedVPCAdditionalPrincipalsForHostedCP == "" && additionalPrincipalRoleArn != "" {
 			flags = append(flags, "--additional-allowed-principals", additionalPrincipalRoleArn)
+		} else if sharedVPCAdditionalPrincipalsForHostedCP != "" && additionalPrincipalRoleArn == "" {
+			flags = append(flags, "--additional-allowed-principals", sharedVPCAdditionalPrincipalsForHostedCP)
```
yuwang-RH, line 403: Before this change, the flag value of --additional-allowed-principals would be wrong with the value like "arn1,arn," with the comma at the end of the value. also add this little fix in this PR.

PR #2745, opened 2025-01-10
OCM-11581 | feat: Introduce SDN -> OVN migration capability

hunterkepley:


bchandra-ocp:
Requesting changes to user-facing verbiage:

1. It's OVN-Kubernetes. If you're mentioning OVN-Kubernetes, SDN must also be completed as OpenShiftSDN.
2. Help: "Changing the network plugin will reboot cluster nodes, can not be interrupted or rolled back, and can not be combined with other operations such as cluster upgrades. Confirm to proceed?"

cmd/edit/cluster/cmd.go
```diff
@@ -186,6 +198,21 @@ func init() {
...
+		"network-type",
+		"",
+		"Migrate a cluster's network type from OpenShiftSDN to OVN-Kubernetes",
+	)
+
+	flags.StringVar(
+		&args.subnetConfiguration,
+		"subnet-configuration",
+		"",
+		"Subnet configuration for migrating 'network-type' from OpenShiftSDN -> OVN-Kubernetes. Choices consist of "+
```
bchandra-ocp, line 213: Subnet configuration is ambiguous because there's VPC subnets.

It should ready 'Internal subnets used by OVN-Kubernetes'.

cmd/edit/cluster/cmd.go
```diff
@@ -985,3 +1136,14 @@ func BuildClusterConfigWithRegistry(clusterConfig ocm.Spec, allowedRegistries []
 	clusterConfig.AllowedRegistriesForImport = allowedRegistriesForImport
 	return clusterConfig, nil
 }
+
+func confirmMigration() (bool, error) {
+	 return interactive.GetBool(interactive.Input{
+		Question: "Changing the network plugin will reboot cluster nodes, can not be interrupted or rolled " +
+			"back, and can not be combined with other operations such as cluster upgrades. Are you sure that " +
+			"you want to proceed with migrating from 'OpenShiftSDN' to 'OVN-Kubernetes",
+		Help: "Confirm that you are wanting to migrate your cluster's network type, it may be safer to do " +
```
bchandra-ocp, line 1145: lgtm - although this is concise:

Confirm that _you want to migrate_ your cluster's network type

cmd/edit/cluster/cmd.go
```diff
@@ -186,6 +198,21 @@ func init() {
 		"",
 		"Account ID used for billing subscriptions purchased through the AWS console for ROSA",
 	)
+
+	flags.StringVar(
```
davidleerh, line 202: Should this be a bool instead since we are not expecting users to revert back to SDN from OVN?
@hunterkepley @AlexVulaj

hunterkepley: gonna let Alex decide, but when you create a cluster, the network-type is a string so I think the idea was parity with that

I'd be fine either way, a bool makes sense as well

AlexVulaj: Might be worth engaging product/UX folks, but my take is that as this is written, a bool probably makes the most sense.

PR #2766, opened 2025-01-29
OCM-11595 | feat: Add support for json/yml output with migrations describe

hunterkepley:


cmd/describe/cluster/cmd.go
```diff
@@ -102,6 +102,30 @@ func run(cmd *cobra.Command, argv []string) {
 	var upgradeState *cmv1.UpgradePolicyState
 	var controlPlaneScheduledUpgrade *cmv1.ControlPlaneUpgradePolicy
 
+	// SDN -> OVN Migration Status
+	migrationsStr := ""
+	collection, err := r.OCMClient.FetchClusterMigrations(cluster.ID())
+	if err != nil {
+		r.Reporter.Errorf("Failed to get cluster migrations: %v", err)
```
davidleerh, line 109: I am not sure if we should log it as an error here. @AlexVulaj CC: @hunterkepley wdyt?
Also i thought that we weren't going to exist the process if it fails to retrieve.
If we are required to exist the process, can we check and see if the error fails due to toggle related error?
maybe warn or debug?

cmd/describe/cluster/cmd.go
```diff
@@ -102,6 +102,29 @@ func run(cmd *cobra.Command, argv []string) {
 	var upgradeState *cmv1.UpgradePolicyState
 	var controlPlaneScheduledUpgrade *cmv1.ControlPlaneUpgradePolicy
 
+	// SDN -> OVN Migration Status
+	migrationsStr := ""
+	collection, err := r.OCMClient.FetchClusterMigrations(cluster.ID())
+	if err != nil {
+		r.Reporter.Warnf("Failed to get cluster migrations: %v", err)
```
gdbranco, line 109: If the reason for the failure is that the toggle isn't enabled I feel like it shouldn't be reported at all :thinking:

cmd/describe/cluster/cmd.go
```diff
@@ -102,6 +102,29 @@ func run(cmd *cobra.Command, argv []string) {
...
+	}
+
+	migrationsToShow := map[string]string{}
+	migrationTypes := map[string]string{}
+	if len(collection.Items().Items()) > 0 {
```
gdbranco, line 114: If an error did happen does an empty collection gets returned or a nil pointer?

cmd/describe/cluster/cmd.go
```diff
@@ -102,6 +102,29 @@ func run(cmd *cobra.Command, argv []string) {
...
 	var controlPlaneScheduledUpgrade *cmv1.ControlPlaneUpgradePolicy
 
+	// SDN -> OVN Migration Status
+	migrationsStr := ""
+	collection, err := r.OCMClient.FetchClusterMigrations(cluster.ID())
+	if err != nil {
+		r.Reporter.Warnf("Failed to get cluster migrations: %v", err)
+	}
+
+	migrationsToShow := map[string]string{}
```
gdbranco, line 112: Any reason to have two maps instead of something like a map to an object of migrations?

hunterkepley: Parallel maps felt simpler and less code for the result here, those reasons would mainly be why I chose that, we don't particularly need all info from the migrations

cmd/describe/cluster/cmd.go
hunterkepley: No clue, will only run this if there's no error

cmd/describe/cluster/cmd.go
```diff
@@ -102,6 +102,27 @@ func run(cmd *cobra.Command, argv []string) {
...
 
+	// SDN -> OVN Migration Status
+	migrationsStr := ""
+	collection, err := r.OCMClient.FetchClusterMigrations(cluster.ID())
+	migrations := map[string]*cmv1.ClusterMigration{}
+	if err != nil {
+		if !strings.Contains(err.Error(), "is not enabled for this organization") {
+			r.Reporter.Warnf("Failed to get cluster migrations: %v", err)
+		}
+	} else if len(collection.Items().Items()) > 0 {
```
hunterkepley, line 113: Intentionally nested if statement, so that the other branch does not run if the error is a feature toggle error.

gdbranco: nit: Do you need the else?

PR #2784, opened 2025-02-11
OCM-13960 | ci: Add a new paramter : OPENSHIFT_VERSION

yingzhanredhat:


tests/utils/handler/profile.go
```diff
@@ -108,6 +108,11 @@ func LoadProfileYamlFileByENV() *Profile {
 			config.Test.GlobalENV.Version)
 		profile.Version = config.Test.GlobalENV.Version
 	}
+	if config.Test.GlobalENV.OpenshiftVersion != "" {
```
xueli181114, line 111: Put this part with `Version` rather than have another duplicated code block.

PR #2791, opened 2025-02-13
OCM-13947 | feat: Enable deleting HCP autoscaler + test modification

hunterkepley:
Allows deletion of HCP autoscaler (tested manually)

Also modifies tests to remove validation check + add test case for HCP

### DIFF
--- a/cmd/dlt/autoscaler/cmd.go
+++ b/cmd/dlt/autoscaler/cmd.go
@@ -62,10 +62,6 @@ func DeleteAutoscalerRunner() rosa.CommandRunner {
 			return err
 		}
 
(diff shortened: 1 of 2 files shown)


cmd/dlt/autoscaler/cmd.go
```diff
@@ -62,10 +62,6 @@ func DeleteAutoscalerRunner() rosa.CommandRunner {
 			return err
 		}
 
-		if cluster.Hypershift().Enabled() {
```
marcolan018, line 65: for hcp autoscaler, we only support edit/get autoscaler. create/dlt are not supported

PR #2795, opened 2025-02-17
OCM-13947 | feat: Support editing HCP cluster autoscaler

hunterkepley:
Supports the following flags:

```
MaxNodesTotal
maxPodGracePeriod: With default of 600 seconds 
maxNodeProvisionTime: With default of 15 minutes
podPriorityThreshold: default is -10
```

Other flags used will result in an unsupported error from CS.

Tests in place to ensure validation works

### DIFF
--- a/pkg/ocm/clusterautoscaler.go
+++ b/pkg/ocm/clusterautoscaler.go
@@ -62,6 +62,15 @@ type ScaleDownConfig struct {
 	DelayAfterFailure    string
 }
 
+func BuildClusterAutoscalerHostedCp(config *AutoscalerConfig) *cmv1.ClusterAutoscalerBuilder {
+	return cmv1.NewClusterAutoscaler().
+		MaxPodGracePeriod(config.MaxPodGracePeriod).
+		PodPriorityThreshold(config.PodPriorityThreshold).
+		MaxNodeProvisionTime(config.MaxNodeProvisionTime).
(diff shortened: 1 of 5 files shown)


hunterkepley:
Example of flag validation error msg:

```❯ ./rosa edit autoscaler -c hk-hcp --log-verbosity 4
E: Unable to use flag 'log-verbosity' when editing a Hosted Control Plane cluster autoscaler.
Supported flags are: 'max-nodes-total', 'max-pod-grace-period', 'max-node-provision-time', 'pod-priority-threshold'```

File: pkg/ocm/clusterautoscaler.go [hunk 1]
```diff
@@ -134,8 +143,17 @@ func (c *Client) CreateClusterAutoscaler(clusterId string, config *AutoscalerCon
 	return response.Body(), nil
 }
 
-func (c *Client) UpdateClusterAutoscaler(clusterId string, config *AutoscalerConfig) (*cmv1.ClusterAutoscaler, error) {
-	object, err := BuildClusterAutoscaler(config).Build()
+func (c *Client) UpdateClusterAutoscaler(clusterId string, isHostedCp bool,
```
gdbranco, line 146: Is cluster id actually used here?

hunterkepley: @gdbranco Yup seems so, when it sends the API request it uses the cluster ID

PR #2798, opened 2025-02-18
OCM-14136 | feat: Add UX informational msgs when create/nodepool

hunterkepley:


pkg/machinepool/machinepool.go
```diff
@@ -948,6 +952,43 @@ func (m *machinePool) CreateNodePools(r *rosa.Runtime, cmd *cobra.Command, clust
 		}
 	}
 
+	sumOfReplicas := replicas
```
marcolan018, line 955: all these changes should only work for HCP, but not for classic.

hunterkepley: They are placed in the `CreateNodePools` and `EditNodePool` functions, which are ran only for HCP

PR #2805, opened 2025-02-20
OCM-14214 | fix: Calculate default NP values when create/edit np

hunterkepley:
Issue was that the default node pool values were not being calculated for the info/warning msgs that print when doing risky scaling choices

This was because the cluster API response does not contain default NPs, had to use a separate API endpoint

So if the default NP had 50 nodes, and the new one had 50, the "total" this code would return would be `50` instead of `100`

### DIFF
--- a/pkg/machinepool/machinepool.go
+++ b/pkg/machinepool/machinepool.go
@@ -957,18 +957,20 @@ func (m *machinePool) CreateNodePools(r *rosa.Runtime, cmd *cobra.Command, clust
 	sumOfMaxReplicas := maxReplicas
 	sumOfMinReplicas := minReplicas
 
-	for _, np := range cluster.NodePools().Items() {
+	nodepools, err := r.OCMClient.GetNodePools(cluster.ID())
+	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/machinepool/machinepool.go
```diff
@@ -957,18 +957,19 @@ func (m *machinePool) CreateNodePools(r *rosa.Runtime, cmd *cobra.Command, clust
...
 
-	for _, np := range cluster.NodePools().Items() {
+	nodepools, err := r.OCMClient.GetNodePools(cluster.ID())
+
+	for _, np := range nodepools {
 		// If autoscaling, calculate min and max, use min and max in separate messages below
-		autoscaling, ok := np.GetAutoscaling()
-		if !ok || autoscaling == nil {
+		npAutoscaling, ok := np.GetAutoscaling()
+		if ok {
```
marcolan018, line 965: confused here
should we get  min/maxReplicas when autoscaling enabled, and get fixed replicas if not?
in current code, you will try to get all of them when autoscaling is enabled, and do nothing if not.

PR #2811, opened 2025-03-03
OCM-14317 | ci: update registry config value for OCP E2E cases

yingzhanredhat:


tests/ci/data/profiles/rosa-hcp.yaml
```diff
@@ -190,7 +190,7 @@ profiles:
     autoscaler_enabled: false
     additional_sg_number: 0
     registries_config: true
-    allowed_registries: true
```
yuwang-RH, line 193: Why remove this config and replace it with blocked_registries: true ?

tests/utils/handler/cluster_handler.go
```diff
@@ -816,7 +816,7 @@ func (ch *clusterHandler) GenerateClusterCreateFlags() ([]string, error) {
 		)
 		if ch.profile.ClusterConfig.AllowedRegistries {
 			flags = append(flags,
-				"--registry-config-allowed-registries", "allowed.example.com,*.test",
+				"--registry-config-allowed-registries", "allowed.example.com,quay.io,*.redhat.com",
```
yuwang-RH, line 819: Could you add a condition by 'else if' for ch.profile.ClusterConfig.BlockedRegistries, rather consider ch.profile.ClusterConfig.BlockedRegistries= true if ch.profile.ClusterConfig.AllowedRegistries = false, as BlockedRegistries and AllowedRegistries are two separated config in profile

yingzhanredhat: Thanks.Update it

tests/ci/data/profiles/rosa-hcp.yaml
yuwang-RH: Make sense. sounds good to me

PR #2820, opened 2025-03-06
OCM-14417 | fix: Bug with upgrading cluster and missing account roles

hunterkepley:
An error occurs when deleting a support role, for example, then trying to upgrade a cluster. The error is a PANIC due to the AWS SDK not returning an error, and instead returning empty role lists + a nil role output. 

Checking for nil after the error check fixes the PANIC

### DIFF
--- a/pkg/aws/policies.go
+++ b/pkg/aws/policies.go
@@ -1428,6 +1428,10 @@ func (c *awsClient) GetAttachedPolicyWithTags(role *string,
 	if err != nil && !awserr.IsNoSuchEntityException(err) {
 		return policies, excludedPolicies, err
 	}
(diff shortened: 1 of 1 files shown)


pkg/aws/policies.go
```diff
@@ -1428,6 +1428,10 @@ func (c *awsClient) GetAttachedPolicyWithTags(role *string,
 	if err != nil && !awserr.IsNoSuchEntityException(err) {
 		return policies, excludedPolicies, err
 	}
+	if attachedPoliciesOutput != nil {
```
marcolan018, line 1431: if attachedPoliciesOutput != nil
or
if attachedPoliciesOutput == nil?

PR #2825, opened 2025-03-11
OCM-14555 | test: fix ids: 76394

jerichokeyne:
~70370,75504,75536 - Increasing timeouts~
76394 - Replace all spaces with commas, not just the first one (to build the correct list string)

### DIFF
--- a/tests/e2e/hcp_cluster_test.go
+++ b/tests/e2e/hcp_cluster_test.go
@@ -441,21 +441,21 @@ var _ = Describe("HCP cluster testing",
 					if v["Allowed Registries"] != nil {
 						allowedList := jsonData.DigString("registry_config", "registry_sources", "allowed_registries")
 						if len(allowedList) > 2 {
-							result := strings.Replace(allowedList, " ", ",", 1)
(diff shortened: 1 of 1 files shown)


yuwang-RH:
75504,75536(rosa-master-e2e-rosa-day1-supplemental-f3) It doesn't need to increase the wait time for the cluster deletion in the cleanup step. for example, 75504 was changed to Error status, the case cannot be passed even we increase the wait time to 1h. By the way, the supplemental job is definied 5h as the job timeout, if we increase wait time for too many step, it will make the job timeout if too many cases are using the long wait time. So my suggestion is, don't increase the wait time for cluster deletion, in our passed experience, the cluster should be finished deprovision within 30 min. And if you met the case failed as timeout, we need to try to reproduce it to check if it is a backend issue.
70370: Have you re-tested after increase the wait time? In the step of this failed checkpoint,it is checking if network verification result is sync to cluster spec and rosacli can show that in `rosa describe cluster`, we need to make if the verification result has been synchronized to cluster spec via API before increase the wait time. Also from the passed experience of running this case, it should be sync the status if there is no issue.

PR #2832, opened 2025-03-21
OCM 6929 | fix: Improved UX for error handling in private clusters

olucasfreitas:
# Details

Improved UX on error handling for privates clusters by adding some verifications to the subnets and better messaging to when a subnet is public and can't be used in this setting

# How to Test

## Teste Case 1

On the terminal, execute the command `./rosa create cluster --hosted-cp --mode auto SUBNET1,SUBNET2`, here you should put a private and public subnet that are located in the same region, go through the process, and after selecting the option for the cluster to be private, you should see a message saying that the command could not be executed, and the program will stop running.

## Teste Case 2

On the terminal, execute the command `./rosa create cluster --hosted-cp --mode auto`, go through the process, and after selecting the option for the cluster to be private, you should see a warning for public subnets that have been excluded from the list, and the command will NOT stop running.

# Ticketing

Closes [OCM-6929](https://issues.redhat.com/browse/OCM-6929)

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -22,6 +22,7 @@ import (
 	"net"
 	"os"
 	"reflect"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -2209,9 +2210,12 @@ func run(cmd *cobra.Command, _ []string) {
 			r.Reporter.Errorf("%s", filterError)
 			os.Exit(1)
 		}
+
+		var excludedPublicSubnets []string
 		if privateLink {
-			subnets = filterPrivateSubnets(subnets, r)
+			subnets, excludedPublicSubnets = filterPrivateSubnets(subnets, r)
 		}
+
 		if len(subnets) == 0 {
 			r.Reporter.Warnf("No subnets found in current region that are valid for the chosen CIDR ranges")
 			if isHostedCP {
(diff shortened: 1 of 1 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -2234,13 +2238,17 @@ func run(cmd *cobra.Command, _ []string) {
...
 		if subnetsProvided {
 			for _, subnetArg := range subnetIDs {
-				verifiedSubnet := false
-				for _, subnet := range subnets {
-					if awssdk.ToString(subnet.SubnetId) == subnetArg {
-						verifiedSubnet = true
-					}
+				// Check if subnet is in the excluded list of public subnets
+				if slices.Contains(excludedSubnetIdsDueToBeingPublic, subnetArg) {
+					r.Reporter.Errorf("The command cannot be executed because %s is public and the cluster is set as private",
```
hunterkepley, line 2243: I would change this error message slightly, something like:
`Cluster is set as private, cannot use public subnet '%s'`

cmd/create/cluster/cmd.go
```diff
@@ -2209,9 +2210,12 @@ func run(cmd *cobra.Command, _ []string) {
 			r.Reporter.Errorf("%s", filterError)
 			os.Exit(1)
 		}
+
+		var excludedSubnetIdsDueToBeingPublic []string
```
hunterkepley, line 2214: NP: Would maybe rename this var to `excludedPublicSubnets` to make the name more concise

hunterkepley: I see that the function you edited below has this name, I think it would be cleaner to name it this regardless, but it's a nitpick so change at your discretion

hunterkepley (commented):
Looks good overall 😄 left a couple comments for wording

cmd/create/cluster/cmd.go
olucasfreitas: changed it here [ed05f9e](https://github.com/openshift/rosa/pull/2832/commits/ed05f9ea971ac0d516c5f142dad1cb49cd25d7ed)

cmd/create/cluster/cmd.go
olucasfreitas: changed it here [ed05f9e](https://github.com/openshift/rosa/pull/2832/commits/ed05f9ea971ac0d516c5f142dad1cb49cd25d7ed)

PR #2857, opened 2025-04-09
OCM-15097 | feat: Un-hide `channel-group` flag in create/cluster

hunterkepley:
Unhides the `channel-group` flag in `create/cluster` - related to EUS region support ([OCM-15097](https://issues.redhat.com//browse/OCM-15097))


olucasfreitas (approved):

vkareh:
I know this is now merged, but can we hold off on tagging a release with this until XCMSTRAT-1119 is out of refinement? There's a possibility that the final refined feature may not actually use the channel group flag.

PR #2866, opened 2025-04-11
OCM-15190 | feat: Remove default channelgroup from edit/cluster

hunterkepley:
Removes the default `"stable"` channel group value from `edit/cluster`


hunterkepley:
yes I can bot

PR #2867, opened 2025-04-17
OCM-4667 | feat: added new warnings and validations to account-role

olucasfreitas:
# Details

This PR improves the UX when handling account roles and introduces a new check to verify whether a role is **managed by Red Hat** or not and some warnings.

---

# How to Test

Before running the test cases, create an account role with version `4.13` using the following command:

```bash
./rosa create account-roles --mode auto --version 4.13
```

---

##  Test Case 1 – Selecting a Different OpenShift Version

1. Run the command:

    ```bash
    ./rosa create cluster --hosted-cp --mode auto
    ```

2. During the interactive flow:
   - Select an account role tied to version **4.13**
   - Then choose a **different OpenShift version**, e.g., `4.14`

3. You should see the following **warnings**, and the command **will NOT stop running**:

    ```
    W: Account roles not created by ROSA CLI cannot be listed, updated, or upgraded. Use ROSA CLI to create them before proceeding.
    W: No suitable accounts with ROSA CLI-created account roles were found. You can manually set them in the next steps or run 'rosa create account-roles' to create them first.
    ```

---

## Test Case 2 – Selecting the Same OpenShift Version

1. Run the command:

    ```bash
    ./rosa create cluster --hosted-cp --mode auto
    ```

2. During the interactive flow:
   - Select an account role tied to version **4.13**
   - Then choose the **same OpenShift version**, `4.13`

3. You should see the following **warning**, and the command **will NOT stop running**:

    ```
    W: Account roles not created by ROSA CLI cannot be listed, updated, or upgraded. Use ROSA CLI to create them before proceeding.
    ```

---

## Additional Info

If the selected role is **not managed by Red Hat**, the following warning should appear:

```
W: The role 'ROLE_NAME' is not a Red Hat managed role
```

---

# Ticket

Closes [[OCM-4667](https://issues.redhat.com/browse/OCM-4667)](https://issues.redhat.com/browse/OCM-4667)

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -1448,6 +1448,8 @@ func run(cmd *cobra.Command, _ []string) {
 			os.Exit(1)
 		}
 
+		r.Reporter.Warnf("Account roles not created by ROSA CLI cannot be listed, updated, or upgraded.")
+
 		if len(roleARNs) > 1 {
 			defaultRoleARN := roleARNs[0]
 			// Prioritize roles with the default prefix
@@ -1497,13 +1499,18 @@ func run(cmd *cobra.Command, _ []string) {
 			if isHostedCP {
 				createAccountRolesCommand = createAccountRolesCommand + " " + hostedCPFlag
 			}
-			r.Reporter.Warnf(fmt.Sprintf("No compatible account roles with version '%s' found. "+
-				"You will need to manually set them in the next steps or run '%s' to create them first.",
-				minor, createAccountRolesCommand))
+			r.Reporter.Warnf("No suitable account with ROSA CLI-created account roles were found. "+
+				"You can manually set them in the next steps or run '%s' to create them first.", createAccountRolesCommand)
 			interactive.Enable()
 		}
 
 		if roleARN != "" {
+			// Check if role has red-hat-managed tag
+			hasTag := roles.CheckHasRedHatManagedTag(roleARN, awsClient)
+			if !hasTag {
+				r.Reporter.Warnf("The role '%s' is not a Red Hat managed role", roleARN)
+			}
+
 			// check if role has hosted cp policy via AWS tag value
 			hostedCPPolicies, err := awsClient.HasHostedCPPolicies(roleARN)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -1497,13 +1500,15 @@ func run(cmd *cobra.Command, _ []string) {
 			if isHostedCP {
 				createAccountRolesCommand = createAccountRolesCommand + " " + hostedCPFlag
 			}
-			r.Reporter.Warnf(fmt.Sprintf("No compatible account roles with version '%s' found. "+
-				"You will need to manually set them in the next steps or run '%s' to create them first.",
-				minor, createAccountRolesCommand))
+			r.Reporter.Warnf("No suitable accounts with ROSA CLI-created account roles were found. "+
+				"You can manually set them in the next steps or run '%s' to create them first.", createAccountRolesCommand)
```
hunterkepley, line 1504: I believe it is only looking in a single account here, rather than multiple, please LMK if that assumption is incorrect

But if one account I think we can remove `accounts with` here

pkg/helper/roles/helpers.go
```diff
@@ -43,6 +43,36 @@ func GetOperatorRoleName(cluster *cmv1.Cluster, missingOperator *cmv1.STSOperato
...
+			if *tag.Key == tags.RedHatManaged {
+				hasRedHatManagedTag = true
+				break tagSearch
+			}
+		}
+	}
+
+	if !hasRedHatManagedTag {
+		r.Reporter.Warnf("The role '%s' is not a Red Hat managed role", roleName)
+	}
```
hunterkepley, line 73: We don't need to check each policy- the role itself will suffice here. The role will have the `red-hat-managed` tag

pkg/helper/roles/helpers.go
```diff
@@ -43,6 +43,36 @@ func GetOperatorRoleName(cluster *cmv1.Cluster, missingOperator *cmv1.STSOperato
 	return awsCommonUtils.TruncateRoleName(role)
 }
 
+func CheckRedHatManagedTag(arn string, awsClient aws.Client, r *rosa.Runtime) {
```
hunterkepley, line 46: NP: `CheckRedHatManagedTag` -> `CheckHasRedHatManagedTag`

cmd/create/cluster/cmd.go
```diff
@@ -1497,13 +1500,15 @@ func run(cmd *cobra.Command, _ []string) {
...
 			interactive.Enable()
 		}
 
 		if roleARN != "" {
+			// Check if role has red-hat-managed tag
+			roles.CheckRedHatManagedTag(roleARN, awsClient, r)
+
```
hunterkepley, line 1511: I think, for this function, we should return a bool, and add the logic here. It's a bit vague at what this does, and what is going on until you look into this function itself (I left a couple small comments on the naming + functionality of it)

pkg/helper/roles/helpers.go
hunterkepley: Also NP about convention; we usually have the rosa runtime first

pkg/helper/roles/helpers.go
```diff
@@ -43,6 +43,36 @@ func GetOperatorRoleName(cluster *cmv1.Cluster, missingOperator *cmv1.STSOperato
...
+		return
+	}
+
+	attachedPolicies, err := awsClient.GetPolicyDetailsFromRole(&roleName)
+	if err != nil {
+		r.Reporter.Errorf("Failed to get policy details for role '%s': %v", roleName, err)
+		return
+	}
+
+	hasRedHatManagedTag := false
```
hunterkepley, line 59: We don't need to get the policies in general I don't believe, related to the other comment about looping through each policy

The role itself will have the tag

cmd/create/cluster/cmd.go
olucasfreitas: should be fixed here [6a52f12](https://github.com/openshift/rosa/pull/2867/commits/6a52f123604cf00d48743681600fb9af9538cf64)

cmd/create/cluster/cmd.go
olucasfreitas: should be fixed here [6a52f12](https://github.com/openshift/rosa/pull/2867/commits/6a52f123604cf00d48743681600fb9af9538cf64)

hunterkepley (approved):

PR #2879, opened 2025-05-05
OCM-15600 | feat: Clean up UX + remove billing account checks for HCP govcloud

hunterkepley:
Removes billing account ID when using HCP on fedramp, as well as forces the FIPS check in the same situation. Also added some UX in the form of warning the user they 1) don't need the billing account ID flag, and 2) informs them where it comes from

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -96,6 +96,8 @@ const (
 	vpcEndpointRoleArnFlag                   = "vpc-endpoint-role-arn"
 	hcpInternalCommunicationHostedZoneIdFlag = "hcp-internal-communication-hosted-zone-id"
 	ingressPrivateHostedZoneIdFlag           = "ingress-private-hosted-zone-id"
+
+	billingAccountFlag = "billing-account"
 )
 
 var args struct {
@@ -459,8 +461,7 @@ func initFlags(cmd *cobra.Command) {
 		&args.fips,
 		"fips",
 		false,
-		"Create cluster that uses FIPS Validated / Modules in Process cryptographic libraries. "+
-			"This is currently only available without the use of the --hosted-cp flag.",
+		"Create cluster that uses FIPS Validated / Modules in Process cryptographic libraries.",
 	)
 
 	flags.StringVar(
@@ -750,7 +751,7 @@ func initFlags(cmd *cobra.Command) {
 
 	flags.StringVar(
 		&args.billingAccount,
-		"billing-account",
+		billingAccountFlag,
 		"",
 		"Account ID used for billing subscriptions purchased through the AWS console for ROSA",
 	)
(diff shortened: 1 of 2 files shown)


AlexVulaj (approved):

cmd/create/cluster/validation.go
```diff
@@ -17,6 +17,8 @@ const (
 
 	isNotGovcloudFeature = "Hosted Control Plane shared VPC clusters are not supported on Govcloud regions; %s"
 	pleaseRemoveFlags    = "Please remove the following flags: %s"
+
+	billingAccountsHcpErrorMsg = "Billing accounts are only supported for commercial Hosted Control Plane clusters"
```
davidleerh, line 21: Would it be better to use "non-Govcloud" instead?
Not sure if customers would understand the difference as the term "commercial" is used internally for distinguishing the env.  :thinking:

cmd/create/cluster/cmd.go
```diff
@@ -1255,10 +1257,20 @@ func run(cmd *cobra.Command, _ []string) {
...
-		r.Reporter.Errorf("Billing accounts are only supported for Hosted Control Plane clusters")
+		r.Reporter.Errorf(billingAccountsHcpErrorMsg)
 		os.Exit(1)
 	}
 
+	if isHostedCP && fedramp.Enabled() && billingAccount != "" {
+		if cmd.Flags().Changed(billingAccountFlag) {
+			r.Reporter.Errorf(billingAccountsHcpErrorMsg)
+		} else {
+			r.Reporter.Warnf("Billing accounts when using Govcloud are associated with commercial accounts, " +
```
davidleerh, line 1268: same here :thinking:

cmd/create/cluster/cmd.go
```diff
@@ -2821,12 +2833,8 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
-	if cmd.Flags().Changed("fips") && isHostedCP {
-		r.Reporter.Errorf("FIPS support not available for Hosted Control Plane clusters")
-		os.Exit(1)
-	}
 	fips := args.fips || fedramp.Enabled()
-	if interactive.Enabled() && !fedramp.Enabled() && !isHostedCP {
+	if interactive.Enabled() {
```
davidleerh, line 2837: if args.fips is set false and fedramp.Enabled is true, fips value would be **true**. (GOOD)
if args.fips is set true and fedramp.Enabled is true, fips value would be **true**. (GOOD)
if args.fips is set true and fedramp.Enabled is false, fips value would be **true**. (GOOD)
if args.fips is set false and fedramp.Enabled is false, fips value would be **false**. (GOOD)
Can we output an informative message when fedramp.Enabled?
Indicating that fips will always be set to true.

davidleerh: Also if we force fips to be enabled for fr hcp, shouldn't we also have this interactive prompt not appear?

davidleerh (approved):
lgtm

davidleerh:
lgtm

PR #2880, opened 2025-05-07
OCM-15158 | feat: Removed duplicated-from-API validation (machinepools)

hunterkepley:
In the API, we have a preflight which covers this validation case: https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/cmd/clusters-service/service/aws_preflight/preflight_service.go#L442

It has been asked to remove from the CLI, so errors are consistent.

2 tests have also been removed, since they relate to the functionality which this validation provided

### DIFF
--- a/pkg/ocm/machines.go
+++ b/pkg/ocm/machines.go
@@ -19,7 +19,6 @@ package ocm
 import (
 	"errors"
 	"fmt"
-	"strings"
 
 	amsv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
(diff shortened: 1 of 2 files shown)


pkg/ocm/machines.go
```diff
@@ -262,12 +260,7 @@ func (mtl *MachineTypeList) ValidateMachineType(machineType string, multiAZ bool
 	v := mtl.Find(machineType)
 
 	if v == nil {
-		allMachineTypes := strings.Join(mtl.IDs(), " ")
-		allAvailabilityZones := strings.Join(mtl.AvailabilityZones, ",")
-		err := fmt.Errorf("Machine type '%s' not found in availability zones '%s' for region: '%s'\n"+
-			"Available machine type list: %s",
-			machineType, allAvailabilityZones, mtl.Region, allMachineTypes)
-		return err
+		return nil // Replaced not-found error with a preflight in CS (validateZoneSupportInstanceType)
```
hunterkepley, line 263: Note: Reasoning for having this is that the quota check after will cause nil ptrs if it is nil. There's no reason to run the validation if it cannot be found immediately by the CLI

davidleerh (approved):

PR #2904, opened 2025-05-27
OCM-15990 | feat: New CF routines in awsClient for RosaNetwork in CAPA

mzazrivec:
This pull request implements several things needed so that rosa-cli golang packages (the `awsClient` in particular) can be (re)used in [CAPA in rosaNetwork implementation](https://github.com/stolostron/cluster-api-installer/pull/178/files): 
* a new `awsclient.CreateStackWithParamsTags()` non-blocking routine through which it will be possible to create a new CF stack with parameters & tags
* a new `awsclient.GetCFStack()` routine to fetch cloudformation stack details
* a new `awsclient.DescribeCFStackResources()` routine to fetch list of CF stack resources
* a new `awsclient.DeleteCFStack()` routine to delete a cloudformation stack
* support for `go-logr/logr` `Logger` in `awsClient`, since that logger is the one being used in CAPA

https://issues.redhat.com/browse/OCM-15990

### DIFF
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -33,6 +33,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/service/cloudformation"
+	cftypes "github.com/aws/aws-sdk-go-v2/service/cloudformation/types"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/aws/aws-sdk-go-v2/service/iam"
@@ -46,6 +47,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/sts"
 	"github.com/aws/smithy-go/middleware"
 	smithyhttp "github.com/aws/smithy-go/transport/http"
+	"github.com/go-logr/logr"
 	awserr "github.com/openshift-online/ocm-common/pkg/aws/errors"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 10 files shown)


pkg/aws/client.go
```diff
@@ -232,14 +238,16 @@ type AccessKeyGetter interface {
 // ClientBuilder contains the information and logic needed to build a new AWS client.
 type ClientBuilder struct {
 	logger              *logrus.Logger
+	logrLogger          *logr.Logger
```
hunterkepley, line 241: do we need this extra logger? Is there a way we could use one or the other?

pkg/aws/policies.go
```diff
@@ -1151,7 +1151,7 @@ func (c *awsClient) DeleteOperatorRole(roleName string, managedPolicies bool,
 					sharedVpcPoliciesNotDeleted[*policyOutput.Policy.Arn] = false
 
 					// Add to list of sharedVpc policies to be actually deleted
-					c.logger.Infof("Deleting policy '%s'", policy)
+					c.logger.Info(fmt.Sprintf("Deleting policy '%s'", policy))
```
hunterkepley, line 1154: Just going along with the earlier question- so far I am unsure I've seen something which requires `logr` logger over the normal `logrus` one we use in the CLI- but I think I am missing something 🙂

hunterkepley (commented):
Im guessing the logger was for compatibility with CAPA - I have a couple of ideas on how to introduce it a bit cleaner, but overall this MR looks good to me. I'll wait for your reply about the logger, so I get the full context, before suggesting anything around it

pkg/aws/client.go
mzazrivec: `logr.Logger` is what CAPA controllers use. When building the aws client, we also need to initialize the logger. And convert CAPA to `logrus.Logger` is just not realistic. So that's why the logger wrapper.

hunterkepley: Understood, maybe we can specify through the name `capaLogger` or something more identifiable, since the AWS client is used in every part of the CLI and new developers may get confused on which one to use 🙂

hunterkepley:
LGTM besides the request to change the name of the new logger

PR #2906, opened 2025-05-28
OCM-1599 | feat: added new flag based columns to list machine pool command

olucasfreitas:
## Details
This PR adds new columns to the `rosa list machinepools` command through the use of 4 new flags to show additional information, and also removes any columns from the final output that have no data to show. The new flags are:
- `--all`
- `--az-type` 
- `--dedicated-host`
- `--win-li`

---

## Prerequisites
Before running the test cases, ensure you have A ROSA cluster with existing machinepools.

---

## Test Case 1 – Default Command (No Flags)
**Command:**
```bash
./rosa list machinepools -c CLUSTER_NAME
```

**Expected Output:**
image
- Should display basic machinepool information with standard columns:
  - ID
  - AUTOSCALING
  - REPLICAS
  - INSTANCE TYPE
  - AVAILABILITY ZONES
  - SPOT INSTANCES
  - DISK SIZE

---

## Test Case 2 – AZ Type Flag
**Command:**
```bash
./rosa list machinepools -c CLUSTER_NAME --az-type
```

**Expected Output:**
image
- Should include all default columns plus:
  - AZ TYPE column showing zone classification based on availability zones

---

## Test Case 3 – Dedicated Host Flag
**Command:**
```bash
./rosa list machinepools -c CLUSTER_NAME --dedicated-host
```

**Expected Output:**
image
- Should include all default columns plus:
  - DEDICATED HOST column showing "false" for non-dedicated instances

---

## Test Case 4 – Windows License Flag
**Command:**
```bash
./rosa list machinepools -c CLUSTER_NAME --win-li
```

**Expected Output:**
image
- Should include all default columns plus:
  - WIN-LI ENABLED column showing "false" for Linux-based machinepools

---

## Test Case 5 – All flag
**Command:**
```bash
./rosa list machinepools -c CLUSTER_NAME --all
```

**Expected Output:**
image
- Should display all available columns including:
  - All standard columns
  - SUBNETS
  - SPOT INSTANCES  
  - DISK SIZE
  - SG IDS
  - AZ TYPE
  - WIN-LI ENABLED
  - DEDICATED HOST

---

## Test Case 6 – Individual Flag Combination
**Command:**
```bash
./rosa list machinepools -c CLUSTER_NAME --az-type --dedicated-host --win-li
```

**Expected Output:**
image
- Should display standard columns plus the three specified additional columns

---

## Expected Behavior Notes
1. **Empty Column Removal**: Columns with no data should be automatically excluded from the output
3. **AZ Type Classification**: Zone types are determined by availability zone naming patterns:
   - **Standard**: Regular AWS zones (e.g., us-west-2a, us-west-2b, us-west-2c)
   - **Wavelength**: Zones containing "wl" in the name
   - **Outpost**: Zones containing "outpost" in the name
   - **LocalZone**: Zones containing "-lz" or having more than 3 dashes in the name
   - **N/A**: When no availability zones are configured

---

## Ticket
Closes [OCM-1599](https://issues.redhat.com/browse/OCM-1599)


pkg/machinepool/machinepool.go
```diff
@@ -49,7 +49,7 @@ var machinePoolKeyRE = regexp.MustCompile(`^[a-z]([-a-z0-9]*[a-z0-9])?$`)
 //go:generate mockgen -source=machinepool.go -package=machinepool -destination=machinepool_mock.go
 type MachinePoolService interface {
 	DescribeMachinePool(r *rosa.Runtime, cluster *cmv1.Cluster, clusterKey string, machinePoolId string) error
-	ListMachinePools(r *rosa.Runtime, clusterKey string, cluster *cmv1.Cluster) error
+	ListMachinePools(r *rosa.Runtime, clusterKey string, cluster *cmv1.Cluster, showAZType bool, showDedicated bool, showWindowsLI bool, showAll bool) error
```
hunterkepley: I think it may be bad practice to include all these bools at the end of the function, and can cause confusion for those working on this function later, as well as when writing/reading tests

I think a good solution here would be to define a structure with these bools, named verbosely so it’s easy to work with (called something like `ListMachinePoolArgs` to be passed into this function so users are forced to include only the information they want to include). This is a clean code practice we are doing as a whole in OCM

pkg/machinepool/machinepool_test.go
```diff
@@ -340,32 +340,265 @@ var _ = Describe("Machinepool and nodepool", func() {
...
+			Expect(out).To(ContainSubstring("TAINTS"))
+			Expect(out).To(ContainSubstring("SUBNETS"))
+			Expect(out).To(ContainSubstring("SPOT INSTANCES"))
+			Expect(out).To(ContainSubstring("DISK SIZE"))
+
+			// Check that the output contains the machine pool data
+			Expect(out).To(ContainSubstring("mp-1"))
+			Expect(out).To(ContainSubstring("m5.large"))
+			Expect(out).To(ContainSubstring("subnet-1, subnet-2"))
+			Expect(out).To(ContainSubstring("test-key=test-value:"))
```
hunterkepley: Could you explain the reasoning behind getting rid of the more concise `Equal()` check for the output and replacing it entirely?

pkg/machinepool/helper.go
```diff
@@ -390,3 +391,52 @@ func (r *ReplicaSizeValidation) MaxReplicaValidatorOnClusterCreate() interactive
...
+		if strings.Contains(zoneLower, "outpost") {
+			return "Outpost"
+		}
+		if strings.Contains(zoneLower, "-lz") || strings.Count(zoneLower, "-") > 3 {
+			return "LocalZone"
+		}
+	}
+
+	return "Standard"
+}
```
hunterkepley: I think it would be best to have all the named string types such as `”LocalZone”` become consts, that way they can be used anywhere and cannot have typos

cmd/list/machinepool/cmd.go
```diff
@@ -34,11 +34,26 @@ const (
...
+  rosa list machinepools --cluster=mycluster
+  
+  # List machine pools with availability zone type information
+  rosa list machinepools --cluster=mycluster --az-type
+  
+  # List machine pools with dedicated host and Windows LI information
+  rosa list machinepools --cluster=mycluster --dedicated-host --win-li
+  
+  # List machine pools showing all additional information
+  rosa list machinepools --cluster=mycluster --all`
```
hunterkepley: we dont need to have examples of every flag, `—help` accomplishes this for us. Maybe put `—all` in on the original example, that should be enough

cmd/list/machinepool/cmd.go
```diff
@@ -52,6 +67,36 @@ func NewListMachinePoolCommand() *cobra.Command {
...
+		false,
+		"Show whether the machine pool is Windows LI enabled",
+	)
+
+	flags.BoolVar(
+		&args.showAll,
+		"all",
+		false,
+		"Show all additional information (equivalent to --az-type --dedicated-host --win-li)",
+	)
```
hunterkepley: these descriptions are a bit conflicting,

First one says `of each machinepool`, second one says `the machinepool` and third says `the machinepool`

I think `each machinepool` would be best for all since it’s part of the `list` command and not the `describe` command

pkg/machinepool/helper.go
```diff
@@ -390,3 +391,52 @@ func (r *ReplicaSizeValidation) MaxReplicaValidatorOnClusterCreate() interactive
...
+	aws := r.AWSClient
+	awsMachinePoolID := machinePool.AWS().ID()
+	if awsMachinePoolID == "" {
+		return "false"
+	}
+
+	hasDedicatedHost, err := aws.CheckIfMachinePoolHasDedicatedHost([]string{awsMachinePoolID})
+	if err != nil {
+		_ = r.Reporter.Errorf("Failed to check dedicated host status: %v", err)
+		return "unknown"
```
hunterkepley: Also would be good to have the normal types such as `unknown`, `true`, `false` be consts as well. I believe `true` and `false` already exist elsewhere as consts in rosa

pkg/machinepool/machinepool.go
```diff
@@ -1228,25 +1228,81 @@ func appendUpgradesIfExist(scheduledUpgrade *cmv1.NodePoolUpgradePolicy, output
...
-			output.PrintStringSlice(machinePool.AvailabilityZones()),
-			output.PrintStringSlice(machinePool.Subnets()),
-			ocmOutput.PrintMachinePoolSpot(machinePool),
-			ocmOutput.PrintMachinePoolDiskSize(machinePool),
-			output.PrintStringSlice(machinePool.AWS().AdditionalSecurityGroupIds()),
-		)
+func getMachinePoolsString(
+	r *rosa.Runtime,
+	machinePools []*cmv1.MachinePool,
+	showAZType, showDedicated, showWindowsLI, showAll bool,
```
hunterkepley: would be best to also use the args struct i mentioned in an earlier comment rather than 5 bools

pkg/machinepool/machinepool.go
```diff
@@ -1228,25 +1228,81 @@ func appendUpgradesIfExist(scheduledUpgrade *cmv1.NodePoolUpgradePolicy, output
...
+		hasValidData := false
+
+		// Extract values for each machine pool and check if there's valid data
+		for i, machinePool := range machinePools {
+			value := col.extract(machinePool)
+			columnValues[i] = value
+			if value != "" && value != "-" {
+				hasValidData = true
+			}
+		}
```
hunterkepley: There are no checks for the machine pool being `nil` here, this could lead to panics

pkg/machinepool/machinepool.go
```diff
@@ -1228,25 +1228,81 @@ func appendUpgradesIfExist(scheduledUpgrade *cmv1.NodePoolUpgradePolicy, output
...
 	}
-	return outputString
+
+	var outputBuilder strings.Builder
+	outputBuilder.WriteString(strings.Join(columnHeaders, "\t"))
+	outputBuilder.WriteString("\n")
+
+	// Add the data rows to the output
+	for rowIndex := range numMachinePools {
+		for colIndex, columnValues := range tableData {
```
hunterkepley: NP: for loop vars named differently for some reason (`col` vs `column`)

pkg/machinepool/machinepool.go
```diff
@@ -1228,25 +1228,81 @@ func appendUpgradesIfExist(scheduledUpgrade *cmv1.NodePoolUpgradePolicy, output
...
-		)
+func getMachinePoolsString(
+	r *rosa.Runtime,
+	machinePools []*cmv1.MachinePool,
+	showAZType, showDedicated, showWindowsLI, showAll bool,
+) string {
+	columnDefinitions := []struct {
+		header  string
+		extract func(*cmv1.MachinePool) string
+		enabled bool
```
hunterkepley: These feel kinda generic for names. I don't think `header` nor `generic` nor `enabled` are proper names for what this is accomplishing, maybe use ones more like what you named these values elsewhere/a bit more verbose in general;

`columnHeader`, `columnValue`, `visible`

pkg/machinepool/machinepool_test.go
```diff
@@ -340,32 +340,265 @@ var _ = Describe("Machinepool and nodepool", func() {
...
+				MachinePools(cmv1.NewMachinePoolList())
+			cluster, err := clusterBuilder.Build()
+			Expect(err).ToNot(HaveOccurred())
+
+			r := &rosa.Runtime{}
+			out := getMachinePoolsString(r, cluster.MachinePools().Slice(), false, false, false, false)
+			Expect(err).ToNot(HaveOccurred())
+
+			// When there are no machine pools, the function returns a single newline
+			Expect(out).To(Equal("\n"))
```
hunterkepley: The output should not return nothing at all, but rather just the column headers. The end user may get confused when nothing happens when they run this command

pkg/machinepool/helper.go
olucasfreitas: Do you want me to store it in this file, or is there a const file with all of them ?

pkg/machinepool/machinepool_test.go
olucasfreitas: When I was researching testing on go, one of the explanations that I came across was that this could be useful when the output format can change and that maybe Equal() would be too strict and prone to breakage with dynamic output

pkg/machinepool/helper.go
hunterkepley: you can check that directory, but generally a `helper` file is used for consts used with the package

pkg/machinepool/machinepool_test.go
hunterkepley: id say that makes sense, in this case we have a strict output we want to follow. This includes specific tabs and spacing, which benefits from exactness. The output may be dynamic in general in prod, but what we are testing should never be dynamic, it should always remain the same since we create the data to test

olucasfreitas: But isn't it more to the dynamic side when I have columns that can or cannot be shown due to being empty  ?

hunterkepley: I do understand what you mean though by the columns which can be hidden;

L344 - L348 show the exact data which will show up in this test case, which will result in the same output every time in this test case (Test printMachinePools with basic functionality)

If this were used as a `AfterEach` check rather than a specific-to-this-test-case, I would say to go with rewriting to a simpler solution

hunterkepley: does this make sense?

olucasfreitas: It does, will do the rewrite here, thanks for explaining

hunterkepley:
Make sure to rebase at some point, it will get rid of most if not all the vendor changes

pkg/aws/client.go
```diff
@@ -302,6 +303,34 @@ func New(
...
+	if err != nil {
+		return false, fmt.Errorf("failed to describe instances: %w", err)
+	}
+
+	for _, reservation := range result.Reservations {
+		for _, instance := range reservation.Instances {
+			if instance.Placement != nil {
+				if instance.Placement.Tenancy == ec2types.TenancyHost {
+					return true, nil
+				}
```
hunterkepley: NP: can combine these into one if statement

pkg/aws/client_mock.go
```diff
@@ -18,9 +18,9 @@ import (
 	types0 "github.com/aws/aws-sdk-go-v2/service/iam/types"
 	servicequotas "github.com/aws/aws-sdk-go-v2/service/servicequotas"
 	sts "github.com/aws/aws-sdk-go-v2/service/sts"
+	gomock "go.uber.org/mock/gomock"
```
hunterkepley: run `make fmt-imports` to make sure imports that were moved like this one weren't incorrectly moved

pkg/machinepool/helper_test.go
```diff
@@ -614,3 +614,272 @@ var _ = Describe("ValidateClusterVersionWithMaxNodesLimit Function", func() {
...
+	})
+
+	It("should return 'Wavelength' when zone contains 'wl'", func() {
+		machinePool, err := cmv1.NewMachinePool().
+			AvailabilityZones("us-east-1-wl-1").
+			Build()
+		Expect(err).ToNot(HaveOccurred())
+
+		zoneType := GetZoneType(machinePool)
+		Expect(zoneType).To(Equal("Wavelength"))
```
hunterkepley: NP: Would be good to replace the hardcoded string values in the tests with the consts which were made

cmd/list/machinepool/cmd_test.go
```diff
@@ -137,6 +132,213 @@ var _ = Describe("List machine pool", func() {
...
 		})
+		Context("Flag behavior tests", func() {
+			It("Shows all columns when --all flag is used", func() {
+				t.ApiServer.AppendHandlers(RespondWithJSON(http.StatusOK, classicClusterReady))
+				t.ApiServer.AppendHandlers(RespondWithJSON(http.StatusOK, mpResponse))
+				runner := ListMachinePoolRunner()
+				err := t.StdOutReader.Record()
+				Expect(err).ToNot(HaveOccurred())
+				cmd := NewListMachinePoolCommand()
+				err = cmd.Flag("cluster").Value.Set(clusterId)
```
hunterkepley, line 143: Instead of having several blocks of copied `Expect`s I think it would be best to make a function containing the `Expects` you use in this test file. Then passing in the vars which can change. Just to reduce line count and increase readability

pkg/aws/client_mock.go
olucasfreitas: I ran the make generate and the go mod tidy, it asked me to run the go mod vendor ?

I'm confused to in what order run all the commands

hunterkepley: `go mod tidy`
`go mod vendor`
`make fmt-imports`

cmd/list/machinepool/cmd.go
```diff
@@ -34,11 +34,15 @@ const (
 	short   = "List cluster machine pools"
 	long    = "List machine pools configured on a cluster."
 	example = `  # List all machine pools on a cluster named "mycluster"
-  rosa list machinepools --cluster=mycluster`
+  rosa list machinepools --cluster=mycluster
+  
+  # List machine pools showing all additional information
```
hunterkepley: NP: `all additional` -> `all`

pkg/aws/client.go
hunterkepley: Is there a reason this is still a nested if statement?

pkg/machinepool/helper_test.go
hunterkepley: these are still hardcoded strings, were these meant to be resolved? You have the consts in the same package

pkg/machinepool/helper.go
```diff
@@ -390,3 +405,54 @@ func (r *ReplicaSizeValidation) MaxReplicaValidatorOnClusterCreate() interactive
...
+	awsMachinePoolID := machinePool.AWS().ID()
+	if awsMachinePoolID == "" {
+		return BoolStringFalse
+	}
+
+	hasDedicatedHost, err := aws.CheckIfMachinePoolHasDedicatedHost([]string{awsMachinePoolID})
+	if err != nil {
+		if r.Reporter != nil {
+			_ = r.Reporter.Errorf("Failed to check dedicated host status: %v", err)
+		}
```
hunterkepley: The reporter is expected to be not-nil in every command, is there a reason it could be nil here? (Is this function called somewhere where the reporter is not passed in?) I think we can remove this check since it's implicit it will exist 🙂

pkg/machinepool/helper.go
```diff
@@ -390,3 +405,54 @@ func (r *ReplicaSizeValidation) MaxReplicaValidatorOnClusterCreate() interactive
...
+		if strings.Contains(zoneLower, "-lz") || strings.Count(zoneLower, "-") > 3 {
+			return ZoneTypeLocalZone
+		}
+	}
+
+	return ZoneTypeStandard
+}
+
+func IsWinLIEnabled(labels map[string]string) string {
```
hunterkepley: These funcs are only used in the `machinepool` package, by the looks of it we could make these private to help reduce clutter in the `machine pool` pkg

pkg/machinepool/helper.go
```diff
@@ -390,3 +405,54 @@ func (r *ReplicaSizeValidation) MaxReplicaValidatorOnClusterCreate() interactive
...
+	if val, ok := labels[LabelImageType]; ok && strings.ToLower(val) == ImageTypeWindows {
+		return BoolStringTrue
+	}
+	return BoolStringFalse
+}
+
+func IsDedicatedHost(machinePool *cmv1.MachinePool, r *rosa.Runtime) string {
```
hunterkepley: this one too

pkg/machinepool/helper.go
```diff
@@ -390,3 +405,54 @@ func (r *ReplicaSizeValidation) MaxReplicaValidatorOnClusterCreate() interactive
 		)
 	}
 }
+
+func GetZoneType(machinePool *cmv1.MachinePool) string {
```
hunterkepley: This one too

pkg/machinepool/helper.go
```diff
@@ -23,6 +24,20 @@ import (
...
+	ZoneTypeStandard   = "Standard"
+	ZoneTypeNA         = "N/A"
+
+	BoolStringTrue    = "true"
+	BoolStringFalse   = "false"
+	BoolStringUnknown = "unknown"
+
+	ImageTypeWindows = "windows"
+
+	LabelImageType = "image_type"
```
hunterkepley: Same with these consts, looks like they can all be private

hunterkepley (commented):
mostly LGTM, some more comments

olucasfreitas:
@hunterkepley Now you can check the fixes for all the comments, let me know if there's any more requests

PR #2909, opened 2025-06-03
OCM-15992 | feat: Added private ingress + private API prompts for HCP

hunterkepley:
Adds the following based on OCM-15992:

* Private API prompt instead of Private Ingress for HCP clusters only
* Adds private ingress value as `ingresses[0].ListeningMethod` for the default ingress when creating a cluster


pkg/ocm/clusters.go
```diff
@@ -1140,6 +1141,13 @@ func (c *Client) createClusterSpec(config Spec) (*cmv1.Cluster, error) {
 			defaultIngress.RouteNamespaceOwnershipPolicy(
 				v1.NamespaceOwnershipPolicy(config.DefaultIngress.NamespaceOwnershipPolicy))
 		}
+
+		// Decide ingress listening method
+		if config.PrivateIngress {
+			defaultIngress.Listening(cmv1.ListeningMethodInternal)
+		} else {
+			defaultIngress.Listening(cmv1.ListeningMethodExternal)
```
hunterkepley: @willkutler my only concern is having this false by default. What should the default be?

gdbranco: Default is external listening method if api is public and internal if api is private

cmd/create/cluster/cmd.go
```diff
@@ -545,6 +546,13 @@ func initFlags(cmd *cobra.Command) {
 			"without exposing your traffic to the public internet.",
 	)
 
+	flags.BoolVar(
+		&args.privateIngress,
+		"private-ingress",
```
gdbranco: Just considering what could be used here `--default-ingress-private` to follow the other default-ingress args

pkg/ocm/clusters.go
willkutler: I agree that we should default to API listening

cmd/create/cluster/cmd.go
```diff
@@ -2035,22 +2043,39 @@ func run(cmd *cobra.Command, _ []string) {
...
 
 	privateLinkWarning := "Once the cluster is created, this option cannot be changed."
 	if isSTS {
 		privateLinkWarning = fmt.Sprintf("STS clusters can only be private if AWS PrivateLink is used. %s ",
 			privateLinkWarning)
 	}
 	if interactive.Enabled() && !fedramp.Enabled() && !isPrivateHostedCP {
+		question := "Private Link cluster"
+		if isHostedCP {
+			question = "Private API cluster"
```
willkutler: I think the flow for this should be that we keep the PrivateLink question as is for classic, and then for HCP we want to set the args.private value

pkg/ocm/clusters.go
hunterkepley: Done

cmd/create/cluster/cmd.go
hunterkepley: Ah, in the original ticket you mentioned changing the prompt to `Private API` when HCP Is used- do you want me to remove that?

Will go ahead and change the value it changes to `args.private` rather than `privateLink`

hunterkepley: Unless you want it to also set `privateLink`- let me know, I'll wait to make this change

willkutler: Setting `privateLink` in this flow shouldn't really matter either way since CS will automatically set it to enabled if the cluster is private and HCP. For readability it might make more sense to set it in the CLI as well, so there is not confusion on whether `privateLink` will be set for HCP clusters.

willkutler: regardless of whether we set the privateLink variable when the user selects `Private API: y` , we should remove the `PrivateLink y/N` prompt for HCP clusters

hunterkepley: @willkutler makes sense, let me narrow my questions some;
1. Do I need to _always_ set the `args.private` var from this prompt for HCP? (it sets it conditionally right now)
2. When you say `remove the PrivateLink y/N prompt for HCP clusters`, are you referring to me changing my implementation to having a separate prompt for `Private API`? 

Currently it replaces the text of the prompt with `Private API?` (going off the ticket), but I can make a separate prompt entirely if needed.

Sorry, a little confused on the exact implementation requested

willkutler: Sorry for the confusion on the ticket - currently both HCP and classic clusters can request the `Privatelink: y/N` prompt from the user. that prompt is only asked of HCP clusters when `args.private` is not set to true, as if a HCP cluster has private API enabled then privateLink will automatically be set to true. when the HCP cluster does not have private API, we should not prompt for privateLink at all since if a HCP cluster does not have private API (which is the same as `args.private`) , then we expect privateLink to be disabled. therefore, it makes sense to remove the privateLink prompt entirely for HCP clusters.

For the second part of this - the `Private API` prompt - we do not currently have a prompt in the UI to set `args.private`. Seeing as this is the only way to set private API on a HCP cluster (since the privateLink value is set to true only if `args.private` is true), we should add a prompt to users so they are able to set private API as part of the interactive flow, rather than them just having to use the `--private` flag.

So it's less replacing the PrivateLink prompt with an API one and more removing the unnecessary prompt and enabling the value that we actually use for HCP (`args.private`) to be set as part of the interactive flow.

hunterkepley: Thanks Will 🙂  I'll clean this up tomorrow and see how it looks then

cmd/create/cluster/cmd.go
```diff
@@ -2064,18 +2061,6 @@ func run(cmd *cobra.Command, _ []string) {
...
-				Help: fmt.Sprintf("Private ingress allows you to change the default level of ingress " +
-					"visibility"),
-				Default: privateIngress,
-			})
-			if err != nil {
-				r.Reporter.Errorf("Expected a valid private ingress value: %s", err)
-				os.Exit(1)
-			}
-		}
 	} else if (privateLink || (isSTS && private)) && !fedramp.Enabled() && !isPrivateHostedCP {
```
willkutler: Private API prompt should probably go above this block and the `isPrivateHostedCP` variable declaration so we get the expected warning that privateLink will default to true because the cluster is private HCP - then you should be able to remove the new code that sets privateLink on lines 2083-2087

cmd/create/cluster/cmd.go
```diff
@@ -2085,6 +2070,34 @@ func run(cmd *cobra.Command, _ []string) {
...
+		private, err = interactive.GetBool(interactive.Input{
+			Question: "Private API",
+			Help: fmt.Sprintf("Private API allows you to change whether or not the cluster will use Private " +
+				"API"),
+			Default: args.private,
+		})
+		if err != nil {
+			r.Reporter.Errorf("Expected a valid Private API value: %s", err)
+		}
+		if private {
```
willkutler: don't think this is needed - see other comment

cmd/create/cluster/cmd.go
hunterkepley: moved above the `else if`

cmd/create/cluster/cmd.go
```diff
@@ -2035,21 +2043,46 @@ func run(cmd *cobra.Command, _ []string) {
 	private := args.private
 	isPrivateHostedCP := isHostedCP && private // all private hosted clusters are private-link
```
willkutler: `isPrivateHostedCP` can be initialized to false here even if the user sets private API to true later, I think we need to move this declaration further down or just set the variable to true if the user selects private API in the prompt

willkutler (approved):

PR #2910, opened 2025-06-04
OCM-15880 | ci: Create new dockerfile for e2e testing component

yuwang-RH:
- Create new dockerfile used for e2e testing component
- update build_cli.sh to build rosacli which will used by the e2e pipeline


hack/build_cli.sh
```diff
@@ -14,10 +14,10 @@ do
     if [[ $os == "windows" ]]; then
```
hunterkepley, line 14: It won't let me comment on the exact line, but I think we need ARM as another `arch` on line 4 for Darwin specifically. maybe other OSes too? not 100% sure but I know Darwin needs ARM 64

Issue #2907, opened 2025-05-29
AWS Region complaints from ROSA CLI when the version isn't applicable


iamkirkbater:
Similar to the issue reported in #1386, I attempted to run `rosa list clusters` and was greeted with an error that my AWS Region was not set.

I then set it via env var to `us-east-2`, even though my cluster is deployed to `us-west-2`. When I again `rosa list clusters` I see my cluster.

I wanted to just re-open #1386, however I don't have permissions and it has closed due to being stale.

From a usability standpoint, I think the concern still exists. If the region is not set, it shouldn't matter for commands like `list clusters` where the cluster's region is not being used to limit results.

hunterkepley:
Thanks Kirk, I'll make a Jira issue for this and try to tackle it for next release if possible

Issue #1386, opened 2023-07-26
rosa list clusters --region is confusing


paulczar:
A rosa user may think they can run `rosa list clusters --region us-east-1` to get a list of clusters deployed in just that region.  However the flag is effectively a no-op as its only applicable to AWS operations (which arguably the ROSA cli should only interact with the OCM api, but that's a different discussion).

```
rosa list clusters -h
List clusters.

Usage:
  rosa list clusters [flags]

Aliases:
  clusters, cluster

Examples:
  # List all clusters
  rosa list clusters

Flags:
  -o, --output string   Output format. Allowed formats are [json yaml]
  -a, --all             List all clusters across different AWS accounts under the same Red Hat organization
  -h, --help            help for clusters

Global Flags:
      --color string     Surround certain characters with escape sequences to display them in color on the terminal. Allowed options are [auto never always] (default "auto")
      --debug            Enable debug mode.
      --profile string   Use a specific AWS profile from your credential file.
      --region string    Use a specific AWS region, overriding the AWS_REGION environment variable.
```

PR #2912, opened 2025-06-04
OCM-16119 | fix: No region set for command which needs no region, `list/clusters`

hunterkepley:
* Fixes https://github.com/openshift/rosa/issues/2907 and https://github.com/openshift/rosa/issues/1386
* Fixes [OCM-16119](https://issues.redhat.com//browse/OCM-16119)

The issue stemmed from the AWS client needed to be used when listing clusters. The command is region-agnostic, but the client was created using `WithAWS` which always exits when no region is set with an error. 

To fix this, I introduced an "overloaded" function called `WithAWSWarnInsteadOfExit` to warn about no region being set and continue, rather than error+exit.

### DIFF
--- a/pkg/rosa/runtime.go
+++ b/pkg/rosa/runtime.go
@@ -32,7 +32,7 @@ func NewRuntime() *Runtime {
 	return &Runtime{Reporter: reporter, Logger: logger, Spinner: spinner}
 }
 
-// Adds an OCM client to the runtime. Requires a deferred call to `.Cleanup()` to close connections.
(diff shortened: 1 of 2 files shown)


pkg/rosa/runtime.go
```diff
@@ -71,7 +93,7 @@ func (r *Runtime) Cleanup() {
 	}
 }
 
-// Load the cluster key provided by the user into the runtime and return it
+// GetClusterKey Load the cluster key provided by the user into the runtime and return it
```
hunterkepley, line 96: Edited these comments because they did not have the leading function name required by automatic docs

PR #2926, opened 2025-06-20
OCM-15992 | fix: Remove check stopping private ingress changes

hunterkepley:
* Removes `if !reflect.DeepEqual(config.DefaultIngress, NewDefaultIngressSpec()) {` check when creating cluster spec
* * Replicates behavior in that it only adds the modified default ingress when something is changed, but now includes Private Ingress in that check
* * Fixes issue where private ingress would not be set unless default ingress had another change

### DIFF
--- a/pkg/ocm/clusters.go
+++ b/pkg/ocm/clusters.go
@@ -103,7 +103,7 @@ type Spec struct {
 	HostPrefix                     int
 	Private                        *bool
 	PrivateLink                    *bool
-	PrivateIngress                 bool
+	PrivateIngress                 *bool
 
 	// Properties
 	CustomProperties map[string]string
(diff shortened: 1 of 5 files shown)


pkg/ocm/clusters.go
```diff
@@ -1130,30 +1130,37 @@ func (c *Client) createClusterSpec(config Spec) (*cmv1.Cluster, error) {
...
-		if !fedramp.Enabled() && config.Hypershift.Enabled {
-			if config.PrivateIngress {
+	// Decide ingress listening method if HCP and not fedramp enabled
+	isHcpNotFedramp := !fedramp.Enabled() && config.Hypershift.Enabled
+	if isHcpNotFedramp {
+		if config.PrivateIngress != nil {
+			if *config.PrivateIngress {
 				defaultIngress.Listening(cmv1.ListeningMethodInternal)
 			} else {
 				defaultIngress.Listening(clusterApiListeningMethod)
```
willkutler: think this should be `defaultIngress.Listening(cmv1.ListeningMethodExternal)`

hunterkepley:
> Creating a cluster with `--private` and `--default-ingress-private=false` will give an error: `Creating cluster 'wk' should fail: status is 400, identifier is '400', code is 'CLUSTERS-MGMT-400', at '2025-06-20T20:32:28Z' and operation identifier is '2ee1f524-a98d-4d39-9958-ceb8e56b8fb3': The following errors occurred in the subnets provided: 1 public subnet is required per VPC. Only private subnets were specified in zones [us-west-2a]. Check the subnets and try again.` - or if a private and public subnet are entered, it will error saying private clusters cannot have public subnets.
> 
> Could we allow users to set public subnets on private clusters if the ingress is public and vice versa?

Checked into this, I can't find where the validation occurs in the CLI- maybe it's part of CS API? Please let me know if I'm wrong here 😅 - The second part definitely sounds like something the CLI has (I followed up on slack with the location of where I think you want the changes made), but the error given is not part of the CLI

PR #2935, opened 2025-07-02
OCM-16864 | fix: Introduce private API to private STS validation

hunterkepley:
Fixes [OCM-16864](https://issues.redhat.com//browse/OCM-16864), a bug where Private API was not taken into account once introduced as the alternative for classic's PrivateLink in HCP in interactive mode.

Creation of a cluster will error after saying `Yes` to the Private API prompt when using HCP clusters. Example failure happens here in the cluster creation process:

```
? AWS region (default = 'us-west-2'): us-west-2
? Private API: Yes
? Private ingress: Yes
E: Private STS clusters are only supported through AWS PrivateLink
```

# Below are test runs surrounding this same step in the cluster creation process using this MR confirming it fixes the issue

### Test runs:

HCP with private API enabled via interactive (OK):

```
? Private API: Yes
? Private ingress: Yes
? Machine CIDR: 10.0.0.0/16
```

HCP with private API and ingress disabled (KO):

```
? Private API: No
? Private ingress: No
E: Private STS clusters are only supported through AWS Private API
```

HCP with private API disabled and ingress enabled (OK):

```
? Private API: No
? Private ingress: Yes
? Machine CIDR: 10.0.0.0/16
```

Classic with AWS Private Link enabled via interactive (OK):

```
? Private Link cluster: Yes
? Machine CIDR: 10.0.0.0/16
```

Classic with AWS Private Link disabled (KO):

```
? Private Link cluster: No
E: Private STS clusters are only supported through AWS PrivateLink
```

* This MR also improves wording of a warning + the help msg for the Private API prompt involved here

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -2052,7 +2052,7 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 	if interactive.Enabled() && !fedramp.Enabled() {
 		if !isHostedCP {
-			question := "Private Link cluster"
+			question := "PrivateLink cluster"
 			privateLink, err = interactive.GetBool(interactive.Input{
 				Question: question,
 				Help:     fmt.Sprintf("%s %s", cmd.Flags().Lookup("private-link").Usage, privateLinkWarning),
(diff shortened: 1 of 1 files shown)


willkutler:
I think an error message change is overkill for this bug - I believe the issue is just that we initialize `privateLink := args.privateLink || fedramp.Enabled() || (private && isHostedCP)` before processing the interactive Private API prompt - so privateLink is false when it should be true.

cmd/create/cluster/cmd.go
```diff
@@ -2086,8 +2085,9 @@ func run(cmd *cobra.Command, _ []string) {
 		}
 	}
 
-	if privateLink {
+	if privateLink || private {
```
willkutler: `if privateLink || (private && isHostedCP)`

willkutler (approved):

PR #2943, opened 2025-07-15
OCM-16926 | feat: Update help message for private-link to 'deprecate' for HCP

hunterkepley:
We cannot deprecate fully, since we don't know if the customer is using HCP when the CLI is built. Deprecating mid-run of the command, which is our only option for CobraCLI deprecation, will not work whatsoever like intended (nothing will happen). 
Therefore, the best solution, as provided by @yuwang-RH in this ticket, is to update the `--private-link` help message itself to let the customer know that it is deprecated specifically for `v1.2.55` and beyond. And the solution to this deprecation is to use the `--private` and `--default-ingress-private` flags for HCP clusters


olucasfreitas (approved):
/lgtm

PR #2944, opened 2025-07-16
OCM-16902 | fix: Fix private+privateIngress clusters + validation

hunterkepley:
Fixes issue where HCP clusters failed to validate with `private` and `private-ingress` set to `true`. Also added validation for users who try to use `--private-link` with HCP clusters

### DIFF
--- a/pkg/interactive/validation.go
+++ b/pkg/interactive/validation.go
@@ -226,7 +226,8 @@ func RegExpBoolean(restr string) Validator {
 
 // SubnetsCountValidator get a slice of `[]core.OptionAnswer` as an interface.
 // e.g. core.OptionAnswer { Value: subnet-04f67939f44a97dbe (us-west-2b), Index: 0 }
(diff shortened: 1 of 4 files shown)


pkg/interactive/validation.go
```diff
@@ -226,7 +226,8 @@ func RegExpBoolean(restr string) Validator {
 
 // SubnetsCountValidator get a slice of `[]core.OptionAnswer` as an interface.
 // e.g. core.OptionAnswer { Value: subnet-04f67939f44a97dbe (us-west-2b), Index: 0 }
-func SubnetsValidator(awsClient aws.Client, multiAZ bool, privateLink bool, hostedCP bool) Validator {
+func SubnetsValidator(awsClient aws.Client, multiAZ bool, privateValue bool, hostedCP bool,
```
gdbranco, line 229: This is not a great method in terms of parameters :melting_face:

hunterkepley: yeahhhhh lol. im not happy with it either

PR #2952, opened 2025-07-23
OCM-17372 | feat: Add support for handling deprecation headers in ROSA CLI

tzvatot:


davidleerh (approved):

pkg/ocm/client.go
```diff
@@ -120,6 +121,9 @@ func (b *ClientBuilder) Build() (result *Client, err error) {
 	builder := sdk.NewConnectionBuilder()
 	builder.Logger(logger)
 
+	// Add deprecation transport wrapper to automatically handle deprecation headers
+	builder.TransportWrapper(deprecation.NewTransportWrapper())
```
hunterkepley, line 125: We need to include this file's changes in a separate MR, and merge the `ocm-common` update before merging this code using the new `deprecation` package, to get the pipeline passing I believe 🙂 

The testing environment does not know this exists yet

hunterkepley: Sorry for the late reply @tzvatot - I miss GitHub notifications easily because I'm involved in some open source projects outside of work

davidleerh:
just an fyi @willkutler that the deprecation header is being handled.

pkg/ocm/client.go
tzvatot: Created https://github.com/openshift/rosa/pull/2959 to do that

PR #2953, opened 2025-07-23
OCM-1520 - Hide empty fields on list commands and sub-commands

olucasfreitas:
# Details

This PR improves the UX when listing resources by introducing a new `--hide-empty-fields` flag that automatically hides columns containing no data across all rows, making the output cleaner and more readable.

---

# How to Test

Before running the test cases, ensure you have a cluster with machine pools that have varying configurations (some with labels/taints, some without).

---

## Test Case 1 – Default Behavior (All Columns Shown)

1. Run the command:

    ```bash
    ./rosa list machinepools --cluster <cluster-name>
    ```

2. You should see **all columns** displayed, including empty ones like `LABELS`, `TAINTS`, `SUBNETS`, and `SG IDS`:

    ```
    ID          AUTOSCALING  REPLICAS  INSTANCE TYPE  LABELS  TAINTS  AVAILABILITY ZONES  SUBNETS  SPOT INSTANCES  DISK SIZE  SG IDS
    workers-1   No           3         m5.xlarge                      us-east-1a                   No              default    
    workers-2   No           2         m5.2xlarge                     us-east-1b                   No              default    
    ```

---

## Test Case 2 – Using --hide-empty-fields Flag

1. Run the command:

    ```bash
    ./rosa list machinepools --cluster <cluster-name> --hide-empty-fields
    ```

2. Empty columns should be **automatically hidden**, showing only columns with data:

    ```
    ID          AUTOSCALING  REPLICAS  INSTANCE TYPE  AVAILABILITY ZONES  SPOT INSTANCES  DISK SIZE
    workers-1   No           3         m5.xlarge      us-east-1a          No              default
    workers-2   No           2         m5.2xlarge     us-east-1b          No              default
    ```

---

## Test Case 3 – With Mixed Data (Some Empty, Some Not)

1. If you have machine pools with labels or taints:

    ```bash
    ./rosa list machinepools --cluster <cluster-name> --hide-empty-fields
    ```

2. Columns with **at least one non-empty value** should still be displayed:

    ```
    ID          AUTOSCALING  REPLICAS  INSTANCE TYPE  LABELS      AVAILABILITY ZONES  SPOT INSTANCES  DISK SIZE
    workers-1   No           3         m5.xlarge                  us-east-1a          No              default
    workers-2   No           2         m5.2xlarge     app=backend us-east-1b          No              default
    ```

---

## Additional Info

The `--hide-empty-fields` flag:
- Only affects **table format** output (default)
- Has **no effect** on JSON or YAML output formats
- Is available for all `rosa list` subcommands

---

# Ticket

Closes [[OCM-1520](https://issues.redhat.com/browse/OCM-1520)]


cmd/list/accessrequests/cmd.go
```diff
@@ -67,12 +65,10 @@ func ListAccessRequestsRunner() rosa.CommandRunner {
...
 			}
-			writer := tabwriter.NewWriter(os.Stdout, 0, 4, 2, ' ', 0)
-			output, hasPending, pendingId := getAccessRequestsOutput(clusterId, accessRequests)
-			fmt.Fprint(writer, output)
-			if err := writer.Flush(); err != nil {
-				return err
-			}
+			// Create the writer that will be used to print the tabulated results:
+			outputStr, hasPending, pendingId := getAccessRequestsOutput(clusterId, accessRequests)
+			fmt.Print(outputStr)
```
hunterkepley: Was this leftover from testing? The original line was printing to a buffer for writing the table, this is just a print

cmd/list/accessrequests/cmd.go
```diff
@@ -67,12 +65,10 @@ func ListAccessRequestsRunner() rosa.CommandRunner {
 				}
 				return nil
 			}
-			writer := tabwriter.NewWriter(os.Stdout, 0, 4, 2, ' ', 0)
-			output, hasPending, pendingId := getAccessRequestsOutput(clusterId, accessRequests)
-			fmt.Fprint(writer, output)
-			if err := writer.Flush(); err != nil {
-				return err
-			}
+			// Create the writer that will be used to print the tabulated results:
```
hunterkepley: You removed the line which creates the writer- was this comment intentional?

cmd/list/accessrequests/cmd.go
```diff
@@ -95,12 +93,13 @@ func getAccessRequestsOutput(clusterId string, accessRequests []*v1.AccessReques
...
+			string(accessRequest.Status().State()),
 			accessRequest.ID(),
 			accessRequest.ClusterId(),
-			accessRequest.UpdatedAt().Format(time.UnixDate))
+			accessRequest.UpdatedAt().Format(time.UnixDate),
+		)
 	}
 
-	return output, hasPending, id
+	return tb.RenderToString(), hasPending, id
```
hunterkepley: We shouldn't need to replace the tab writer with a custom service + builder, the tab writer is meant for writing tables

hunterkepley: Also I am unsure if `Render` is distinct/clear enough, we may want something a bit more clear as to what it does- but i would prefer not to go with replacing the current flow

cmd/list/machinepool/cmd.go
```diff
@@ -79,13 +76,6 @@ func NewListMachinePoolCommand() *cobra.Command {
 		"Show whether each machine pool is Windows LI enabled",
 	)
```
hunterkepley: Why are we removing `--all` in this MR?

cmd/list/machinepool/cmd_test.go
```diff
@@ -299,7 +276,7 @@ var _ = Describe("List machine pool", func() {
 				Expect(stdout).To(ContainSubstring("sg-1, sg-2"))
 			})
 
-			It("Hides empty columns by default", func() {
+			It("Shows all base columns by default", func() {
```
hunterkepley: What are the 'base columns'? We should identify it in a comment I think

cmd/list/rhRegion/cmd.go
```diff
@@ -68,7 +67,10 @@ func ListRhRegions(discoveryURL string, r *rosa.Runtime) error {
 		return fmt.Errorf("Failed to determine gateway URL: %v", err)
 	}
 
-	fmt.Fprintf(writer, "Discovery URL: %s\n\n", gatewayURL)
+	// Create the writer that will be used to print the tabulated results:
+	tb := output.NewTableBuilder()
+
+	fmt.Printf("Discovery URL: %s\n\n", gatewayURL)
```
hunterkepley: Why are we printing this out?

cmd/list/rhRegion/cmd.go
```diff
@@ -79,10 +81,11 @@ func ListRhRegions(discoveryURL string, r *rosa.Runtime) error {
 		r.Reporter.Warnf("No regions found")
 		return nil
 	}
-	fmt.Fprintf(writer, "RH Region\t\tGateway URL\n")
+
+	tb.SetHeaders("RH Region", "Gateway URL")
 	for regionName, region := range regions {
-		fmt.Fprintf(writer, "%s\t\t%v\n", regionName, region.URL)
+		tb.AddRow(regionName, region.URL)
```
hunterkepley: Does this account for the fact that the original table was tabbed twice between the values? `\t\t`. The table builder doesnt have any configuration for this it seems. This means every table, even if it is formatted differently in a specific list command, is going to be formatted the same way

cmd/list/user/cmd.go
```diff
@@ -117,11 +115,11 @@ func run(_ *cobra.Command, _ []string) {
 	}
 
 	// Create the writer that will be used to print the tabulated results:
-	writer := tabwriter.NewWriter(os.Stdout, int(longestUserId)+2, 4, 2, ' ', 0)
-	fmt.Fprintf(writer, "ID\tGROUPS\t\n")
+	tb := output.NewTableBuilder()
```
hunterkepley: An example of how the writers themselves are misconfigured in the new table builder- the args are always the same in the table builder, but were different for each list command table originally

cmd/list/version/cmd.go
```diff
@@ -128,12 +127,19 @@ func run(_ *cobra.Command, _ []string) {
...
+				version.RawID(),
+				isDefault,
+			)
+		} else {
+			tb.AddRow(
+				version.RawID(),
+				isDefault,
+				availableUpgrades,
+			)
+		}
```
hunterkepley: theres no need for this if the table writer gets rid of empty columns using that new flag

cmd/rosa/structure_test/command_args/rosa/list/machinepools/command_args.yml
```diff
@@ -5,4 +5,4 @@
 - name: az-type
 - name: dedicated-host
 - name: win-li
-- name: all
+- name: hide-empty-fields
```
hunterkepley: Is there a reason we are getting rid of all? Was it an ask? We should reflect this in another ticket if so, and not remove flags without telling the appropriate teams such as docs (`--all` is already out and in the new ROSA versions)

pkg/machinepool/machinepool.go
```diff
@@ -1058,15 +1056,11 @@ func (m *machinePool) ListMachinePools(r *rosa.Runtime, clusterKey string, clust
...
 
-	// Create the writer that will be used to print the tabulated results:
-	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
-
-	finalStringToOutput := getMachinePoolsString(r, machinePools, args)
 	if isHypershift {
-		finalStringToOutput = getNodePoolsString(nodePools)
+		fmt.Print(getNodePoolsString(nodePools))
+	} else {
+		fmt.Print(getMachinePoolsString(r, machinePools, args))
```
hunterkepley: Why are we sometimes deleting the buffer writes, and sometimes replacing them with normal prints to the terminal? It's not consistent so I'm unsure on what's happening here

hunterkepley: `Fprint` is not the same thing as a normal print

pkg/machinepool/machinepool.go
```diff
@@ -1240,90 +1234,65 @@ func getMachinePoolsString(
...
-
-		columnValues := make([]string, numPools)
-		hasNonEmptyValue := false
-
-		for i, pool := range machinePools {
-			if pool == nil {
-				columnValues[i] = "-"
-				continue
-			}
+	tb := output.NewTableBuilder()
```
hunterkepley: Is there a reason we are replacing most of the content of https://github.com/openshift/rosa/pull/2906 here? 

By design, you have column definitions + the functions which are called to print these columns. Why can't we remove one of these definitions when the data is empty?

I do agree, that maybe the old solution shouldn't have been introduced, and should have been kept the same as the other list commands. But this should be addressed in a refactor MR, separate from this one (I will make a ticket to do this soon)

pkg/machinepool/machinepool.go
```diff
@@ -1240,90 +1234,65 @@ func getMachinePoolsString(
...
-		tableBuilder.WriteString("\n")
+		if args.ShowDedicated {
+			values = append(values, isDedicatedHost(mp, runtime))
+		}
+
+		tb.AddRow(values...)
 	}
 
-	return tableBuilder.String()
+	return tb.RenderToString()
```
hunterkepley: NP: This would be a bit more clear if it were named `.ToString()` or something similar

pkg/machinepool/machinepool.go
```diff
@@ -1240,90 +1234,65 @@ func getMachinePoolsString(
...
 func getNodePoolsString(nodePools []*cmv1.NodePool) string {
-	outputString := "ID\tAUTOSCALING\tREPLICAS\t" +
-		"INSTANCE TYPE\tLABELS\t\tTAINTS\t\tAVAILABILITY ZONE\tSUBNET\tDISK SIZE\tVERSION\tAUTOREPAIR\t\n"
+	tb := output.NewTableBuilder()
+	tb.SetHeaders("ID", "AUTOSCALING", "REPLICAS", "INSTANCE TYPE", "LABELS", "TAINTS",
+		"AVAILABILITY ZONE", "SUBNET", "DISK SIZE", "VERSION", "AUTOREPAIR")
+
 	for _, nodePool := range nodePools {
-		outputString += fmt.Sprintf("%s\t%s\t%s\t%s\t%s\t\t%s\t\t%s\t%s\t%s\t%s\t%s\t\n",
+		tb.AddRow(
```
hunterkepley: Another example of when the table is now formatted differently from the original table, notice the `\t\t` in a couple spots in the original rows

pkg/machinepool/machinepool_test.go
```diff
@@ -190,24 +190,11 @@ var _ = Describe("Machinepool and nodepool", func() {
...
-				cluster.NodePools().Get(0).AvailabilityZone(),
-				cluster.NodePools().Get(0).Subnet(),
-				ocmOutput.PrintNodePoolDiskSize(cluster.NodePools().Get(0).AWSNodePool()),
-				ocmOutput.PrintNodePoolVersion(cluster.NodePools().Get(0).Version()),
-				ocmOutput.PrintNodePoolAutorepair(cluster.NodePools().Get(0).AutoRepair()))))
+
+			// Expected output with TableBuilder format (space-aligned columns)
+			expectedOutput := "ID  AUTOSCALING  REPLICAS  INSTANCE TYPE  LABELS  TAINTS  AVAILABILITY ZONE  SUBNET  DISK SIZE  VERSION  AUTOREPAIR\n" +
+				"np  No           /8                                       az                 sn      256 GiB    1        No\n"
+			Expect(out).To(Equal(expectedOutput))
```
hunterkepley: I'm unsure I see a reason to use spaces, the reason this test broke is not because your TableBuilder has a specific format (`// Expected output with TableBuilder format (space-aligned columns)`) - it's because the format is just different;

`%s\t%s\t%s\t%s\t%s\t\t%s\t\t%s\t%s\t%s\t%s\t%s\t\n` is the original table format (sometimes has `\t\t`)

`%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\t\n` is yours if we dont count any removed columns

`\t` tabs are arbitrary, and are represented differently depending on the source rendering it. Some terminals will show 2 spaces, some will show 8, some will show 4, whatever it may be. Using them allows ease of development/test writing/ensuring the table is lined up correctly (which is also why it may need `\t\t`! If the data is too long for a single `\t`, then it will overflow the rest of the row, and look bad. Try it yourself, print out a long string in a table made with this table writer, it will tab everything in that row to the right by however many it takes

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
+package output
```
hunterkepley: I think the name of the file is a bit too broad. `table_builder.go` would make sense for example, since this houses the table builder, not hiding empty things from the output in general

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
...
+			writer: tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0),
+		}
+	}
+
+	buffer := &bytes.Buffer{}
+	return &TableBuilder{
+		buffer: buffer,
+		writer: tabwriter.NewWriter(buffer, 0, 0, 2, ' ', 0),
+	}
+}
```
hunterkepley: Is there a reason to use a buffer when the flag is used? I also see these are not used anyway, so im not entirely sure the goal

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
...
+func (tb *TableBuilder) RenderToString() string {
+	return tb.buildOutput()
+}
+
+// buildOutput builds the table output string
+func (tb *TableBuilder) buildOutput() string {
+	if !HasHideEmptyFieldsFlag() {
+		// Direct output without filtering
+		var outputBuilder strings.Builder
+		outputWriter := tabwriter.NewWriter(&outputBuilder, 0, 0, 2, ' ', 0)
```
hunterkepley: Why are we recreating the tabwriter and buffer if they are both already in the TableBuilder struct and set up already based on the flag? `Render` and `RenderToString` can be compressed into a single function, with the same code, and it will only print when the `stdOut` buffer is used

I'm not sure I understand

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
...
+
+		if len(tb.headers) > 0 {
+			fmt.Fprintf(outputWriter, "%s\n", strings.Join(tb.headers, "\t"))
```
hunterkepley: `strings.Join` already handles the empty check for you, but we should fail fast and return early if the length of the headers is `0` anyway. There's no table to build without headers.

[Also, we have now introduced 2 more for loops with `strings.Join` on top of the for loop already used for adding each row to this table builder in the first place](https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/strings/strings.go;l=483)

hunterkepley: that link is `strings.Join` in the go source code

cmd/list/accessrequests/cmd.go
```diff
@@ -85,7 +81,9 @@ func ListAccessRequestsRunner() rosa.CommandRunner {
 }
 
 func getAccessRequestsOutput(clusterId string, accessRequests []*v1.AccessRequest) (string, bool, string) {
-	output := "STATE\tID\tCLUSTER ID\tUPDATED AT\n"
+	tb := output.NewTableBuilder()
+	tb.SetHeaders("STATE", "ID", "CLUSTER ID", "UPDATED AT")
+
```
hunterkepley: Is there a reason you are turning this into a table before returning it? Originally, the writer was not inside this function, since the purpose of `getAccessRequestsOutput` is to return the 3 strings used for output (based on the name+original function)

cmd/list/accessrequests/cmd.go
hunterkepley: it was added after removing it so im a bit confused :sweat_smile:

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
...
+		}
+		for _, row := range tb.rows {
+			fmt.Fprintf(outputWriter, "%s\n", strings.Join(row, "\t"))
```
hunterkepley: I think there's some confusion on the use case of the new `RenderToString` function. We are using the tab writer in the general function thats used for both printing the table (`Render`) and returning a string of the data in the table ('RenderToString').
The original design is that there's some situations where we want the `\t`'s included, in a string, to potentially be printed or used in another way later, or maybe even discarded.

Looking above, it seems you made `RenderToString()` return the actual table, with tabs already processed, before printing with `fmt.Print`. This kind of defeats the purpose of `RenderToString` with this new design, since the only difference is that we dont immediately print with it. (Though, in the implementation of that function, it seems there are `fmt.Print`'s immediately following it)

Just as an example, if we parsed the string locally or through an API, and it was expecting a string formatted like so: `DATA\tDATA\tDATA` but it came as `DATA    DATA    DATA    DATA` it may break

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
...
+		}
+		outputWriter.Flush()
+		return outputBuilder.String()
+	}
+
+	columnsToKeep := make([]bool, len(tb.headers))
+	for colIdx := range tb.headers {
+		columnsToKeep[colIdx] = false
```
hunterkepley: Default value is already false for bools. We can remove this line with no impact

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
...
+		// Check if any row has non-empty value for this column
+		for _, row := range tb.rows {
+			if colIdx < len(row) {
```
hunterkepley: Is there a case in which we would be missing data entirely in a column to make this `if` statement false? What's the plan if it were to be out of index? It looks like the function would continue and return an empty table since every value would be false in `columnsToKeep`

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
...
+		var filteredRow []string
+		for i, keep := range columnsToKeep {
+			if keep && i < len(row) {
+				filteredRow = append(filteredRow, row[i])
+			}
+		}
+		if len(filteredRow) > 0 {
+			fmt.Fprintf(outputWriter, "%s\n", strings.Join(filteredRow, "\t"))
+		}
+	}
```
hunterkepley: This is a bit confusing to read with the fact that `tb.rows` is actually a nested array of the entire table, not just strings of the rows

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
...
+
+	// Build filtered output
+	var outputBuilder strings.Builder
+	outputWriter := tabwriter.NewWriter(&outputBuilder, 0, 0, 2, ' ', 0)
+
+	// Write filtered headers
+	if len(filteredHeaders) > 0 {
+		fmt.Fprintf(outputWriter, "%s\n", strings.Join(filteredHeaders, "\t"))
+	}
```
hunterkepley: we already have a tabwriter+buffer in the TableBuilder

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
...
+		if len(filteredRow) > 0 {
+			fmt.Fprintf(outputWriter, "%s\n", strings.Join(filteredRow, "\t"))
+		}
+	}
+
+	outputWriter.Flush()
+	return outputBuilder.String()
+}
+
+// ProcessTabularOutput processes tabwriter output to hide empty columns when --hide-empty-fields is set
```
hunterkepley: dont we do this above already?

hunterkepley: I dont see this function used anywhere..

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
...
+			} else {
+				first = false
+			}
+			result.WriteString(strings.TrimSpace(colValue))
+		}
+		result.WriteString("\n")
+	}
+
+	return result.String()
+}
```
hunterkepley: What is the usecase of this function? It looks like it's meant to process the output from a `tabwriter` back into a tabulated string?

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
...
+
+// findColumnStarts finds the starting positions of columns in a header line
+func findColumnStarts(header string) []int {
```
hunterkepley: This function is only used in the non-used `ProcessTabularOutput` function

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
...
+				starts = append(starts, i)
+				inColumn = true
+			}
+		} else {
+			inColumn = false
+		}
+	}
+
+	return starts
+}
```
hunterkepley: Since it's not used anywhere I am not sure you want it now, but I'm not sure this is needed overall even if so. This is why the original tabwriter + way of tackling this issue was done the way it was, we don't want to iterate over these strings as if a user pasted them in, it's tedious and error prone

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
...
+
+// extractColumn extracts a column value from a line based on column positions
```
hunterkepley: Another function only used in `ProcessTabularOutput`

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
...
+
+	if end > len(line) {
+		end = len(line)
+	}
+
+	return line[start:end]
+}
+
+// isEmptyValue checks if a value is considered empty
+func isEmptyValue(value string) bool {
```
hunterkepley: Another function only used in `ProcessTabularOutput`

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
...
+	return value == "" || value == "-"
+}
+
+// RemoveEmptyFieldsFromStructuredData removes empty fields from JSON/YAML data
+func RemoveEmptyFieldsFromStructuredData(data any) any {
```
hunterkepley: I dont believe this was an ask in the ticket

The reason we are removing from the tables is because they are the human readable output, which is meant to be easy to read

hunterkepley: This is a very broadly named function + parameter. It's also a generic function to deal with both JSON and YAML, I don't think the purpose+result+action is clear to someone reading this function name somewhere else in the code
 It may be helpful to look through how we print JSON or YAML in these commands

pkg/output/hide_empty.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	var parsed any
+	err = json.Unmarshal(jsonBytes, &parsed)
+	if err != nil {
+		return nil, err
+	}
+
+	return RemoveEmptyFieldsFromStructuredData(parsed), nil
+}
```
hunterkepley: I do not think this will work, it says it's for JSON/YAML and is generic, but converts the data to JSON and back before parsing it. Why would we need to do this to JSON data before parsing anyway? Were the API responses which were fed into it by the above commands inconsistent in testing? I'm worried it will not be able to parse a yaml object into a json object cleanly, or at all

hunterkepley: Tests would be very useful for these funcs if they were kept, but I don't think we should be changing the json/yaml output anyway

pkg/output/output.go
```diff
@@ -42,9 +43,35 @@ import (
...
 func Print(resource interface{}) error {
 	var b bytes.Buffer
 
+	// Process the resource to remove empty fields if flag is set
+	if HasHideEmptyFieldsFlag() {
+		processedResource, err := ProcessStructuredOutput(resource)
+		if err == nil {
+			resource = processedResource
+		}
+		// If error, continue with original resource
```
hunterkepley: NP: is the comment meant to be above the error check? Or was it left over? We dont need to have a comment for 1 line error checks, they are self explanatory

hunterkepley: I would highly recommend looking through the rest of this file, at functions like `parseResource`, to get an idea of how we parse JSON and YAML, im afraid the function introduced for `Processing Structured Output` will produce some issues

cmd/list/addon/cmd.go
```diff
@@ -143,12 +142,13 @@ func listClusterAddOns(clusterKey string, r *rosa.Runtime) {
...
 
 	// Create the writer that will be used to print the tabulated results:
-	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
-	fmt.Fprintf(writer, "ID\t\tNAME\t\tSTATE\n")
+	tb := output.NewTableBuilder()
+	tb.SetHeaders("ID", "NAME", "STATE")
+
 	for _, clusterAddOn := range clusterAddOns {
-		fmt.Fprintf(writer, "%s\t\t%s\t\t%s\n", clusterAddOn.ID, clusterAddOn.Name, clusterAddOn.State)
+		tb.AddRow(clusterAddOn.ID, clusterAddOn.Name, clusterAddOn.State)
```
hunterkepley: `\t\t`

pkg/output/hide_empty.go
hunterkepley: maybe some more clear variable names could be used

pkg/machinepool/machinepool_test.go
```diff
@@ -535,9 +506,9 @@ var _ = Describe("Machinepool and nodepool", func() {
 			args := ListMachinePoolArgs{}
 			out := getMachinePoolsString(r, cluster.MachinePools().Slice(), args)
 
-			// Only columns with actual data should be included
-			expectedOutput := "ID\tAUTOSCALING\tREPLICAS\tINSTANCE TYPE\tSPOT INSTANCES\tDISK SIZE\n" +
-				"mp-1\tNo\t2\tm5.medium\tNo\tdefault\n"
+			// All base columns are included by default, even if empty
```
hunterkepley: was this an ask?

pkg/machinepool/machinepool.go
hunterkepley: After looking at the actual function, this would be much more useful as a ToString which just returned the pre-tabwriter string.

cmd/list/rhRegion/cmd.go
hunterkepley: Commands like this with 2 columns dont really need this feature, it's definitely fine to have, but I'm thinking each RH Region should have a gateway URL so it may not ever function here

hunterkepley (commented):
Overall, my thoughts are that this solution is a bit of overengineering for the problem, I did however provide a round of review on the entire MR to give you some stuff to look at.

Anyway, as far as this MR goes,

`text/tabwriter`'s `tabwriter` has a constant called `DiscardEmptyColumns` which will accomplish hiding empty columns for us

Just need 1 line for each list command to change this attribute, when that new flag is used.  I'd recommend rewriting with this in mind :slightly_smiling_face: 

Documentation on tabwriter: https://pkg.go.dev/text/tabwriter

P.S.; I think we should rename this MR, `Hide empty fields on list commands and sub-commands` -> `Hide empty columns from list command tables` , WDYT?

And we need to update the commit+MR titles, to fit the required formatting of `TICKET-#### | type: MSG` in this repo

pkg/output/output.go
```diff
@@ -42,9 +43,35 @@ import (
...
+
+// AddHideEmptyFieldsFlag adds the --hide-empty-fields flag to a command
+func AddHideEmptyFieldsFlag(fs *pflag.FlagSet) {
+	fs.BoolVar(
+		&hideEmptyFields,
+		"hide-empty-fields",
+		false,
+		"Hide fields that do not contain any data",
+	)
+}
```
hunterkepley: From the ticket:

```
A one-letter or two-letter flag would be appropriate for this, and visible as an option in the help messages. 

The flag for this might be -cc (column collapse?) or -hc (hide columns?).
```

cmd/list/machinepool/cmd.go
olucasfreitas: This was removed because now as default we would show all the columns, and hide them with the new flag, so the ''all" flag would become redundant

cmd/list/rhRegion/cmd.go
olucasfreitas: Could you maybe point out whats commands I should not worry of implementing this is, because in this ticket, there weren't any commands laid out for me to focus on

pkg/machinepool/machinepool.go
olucasfreitas: I refactored this because all the code here would be unnecessary due to the new implementation, because right now the machine pools is filtering out the content, witch would make no sense having the new flag

pkg/output/output.go
olucasfreitas: I tried adding hc to the flags but searching in the documentation it would not accept a two letter flag

cmd/list/accessrequests/cmd.go
hunterkepley: Hey, did you read my question from the first comment in this thread here? I see a thumbs up but I was looking for an answer 🙂

cmd/list/machinepool/cmd.go
hunterkepley: We still have 3 new columns in the `list/machinepool` command- shouldn't we have `--all` to show each of these new columns, without having to type in the 3 separate flags? I believe this is why it was asked to be introduced by Aaren in #2906 but i can't fully remember the context to be honest

cmd/list/rhRegion/cmd.go
hunterkepley: > Commands like this with 2 columns dont really need this feature, it's definitely fine to have, but I'm thinking each RH Region should have a gateway URL so it may not ever function here

Majority are fine, here I mention that this command has 2 columns, and neither should ever be empty- it's fine to keep it here though since it won't affect anything, it just may never get use

hunterkepley: > Why are we printing this out?

Also did you see my first comment?

cmd/list/accessrequests/cmd.go
hunterkepley: > We shouldn't need to replace the tab writer with a custom service + builder, the tab writer is meant for writing tables

Hey, just making sure you saw this- I see the thumbs up below but I'm curious what you are thinking about that comment + the summary above the all the review comments

pkg/machinepool/machinepool.go
hunterkepley: Sorry im a bit confused- how does this remove the need for `--all`?

We introduced 3 new columns- as optionally viewed columns- in #2906 . The purpose of `--all` I thought was to show all 3 of these new columns when used, so the fact we are hiding empty columns should have no effect on this functionality besides it may not show one of the 3 new columns if it is indeed empty

Looking through your tests I saw that we expect only the **base** columns to show, I.E., not these 3 new ones. I think there is a valid reason to keep `--all` and I don't think it's been affected by #2953

cmd/list/machinepool/cmd.go
olucasfreitas: I see, looking from that point of view I would agree its better to have the all flag

cmd/list/accessrequests/cmd.go
olucasfreitas: Sorry, I just did not send the response 
In this case, I'm probably gonna get rid of it due to the new implementation that you mentioned

cmd/list/accessrequests/cmd.go
olucasfreitas: This was an error, removed by accident

cmd/list/accessrequests/cmd.go
olucasfreitas: Yes, leftover

cmd/list/machinepool/cmd_test.go
olucasfreitas: The columns that are not activated by flags

cmd/list/rhRegion/cmd.go
olucasfreitas: That print was for testing, so leftover

cmd/list/rhRegion/cmd.go
olucasfreitas: Definitely wrong here, will look for a fix in the changes I'm making

cmd/rosa/structure_test/command_args/rosa/list/machinepools/command_args.yml
olucasfreitas: This I addressed in other comment

pkg/output/hide_empty.go
olucasfreitas: This is gonna get removed with the new approach

pkg/output/hide_empty.go
olucasfreitas: Used it for other purposes, but forgot to clean up

pkg/output/output.go
hunterkepley: > A one-letter or two-letter flag would be appropriate for this

I think a 1 letter alias would be fine based on what was on the ticket. You can still have a 2 letter alias, it just won't be using a single dash like `-cc`. It'll be `--cc` in practice. the single dash is known as `shorthand`, an alias is just as useful for this as long as it is visible in the help like the ticket says

pkg/output/hide_empty.go
hunterkepley: I think in future work, it's good for you to know that your code I mentioned here was the right way of going about it, I think the fact that it was shadowed later and not used in the end was what should be cleaned up. Maybe I worded that in a way that made it sound like the opposite- but the above code shown above was good

I like the idea of these writer/buffer in the TableBuilder itself (which is decided on creation) lets it know it's job- but at the same time, I think that we should only need one buffer+writer for this in general

It's good for reusing in different member functions though, even if we had that single writer+buffer, to keep it in the struct

PR #2966, opened 2025-08-07
OCM-17643 | fix: Fix retrieving versions for listing CLI versions

hunterkepley:
Fixes issue brought up by customers and internal users, where after moving from ART to CGW for the CLI's download mirror, redirects were not working as expected. We must migrate the links within the CLI, as well as get the version retrieval service working with the new website

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -16,8 +16,8 @@ import (
 )
 
 const (
-	DownloadLatestMirrorFolder = "https://mirror.openshift.com/pub/openshift-v4/clients/rosa/latest/"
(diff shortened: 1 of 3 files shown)


pkg/version/version.go
```diff
@@ -16,8 +16,8 @@ import (
 )
 
 const (
-	DownloadLatestMirrorFolder = "https://mirror.openshift.com/pub/openshift-v4/clients/rosa/latest/"
-	baseReleasesFolder         = "https://mirror.openshift.com/pub/openshift-v4/clients/rosa/"
+	DownloadLatestMirrorFolder = "https://mirror.openshift.com/pub/cgw/rosa/latest/"
```
gdbranco, line 19: Create a ticket to have this be consumed from API at some point?

PR #2974, opened 2025-08-14
OCM-17719 | feat: Deprecate user workload monitoring toggle

hunterkepley:
* Deprecates and hides the disable-workload-monitoring flag in both edit+create -> cluster
* Also removes the printout from describe -> cluster
* Removes all logic surrounding disable-workload-monitoring besides the existence of it within the cluster spec struct + flags (though, flags are hidden, and there is no usage of the user-given value)

Example of create/edit:

```
❯ ./rosa create cluster --sts --mode auto --disable-workload-monitoring
Flag --disable-workload-monitoring has been deprecated, Disabling user workload monitoring (--disable-workload-monitoring) is deprecated and will be removed in a future version of ROSA CLI.
Please update your usage to avoid issues when this endpoint is removed.
I: Enabling interactive mode
? Cluster name: [? for help]
E: Expected a valid cluster name: interrupt

❯ ./rosa edit cluster --disable-workload-monitoring -c fake-cluster
Flag --disable-workload-monitoring has been deprecated, Disabling user workload monitoring (--disable-workload-monitoring) is deprecated and will be removed in a future version of ROSA CLI.
Please update your usage to avoid issues when this endpoint is removed.
E: Failed to get cluster 'fake-cluster': There is no cluster with identifier or name 'fake-cluster'
```

Describe:

```
❯ ./rosa describe cluster -c vkareh-iam

Name:                       vkareh-iam
Domain Prefix:              vkareh-iam
Display Name:               vkareh-iam
ID:                         XXX
External ID:                XXX
OpenShift Version:          4.17.4
DNS:                        XXX
AWS Account:                XXX
API URL:                    XXX
Console URL:                XXX
Region:                     ca-central-1

Nodes:
 - Control plane:           3
 - Infra:                   2
 - Compute:                 2
Network:
 - Type:                    OVNKubernetes
 - Service CIDR:            XXX
 - Machine CIDR:            XXX
 - Pod CIDR:                XXX
 - Host Prefix:             XXX
Infra ID:                   XXX
EC2 Metadata Http Tokens:   optional
State:                      ready
Delete Protection:          Disabled
Created:                    Nov 22 2024 18:01:45 UTC
Etcd Encryption:            Disabled
```

### DIFF
--- a/cmd/edit/cluster/cmd.go
+++ b/cmd/edit/cluster/cmd.go
@@ -26,6 +26,7 @@ import (
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 
+	"github.com/openshift/rosa/pkg/arguments"
 	"github.com/openshift/rosa/pkg/aws"
 	"github.com/openshift/rosa/pkg/clusterregistryconfig"
 	"github.com/openshift/rosa/pkg/fedramp"
@@ -39,6 +40,8 @@ import (
 )
 
 const enableDeleteProtectionFlagName = "enable-delete-protection"
+
+const disableWorkloadMonitoringFlagName = "disable-workload-monitoring"
 
 var args struct {
 	// Basic options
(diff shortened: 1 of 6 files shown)


hunterkepley:
Pipeline for conflux will be fixed in https://github.com/openshift/rosa/pull/2973

cmd/edit/cluster/cmd.go
```diff
@@ -230,14 +234,13 @@ func run(cmd *cobra.Command, _ []string) {
 	// Enable interactive mode if no flags have been set
 	if !interactive.Enabled() {
 		changedFlags := false
-		for _, flag := range []string{"expiration-time", "expiration", "private",
```
hunterkepley, line 233: Note for this; all I did was remove `disable-workload-monitoring`, each line change was due to reformatting each line manually

PR #2982, opened 2025-08-20
OCM-17631 | feat: Introduce capacity reservation to nodepools

hunterkepley:
Introduces Capacity Reservation ID to ROSA CLI in 3 ways;
1. Create/nodepool
2. Edit/nodepool
3. Describe/nodepool

Following the DDR specs for XCMStrat-757 milestone 1


pkg/machinepool/machinepool.go
```diff
@@ -764,6 +764,26 @@ func (m *machinePool) CreateNodePools(r *rosa.Runtime, cmd *cobra.Command, clust
...
+				machinepools.ValidateNodeDrainGracePeriod,
+			},
+		})
+		if err != nil {
+			return fmt.Errorf("Expected a valid value for Capacity Reservation ID: %s", err)
+		}
+	}
+
+	capacityReservation := cmv1.NewAWSCapacityReservation().Id(capacityReservationId)
+
```
BraeTroutman, line 786: Lucas had to point this same thing out to me in the CS code change: we can't support editing a machinepool to add a capacity reservation (because it will trigger a cycling of the nodes, which is undesirable)

So sorry I didn't catch that when you were asking about the cases earlier, I didn't see it listed in these 2 points in OCM-17631 so I thought it was documented properly, that's my bad:
```
    Create a machine pool in an existing HCP cluster that uses on-demand Capacity Reservation ID
    Create a machine pool in an existing HCP cluster that uses a Capacity Block for ML.
```

bchandra-ocp: Are we also missing market type or something required for Capacity Blocks for ML?

BraeTroutman: Right now, MarketType is inferred by OCM by retrieving the capacity reservation, which has an embedded field that specifies whether it is a capacity blocks for ML or On Demand reservation. When the customer retrieves their nodepool they can still see the type of the reservation as populated by OCM

bchandra-ocp: oh yea. We discussed this two weeks ago, sorry I forgot.

BraeTroutman: No worries! There's a lot to keep up with

PR #2983, opened 2025-08-21
OCM-1520 | feat: Add --hide-empty-columns flag to rosa list commands

olucasfreitas:
# Details

This PR improves the UX when listing resources by introducing a new `--hide-empty-columns` flag that automatically hides columns containing no data across all rows, making the output cleaner and more readable.

The flag has been added to the following `rosa list` commands:
- accessrequests
- accountroles
- addon
- breakglasscredential
- cluster
- dnsdomains
- gates
- iamserviceaccounts
- idp
- ingress
- instancetypes
- ocmroles
- oidcconfig
- operatorroles
- region
- service
- upgrade

**Note:** The following commands do NOT have this flag:
- `list machinepools` - Already has built-in auto-hide empty columns logic
- `list version`, `list userroles`, `list user`, `list rhRegion`, `list tuningconfigs`, `list oidcprovider`, `list kubeletconfig`, `list externalauthprovider` - These maintain their original behavior

---

## Default Behavior (All Columns Shown)

1. Run any supported list command without the flag:

    ```bash
    ./rosa list clusters
    ```

2. You should see **all columns** displayed, including empty ones:

    ```
    ID          NAME        STATE  TOPOLOGY  REGION      PROVIDER  VERSION
    1a2b3c4d    my-cluster  ready            us-east-1   aws       4.14.0
    5e6f7g8h    test-rosa   ready            us-west-2   aws       4.13.5
    ```

---

## Using --hide-empty-columns Flag

1. Run the same command with the flag:

    ```bash
    ./rosa list clusters --hide-empty-columns
    ```

2. Empty columns (like TOPOLOGY in this example) should be **automatically hidden**:

    ```
    ID          NAME        STATE  REGION      PROVIDER  VERSION
    1a2b3c4d    my-cluster  ready  us-east-1   aws       4.14.0
    5e6f7g8h    test-rosa   ready  us-west-2   aws       4.13.5
    ```

---
## New behavior of the '--all' 

1. The `--all` flag shows the 3 additional columns (AZ TYPE, WIN-LI ENABLED, DEDICATED HOST) but still hides empty columns:

    ```bash
    ./rosa list machinepools --cluster <cluster-name> --all
    ```

    Example output:
    ```
    ID       AUTOSCALING  REPLICAS  INSTANCE TYPE  SPOT INSTANCES  DISK SIZE  AZ TYPE   WIN-LI ENABLED  DEDICATED HOST
    workers  No           3         m5.xlarge      No              default    Standard  No              No
    ```

---

# Ticket

Closes [[OCM-1520](https://issues.redhat.com/browse/OCM-1520)]

### DIFF
--- a/cmd/list/accessrequests/cmd.go
+++ b/cmd/list/accessrequests/cmd.go
@@ -2,7 +2,6 @@ package accessrequests
 
 import (
 	"context"
-	"fmt"
 	"os"
 	"text/tabwriter"
 	"time"
@@ -38,6 +37,7 @@ func NewListAccessRequestsCommand() *cobra.Command {
 	}
 
 	output.AddFlag(cmd)
+	output.AddHideEmptyColumnsFlag(cmd)
 	ocm.AddOptionalClusterFlag(cmd)
 	return cmd
 }
@@ -67,12 +67,34 @@ func ListAccessRequestsRunner() rosa.CommandRunner {
 				}
 				return nil
 			}
+
+			headers := []string{"STATE", "ID", "CLUSTER ID", "UPDATED AT"}
+
+			var tableData [][]string
+			for _, accessRequest := range accessRequests {
+				row := []string{
+					string(accessRequest.Status().State()),
+					accessRequest.ID(),
+					accessRequest.ClusterId(),
+					accessRequest.UpdatedAt().UTC().Format(time.UnixDate),
+				}
+				tableData = append(tableData, row)
+			}
+
+			if output.ShouldHideEmptyColumns() {
+				tableData = output.RemoveEmptyColumns(headers, tableData)
+			} else {
+				tableData = append([][]string{headers}, tableData...)
+			}
+
 			writer := tabwriter.NewWriter(os.Stdout, 0, 4, 2, ' ', 0)
-			output, hasPending, pendingId := getAccessRequestsOutput(clusterId, accessRequests)
-			fmt.Fprint(writer, output)
+			output.BuildTable(writer, "\t", tableData)
 			if err := writer.Flush(); err != nil {
 				return err
 			}
+
+			hasPending, pendingId := checkForPendingRequests(clusterId, accessRequests)
+
 			if hasPending {
 				r.Reporter.Infof("Run the following command to approve or deny the Access Request:\n\n"+
 					"   rosa create decision --access-request %s --decision Approved\n"+
@@ -84,23 +106,22 @@ func ListAccessRequestsRunner() rosa.CommandRunner {
 	}
 }
 
-func getAccessRequestsOutput(clusterId string, accessRequests []*v1.AccessRequest) (string, bool, string) {
-	output := "STATE\tID\tCLUSTER ID\tUPDATED AT\n"
+func checkForPendingRequests(clusterId string, accessRequests []*v1.AccessRequest) (bool, string) {
 	hasPending := false
 	id := "<ID>"
 	for _, accessRequest := range accessRequests {
-		if accessRequest.Status().State() == v1.AccessRequestStatePending {
+		if accessRequest.Status().State() == v1.AccessRequestStatePending ||
+			accessRequest.Status().State() == v1.AccessRequestStateApproved {
 			hasPending = true
 			if clusterId != "" {
 				id = accessRequest.ID()
 			}
+			// Once we find the first pending/approved request for the cluster, we can break
+			// since we only need one ID for the suggestion message
+			if clusterId != "" && hasPending {
+				break
+			}
 		}
-		output += fmt.Sprintf("%s\t%s\t%s\t%s\n",
-			accessRequest.Status().State(),
-			accessRequest.ID(),
-			accessRequest.ClusterId(),
-			accessRequest.UpdatedAt().Format(time.UnixDate))
 	}
-
-	return output, hasPending, id
+	return hasPending, id
 }
--- a/cmd/list/accessrequests/cmd_test.go
+++ b/cmd/list/accessrequests/cmd_test.go
@@ -2,7 +2,6 @@ package accessrequests
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"time"
 
@@ -126,7 +125,6 @@ var _ = Describe("rosa attach policy", func() {
 			Expect(err).NotTo(HaveOccurred())
 
 			stdOut, _ := t.StdOutReader.Read()
-			fmt.Println(stdOut)
 			Expect(stdOut).To(Equal(accessRequestsOutput))
 		})
 
--- a/cmd/list/accountroles/cmd.go
+++ b/cmd/list/accountroles/cmd.go
@@ -17,7 +17,6 @@ limitations under the License.
 package accountroles
 
 import (
-	"fmt"
 	"os"
 	"text/tabwriter"
 	"time"
@@ -55,6 +54,7 @@ func init() {
 		"List only account-roles that are associated with the given version.",
 	)
 	output.AddFlag(Cmd)
+	output.AddHideEmptyColumnsFlag(Cmd)
 }
 
 func run(_ *cobra.Command, _ []string) {
@@ -108,23 +108,34 @@ func run(_ *cobra.Command, _ []string) {
 		os.Exit(0)
 	}
 
-	// Create the writer that will be used to print the tabulated results:
-	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
-	fmt.Fprintf(writer, "ROLE NAME\tROLE TYPE\tROLE ARN\tOPENSHIFT VERSION\tAWS Managed\n")
+	headers := []string{"ROLE NAME", "ROLE TYPE", "ROLE ARN", "OPENSHIFT VERSION", "AWS Managed"}
+
+	var tableData [][]string
 	for _, accountRole := range accountRoles {
 		awsManaged := "No"
 		if accountRole.ManagedPolicy {
 			awsManaged = "Yes"
 		}
-		fmt.Fprintf(
-			writer,
-			"%s\t%s\t%s\t%s\t%s\n",
+		tableData = append(tableData, []string{
 			accountRole.RoleName,
 			accountRole.RoleType,
 			accountRole.RoleARN,
 			accountRole.Version,
 			awsManaged,
-		)
+		})
 	}
-	writer.Flush()
+
+	if output.ShouldHideEmptyColumns() {
+		tableData = output.RemoveEmptyColumns(headers, tableData)
+	} else {
+		tableData = append([][]string{headers}, tableData...)
+	}
+
+	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
+	output.BuildTable(writer, "\t", tableData)
+	if err := writer.Flush(); err != nil {
+		_ = r.Reporter.Errorf("Failed to flush output: %v", err)
+		os.Exit(1)
+	}
+
 }
--- a/pkg/machinepool/machinepool.go
+++ b/pkg/machinepool/machinepool.go
@@ -1031,9 +1031,16 @@ func (m *machinePool) CreateNodePools(r *rosa.Runtime, cmd *cobra.Command, clust
 	return nil
 }
 
-// ListMachinePools lists all machinepools (or, nodepools if hypershift) in a cluster
+// createSeparators creates a separator array with single tabs for each column
+func createSeparators(count int) []string {
+	separators := make([]string, count)
+	for i := range separators {
+		separators[i] = "\t"
+	}
+	return separators
+}
+
 func (m *machinePool) ListMachinePools(r *rosa.Runtime, clusterKey string, cluster *cmv1.Cluster, args ListMachinePoolArgs) error {
-	// Load any existing machine pools for this cluster
 	r.Reporter.Debugf("Loading machine pools for cluster '%s'", clusterKey)
 	isHypershift := cluster.Hypershift().Enabled()
 	var err error
@@ -1058,19 +1065,49 @@ func (m *machinePool) ListMachinePools(r *rosa.Runtime, clusterKey string, clust
 		return output.Print(machinePools)
 	}
 
-	// Create the writer that will be used to print the tabulated results:
 	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
 
-	finalStringToOutput := getMachinePoolsString(r, machinePools, args)
+	var headers []string
+	var tableData [][]string
+
 	if isHypershift {
-		finalStringToOutput = getNodePoolsString(nodePools)
+		headers, tableData = getNodePoolsData(nodePools)
+	} else {
+		headers, tableData = getMachinePoolsData(r, machinePools, args)
 	}
-	fmt.Fprint(writer, finalStringToOutput)
-	writer.Flush()
+
+	var separators []string
+
+	if output.ShouldHideEmptyColumns() {
+		if !isHypershift {
+			specialColumns := make(map[int]bool)
+			for i, header := range headers {
+				if (header == "AZ TYPE" && (args.ShowAZType || args.ShowAll)) ||
+					(header == "WIN-LI ENABLED" && (args.ShowWindowsLI || args.ShowAll)) ||
+					(header == "DEDICATED HOST" && (args.ShowDedicated || args.ShowAll)) {
+					specialColumns[i] = true
+				}
+			}
+			headers, tableData = output.FilterColumnsWithConditions(headers, tableData, specialColumns)
+			tableData = append([][]string{headers}, tableData...)
+			separators = createSeparators(len(headers))
+		} else {
+			separators = createSeparators(len(headers))
+			tableData, separators = output.RemoveEmptyColumnsWithSeparators(headers, tableData, separators)
+		}
+	} else {
+		tableData = append([][]string{headers}, tableData...)
+		separators = createSeparators(len(headers))
+	}
+
+	output.BuildTableWithSeparators(writer, separators, tableData)
+
+	if err := writer.Flush(); err != nil {
+		return err
+	}
 	return nil
 }
 
-// DescribeMachinePool describes either a machinepool, or, a nodepool (if hypershift)
 func (m *machinePool) DescribeMachinePool(r *rosa.Runtime, cluster *cmv1.Cluster, clusterKey string,
 	machinePoolId string) error {
 	if cluster.Hypershift().Enabled() {
@@ -1235,11 +1272,11 @@ func appendUpgradesIfExist(scheduledUpgrade *cmv1.NodePoolUpgradePolicy, output 
 	return output
 }
 
-func getMachinePoolsString(
+func getMachinePoolsData(
 	runtime *rosa.Runtime,
 	machinePools []*cmv1.MachinePool,
 	args ListMachinePoolArgs,
-) string {
+) ([]string, [][]string) {
 	type columnDefinition struct {
 		header      string
 		isVisible   bool
@@ -1267,63 +1304,41 @@ func getMachinePoolsString(
 		{"DEDICATED HOST", args.ShowDedicated || args.ShowAll, func(mp *cmv1.MachinePool) string { return isDedicatedHost(mp, runtime) }},
 	}
 
-	var visibleColumnHeaders []string
-	var visibleColumnData [][]string
-	numPools := len(machinePools)
-
+	var headers []string
+	var tableData [][]string
 	for _, column := range allColumnDefinitions {
-		if !column.isVisible {
-			continue
+		if column.isVisible {
+			headers = append(headers, column.header)
 		}
+	}
 
-		columnValues := make([]string, numPools)
-		hasNonEmptyValue := false
-
-		for i, pool := range machinePools {
-			if pool == nil {
-				columnValues[i] = "-"
+	for _, pool := range machinePools {
+		var row []string
+		for _, column := range allColumnDefinitions {
+			if !column.isVisible {
 				continue
 			}
 
-			value := column.extractData(pool)
-			columnValues[i] = value
-			if value != "" && value != "-" {
-				hasNonEmptyValue = true
+			if pool == nil {
+				row = append(row, "")
+			} else {
+				value := column.extractData(pool)
+				row = append(row, value)
 			}
 		}
-
-		if args.ShowAll || hasNonEmptyValue || numPools == 0 {
-			visibleColumnHeaders = append(visibleColumnHeaders, column.header)
-			visibleColumnData = append(visibleColumnData, columnValues)
-		}
+		tableData = append(tableData, row)
 	}
 
-	var tableBuilder strings.Builder
-
-	// Write header
-	if len(visibleColumnHeaders) > 0 {
-		tableBuilder.WriteString(strings.Join(visibleColumnHeaders, "\t") + "\n")
-	}
-
-	// Write data rows
-	for rowIndex := range numPools {
-		for colIndex, columnValues := range visibleColumnData {
-			if colIndex > 0 {
-				tableBuilder.WriteString("\t")
-			}
-			tableBuilder.WriteString(columnValues[rowIndex])
-		}
-		tableBuilder.WriteString("\n")
-	}
-
-	return tableBuilder.String()
+	return headers, tableData
 }
 
-func getNodePoolsString(nodePools []*cmv1.NodePool) string {
-	outputString := "ID\tAUTOSCALING\tREPLICAS\t" +
-		"INSTANCE TYPE\tLABELS\t\tTAINTS\t\tAVAILABILITY ZONE\tSUBNET\tDISK SIZE\tVERSION\tAUTOREPAIR\t\n"
+func getNodePoolsData(nodePools []*cmv1.NodePool) ([]string, [][]string) {
+	headers := []string{"ID", "AUTOSCALING", "REPLICAS", "INSTANCE TYPE", "LABELS", "TAINTS",
+		"AVAILABILITY ZONE", "SUBNET", "DISK SIZE", "VERSION", "AUTOREPAIR"}
+
+	var tableData [][]string
 	for _, nodePool := range nodePools {
-		outputString += fmt.Sprintf("%s\t%s\t%s\t%s\t%s\t\t%s\t\t%s\t%s\t%s\t%s\t%s\t\n",
+		row := []string{
 			nodePool.ID(),
 			ocmOutput.PrintNodePoolAutoscaling(nodePool.Autoscaling()),
 			ocmOutput.PrintNodePoolReplicasShort(
@@ -1338,9 +1353,11 @@ func getNodePoolsString(nodePools []*cmv1.NodePool) string {
 			ocmOutput.PrintNodePoolDiskSize(nodePool.AWSNodePool()),
 			ocmOutput.PrintNodePoolVersion(nodePool.Version()),
 			ocmOutput.PrintNodePoolAutorepair(nodePool.AutoRepair()),
-		)
+		}
+		tableData = append(tableData, row)
 	}
-	return outputString
+
+	return headers, tableData
 }
 
 func (m *machinePool) EditMachinePool(cmd *cobra.Command, machinePoolId string, clusterKey string,
--- a/pkg/machinepool/machinepool_test.go
+++ b/pkg/machinepool/machinepool_test.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"reflect"
+	"strings"
 	"time"
 
 	"go.uber.org/mock/gomock"
@@ -34,6 +35,54 @@ import (
 
 var policyBuilder cmv1.NodePoolUpgradePolicyBuilder
 var date time.Time
+
+func formatTableString(headers []string, tableData [][]string) string {
+	return formatTableStringWithSeparators(headers, tableData, "\t")
+}
+
+func formatTableStringWithSeparators(headers []string, tableData [][]string, separatorSpec string) string {
+	var result strings.Builder
+
+	var separators []string
+	if strings.Contains(separatorSpec, ",") {
+		separators = strings.Split(separatorSpec, ",")
+		maxCols := len(headers)
+		if len(separators) < maxCols {
+			padSeparator := "\t"
+			if len(separators) > 0 {
+				padSeparator = separators[len(separators)-1]
+			}
+			for len(separators) < maxCols {
+				separators = append(separators, padSeparator)
+			}
+		}
+	} else {
+		separators = make([]string, len(headers))
+		for i := range separators {
+			separators[i] = separatorSpec
+		}
+	}
+
+	for i, header := range headers {
+		result.WriteString(header)
+		if i < len(headers)-1 && i < len(separators) {
+			result.WriteString(separators[i])
+		}
+	}
+	result.WriteString("\n")
+
+	for _, row := range tableData {
+		for i, col := range row {
+			result.WriteString(col)
+			if i < len(row)-1 && i < len(separators) {
+				result.WriteString(separators[i])
+			}
+		}
+		result.WriteString("\n")
+	}
+
+	return result.String()
+}
 
 var _ = Describe("Machinepool and nodepool", func() {
 	var (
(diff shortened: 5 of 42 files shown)


cmd/list/accessrequests/cmd.go
```diff
@@ -67,40 +69,57 @@ func ListAccessRequestsRunner() rosa.CommandRunner {
...
+			pendingId := "<ID>"
+			for _, accessRequest := range accessRequests {
+				if accessRequest.Status().State() == v1.AccessRequestStatePending ||
+					accessRequest.Status().State() == v1.AccessRequestStateApproved {
+					hasPending = true
+					if clusterId != "" {
+						pendingId = accessRequest.ID()
+					}
+				}
 			}
```
hunterkepley: Is there a specific reason to replace the output function with a new for loop?

cmd/list/accessrequests/cmd.go
```diff
@@ -67,40 +69,57 @@ func ListAccessRequestsRunner() rosa.CommandRunner {
...
 
-func getAccessRequestsOutput(clusterId string, accessRequests []*v1.AccessRequest) (string, bool, string) {
-	output := "STATE\tID\tCLUSTER ID\tUPDATED AT\n"
-	hasPending := false
-	id := "<ID>"
-	for _, accessRequest := range accessRequests {
-		if accessRequest.Status().State() == v1.AccessRequestStatePending {
-			hasPending = true
-			if clusterId != "" {
-				id = accessRequest.ID()
```
hunterkepley, line 95: I think we can keep this code in a separate function, moving the code to the function above increases the complexity of that function

cmd/list/accessrequests/cmd.go
```diff
@@ -67,40 +69,57 @@ func ListAccessRequestsRunner() rosa.CommandRunner {
...
+				}
+				output.PrintTable(writer, newHeaders, newData, config)
+			} else {
+				fmt.Fprint(writer, "STATE\tID\tCLUSTER ID\tUPDATED AT\n")
+				for _, row := range tableData {
+					fmt.Fprintf(writer, "%s\n", strings.Join(row, "\t"))
+				}
+			}
+
+			writer.Flush()
```
hunterkepley: we should check for error on flush like the original code did

cmd/list/accessrequests/cmd.go
```diff
@@ -67,40 +69,57 @@ func ListAccessRequestsRunner() rosa.CommandRunner {
...
+			if output.ShouldHideEmptyColumns() {
+				newHeaders, newData := output.RemoveEmptyColumns(headers, tableData)
+				config := output.TableConfig{
+					Separator:            "\t",
+					HasTrailingSeparator: false,
+					UseFprintln:          false,
+				}
+				output.PrintTable(writer, newHeaders, newData, config)
+			} else {
+				fmt.Fprint(writer, "STATE\tID\tCLUSTER ID\tUPDATED AT\n")
```
hunterkepley: We've already stored the headers above

cmd/list/accessrequests/cmd_test.go
```diff
@@ -18,23 +17,23 @@ import (
...
    rosa create decision --access-request <ID> --decision Denied --justification "justification"
 
+STATE     ID         CLUSTER ID         UPDATED AT
+Approved  mock-id-1  mock-cluster-id-1  Sat Sep 28 20:35:00 UTC 2024
+Pending   mock-id-2  mock-cluster-id-2  Sat Sep 28 20:35:00 UTC 2024
 `
-	clusterAccessRequestsOutput = `STATE    ID         CLUSTER ID       UPDATED AT
-Pending  mock-id-1  mock-cluster-id  Sat Sep 28 21:35:00 UTC 2024
-Denied   mock-id-2  mock-cluster-id  Sat Sep 28 20:35:00 UTC 2024
-INFO: Run the following command to approve or deny the Access Request:
```
hunterkepley, line 33: I think it would be best to not affect command output through this feature beyond hiding empty columns

hunterkepley: The reason this function's output changed is because the command was edited to reverse the two prints (table and info msg about `create decision`)

cmd/list/accountroles/cmd.go
```diff
@@ -108,23 +110,39 @@ func run(_ *cobra.Command, _ []string) {
...
+			Separator:            "\t",
+			HasTrailingSeparator: false,
+			UseFprintln:          false,
+		}
+		output.PrintTable(writer, newHeaders, newData, config)
+	} else {
+		fmt.Fprintf(writer, "ROLE NAME\tROLE TYPE\tROLE ARN\tOPENSHIFT VERSION\tAWS Managed\n")
+		for _, row := range tableData {
+			fmt.Fprintf(writer, "%s\n", strings.Join(row, "\t"))
+		}
```
hunterkepley: If you prepare the headers the same way outside of the if statement, and store the table data in a var outside of this if statement, you can reduce the code here by a lot and remove the `else` statement, as well as the `PrintTable` function entirely

hunterkepley: This goes for every command

hunterkepley: You can always split the headers definition up by removing all `\t`'s and splitting the string for use with the empty column preprocessing

pkg/machinepool/machinepool_test.go
```diff
@@ -385,8 +385,8 @@ var _ = Describe("Machinepool and nodepool", func() {
 			args := ListMachinePoolArgs{ShowAll: true}
 			out := getMachinePoolsString(r, cluster.MachinePools().Slice(), args)
 
-			expectedOutput := "ID\tAUTOSCALING\tREPLICAS\tINSTANCE TYPE\tLABELS\tTAINTS\tAVAILABILITY ZONES\tSUBNETS\tSPOT INSTANCES\tDISK SIZE\tSG IDS\tAZ TYPE\tWIN-LI ENABLED\tDEDICATED HOST\n" +
-				"mp-1\tNo\t3\tm5.large\t\t\t\t\tNo\tdefault\t\tN/A\tNo\tNo\n"
```
hunterkepley, line 389: I'm a little confused, the test name is `Test printMachinePools with showAll flag` - shouldn't it show all 3 columns instead of hiding them?

pkg/machinepool/machinepool.go
```diff
@@ -1292,7 +1291,7 @@ func getMachinePoolsString(
 			}
 		}
 
-		if args.ShowAll || hasNonEmptyValue || numPools == 0 {
```
hunterkepley, line 1318: I think there's still a misunderstanding on this,

```
Yeah I think it would be best to keep machinepools the same as the others in terms of functionality. On the MR itself, I moreso meant I think it would be best to keep the --all flag, but let it be used for it's described purpose rather than showing/hiding empty columns as well
```

From a recent DM I sent you a week or 2 back ^

The functionality of `--all` _should not change_. Its purpose to is to show the automatically-hidden columns from the feature which introduced them. This is unrelated entirely to this feature

This feature's goal is to hide columns which are empty, when a flag is given. This is unrelated to those special columns which are _already hidden even if they have values_ unless flags are supplied

`--all` should show those special columns, even if the hide empty columns flag is supplied and they are empty

`--hide-empty-columns` should hide any other column which is empty but visible


```
Show all additional information for each machine pool (equivalent to --az-type --dedicated-host --win-li)
```

The description of the `--all` flag that you made is correct, Im not sure I understand why you're confused about this, since you made the description + functionality. But, it is equivalent to running those 3 flags mentioned right there. So, without `--all`, or those 3 flags, those three columns are already hidden

This means if we use `--all`, we **want** to see those 3 columns. Even if empty. Otherwise it's confusing to the end user

Overall, `--all` should show those 3 columns, `--hide-empty-columns` should hide any **other** column which is empty

pkg/machinepool/machinepool_test.go
hunterkepley: Also we should update the test name no matter what to have `--all` instead of `showAll` if you don't mind 🙂

cmd/list/accessrequests/cmd.go
```diff
@@ -38,10 +39,32 @@ func NewListAccessRequestsCommand() *cobra.Command {
...
+	output.AddHideEmptyColumnsFlag(cmd)
 	ocm.AddOptionalClusterFlag(cmd)
 	return cmd
 }
 
+func getAccessRequestsOutput(clusterId string, accessRequests []*v1.AccessRequest) (string, bool, string) {
+	output := "STATE\tID\tCLUSTER ID\tUPDATED AT\n"
+	hasPending := false
+	id := "<ID>"
+	for _, accessRequest := range accessRequests {
```
hunterkepley: Could we move this function below the runner (just for consistency with other commands, and the pre-existing structure 🙂 )

cmd/list/accessrequests/cmd.go
```diff
@@ -67,12 +90,36 @@ func ListAccessRequestsRunner() rosa.CommandRunner {
...
+			fmt.Fprintf(writer, "%s\n", strings.Join(headers, "\t"))
+			for _, row := range tableData {
+				fmt.Fprintf(writer, "%s\n", strings.Join(row, "\t"))
+			}
+
 			if err := writer.Flush(); err != nil {
 				return err
 			}
+
+			_, hasPending, pendingId := getAccessRequestsOutput(clusterId, accessRequests)
```
hunterkepley: Why wouldn't we use the output like the original file did here?

cmd/list/accountroles/cmd.go
```diff
@@ -108,23 +110,37 @@ func run(_ *cobra.Command, _ []string) {
...
+
+	if output.ShouldHideEmptyColumns() {
+		headers, tableData = output.RemoveEmptyColumns(headers, tableData)
+	}
+
+	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
+	fmt.Fprintf(writer, "%s\n", strings.Join(headers, "\t"))
+	for _, row := range tableData {
+		fmt.Fprintf(writer, "%s\n", strings.Join(row, "\t"))
+	}
```
hunterkepley: I think this could be simplified a liiiittle bit more;

Since all that is done after outputting the headers and tableData is adding them to the write here, why don't we combine them in the `RemoveEmptyColumns` function into a single `tableData` variable?


Then, you could make a function in the same file as `RemoveEmptyColumns` that takes in a tabwriter, the format string (such as `\t%s\t\t%s\t%s\n`), and the table data, and returns the tabwriter with this data added (maybe it doesn't need to be returned? I think tab writers would be passed by reference)

So the final output in each command could look like:

```
formatString = "%s\t%s\t\t\t%s\t\t%s\t\n"
writer := tabwriter.NewWriter(...)

if output.ShouldHideEmptyColumns() {
    tableData = output.RemoveEmptyColumns // We don't need to attach headers since they are attached within `RemoveEmptyColumns`
} else {
    fmt.Fprintf(writer, "format_string\n", headers...) // Attach headers separately when `RemoveEmptyColumns` is not called
}


err := output.BuildTable(writer, formatString, tableData).Flush() 
// If returning the writer from the function. Please check if this can be done WITHOUT returning the writer!! 
//Below is an example of how we could check if writer needs to be retuned (whether or not it copies it over when calling the function, or passes the reference)

// without returning writer
writer = output.BuildTableFromTableData(writer, formatString, tableData)
if err := writer.Flush(); err != nil {
    ...
    os.Exit(1)
}
```

cmd/list/accountroles/cmd.go
```diff
@@ -108,23 +110,37 @@ func run(_ *cobra.Command, _ []string) {
...
+		}
+		tableData = append(tableData, row)
+	}
+
+	if output.ShouldHideEmptyColumns() {
+		headers, tableData = output.RemoveEmptyColumns(headers, tableData)
+	}
+
+	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
+	fmt.Fprintf(writer, "%s\n", strings.Join(headers, "\t"))
```
hunterkepley: Go supports `...` when calling functions, which is basically a way to unwrap slices within function calls. Usually it's used for arguments, or string formatting, like we are doing here

 For example, this would be better written as:

```
formatString := "%s\t%s\t%s\t%s\t%s\n"
fmt.Fprintf(writer, formatString, headers...)
```

Or for the for loops you have for `tableData`, you could do the same thing with `fmt.Fprintf(writer, formatString, row...)`

hunterkepley (commented):
Some more comments, looking good 🙂 Noticed a couple of improvements which could be made to lessen the LoC committed for this effort

cmd/list/accessrequests/cmd.go
olucasfreitas: We're no longer using the output string from getAccessRequestsOutput because we've refactored the code to use the dynamic headers approach with output.BuildTable()

cmd/list/accessrequests/cmd_test.go
```diff
@@ -126,7 +125,6 @@ var _ = Describe("rosa attach policy", func() {
 			Expect(err).NotTo(HaveOccurred())
 
 			stdOut, _ := t.StdOutReader.Read()
-			fmt.Println(stdOut)
```
hunterkepley, line 129: unnecessary change

cmd/list/accessrequests/cmd.go
```diff
@@ -84,23 +106,22 @@ func ListAccessRequestsRunner() rosa.CommandRunner {
...
 		}
-		output += fmt.Sprintf("%s\t%s\t%s\t%s\n",
-			accessRequest.Status().State(),
-			accessRequest.ID(),
-			accessRequest.ClusterId(),
-			accessRequest.UpdatedAt().Format(time.UnixDate))
 	}
-
-	return output, hasPending, id
+	return hasPending, id
```
hunterkepley, line 126: This command has been altered pretty heavily in terms of how it gathers data and im worried it will have different a different table for certain clusters after these changes. Could you provide some output from it to show that it works the same?

cmd/list/accountroles/cmd.go
```diff
@@ -108,23 +108,35 @@ func run(_ *cobra.Command, _ []string) {
...
-			"%s\t%s\t%s\t%s\t%s\n",
+		row := []string{
 			accountRole.RoleName,
 			accountRole.RoleType,
 			accountRole.RoleARN,
 			accountRole.Version,
 			awsManaged,
-		)
+		}
+		tableData = append(tableData, row)
```
hunterkepley: Can be simplified:

```
tableData = append(tableData, []string{
    accountRole.RoleName,
    ...
    ...
})
```

cmd/list/accountroles/cmd.go
```diff
@@ -108,23 +108,35 @@ func run(_ *cobra.Command, _ []string) {
...
+		tableData = append([][]string{headers}, tableData...)
 	}
-	writer.Flush()
+
+	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
+	output.BuildTable(writer, "\t", tableData)
+	if err := writer.Flush(); err != nil {
+		_ = r.Reporter.Errorf("Failed to flush output: %v", err)
+		os.Exit(1)
+	}
```
hunterkepley, line 139: This file looks great

pkg/machinepool/machinepool.go
```diff
@@ -1338,9 +1343,49 @@ func getNodePoolsString(nodePools []*cmv1.NodePool) string {
...
+func removeEmptyColumnsExceptSpecial(headers []string, tableData [][]string, specialColumns map[int]bool) ([]string, [][]string) {
+	var newHeaders []string
+	var columnsToKeep []int
+
+	for i, header := range headers {
+		// Keep special columns regardless of whether they're empty
+		if specialColumns[i] {
+			newHeaders = append(newHeaders, header)
+			columnsToKeep = append(columnsToKeep, i)
+			continue
```
hunterkepley: We should be able to come up with a solution which can take in special cases, rather than this.

Also the name of this function is not clear

pkg/output/table_filter.go
```diff
@@ -0,0 +1,88 @@
...
+// buildFormatString creates a format string with the appropriate number of %s placeholders
+func buildFormatString(columnCount int, separator string) string {
+	if columnCount == 0 {
+		return "\n"
+	}
+
+	format := ""
+	for i := range columnCount {
+		format += "%s"
+		if i < columnCount-1 {
```
hunterkepley: I think we could simplify this somewhat

```
finalString := ""
for i, row := range c {
		finalString += strings.Join(row, "\t")
         if i < len(c) {
             finalString+="\n"
    }
}
```

This accomplishes joining rows using a separator `\t`

Now, if we want to use a custom separator (such as `col1\t\tcol2\t\tcol3\tcol4\tcol5\n` which has `\t\t` and `\t`) to avoid making an edge case function like you have, we can modify this slightly

```
separators := ["\t\t", "\t\t", "\t", "\t"] // Pass in rather than `separator` string
// To make it simple, you can edit this slice in `RemoveEmptyColumns` like you do with the tableData

// Combine rows with separator simply, using only a single nested for loop
for i, row := range c {
		for j, col := range row {
			finalStr += col + separators[j]
		}
		if i < len(c) {
			finalStr += "\n"
		}
}
```

hunterkepley: We need to make sure the separators list is the same length as each row. Probably could also add an if statement in the loop to break if it is >= the length of the separator list

pkg/output/table_filter.go
```diff
@@ -0,0 +1,88 @@
...
+
+	var result [][]string
+	result = append(result, newHeaders)
+
+	for _, row := range tableData {
+		var newRow []string
+		for _, colIdx := range columnsToKeep {
+			if colIdx < len(row) {
+				newRow = append(newRow, row[colIdx])
+			} else {
```
hunterkepley, line 113: `colIdx` -> `col` (lets keep naming consistent..)

hunterkepley (commented):
Some suggestions on the implementation of `BuildTable` as well as the edge case

cmd/list/accessrequests/cmd.go
hunterkepley: If this is the only place that function is used, we should rename the function + remove that return, `get...Output` doesn't sound like a `print` or `write` function

I was thinking we could do what it did originally, and return the data, then print below the function call after it is returned. This shouldn't be affected by the refactor from what I can tell

cmd/list/accessrequests/cmd.go
olucasfreitas: here goes a screenshot  of how the behavior is 
ar1

cmd/list/accountroles/cmd.go
olucasfreitas: fixed in the commit

cmd/list/accountroles/cmd.go
olucasfreitas: thanks :)

pkg/machinepool/machinepool.go
olucasfreitas: fixed in the last commit

pkg/output/table_filter.go
olucasfreitas: implemented this on the last commit

pkg/output/table_filter.go
olucasfreitas: fixed in last commit

cmd/list/accessrequests/cmd.go
olucasfreitas: fixed on the last commit

pkg/output/table_filter.go
```diff
@@ -0,0 +1,159 @@
...
+
+	for rowIdx, row := range tableData {
+		if len(row) == 0 {
+			continue
+		}
+
+		for colIdx, col := range row {
+			output += col
+			// Add separator after column if not the last column
+			if colIdx < len(row)-1 && colIdx < len(separators) {
```
hunterkepley: This function looks much better, one suggestion is maybe we don't need the first part of this if statement (please test some hidden columns to make sure im right!)

My thinking is that if the separator list and each row are parallel (same number of items in each slice) then there should be no issues using that alone

hunterkepley: Otherwise this is looking great now

pkg/output/table_filter.go
```diff
@@ -0,0 +1,159 @@
...
+		// Add newline if not the last row
+		if rowIdx < len(tableData)-1 {
+			output += "\n"
+		}
+	}
+
+	// Add final newline
+	if len(tableData) > 0 {
+		output += "\n"
+	}
```
hunterkepley: One more comment for this function; I see that we add a newline if it is not the last row, but we also add it for the final row at the bottom, right?

I might be wrong, please correct me if I am, but I think we can just add one to every single row based on what I am seeing here

pkg/output/table_filter.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// BuildTableWithSeparators writes table data using a list of separators
+func BuildTableWithSeparators(writer *tabwriter.Writer, separators []string, tableData [][]string) {
+	var output string
+
+	for rowIdx, row := range tableData {
+		if len(row) == 0 {
+			continue
+		}
```
hunterkepley: I'm curious, is there a case where we would expect the length of a row to be 0 in the middle of a table?

pkg/output/table_filter.go
```diff
@@ -0,0 +1,159 @@
...
+			maxCols = len(row)
+		}
+	}
+
+	separators := make([]string, maxCols)
+	for i := range separators {
+		separators[i] = separator
+	}
+
+	BuildTableWithSeparators(writer, separators, tableData)
```
hunterkepley: https://github.com/openshift/rosa/pull/2983#discussion_r2325290550

In this comment I talked about having support for separators which have different separators at different parts of the table. For example, `val1\tval2\t\tval3`

It seems like you have support for it now, but did not actually support it, instead wrapping it in a function which makes a copy of a single separator that was passed in

If we could make that functional, and get the original separators back in the command which have unique separator values per column, that would be great

hunterkepley: For example, `list machinepools`

hunterkepley (commented):
some more comments, will do a more in-depth review after these are addressed since it may change up other things im worried about

hunterkepley:
@olucasfreitas 
https://github.com/openshift/rosa/pull/2983#discussion_r2342191100

What is `MOCK MODE`? I don't see it in this PR or in the CLI? 😄

olucasfreitas:
@hunterkepley the mock mode was something I added only for testing, to activate the mock responses on the command

olucasfreitas:
@hunterkepley was actually pretty simple, not an ENTIRE mock mode as you maybe thought

pkg/output/table_filter.go
olucasfreitas: Improved this on the last commit

pkg/output/table_filter.go
olucasfreitas: you caught something here, fixed it on the last commit

pkg/output/table_filter.go
olucasfreitas: tested and saw some improvements there, the behavior was buggy, so thanks for the suggestion

pkg/output/table_filter.go
olucasfreitas: for what i understood here, empty rows can occur in real-world data

hunterkepley: Could you provide an example?

pkg/output/table_filter.go
hunterkepley: Could you describe what you mean?

pkg/machinepool/machinepool_test.go
```diff
@@ -189,16 +207,16 @@ var _ = Describe("Machinepool and nodepool", func() {
...
 			Expect(err).ToNot(HaveOccurred())
-			out := getNodePoolsString(cluster.NodePools().Slice())
-			Expect(err).ToNot(HaveOccurred())
-			Expect(out).To(Equal(fmt.Sprintf("ID\tAUTOSCALING\tREPLICAS\t"+
-				"INSTANCE TYPE\tLABELS\t\tTAINTS\t\tAVAILABILITY ZONE\tSUBNET\tDISK SIZE\tVERSION\tAUTOREPAIR\t\n"+
-				"%s\t%s\t%s\t%s\t%s\t\t%s\t\t%s\t%s\t%s\t%s\t%s\t\n",
+			headers, tableData := getNodePoolsData(cluster.NodePools().Slice())
+			out := formatTableString(headers, tableData)
+
+			expectedOutput := fmt.Sprintf("ID\tAUTOSCALING\tREPLICAS\tINSTANCE TYPE\tLABELS\tTAINTS\tAVAILABILITY ZONE\tSUBNET\tDISK SIZE\tVERSION\tAUTOREPAIR\n"+
```
hunterkepley: You are still not taking into account different separators within a single table

pkg/output/table_filter.go
```diff
@@ -0,0 +1,176 @@
...
+func CheckIfColumnIsEmpty(columnIdx int, tableData [][]string) bool {
+	for _, row := range tableData {
+		if columnIdx < len(row) && row[columnIdx] != "" {
+			return false
+		}
+	}
+	return true
+}
+
+type ColumnFilterFunc func(columnIdx int) bool
```
hunterkepley: Why was this introduced?

pkg/output/table_filter.go
hunterkepley: @olucasfreitas im afraid this was virtually unchanged from the last commit

pkg/machinepool/machinepool.go
```diff
@@ -1061,12 +1061,39 @@ func (m *machinePool) ListMachinePools(r *rosa.Runtime, clusterKey string, clust
 	// Create the writer that will be used to print the tabulated results:
 	writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
 
-	finalStringToOutput := getMachinePoolsString(r, machinePools, args)
+	// Get headers and data
+	headers, tableData := getMachinePoolsData(r, machinePools, args)
 	if isHypershift {
-		finalStringToOutput = getNodePoolsString(nodePools)
+		headers, tableData = getNodePoolsData(nodePools)
+	}
```
hunterkepley: In the original functions (`getMachinePoolsString` and `getNodePoolsString`), we return a string such as

`test1\ttest2\ttest3\nval1\tval2\tval3\n`

Couldn't we convert this into a list and not change the original functions?

hunterkepley (commented):
seems there were a missed comment or 2 from the last round of review, id like to address those first before going to deep on another (but I did add some comments)

pkg/output/table_filter.go
olucasfreitas: When there were fewer separators than columns, for example 2 separators for 4 columns, the && condition would fail for columns 2 and 3, causing no separator at all to be added

pkg/output/table_filter.go
olucasfreitas: Data Import with Blank Lines
```go
// CSV parsing or API responses might produce:
tableData := [][]string{
    {"ID", "Name", "Status"},
    {"1", "Alice", "Active"},
    {},  // Blank line in source data
    {"2", "Bob", "Pending"},
}
```
Important Distinction
```go
// These are different:
emptyRow := []string{}        // Length 0 - gets skipped ✓
blankRow := []string{"", "", ""}  // Length 3 - displays with spacing
```

pkg/output/table_filter.go
olucasfreitas: This was created to support the new FilterColumns function, which needed a way to let callers specify custom logic for which columns to keep or remove. I did this for the flexibility, reusability and composability

pkg/machinepool/machinepool.go
olucasfreitas: Worked on this on this commit https://github.com/openshift/rosa/pull/2983/commits/3f5390f0fcb7d52c619f600aa27320e2dfe4f7c0

pkg/machinepool/machinepool_test.go
olucasfreitas: Worked on this on this commit https://github.com/openshift/rosa/pull/2983/commits/3f5390f0fcb7d52c619f600aa27320e2dfe4f7c0

pkg/output/table_filter.go
olucasfreitas: Worked on this on this commit https://github.com/openshift/rosa/pull/2983/commits/3f5390f0fcb7d52c619f600aa27320e2dfe4f7c0

pkg/output/table_filter.go
hunterkepley: Hey Lucas, thanks for explaining the logic here;

I was more-so asking for a 'real world' example, such as one coming from the ROSA CLI. My thoughts are, that this may not be possible for our tables in the ROSA CLI, just by design. We add data generally when data exists, so there should be at least _one_ value in each row (for loops over resource lists add data by nature causes this)

BUT- im asking for examples because I want to see if I might be wrong here, and if there's a list command which can produce empty rows

hunterkepley: Also we do not use CSVs, why is that in your example?

hunterkepley: Please use your own words

hunterkepley: For example, if we list `access requests` - we use the API to fetch a list of the cluster's access requests. Then we loop through those

This pattern is pretty common across all list commands, where we fetch data before listing, since that data is generally not available locally on the user's machine

Hopefully that makes senes, LMK if you have questions

pkg/output/table_filter.go
hunterkepley: Cool yeah that makes sense,

I did have a question about this:

```go
	// Use FilterColumns with a predicate that keeps non-empty columns
	newHeaders, newTableData := FilterColumns(headers, tableData, func(columnIdx int) bool {
		return !CheckIfColumnIsEmpty(columnIdx, tableData)
	})
```

I see that this is used twice, to filter columns based on `!CheckIfColumnIsEmpty`

I think this is greatly useful if we had different types of filtering to be done, but in our case here I see we have only `CheckIfColumnIsEmpty`

My thoughts are, that we can get rid of the type here, and instead of passing a function into `FilterColumns`, we can directly call `CheckIfColumnIsEmpty`. WDYT?

OH- another note;

The current design has it so `tableData` is passed into _**both**_ `FilterColumns` _and_ the function being passed in.

```go
	// Use FilterColumns with a predicate that keeps non-empty columns
	newHeaders, newTableData := FilterColumns(headers, tableData, func(columnIdx int) bool {
		return !CheckIfColumnIsEmpty(columnIdx, tableData)
	})
```

Looking at this again, we can see that. Sadly, this means this data (no matter how large or small) is tripled in memory every time this is called (once for the first duplication to pass into the function, and a second one for the `CheckIfColumnIsEmpty` func

hunterkepley: While the memory part may not seem like a big deal at first, we have to think about the fact that some tables can be very long. And we have the ability to control the pagination with the CLI commands usually

pkg/output/table_filter.go
olucasfreitas: Oh, now I got it, I'm gonna run through some commands a send you the data here

olucasfreitas: So, I ran through the commands, and I stand corrected, no cases in where this can happen showed up for me, thanks for the explanation here, I addressed the changes on this commit [e397f65](https://github.com/openshift/rosa/pull/2983/commits/e397f65b841261ccf68eb7ec8235e9ff9d135f01)

pkg/output/table_filter.go
olucasfreitas: > My thoughts are, that we can get rid of the type here, and instead of passing a function into FilterColumns, we can directly call CheckIfColumnIsEmpty. WDYT?

I do agree with this, makes this much simpler to understand and the implementation cleaner too

> While the memory part may not seem like a big deal at first, we have to think about the fact that some tables can be very long. And we have the ability to control the pagination with the CLI commands usually

Did not think of this, thanks for the explanation, will take a look into to fixing it on a new commit

olucasfreitas: Hopefully addressed on [75d64ee](https://github.com/openshift/rosa/pull/2983/commits/75d64ee84dd1f1e802d4fffd2d1750454fd488ce)

pkg/output/table_filter.go
hunterkepley: thanks 🙂

pkg/machinepool/machinepool.go
```diff
@@ -1290,63 +1341,58 @@ func getMachinePoolsString(
...
-			visibleColumnData = append(visibleColumnData, columnValues)
-		}
+		tableData = append(tableData, row)
 	}
 
-	var tableBuilder strings.Builder
+	return headers, tableData
+}
 
-	// Write header
```
hunterkepley, line 1326: Hey I think we need to include unique separators for the nodepool output- you can see here that the original separator list has a mix of `\t\t` and `\t`

It seems support was added for it in the last commit but it was not implemented in the actual command yet (still using a single separator `\t` for this command for example)

olucasfreitas: @hunterkepley adress this on the new commit, please take a look when you can

hunterkepley: @olucasfreitas Hey it looks like this wasn't changed- the `BuildTableWithSeparators` function is not called in the machinepool command still

hunterkepley: it looks like it's only used in tests

olucasfreitas: Sorry, I forgot to push the rest of the changes, please check it again

PR #3013, opened 2025-09-08
OCM-18340 | fix: Fix delete interactive + auto modes

hunterkepley:
This MR is not the prettiest, it was a lot of refactoring where the prompts lived and when they were called. Output from testing is below, which matches the behavior desired

* No flags provided asks whether or not user wants to provide explicit role name:

```bash
❯ ./rosa delete iamserviceaccount -c hkepley
? Do you want to provide an explicit role name: No
? Service account name: [? for help]
...

❯ ./rosa delete iamserviceaccount -c hkepley
? Do you want to provide an explicit role name: Yes
? IAM role name: [? for help]
...
```

* role name flag without interactive mode runs command automatically

```bash
❯ ./rosa delete iamserviceaccount -c hkepley --role-name test
W: Role 'test' does not exist
```

* service account name + namespace without interactive mode runs command automatically

```bash
❯ ./rosa delete iamserviceaccount -c hkepley --name test --namespace ns
W: Role 'hkepley-ns-test-role' does not exist
```

* service account name without namespace // namespace without service account name falls back to interactive mode:

```bash
❯ ./rosa delete iamserviceaccount -c hkepley --name test
? Service account name: test
? Namespace: default
W: Role 'hkepley-default-test-role' does not exist

❯ ./rosa delete iamserviceaccount -c hkepley --namespace ns
? Service account name: testtt
? Namespace: ns
W: Role 'hkepley-ns-testtt-role' does not exist
```

* Interactive mode with namespace + service account name flags still goes to interactive mode properly:

```bash
❯ ./rosa delete iamserviceaccount -c hkepley --name test --namespace ns -i
? Service account name: [? for help] (test)
```

* Interactive mode with role name still goes to interactive mode:

```bash
❯ ./rosa delete iamserviceaccount -c hkepley --role-name test -i
? IAM role name: [? for help] (test)
```

### DIFF
--- a/cmd/dlt/iamserviceaccount/cmd.go
+++ b/cmd/dlt/iamserviceaccount/cmd.go
@@ -51,7 +51,8 @@ func DeleteIamServiceAccountRunner(userOptions *iamServiceAccountOpts.DeleteIamS
 		}
 
 		// Determine if interactive mode is needed
-		if !interactive.Enabled() && !cmd.Flags().Changed("mode") {
+		if !interactive.Enabled() && !cmd.Flags().Changed("mode") && !cmd.Flags().Changed("role-name") &&
+			(!cmd.Flags().Changed("name") || !cmd.Flags().Changed("namespace")) {
 			interactive.Enable()
 		}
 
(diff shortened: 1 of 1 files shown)


cmd/dlt/iamserviceaccount/cmd.go
```diff
@@ -99,18 +108,12 @@ func DeleteIamServiceAccountRunner(userOptions *iamServiceAccountOpts.DeleteIamS
 				if err != nil {
 					return fmt.Errorf("expected a valid service account name: %s", err)
 				}
-			}
-
-			if serviceAccountName == "" {
-				return fmt.Errorf("service account name is required when role name is not specified")
-			}
 
-			if err := iamserviceaccount.ValidateServiceAccountName(serviceAccountName); err != nil {
```
AlexVulaj, line 108: Out of curiosity, what else was "ValidateServiceAccountName" doing here?

AlexVulaj: Ah never mind - I see we're still using it below, just differently in the flow.

hunterkepley: Things which AWS does for us already but we put in the CLI to make the UX a little better (not empty, not >253 chars, must be alphanumeric or '-'/'.'

The validation exists on the interactive prompt, to be honest didn't notice I removed it for non-interactive

There was a discussion a few months ago about whether or not this was useful to have and it was deemed useful so I do not want to remove it 😓  But honestly not an issue and doesn't break anything

PR #3017, opened 2025-09-10
OCM-18376 | fix: Delete inline policies

hunterkepley:
Fixes bug with [XCMSTRAT-1311](http://issues.redhat.com/browse/XCMSTRAT-1311) where inline policies were not deleted
* Bug seems to be that it was never introduced, so inline policies were still attached when the role was attempted to be deleted


olucasfreitas (approved):
/approve

PR #3020, opened 2025-09-14
OCM-17621 | Add support for enabling AutoNode to 'rosa edit cluster'

olucasfreitas:


cmd/edit/cluster/cmd.go
```diff
@@ -1142,3 +1189,163 @@ func confirmMigration() (bool, error) {
 		Default: false,
 	})
 }
+
+// AutoNode helper functions
+func autonodeExists(cluster *cmv1.Cluster) bool {
+	return cluster.AutoNode() != nil && cluster.AutoNode().Mode() == "enabled"
+}
```
hunterkepley: helper functions should be in a helper and/or pkg file

cmd/edit/cluster/cmd.go
```diff
@@ -1142,3 +1189,163 @@ func confirmMigration() (bool, error) {
...
+
+func autonodeRoleExists(cluster *cmv1.Cluster) string {
+	if cluster.AWS() != nil && cluster.AWS().AutoNode() != nil {
+		return cluster.AWS().AutoNode().RoleArn()
+	}
+	return ""
+}
+
+func setAutoNode(r *rosa.Runtime, cmd *cobra.Command, cluster *cmv1.Cluster, autonode string, roleArn string) (
```
hunterkepley: all of these functions should go into a helper file and/or in the pkg/ dir

cmd/edit/cluster/cmd.go
```diff
@@ -1142,3 +1189,163 @@ func confirmMigration() (bool, error) {
...
+// AutoNode helper functions
+func autonodeExists(cluster *cmv1.Cluster) bool {
+	return cluster.AutoNode() != nil && cluster.AutoNode().Mode() == "enabled"
+}
+
+func autonodeRoleExists(cluster *cmv1.Cluster) string {
+	if cluster.AWS() != nil && cluster.AWS().AutoNode() != nil {
+		return cluster.AWS().AutoNode().RoleArn()
+	}
+	return ""
```
hunterkepley: This isn't a proper Go 'exists' function- we should return a bool named `ok` here as well and use that to check if it exists

cmd/edit/cluster/cmd.go
```diff
@@ -1142,3 +1189,163 @@ func confirmMigration() (bool, error) {
...
+		if autonode != "enabled" {
+			return "", "", fmt.Errorf("Invalid value for --autonode. Currently only 'enabled' is supported")
+		}
+		mode = autonode
+	}
+
+	// Handle role ARN
+	if roleArnChanged {
+		if roleArn != "" && !aws.RoleArnRE.MatchString(roleArn) {
+			return "", "", fmt.Errorf("Invalid IAM Role ARN format: %s", roleArn)
```
hunterkepley: This error doesn't tell the user why the format is wrong, only that it is incorrect and what it is. For example, what if the RE doesn't match? The user wouldn't know this

cmd/edit/cluster/cmd.go
```diff
@@ -1142,3 +1189,163 @@ func confirmMigration() (bool, error) {
...
+	}
+
+	// If enabling AutoNode, role ARN is required
+	if mode == "enabled" && !currentAutoNodeEnabled {
+		if role == "" && currentRoleArn == "" {
+			return "", "", fmt.Errorf("IAM Role ARN is required when enabling AutoNode. Use --autonode-iam-role-arn to specify the role")
+		}
+		if role == "" {
+			role = currentRoleArn
+		}
```
hunterkepley:
```
if role == "" {
    if currentRoleArn == "" {
        return "", "", fmt.Errorf(...)
    }
    role = currentRoleArn
}
```

cmd/edit/cluster/cmd.go
```diff
@@ -1142,3 +1189,163 @@ func confirmMigration() (bool, error) {
...
+		return cluster.AWS().AutoNode().RoleArn()
+	}
+	return ""
+}
+
+func setAutoNode(r *rosa.Runtime, cmd *cobra.Command, cluster *cmv1.Cluster, autonode string, roleArn string) (
+	mode string, role string, err error) {
+
+	// Check if flags are changed
+	autonodeChanged := cmd.Flags().Changed("autonode")
```
hunterkepley: There's some inconsistency with the naming with `autonode` in the MR- would you mind picking whichever one (`autonode` or `autoNode`) and changing the rest to match?

cmd/edit/cluster/cmd.go
```diff
@@ -179,6 +183,20 @@ func init() {
 		"The ARN of the role that is used to forward audit logs to AWS CloudWatch.",
 	)
 
+	flags.StringVar(
+		&args.autonode,
+		"autonode",
+		"",
+		"Enable AutoNode for automatic node management.",
```
hunterkepley: We should inform the user that the only value available is `enabled`, since it's in the value maybe we could reword entirely

`Supply the value 'enabled' to enable AutoNode for automatic node management.`

cmd/edit/cluster/cmd.go
```diff
@@ -1142,3 +1189,163 @@ func confirmMigration() (bool, error) {
...
+
+func autonodeRoleExists(cluster *cmv1.Cluster) string {
+	if cluster.AWS() != nil && cluster.AWS().AutoNode() != nil {
+		return cluster.AWS().AutoNode().RoleArn()
+	}
+	return ""
+}
+
+func setAutoNode(r *rosa.Runtime, cmd *cobra.Command, cluster *cmv1.Cluster, autonode string, roleArn string) (
+	mode string, role string, err error) {
```
hunterkepley: Can we change the name of `mode string` to `autonodeMode string`? it's a bit confusing later on in the function

cmd/edit/cluster/cmd.go
```diff
@@ -1142,3 +1189,163 @@ func confirmMigration() (bool, error) {
...
+
+	// If updating role ARN with AutoNode already enabled, maintain enabled state
+	if roleArnChanged && currentAutoNodeEnabled && !autonodeChanged {
+		mode = "enabled"
+	}
+
+	// Confirm with user
+	if mode != "" || role != "" {
+		confirmAutoNode(r, mode, role, currentAutoNodeEnabled, currentRoleArn)
+	}
```
hunterkepley: Curious- should role ever be empty if mode isn't with the validations above?

hunterkepley: Either way shouldn't affect anything, but I think we could get away with removing `|| role != ""`

cmd/edit/cluster/cmd.go
```diff
@@ -1142,3 +1189,163 @@ func confirmMigration() (bool, error) {
...
+		r.Reporter.Warnf("You are choosing to enable AutoNode for automatic node management")
+		if !confirm.Confirm("enable AutoNode for cluster with the provided role arn '%s'", roleArn) {
+			os.Exit(0)
+		}
+	} else if roleArn != "" && currentRoleArn != "" && roleArn != currentRoleArn {
+		r.Reporter.Warnf("You are choosing to update the AutoNode IAM role ARN")
+		if !confirm.Confirm("update AutoNode IAM role ARN from '%s' to '%s'", currentRoleArn, roleArn) {
+			os.Exit(0)
+		}
+	}
```
hunterkepley: while encapsulating code is good for not duplicating it- I think we should move this code to where the function is called and delete the function.

I have a couple of reasons;

1) This function is only an if statement controlled by a boolean passed in,[ I think this article can explain it better than I can though (4 minute read)](https://alexkondov.com/should-you-pass-boolean-to-functions/)
2) This code is not going to be reusable really, we only need to call this confirmation once and it's very specific. But, if we were to reuse it, we should make 2 functions and put each branch into a separate function

WDYT?

cmd/edit/cluster/cmd.go
```diff
@@ -1142,3 +1189,163 @@ func confirmMigration() (bool, error) {
...
+}
+
+func autonodeInteractivePrompt(r *rosa.Runtime, cmd *cobra.Command, cluster *cmv1.Cluster) (
+	mode string, roleArn string, err error) {
+
+	// Only prompt if AutoNode is not already enabled
+	currentAutoNodeEnabled := autonodeExists(cluster)
+	if currentAutoNodeEnabled {
```
hunterkepley: I don't think we don't need to make this a var since it's only used in the if statement right below it, WDYT?

cmd/edit/cluster/cmd.go
```diff
@@ -1142,3 +1189,163 @@ func confirmMigration() (bool, error) {
...
+
+	if !enableAutoNode {
+		return "", "", nil
+	}
+
+	r.Reporter.Infof("AutoNode requires an IAM role with appropriate permissions for node management.")
+	r.Reporter.Infof("Please ensure the IAM role is created before proceeding.")
+
+	roleArnValue, err := interactive.GetString(interactive.Input{
+		Question: "AutoNode IAM Role ARN",
```
hunterkepley: I think we could get away with 2 prompts rather than 4, WDYT? Even if we want the different prompt messages provided in the above prompts, we can do so while only having 2 prompts in the code

hunterkepley: Would reduce the LoC by 30-40 lines by a quick estimate

cmd/edit/cluster/cmd.go
```diff
@@ -654,6 +673,26 @@ func run(cmd *cobra.Command, _ []string) {
...
 	}
 
+	// AutoNode configuration
+	autonodeMode, autonodeRoleArn, err := setAutoNode(r, cmd, cluster, args.autonode, args.autonodeRoleARN)
+	if err != nil {
+		r.Reporter.Errorf("%s", err)
+		os.Exit(1)
+	}
+	if interactive.Enabled() && aws.IsHostedCP(cluster) {
+		interactiveMode, interactiveRoleArn, err := autonodeInteractivePrompt(r, cmd, cluster)
```
hunterkepley: Could we change the names of these to just `autonodeMode` and `autonodeRoleArn` like the above ones? `interactiveMode` makes me immediately think of `-i` or `--interactive` mode rather than the auto node gathered from interactive

pkg/ocm/clusters.go
```diff
@@ -712,9 +717,20 @@ func (c *Client) UpdateCluster(clusterKey string, creator *aws.Creator, config S
...
+		if config.AutoNodeRoleARN != "" {
+			autoNodeBuilder := cmv1.NewAwsAutoNode().RoleArn(config.AutoNodeRoleARN)
+			awsBuilder = awsBuilder.AutoNode(autoNodeBuilder)
+		}
 		clusterBuilder.AWS(awsBuilder)
 	}
 
+	// Set AutoNode mode if enabled
+	if config.AutoNodeEnabled {
+		autoNodeBuilder := cmv1.NewClusterAutoNode().Mode("enabled")
```
hunterkepley: We should set it to the actual value since the ticket specifies "for now" we only have the value `enabled`

pkg/ocm/clusters.go
```diff
@@ -153,6 +153,10 @@ type Spec struct {
 	// Audit Log Forwarding
 	AuditLogRoleARN *string
 
+	// AutoNode configuration
+	AutoNodeEnabled bool
```
hunterkepley: should be a string to match the command

cmd/edit/cluster/cmd.go
olucasfreitas: fixed on last commit

cmd/edit/cluster/cmd.go
olucasfreitas: fixed on last commit

cmd/edit/cluster/cmd.go
olucasfreitas: fixed on last commit

cmd/edit/cluster/cmd.go
olucasfreitas: fixed on last commit

cmd/edit/cluster/cmd.go
olucasfreitas: fixed on last commit

cmd/edit/cluster/cmd.go
olucasfreitas: fixed on last commit

cmd/edit/cluster/cmd.go
olucasfreitas: fixed on last commit

cmd/edit/cluster/cmd.go
olucasfreitas: fixed on last commit

cmd/edit/cluster/cmd.go
olucasfreitas: fixed on last commit

cmd/edit/cluster/cmd.go
olucasfreitas: True, can be simplified, fixed on last commit

cmd/edit/cluster/cmd.go
olucasfreitas: You're right, the 2 prompts are more concise, fixed on the last commit

cmd/edit/cluster/cmd.go
olucasfreitas: fixed on last commit

pkg/ocm/clusters.go
olucasfreitas: fixed on last commit

pkg/ocm/clusters.go
olucasfreitas: fixed on last commit

cmd/edit/cluster/cmd.go
olucasfreitas: You're right, the article makes a good case, this is indeed a code smell, fixed on the last commit

cmd/edit/cluster/cmd.go
hunterkepley: @olucasfreitas Still not fixed

cmd/edit/cluster/cmd.go
hunterkepley: @olucasfreitas Did you test it? Did it affect anything?

pkg/helper/autonode/helpers.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/openshift/rosa/pkg/aws"
+	"github.com/openshift/rosa/pkg/interactive"
+	"github.com/openshift/rosa/pkg/interactive/confirm"
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+// SetAutoNode validates and sets AutoNode configuration for a cluster
+func SetAutoNode(r *rosa.Runtime, cmd *cobra.Command, cluster *cmv1.Cluster, autonode string, roleArn string) (
+	autonodeMode string, autonodeRole string, err error) {
```
hunterkepley: what's the difference between autonode and autonodemode?

hunterkepley: And `roleArn` vs `autonodeRole`?

hunterkepley: Since this is 5 parameters + 3 return values, and since the below function uses some of them, and this is in it's own package, I think we should put them into a struct and use that instead

pkg/helper/autonode/helpers.go
```diff
@@ -0,0 +1,156 @@
...
+		if currentAutonodeEnabled {
+			return "", "", fmt.Errorf("AutoNode is already enabled for this cluster")
+		}
+
+		// When enabling AutoNode, IAM role ARN is required
+		if !roleArnChanged || roleArn == "" {
+			return "", "", fmt.Errorf("IAM role ARN is required when enabling AutoNode")
+		}
+
+		autonodeMode = autonode
```
hunterkepley: this should be in a validation function in the proper file IMO

pkg/helper/autonode/helpers.go
```diff
@@ -0,0 +1,156 @@
...
+		// If AutoNode is not being enabled in this request, it must already be enabled
+		if !autonodeChanged && !currentAutonodeEnabled {
+			return "", "", fmt.Errorf("Cannot update IAM role ARN when AutoNode is not enabled. " +
+				"Enable AutoNode first with --autonode=enabled")
+		}
+
+		// If AutoNode is already enabled and only updating the role
+		if currentAutonodeEnabled && !autonodeChanged {
+			autonodeMode = "enabled" // Maintain the current state
+		}
```
hunterkepley, line 111: this should be in a validation function in the proper file IMO

pkg/helper/autonode/helpers.go
```diff
@@ -0,0 +1,156 @@
...
+
+		autonodeRole = roleArn
+	}
+
+	// Display confirmation prompts based on the changes being made
+	if autonodeMode == "enabled" && !currentAutonodeEnabled {
+		// Enabling AutoNode for the first time
+		r.Reporter.Warnf("You are choosing to enable automatic node management (AutoNode)")
+		if !confirm.Confirm("enable AutoNode for cluster with the provided role arn '%s'", autonodeRole) {
```
hunterkepley: is that role arn the cluster's role arn? Or the auto node role arn?

cmd/edit/cluster/cmd.go
hunterkepley: @olucasfreitas I don't think it is fixed

hunterkepley:
> @hunterkepley Sorry about that, was not aware it did that

Github automatically resolves conversations for you?

hunterkepley:
> Can I ask why we're prompting the user with this message when enabling Autonode?

Thanks @robpblake - @olucasfreitas we should remove this prompt

cmd/edit/cluster/cmd.go
```diff
@@ -179,6 +184,20 @@ func init() {
 		"The ARN of the role that is used to forward audit logs to AWS CloudWatch.",
 	)
 
+	flags.StringVar(
+		&args.autonode,
+		"autonode",
+		"",
+		"Supply the value 'enabled' to enable AutoNode for automatic node management.",
```
robpblake: @olucasfreitas Can we change this to:

> Configure AutoNode for the cluster. Valid values are: enabled

cmd/edit/cluster/cmd.go
```diff
@@ -179,6 +184,20 @@ func init() {
...
+	)
+
+	flags.StringVar(
+		&args.autonodeRoleARN,
+		"autonode-iam-role-arn",
+		"",
+		"The AWS ARN of the IAM Role that has the permissions required for the AutoNode controller.",
```
robpblake: @olucasfreitas Can this please be:

> The AWS ARN of the IAM Role that has permissions for AutoNode

pkg/helper/autonode/helpers.go
```diff
@@ -0,0 +1,156 @@
...
+		// If AutoNode is already enabled and only updating the role
+		if currentAutonodeEnabled && !autonodeChanged {
+			autonodeMode = "enabled" // Maintain the current state
+		}
+
+		autonodeRole = roleArn
+	}
+
+	// Display confirmation prompts based on the changes being made
+	if autonodeMode == "enabled" && !currentAutonodeEnabled {
```
robpblake: @olucasfreitas / @hunterkepley Both of these prompts should be removed. We typically only prompt the user when their action will cause disruption in their cluster e.g. cause all data plane nodes to be recreated. This is not the case for enabling AutoNode.

pkg/helper/autonode/helpers.go
robpblake: @hunterkepley It is the AutoNode role ARN.

pkg/helper/autonode/helpers.go
```diff
@@ -0,0 +1,156 @@
...
+
+	autonodeEnabled := ocm.AutoNodeExists(cluster)
+	currentRoleArn, _ := ocm.GetAutoNodeRoleArn(cluster)
+
+	// Build the appropriate question based on current state
+	var question string
+	if autonodeEnabled {
+		question = fmt.Sprintf("Update AutoNode IAM role ARN (current: %s)", currentRoleArn)
+	} else {
+		question = "Enable automatic node management (AutoNode)"
```
robpblake: @olucasfreitas / @hunterkepley Can this please just be `Enable AutoNode`

cmd/edit/cluster/cmd.go
hunterkepley: @olucasfreitas doesnt seem this changed

pkg/helper/autonode/helpers.go
hunterkepley: @olucasfreitas Would you mind checking these comments? Had a couple questions + a request 🙂

cmd/edit/cluster/cmd.go
olucasfreitas: image
 the changes are there, if you check the last commit you can see them, I did not squash them to no afftect them reviews here

pkg/ocm/helpers.go
```diff
@@ -1122,3 +1122,17 @@ func ValidateClaimValidationRules(input interface{}) error {
 	}
 	return nil
 }
+
+// AutoNodeExists checks if AutoNode is enabled on the cluster
+func AutoNodeExists(cluster *cmv1.Cluster) bool {
+	return cluster.AutoNode() != nil && cluster.AutoNode().Mode() == "enabled"
+}
```
hunterkepley: I left a comment on the original changes, but it seems this still hasn't changed. I would prefer if this were different

1. `enabled` is only one possible value for autonode
2. This only returns a bool, we will only get `false` in a case where `cluster.AutoNode()` is `nil`

As stated on the original comment, I think we should use a more Go-style `exists` function pattern.

Take a look at `isHcpBillingTechPreview` in the OCM Client. Notice how we are accomplishing the same thing, but we get a lot more useful information out of it with that error return

To add onto the first point: in the future, what would happen if the auto node value is valid, not `enabled`? We would always get `false` here

hunterkepley (commented):
@olucasfreitas ~~Please check and make sure you don't miss comments when asking for review 🙂 there's several which were not addressed that Rob left~~

cmd/edit/cluster/cmd.go
hunterkepley: @olucasfreitas it was a caching issue with GitHub, apologies for those comments

pkg/helper/autonode/helpers.go
```diff
@@ -2,28 +2,89 @@ package autonode
...
 
 	"github.com/openshift/rosa/pkg/aws"
 	"github.com/openshift/rosa/pkg/interactive"
-	"github.com/openshift/rosa/pkg/interactive/confirm"
 	"github.com/openshift/rosa/pkg/ocm"
 	"github.com/openshift/rosa/pkg/rosa"
 )
 
+// Config holds AutoNode configuration parameters and results
+type Config struct {
```
hunterkepley: This is a tad too generic of a name

hunterkepley: Also this isn't used anywhere ?

pkg/helper/autonode/helpers.go
```diff
@@ -2,28 +2,89 @@ package autonode
...
+
+// DetermineAutoNodeMode determines the AutoNode mode based on the current state and changes
+func DetermineAutoNodeMode(autoNodeChanged, currentEnabled bool, flagValue string) string {
+	if autoNodeChanged {
+		return flagValue
+	}
+	// If AutoNode is already enabled and only updating the role, maintain current state
+	if currentEnabled {
+		return "enabled"
+	}
```
hunterkepley: I'm not too sure I see a use for this, we only allow `auto-node` to be `enabled` at the moment. Eventually, this won't be the case though

Either way, all this code seems to do is set the value to `enabled` if the value is already set to `enabled`

So, we're essentially saying `is autonode equal to "enabled?" > set autonode to "enabled"`

hunterkepley: We also have several checks for `autonode == "enabled"` in each command. Realistically, we only need to validate this one time (after user input is taken)

pkg/helper/autonode/helpers.go
```diff
@@ -129,11 +153,11 @@ func InteractivePrompt(r *rosa.Runtime, cmd *cobra.Command, cluster *cmv1.Cluste
 	}
 
 	// Set mode to enabled
-	autonodeMode = "enabled"
+	mode = "enabled"
```
hunterkepley: I would prefer this to be `autoNodeMode` like it was; purely because there are CLI operating modes we call `mode` (auto mode + interactive mode`

I think I brought that up in a past comment, but we avoid calling it something as generic as `mode` since that var name is used in every command to signify CLI operating mode

hunterkepley: For example, in `create/cluster`:

```go
if isSTS && mode != "" {
		isValidMode := arguments.IsValidMode(interactive.Modes, mode)
		if !isValidMode {
			r.Reporter.Errorf("Invalid --mode '%s'. Allowed values are %s", mode, interactive.Modes)
			os.Exit(1)
		}
	}
```

hunterkepley (commented):
Looking better, left a couple more comments and ill wait on Rob to review again tomorrow as well

pkg/helper/autonode/helpers.go
olucasfreitas: Agreed and did these in the last commit

pkg/helper/autonode/helpers.go
olucasfreitas: Agreed and did these in the last commit

cmd/edit/cluster/cmd.go
olucasfreitas: @robpblake did these changes in the last commit, please check

cmd/edit/cluster/cmd.go
olucasfreitas: @robpblake did these changes in the last commit, please check

pkg/helper/autonode/helpers.go
olucasfreitas: @robpblake did these changes in the last commit, please check

pkg/helper/autonode/helpers.go
olucasfreitas: @robpblake did these changes in the last commit, please check

pkg/helper/autonode/helpers.go
olucasfreitas: I forgot to respond, but I implement the request on the last commit, and to answer the questions

Input vs Output parameters:
autonode (input) = raw value from the --autonode flag (e.g., "enabled")
autonodeMode (output) = processed/validated mode to actually set on the cluster
roleArn (input) = raw value from the --autonode-iam-role-arn flag
autonodeRole (output) = processed/validated IAM role ARN to use

Probably a bit redundant, so I made some changes in the commit

pkg/helper/autonode/helpers.go
olucasfreitas: Updated on this commit [9f4c0b0](https://github.com/openshift/rosa/pull/3020/commits/9f4c0b096510fc069cf2531425b3e395b5c4d507)

pkg/helper/autonode/helpers.go
olucasfreitas: Updated on this commit [9f4c0b0](https://github.com/openshift/rosa/pull/3020/commits/9f4c0b096510fc069cf2531425b3e395b5c4d507)

pkg/helper/autonode/helpers.go
olucasfreitas: Updated on this commit [9f4c0b0](https://github.com/openshift/rosa/pull/3020/commits/9f4c0b096510fc069cf2531425b3e395b5c4d507)

pkg/ocm/helpers.go
olucasfreitas: Updated on this commit [9f4c0b0](https://github.com/openshift/rosa/pull/3020/commits/9f4c0b096510fc069cf2531425b3e395b5c4d507)

pkg/helper/autonode/helpers.go
```diff
@@ -0,0 +1,185 @@
...
+
+// SetAutoNode validates and sets AutoNode configuration for a cluster
+func SetAutoNode(r *rosa.Runtime, cmd *cobra.Command, cluster *cmv1.Cluster, flagValue string, roleArnValue string) (*AutoNodeConfig, error) {
+	config := &AutoNodeConfig{
+		AutoNodeFlag: flagValue,
+		RoleARNFlag:  roleArnValue,
+	}
+
+	// Check if flags are changed
+	autoNodeChanged := cmd.Flags().Changed("autonode")
```
robpblake: @olucasfreitas Minor, but you could extract the flag names to `const` so you don't need to keep repeating them an introduce the risk of typos.

pkg/ocm/helpers.go
```diff
@@ -1122,3 +1122,21 @@ func ValidateClaimValidationRules(input interface{}) error {
 	}
 	return nil
 }
+
+// GetAutoNodeMode returns the AutoNode mode and whether AutoNode configuration exists
```
robpblake, line 1126: @olucasfreitas / @hunterkepley I'd personally add unit tests for this.

pkg/helper/autonode/helpers.go
```diff
@@ -0,0 +1,185 @@
...
+	AutoNodeFlag string // Value from --autonode flag
+	RoleARNFlag  string // Value from --autonode-iam-role-arn flag
+
+	// Output values
+	AutoNodeMode    string // AutoNode mode to set (e.g., "enabled")
+	AutoNodeRoleARN string // IAM role ARN to use
+}
+
+// ValidateAutoNodeValue validates the autonode flag value
+func ValidateAutoNodeValue(value string) error {
```
robpblake, line 35: @olucasfreitas / @hunterkepley I'd personally add unit tests for all these `Validate*` functions.

hunterkepley:
@robpblake Thanks rob, I do think that all your comments look valid- we definitely should have some unit tests especially for those validate funds

pkg/helper/autonode/helpers.go
olucasfreitas: adressed this on [38b402f](https://github.com/openshift/rosa/pull/3020/commits/38b402fbc67c8db4dd23820e0dbfa12ba68b927d)

pkg/ocm/helpers.go
olucasfreitas: adressed on [234f056](https://github.com/openshift/rosa/pull/3020/commits/234f056a9425a5e7b4d4c45c6af4337b6d16baf0)

pkg/helper/autonode/helpers.go
olucasfreitas: adressed on [234f056](https://github.com/openshift/rosa/pull/3020/commits/234f056a9425a5e7b4d4c45c6af4337b6d16baf0)

robpblake (approved):

PR #3029, opened 2025-09-23
OCM-17611 | feat: Introduce billing account selection for classic

hunterkepley:
Introduces day 1 and 2 billing account selection for classic clusters. Simple introduction since it is available for HCP already, the goal here was to replace the `if isHostedCp` gates with `if isHostedCp || (isClassic and isCapabilityEnabledForOrg)` so that billing account selection is available for customers with the correct capability for classic clusters

```swift
❯ ./rosa edit cluster -c hkepley --billing-account 123
I: Updated cluster 'hkepley'
❯ make
go build -ldflags="-X github.com/openshift/rosa/pkg/info.Build=a587d1642" ./cmd/rosa
❯ ./rosa edit cluster -c hkepley --billing-account 123
E: Provided billing account number 123 is not valid. Rerun the command with a valid billing account number
```

^ Example of editing, first command run was with a `false` org capability. Second run is with a `true` capability, but an incorrect billing account number

```swift
❯ ./rosa create cluster --mode auto --sts
hkeI: Enabling interactive mode
? Cluster name: hkepley
? Domain prefix (optional):
? Deploy cluster with Hosted Control Plane: No
? Create cluster admin user: No
? Billing Account:  [Use arrows to move, type to filter, ? for more help]
> ....
....
```

^ Creating (when you do not have capability, it simply skips the prompt, and ignores the flag if passed in. This way, we won't break clients for customers with or without the capability during private preview)

```swift
❯ ./rosa describe cluster -c hkepley

Name:                       hkepley
Domain Prefix:              hkepley
Display Name:               hkepley
ID:                         <CLUSTER_ID>
External ID:                <EXTERNAL_ID>
OpenShift Version:          4.19.12
DNS:                        <DNS>
AWS Account:                <REAL_BILLING_ACCOUNT_NUMBER>
AWS Billing Account:        <REAL_BILLING_ACCOUNT_NUMBER>
```

^ Describing

### DIFF
--- a/cmd/create/cluster/validation.go
+++ b/cmd/create/cluster/validation.go
@@ -23,7 +23,7 @@ const (
 	isNotGovcloudFeature = "Hosted Control Plane shared VPC clusters are not supported on Govcloud regions; %s"
 	pleaseRemoveFlags    = "Please remove the following flags: %s"
 
-	billingAccountsHcpErrorMsg = "Billing accounts are only supported for non-govcloud Hosted Control Plane clusters"
+	billingAccountsGovErrorMsg = "Billing accounts are only supported for non-govcloud clusters"
 )
 
 func validateHcpSharedVpcArgs(route53RoleArn string, vpcEndpointRoleArn string,
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -1208,14 +1208,22 @@ func run(cmd *cobra.Command, _ []string) {
 
 	// Billing Account
 	billingAccount := args.billingAccount
-	if isHostedCP && !fedramp.Enabled() {
+	if !fedramp.Enabled() {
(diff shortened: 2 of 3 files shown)


cmd/create/cluster/validation.go
```diff
@@ -23,7 +23,7 @@ const (
 	isNotGovcloudFeature = "Hosted Control Plane shared VPC clusters are not supported on Govcloud regions; %s"
 	pleaseRemoveFlags    = "Please remove the following flags: %s"
 
-	billingAccountsHcpErrorMsg = "Billing accounts are only supported for non-govcloud Hosted Control Plane clusters"
+	billingAccountsHcpErrorMsg = "Billing accounts are only supported for non-govcloud clusters"
```
marcolan018: better to update name of the var to reflect its usage

PR #3035, opened 2025-10-02
OCM-18785 | fix: added verifications to exclude fedramp from autonode feature

olucasfreitas:


pkg/helper/autonode/helpers_test.go
```diff
@@ -145,3 +147,19 @@ var _ = Describe("DetermineAutoNodeMode", func() {
...
+
+		AfterEach(func() {
+			fedramp.Disable()
+		})
+
+		It("InteractivePrompt returns nil for FedRAMP clusters", func() {
+			Expect(fedramp.Enabled()).To(BeTrue())
+		})
+	})
+})
```
hunterkepley: no need to test this with written tests, we know that `fedramp.Enabled()` works from the tests made for it

hunterkepley: Feel free to just delete this test entirely

pkg/helper/autonode/helpers.go
```diff
@@ -101,6 +102,11 @@ func SetAutoNode(r *rosa.Runtime, cmd *cobra.Command, cluster *cmv1.Cluster, fla
 		return nil, fmt.Errorf("AutoNode is only supported for Hosted Control Plane clusters")
 	}
 
+	// AutoNode is not supported in FedRAMP
+	if fedramp.Enabled() {
+		return nil, fmt.Errorf("AutoNode is not supported for FedRAMP clusters")
+	}
+
```
hunterkepley: [This function runs no matter what](https://github.com/openshift/rosa/blob/7903b7e2b4766e01e42c36fadc2469a7f6f4ad5a/cmd/edit/cluster/cmd.go#L678) (SetAutoNode)- this means this fedramp check will run no matter what

effectively breaking `edit cluster` for fedramp users

pkg/helper/autonode/helpers.go
```diff
@@ -134,6 +140,11 @@ func SetAutoNode(r *rosa.Runtime, cmd *cobra.Command, cluster *cmv1.Cluster, fla
 
 // InteractivePrompt handles interactive mode for AutoNode configuration
 func InteractivePrompt(r *rosa.Runtime, cmd *cobra.Command, cluster *cmv1.Cluster) (*AutoNodeConfig, error) {
+	// AutoNode is not supported in FedRAMP - skip prompting
+	if fedramp.Enabled() {
+		return nil, nil
+	}
+
```
hunterkepley: looks fine, I would prefer to never reach this point anyway, but if we do I see no harm in an extra safeguard for now

pkg/helper/autonode/helpers.go
hunterkepley: We should maybe put this check somewhere else, or make sure the user is actually attempting to set an autonode value before erroring

pkg/helper/autonode/helpers_test.go
hunterkepley: either way, this is what the test is testing:

1. You set `fedramp` to `Enabled` in the `before each`
2. You test if `fedramp.Enabled()` is true

like so;

```swift
var a = 0
if a == 0:
    test_passed()
```

Hopefully this makes sense as to why I think we don't need this test, LMK if theres questions

pkg/helper/autonode/helpers.go
olucasfreitas: Moved this on https://github.com/openshift/rosa/pull/3035/commits/9038f47fc9874998aabb4cee3828a9f44b5828c0

pkg/helper/autonode/helpers_test.go
olucasfreitas: removed this on https://github.com/openshift/rosa/pull/3035/commits/9038f47fc9874998aabb4cee3828a9f44b5828c0

pkg/helper/autonode/helpers.go
hunterkepley: Take a look at the link I sent in the first comment,  the `SetAutoNode` func is called on line 678, but the change you made on this MR is below that on line 682

The line you changed on 682 affects only the interactive prompt

When gating fedramp usage in this regression fix, lets make sure to capture all paths 🙂 . LMK if there's any questions on this

hunterkepley: Also, we already have a check for fedramp in the interactive prompt, we do not need one in the edit/cluster file as well 🙂

hunterkepley: Could we remove the fedramp check inside this function, and put it above line 678? And remove the line 682 change. That way we solve both problems

hunterkepley (commented):
LGTM, nitpick on wording and we can merge (in the error msg, lets change `FedRAMP` -> `govcloud` .. customer-facing wording and all that)

PR #3041, opened 2025-10-03
OCM-17465 | fix: created verification to when subnets are not set on proxy

olucasfreitas:
# Details

This PR fixes an issue where ROSA incorrectly attempts to use all existing subnets in a region when proxy settings are configured without explicitly providing subnet IDs via the `--subnet-ids` flag.

The implementation includes:
- Modified the subnet initialization logic to only fetch subnets when subnet IDs are explicitly provided
- Prevents automatic listing of all region subnets when proxy is configured without subnets
- Added test case to verify the correct behavior

---

## Reproducing the Issue

1. Run the same command:
    ```./rosa create cluster -c test-proxy-bug --http-proxy http://proxy.com --https-proxy https://proxy.com ```

2. **New behavior:**

Screenshot From 2025-10-08 14-53-21
Screenshot From 2025-10-08 14-53-55

---

# Ticket

Closes [ROSA-17465](https://issues.redhat.com/browse/ROSA-17465)

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -2267,12 +2267,13 @@ func run(cmd *cobra.Command, _ []string) {
 	var subnets []ec2types.Subnet
 	mapSubnetIDToSubnet := make(map[string]aws.Subnet)
 	if useExistingVPC || subnetsProvided {
-		initialSubnets, err := getInitialValidSubnets(awsClient, subnetIDs, r.Reporter)
-		if err != nil {
-			_ = r.Reporter.Errorf("Failed to get the list of subnets: %s", err)
-			os.Exit(1)
-		}
+		var initialSubnets []ec2types.Subnet
 		if subnetsProvided {
+			initialSubnets, err = getInitialValidSubnets(awsClient, subnetIDs, r.Reporter)
+			if err != nil {
+				_ = r.Reporter.Errorf("Failed to get the list of subnets: %s", err)
+				os.Exit(1)
+			}
 			useExistingVPC = true
 		}
 		_, machineNetwork, err := net.ParseCIDR(machineCIDR.String())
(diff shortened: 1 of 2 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -2267,10 +2267,13 @@ func run(cmd *cobra.Command, _ []string) {
 	var subnets []ec2types.Subnet
 	mapSubnetIDToSubnet := make(map[string]aws.Subnet)
 	if useExistingVPC || subnetsProvided {
-		initialSubnets, err := getInitialValidSubnets(awsClient, subnetIDs, r.Reporter)
-		if err != nil {
-			_ = r.Reporter.Errorf("Failed to get the list of subnets: %s", err)
-			os.Exit(1)
+		var initialSubnets []ec2types.Subnet
+		if len(subnetIDs) > 0 || subnetsProvided {
```
hunterkepley: I think we can change this to simplify, is there a specific reason we need `if len(subnetIDs) > 0`?

`subnetsProvided` is already the result of `len(subnetIDs) > 0`, so we can just use `subnetsProvided` instead

Beyond this, the pre-existing if statement on line 2269 already has `subnetsProvided`

So.. I think we can remove the new if statement, and the new `var initialSubnets []ec2types.Subnet` definition, since after removing this new if statement, the definition on 2272 will be in the proper scope again

hunterkepley (commented):
```swift
Expected results:
1. it should report "No subnets found in current region that are valid for the chosen CIDR ranges" then questionair or quite as design
Additional info:
the root cause is this line 'initialSubnets, err := getInitialValidSubnets(awsClient, subnetIDs, r.Reporter)', when there is no subnetIDs passed by flag, getInitialValidSubnets will return all subnets in this region. 
The expected result should be returning error/warning when no subnetID passed by flag, rather than using existing subnets in the region
```

From the expected results and additional info on the ticket, it seems that the goal is to 1) change the error msg that is reported in this case, and 2) check specifically if there is a subnetID passed by flag or not and react accordingly

It seems the error mentioned in the ticket is `E: The number of subnets for a 'single AZ' 'cluster' should be '2', instead received: '0'`

This error comes from the function `ValidateSubnetsCount` function specifically:

```swift
"The number of subnets for a '%s' '%s' should be '%d', "+
			"instead received: '%d'"
```

I also see that the error you put in the screenshot on the description of this MR is a 400 response from the EC2 API from AWS. I am unsure that this is correct

Image

PR #3047, opened 2025-10-15
OCM-12612 | fix: Filter DNS domains by organization in list command

olucasfreitas:
# Details

This PR fixes a security issue where `rosa list dns-domain` was displaying DNS domains from all organizations instead of filtering by the current logged-in organization, potentially exposing resources from other orgs.

---

## Fixed Behavior (Organization Filtered)

1. Run the same command with the fix:

    ```bash
    ./rosa list dns-domain
    ```

2. Now shows **only DNS domains from the current organization** (3 domains):

    ```
    ID                         CLUSTER ID  RESERVED TIME         USER DEFINED  ARCHITECTURE
    531t.p1.openshiftapps.com              0001-01-01T00:00:00Z  Yes           classic
    rgct.p1.openshiftapps.com              0001-01-01T00:00:00Z  Yes           classic
    ypwi.p4.openshiftapps.com              0001-01-01T00:00:00Z  Yes           
    ```

---

## The --all Flag Behavior

1. The `--all` flag now shows all DNS domains (user-defined + system) but **still filtered by current organization**:

    ```bash
    ./rosa list dns-domain --all
    ```

    Example output (82 domains, all from current org):
    ```
    ID                         CLUSTER ID                        RESERVED TIME         USER DEFINED  ARCHITECTURE
    531t.p1.openshiftapps.com              0001-01-01T00:00:00Z  Yes           classic
    rgct.p1.openshiftapps.com              0001-01-01T00:00:00Z  Yes           classic
    ypwi.p4.openshiftapps.com              0001-01-01T00:00:00Z  Yes           
    auto.p1.openshiftapps.com  2abc3def4...                      2024-01-15T10:30:00Z  No            classic
    ```

---

# Ticket

Closes [OCM-12612](https://issues.redhat.com/browse/OCM-12612)

### DIFF
--- a/cmd/list/dnsdomains/cmd.go
+++ b/cmd/list/dnsdomains/cmd.go
@@ -71,13 +71,18 @@ func run(_ *cobra.Command, _ []string) {
 	defer r.Cleanup()
 
 	r.Reporter.Debugf("Loading dns domains for current org id")
-	search := "user_defined='true'"
+	orgID, _, err := r.OCMClient.GetCurrentOrganization()
+	if err != nil {
+		_ = r.Reporter.Errorf("Failed to get current organization: %s", err)
+		os.Exit(1)
+	}
+	search := fmt.Sprintf("user_defined='true' AND organization.id='%s'", orgID)
 	if args.all {
 		search = ""
 	}
 	dnsDomains, err := r.OCMClient.ListDNSDomains(search)
 	if err != nil {
-		r.Reporter.Errorf("Failed to list DNS Domains: %v", err)
+		_ = r.Reporter.Errorf("Failed to list DNS Domains: %v", err)
 		os.Exit(1)
 	}
 
@@ -88,7 +93,7 @@ func run(_ *cobra.Command, _ []string) {
 	if output.HasFlag() {
 		err = output.Print(dnsDomains)
 		if err != nil {
-			r.Reporter.Errorf("%s", err)
+			_ = r.Reporter.Errorf("%s", err)
 			os.Exit(1)
 		}
 		os.Exit(0)
--- a/pkg/ocm/dns_domain.go
+++ b/pkg/ocm/dns_domain.go
@@ -22,6 +22,7 @@ func (c *Client) ListDNSDomains(search string) ([]*cmv1.DNSDomain, error) {
 		DNSDomains().
 		List().
 		Parameter("search", search).
+		Parameter("order", "organization.id asc").
 		Page(1).Size(-1).
 		Send()
 	if err != nil {


cmd/list/dnsdomains/cmd.go
```diff
@@ -70,14 +70,22 @@ func run(_ *cobra.Command, _ []string) {
 	r := rosa.NewRuntime().WithOCM()
 	defer r.Cleanup()
 
-	r.Reporter.Debugf("Loading dns domains for current org id")
-	search := "user_defined='true'"
```
gdbranco: Isn't it pretty much the same to just tie in the org ID filter here instead?
Then when you print each entry just include the org if all was supplied, just to differentiate when entries change

Might be also good to include order by org when fetching

olucasfreitas: I'm actually following the cluster pattern where --all stays within the organization boundary.
I did not cross org boundaries so I thought there's no need to display org IDs or order by org.

olucasfreitas: @gdbranco Addressed the comments here and on slack on the last commit, please take a look when you can

cmd/list/dnsdomains/cmd.go
```diff
@@ -70,14 +70,24 @@ func run(_ *cobra.Command, _ []string) {
...
 	defer r.Cleanup()
 
-	r.Reporter.Debugf("Loading dns domains for current org id")
+	// Default: show user-defined domains from current org
 	search := "user_defined='true'"
-	if args.all {
-		search = ""
+
+	if !args.all {
+		organizationID, _, err := r.OCMClient.GetCurrentOrganization()
```
gdbranco: Any reason not to just do this by default and if args.all remove the org?
I get it that it would save the request, but most users wouldn't have access to see anything other than their own org anyways

olucasfreitas: did not think of doing this, made the changes in the last commit

cmd/list/dnsdomains/cmd.go
hunterkepley: +1 to using the Org ID instead, that way it will automatically not filter an Org ID when one is not supplied

hunterkepley:
@olucasfreitas Looks like this PR has a merge conflict

hunterkepley (commented):
LGTM, will merge once @gdbranco is satisfied 🙂 I see this wasn't addressed yet: `Might be also good to include order by org when fetching`

olucasfreitas:
@hunterkepley how can I do this ? `Might be also good to include order by org when fetching`

PR #3048, opened 2025-10-15
OCM-18906 | fix: Handle HTTP errors in rosa download command and improve error messages

olucasfreitas:
# Details

This PR fixes a critical bug where the `rosa download` command would incorrectly report success when receiving HTTP error responses (like 404), downloading error pages instead of failing appropriately. It also improves error message formatting to be more user-friendly.

---

## Fixed Behavior

1. Run the same command after the fix:

    ```bash
    ./rosa download rosa-client
    ```

2. With a 404 response, the command now **properly fails** with a clear error:

    ```
    I: Downloading https://mirror.openshift.com/pub/cgw/rosa/latest/rosa-linux.tar.gz to your current directory
    E: unable to resolve host for https://mirror.openshift.com/pub/cgw/rosa/latest/rosa-linux.tar.gz 
    Please check your internet connection and DNS settings
    ```

---

# Ticket

Closes [OCM-18906](https://issues.redhat.com/browse/OCM-18906)

### DIFF
--- a/pkg/helper/download/download.go
+++ b/pkg/helper/download/download.go
@@ -3,8 +3,10 @@ package helper
 import (
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"os"
+	"path/filepath"
 	"runtime"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/helper/download/download.go
```diff
@@ -15,27 +16,52 @@ import (
 // write as it downloads and not load the whole file into memory. We pass an io.TeeReader
 // into Copy() to report progress on the download.
 func Download(url string, filename string) error {
+	tmpFile := filename + ".tmp"
+
 	// Create the file, but give it a tmp file extension, this means we won't overwrite a
 	// file until it's downloaded, but we'll remove the tmp extension once downloaded.
-	out, err := os.Create(filename + ".tmp")
+	out, err := os.Create(tmpFile)
```
gdbranco: If it is a temp file might be better to use CreateTemp instead :thinking:

pkg/helper/download/download.go
```diff
@@ -15,27 +16,52 @@ import (
...
 	if err != nil {
-		out.Close()
-		return err
+		cleanupTempFile()
+		return formatDownloadError(err, url)
 	}
 	defer resp.Body.Close()
 
+	// Check HTTP response status code
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
```
gdbranco: Could also be statusCode / 100 != 2
Not sure if that is more confusing or not, but it is just a single comparison instead

pkg/helper/download/download.go
olucasfreitas: Addressed here [c4eb4cd](https://github.com/openshift/rosa/pull/3048/commits/c4eb4cd031e0d2ef26c619fdeac798964187a9b4)

pkg/helper/download/download.go
olucasfreitas: Addressed here [c4eb4cd](https://github.com/openshift/rosa/pull/3048/commits/c4eb4cd031e0d2ef26c619fdeac798964187a9b4)

PR #3051, opened 2025-10-20
OCM-19565 | feat: Deprecate available upgrades for `rosa list versions`

hunterkepley:
Adds deprecation warnings for `rosa list versions`; recommending the `rosa list upgrades` instead for listing cluster-based upgrade recommendations

### DIFF
--- a/cmd/list/version/cmd.go
+++ b/cmd/list/version/cmd.go
@@ -40,7 +40,8 @@ var Cmd = &cobra.Command{
 	Use:     "versions",
 	Aliases: []string{"version"},
 	Short:   "List available versions",
-	Long:    "List versions of OpenShift that are available for creating clusters.",
+	Long: "List versions of OpenShift that are available for creating clusters.\n\nNOTE: Available upgrades " +
+		"shown in this command are deprecated.",
 	Example: `  # List all OpenShift versions
   rosa list versions`,
 	Run:  run,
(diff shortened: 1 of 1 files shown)


cmd/list/version/cmd.go
```diff
@@ -40,9 +40,14 @@ var Cmd = &cobra.Command{
...
-	Long:    "List versions of OpenShift that are available for creating clusters.",
+	Long: "List versions of OpenShift that are available for creating clusters.\n\nNOTE: Available upgrades " +
+		"shown in this command are deprecated. Use 'rosa list upgrades --cluster <cluster_id>' for cluster-specific " +
+		"upgrade information.",
 	Example: `  # List all OpenShift versions
-  rosa list versions`,
+  rosa list versions
+  
+  # List available upgrades for a specific cluster (recommended)
+  rosa list upgrades --cluster <cluster_id>`,
```
yuwang-RH: @hunterkepley Do we want to add this recommended command of 'rosa list upgrades -cluster' in the help message of "rosa list versions" as an example? Maybe this is better to update in some document for guide, rather in the help message of a different command.

hunterkepley: Hey @yuwang-RH - from the DDR we want to tell users about this recommended command from within the `list versions` command

https://issues.redhat.com/browse/OCM-19565

The ticket description has a snippet from the DDR, I do however think that maybe the help message is a bit much, I will keep it within only the deprecation warning

PR #3052, opened 2025-10-21
OCM-18900 | fix: Improve proxy validation error messages for special characters in passwords

olucasfreitas:
# Details

This PR fixes a usability issue where `rosa create cluster --http-proxy` was displaying unhelpful error messages when proxy passwords contained special characters like forward slashes, making it difficult for users to understand what was wrong with their input.

---

## Problem (Before)

```bash
rosa create cluster --http-proxy "http://proxyuser:QvoZjyy/trkCiY5@10.0.0.161:8080"
```

Output:
```
E: Invalid http-proxy value 'http://proxyuser:QvoZjyy/trkCiY5@10.0.0.161:8080'
```

The user has no idea what's wrong with the URL.

---

## Fixed Behavior (After)

```bash
./rosa create cluster --http-proxy "http://proxyuser:QvoZjyy/trkCiY5@10.0.0.161:8080"
```

Output:
```
ERR: invalid http-proxy value: password contains invalid character '/'
```

Now the user knows exactly what's wrong.

---

## Additional Validations

The implementation validates special characters in both username and password fields:

1. **Forward slash in password**:
    ```bash
    rosa edit cluster --cluster=my-cluster --http-proxy "http://user:pass/word@10.0.0.161:8080"
    ```
    Output:
    ```
    ERR: invalid http-proxy value: password contains invalid character '/'
    ```

2. **@ symbol in password**:
    ```bash
    rosa edit cluster --cluster=my-cluster --http-proxy "http://user:p@ssword@10.0.0.161:8080"
    ```
    Output:
    ```
    ERR: invalid http-proxy value: password contains invalid character '@'
    ```

3. **Question mark in password**:
    ```bash
    rosa edit cluster --cluster=my-cluster --http-proxy "http://user:pass?word@10.0.0.161:8080"
    ```
    Output:
    ```
    ERR: invalid http-proxy value: password contains invalid character '?'
    ```

4. **Hash in password**:
    ```bash
    rosa edit cluster --cluster=my-cluster --http-proxy "http://user:pass#word@10.0.0.161:8080"
    ```
    Output:
    ```
    ERR: invalid http-proxy value: password contains invalid character '#'
    ```

5. **Invalid character in username**:
    ```bash
    rosa edit cluster --cluster=my-cluster --http-proxy "http://us/er:password@10.0.0.161:8080"
    ```
    Output:
    ```
    ERR: invalid http-proxy value: username contains invalid character '/'
    ```

6. **Missing scheme**:
    ```bash
    rosa edit cluster --cluster=my-cluster --http-proxy "user:password@10.0.0.161:8080"
    ```
    Output:
    ```
    ERR: invalid http-proxy value: URL is missing scheme (expected '://')
    ```

7. **HTTPS proxy validation**:
    ```bash
    rosa edit cluster --cluster=my-cluster --https-proxy "https://user:pass/word@10.0.0.161:8080"
    ```
    Output:
    ```
    ERR: invalid https-proxy value: password contains invalid character '/'
    ```

---

# Ticket

Closes [OCM-18900](https://issues.redhat.com/browse/OCM-18900)

### DIFF
--- a/pkg/helper/url/validation.go
+++ b/pkg/helper/url/validation.go
@@ -1,0 +1,63 @@
+/*
+Copyright (c) 2025 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package url
+
+import (
+	"fmt"
+	"strings"
+)
+
+func ValidateURLCredentials(urlString string) error {
+	parts := strings.SplitN(urlString, "://", 2)
+	if len(parts) < 2 {
+		return nil
+	}
+	rest := parts[1]
+
+	if strings.Count(rest, "@") > 1 {
+		return fmt.Errorf("password contains invalid character '@'")
+	}
+
+	userinfo, _, found := strings.Cut(rest, "@")
+	if !found {
+		return nil
+	}
+
+	username, password, hasPassword := strings.Cut(userinfo, ":")
+
+	if err := checkForInvalidChars(username, "username"); err != nil {
+		return err
+	}
+
+	if !hasPassword {
+		return nil
+	}
+
+	return checkForInvalidChars(password, "password")
+}
+
+func checkForInvalidChars(value, field string) error {
+	invalidChars := "/:#?[]@!$&'()*+,;="
+
+	for _, char := range value {
+		if strings.ContainsRune(invalidChars, char) {
+			return fmt.Errorf("%s contains invalid character '%c'", field, char)
+		}
+	}
+
+	return nil
+}
--- a/pkg/helper/url/validation_test.go
+++ b/pkg/helper/url/validation_test.go
@@ -1,0 +1,107 @@
+/*
+Copyright (c) 2025 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package url
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestURLValidation(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "URL Validation Suite")
+}
+
+var _ = Describe("ValidateURLCredentials", func() {
+	Context("when URL has no scheme separator", func() {
+		It("returns nil for URL without scheme", func() {
+			err := ValidateURLCredentials("example.com")
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("returns nil for URL with partial scheme", func() {
+			err := ValidateURLCredentials("http:/example.com")
+			Expect(err).NotTo(HaveOccurred())
+		})
+	})
+
+	Context("when URL has no credentials", func() {
+		It("returns nil for URL without @", func() {
+			err := ValidateURLCredentials("http://example.com")
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("returns nil for URL with port but no credentials", func() {
+			err := ValidateURLCredentials("http://example.com:8080")
+			Expect(err).NotTo(HaveOccurred())
+		})
+	})
+
+	Context("when URL has valid credentials", func() {
+		It("returns nil for URL with username only", func() {
+			err := ValidateURLCredentials("http://user@example.com")
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("returns nil for URL with username and password", func() {
+			err := ValidateURLCredentials("http://user:pass@example.com")
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("returns nil for URL with empty password", func() {
+			err := ValidateURLCredentials("http://user:@example.com")
+			Expect(err).NotTo(HaveOccurred())
+		})
+	})
+
+	Context("when username contains invalid characters", func() {
+		DescribeTable("returns error for invalid username character",
+			func(url string, expectedChar rune) {
+				err := ValidateURLCredentials(url)
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(Equal("username contains invalid character '" + string(expectedChar) + "'"))
+			},
+			Entry("slash in username", "http://us/er:pass@example.com", '/'),
+			Entry("question mark in username", "http://us?er:pass@example.com", '?'),
+			Entry("hash in username", "http://us#er:pass@example.com", '#'),
+		)
+	})
+
+	Context("when password contains invalid characters", func() {
+		DescribeTable("returns error for invalid password character",
+			func(url string, expectedChar rune) {
+				err := ValidateURLCredentials(url)
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(Equal("password contains invalid character '" + string(expectedChar) + "'"))
+			},
+			Entry("slash in password", "http://user:pa/ss@example.com", '/'),
+			Entry("question mark in password", "http://user:pa?ss@example.com", '?'),
+			Entry("hash in password", "http://user:pa#ss@example.com", '#'),
+			Entry("bracket in password", "http://user:pa[ss@example.com", '['),
+		)
+	})
+
+	Context("when URL has multiple @ signs", func() {
+		It("returns error indicating @ in password", func() {
+			err := ValidateURLCredentials("http://user:p@ss@example.com")
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(Equal("password contains invalid character '@'"))
+		})
+	})
+})
--- a/pkg/interactive/validation_test.go
+++ b/pkg/interactive/validation_test.go
@@ -131,7 +131,7 @@ var _ = Describe("SubnetsValidator", func() {
 
 			err := validator(answers)
 			Expect(err).To(HaveOccurred())
-			Expect(err.Error()).To(ContainSubstring("The number of subnets for a public hosted " +
+			Expect(err.Error()).To(ContainSubstring("the number of subnets for a public hosted " +
 				"cluster should be at least two"))
 		})
 
@@ -153,7 +153,7 @@ var _ = Describe("SubnetsValidator", func() {
 
 			err := validator(answers)
 			Expect(err).To(HaveOccurred())
-			Expect(err.Error()).To(ContainSubstring("Must have at least one public subnet when not using both " +
+			Expect(err.Error()).To(ContainSubstring("must have at least one public subnet when not using both " +
 				"'private API' and 'private ingress'"))
 		})
 
@@ -169,7 +169,7 @@ var _ = Describe("SubnetsValidator", func() {
 
 			err := validator(answers)
 			Expect(err).To(HaveOccurred())
-			Expect(err.Error()).To(ContainSubstring("The number of subnets for a 'multi-AZ'"+
+			Expect(err.Error()).To(ContainSubstring("the number of subnets for a 'multi-AZ'"+
 				" 'private link cluster' should be '3', instead received: '%d'", len(answers)))
 		})
 
--- a/pkg/ocm/helpers.go
+++ b/pkg/ocm/helpers.go
@@ -43,6 +43,7 @@ import (
 
 	"github.com/openshift/rosa/pkg/aws"
 	"github.com/openshift/rosa/pkg/helper"
+	urlHelper "github.com/openshift/rosa/pkg/helper/url"
 	"github.com/openshift/rosa/pkg/output"
 	"github.com/openshift/rosa/pkg/reporter"
 )
@@ -108,9 +109,9 @@ func ClusterNameValidator(name interface{}) error {
 	if str, ok := name.(string); ok {
 		str := strings.Trim(str, " \t")
 		if !IsValidClusterName(str) {
-			return fmt.Errorf("Cluster name must consist of no more than %d lowercase "+
+			return fmt.Errorf("cluster name must consist of no more than %d lowercase "+
 				"alphanumeric characters or '-', start with a letter, and end with an "+
-				"alphanumeric character.", MaxClusterNameLength)
+				"alphanumeric character", MaxClusterNameLength)
 		}
 		return nil
 	}
@@ -125,13 +126,48 @@ func ClusterDomainPrefixValidator(domainPrefix interface{}) error {
 	if str, ok := domainPrefix.(string); ok {
 		str := strings.Trim(str, " \t")
 		if str != "" && !IsValidClusterDomainPrefix(str) {
-			return fmt.Errorf("Cluster domain prefix must consist of no more than %d lowercase "+
+			return fmt.Errorf("cluster domain prefix must consist of no more than %d lowercase "+
 				"alphanumeric characters or '-', start with a letter, and end with an "+
-				"alphanumeric character.", MaxClusterDomainPrefixLength)
+				"alphanumeric character", MaxClusterDomainPrefixLength)
 		}
 		return nil
 	}
 	return fmt.Errorf("can only validate strings, got '%v'", domainPrefix)
+}
+
+// validateProxyURL validates common proxy URL requirements
+func validateProxyURL(proxyURL string, proxyType string, expectedScheme string) error {
+	if strings.Contains(proxyURL, " ") {
+		return fmt.Errorf("invalid %s value: URL cannot contain spaces", proxyType)
+	}
+
+	if !strings.Contains(proxyURL, "://") {
+		return fmt.Errorf("invalid %s value: URL is missing scheme (expected '://')", proxyType)
+	}
+
+	if err := urlHelper.ValidateURLCredentials(proxyURL); err != nil {
+		return fmt.Errorf("invalid %s value: %s", proxyType, err)
+	}
+
+	parsedURL, err := url.Parse(proxyURL)
+	if err != nil {
+		return fmt.Errorf("invalid %s value: %v", proxyType, err)
+	}
+
+	if parsedURL.Scheme != expectedScheme {
+		return fmt.Errorf("expected '%s' to have an '%s://' scheme", proxyType, expectedScheme)
+	}
+
+	if parsedURL.Host == "" {
+		return fmt.Errorf("invalid %s value: host is missing", proxyType)
+	}
+
+	// Proxy URLs shouldn't have a path
+	if parsedURL.Path != "" && parsedURL.Path != "/" {
+		return fmt.Errorf("invalid %s value: proxy URL should not contain a path '%s'", proxyType, parsedURL.Path)
+	}
+
+	return nil
 }
 
 func ValidateHTTPProxy(val interface{}) error {
@@ -139,14 +175,17 @@ func ValidateHTTPProxy(val interface{}) error {
 		if httpProxy == "" {
 			return nil
 		}
-		url, err := url.ParseRequestURI(httpProxy)
-		if err != nil {
-			return fmt.Errorf("Invalid http-proxy value '%s'", httpProxy)
+		return validateProxyURL(httpProxy, "http-proxy", "http")
+	}
+	return fmt.Errorf("can only validate strings, got '%v'", val)
+}
+
+func ValidateHTTPSProxy(val interface{}) error {
+	if httpsProxy, ok := val.(string); ok {
+		if httpsProxy == "" {
+			return nil
 		}
-		if url.Scheme != "http" {
-			return errors.Errorf("%s", "Expected http-proxy to have an http:// scheme")
-		}
-		return nil
+		return validateProxyURL(httpsProxy, "https-proxy", "https")
 	}
 	return fmt.Errorf("can only validate strings, got '%v'", val)
 }
@@ -174,9 +213,9 @@ func ValidateAllowedRegistriesForImport(input interface{}) error {
 		registries = strings.Split(input.(string), ",")
 		for _, registry := range registries {
 			if !idRE.MatchString(registry) {
-				return fmt.Errorf("invalid identifier '%s' for 'allowed registries for import.' "+
+				return fmt.Errorf("invalid identifier '%s' for 'allowed registries for import'. "+
 					"Should be in a <registry>:<boolean> format. "+
-					"The boolean indicates whether the registry is secure or not.", registry)
+					"The boolean indicates whether the registry is secure or not", registry)
 			}
 		}
 	case reflect.Slice:
@@ -745,7 +784,7 @@ func ValidateSubnetsCount(multiAZ bool, privateLink bool, subnetsInputCount int)
 		if multiAZ {
 			azPrefix = "multi-AZ"
 		}
-		return fmt.Errorf("The number of subnets for a '%s' '%s' should be '%d', "+
+		return fmt.Errorf("the number of subnets for a '%s' '%s' should be '%d', "+
 			"instead received: '%d'", azPrefix, clusterPrefix, expected, subnetsInputCount)
 	}
 	return nil
@@ -755,10 +794,10 @@ func ValidateHostedClusterSubnets(awsClient aws.Client, isPrivate bool, subnetID
 	privateIngress bool) (int, error) {
 
 	if isPrivate && len(subnetIDs) < 1 {
-		return 0, fmt.Errorf("The number of subnets for a private hosted cluster should be at least one")
+		return 0, fmt.Errorf("the number of subnets for a private hosted cluster should be at least one")
 	}
 	if !isPrivate && len(subnetIDs) < 2 {
-		return 0, fmt.Errorf("The number of subnets for a public hosted cluster should be at least two")
+		return 0, fmt.Errorf("the number of subnets for a public hosted cluster should be at least two")
 	}
 	vpcSubnets, vpcSubnetsErr := awsClient.GetVPCSubnets(subnetIDs[0])
 	if vpcSubnetsErr != nil {
@@ -785,11 +824,11 @@ func ValidateHostedClusterSubnets(awsClient aws.Client, isPrivate bool, subnetID
 
 	if isPrivate && privateIngress {
 		if publicSubnetsCount > 0 {
-			return 0, fmt.Errorf("The number of public subnets for a private hosted cluster should be zero")
+			return 0, fmt.Errorf("the number of public subnets for a private hosted cluster should be zero")
 		}
 	} else {
 		if publicSubnetsCount == 0 {
-			return 0, fmt.Errorf("Must have at least one public subnet when not using both " +
+			return 0, fmt.Errorf("must have at least one public subnet when not using both " +
 				"'private API' and 'private ingress'")
 		}
 	}
@@ -1106,8 +1145,8 @@ func ValidateClaimValidationRules(input interface{}) error {
 		inputRules = strings.Split(input.(string), ",")
 		for _, inputRule := range inputRules {
 			if !idRE.MatchString(inputRule) {
-				return fmt.Errorf("invalid identifier '%s' for 'claim validation rule. '"+
-					"Should be in a <claim>:<required_value> format.", inputRule)
+				return fmt.Errorf("invalid identifier '%s' for 'claim validation rule'. "+
+					"Should be in a <claim>:<required_value> format", inputRule)
 			}
 		}
 	case reflect.Slice:
--- a/pkg/ocm/helpers_test.go
+++ b/pkg/ocm/helpers_test.go
@@ -395,7 +395,7 @@ var _ = Describe("ValidateSubnetsCount", func() {
 				Expect(err).To(HaveOccurred())
 				Expect(
 					err.Error(),
-				).To(Equal(fmt.Sprintf("The number of subnets for a 'multi-AZ' 'private link cluster' should be"+
+				).To(Equal(fmt.Sprintf("the number of subnets for a 'multi-AZ' 'private link cluster' should be"+
 					" '%d', instead received: '%d'", privateLinkMultiAZSubnetsCount, privateLinkMultiAZSubnetsCount+1)))
 			})
 
@@ -411,7 +411,7 @@ var _ = Describe("ValidateSubnetsCount", func() {
 				Expect(err).To(HaveOccurred())
 				Expect(
 					err.Error(),
-				).To(Equal(fmt.Sprintf("The number of subnets for a 'single AZ' 'private link cluster' should be"+
+				).To(Equal(fmt.Sprintf("the number of subnets for a 'single AZ' 'private link cluster' should be"+
 					" '%d', instead received: '%d'", privateLinkSingleAZSubnetsCount, privateLinkSingleAZSubnetsCount+1)))
 			})
 
@@ -427,7 +427,7 @@ var _ = Describe("ValidateSubnetsCount", func() {
 			It("should return an error if subnetsInputCount is not equal to BYOVPCMultiAZSubnetsCount", func() {
 				err := ValidateSubnetsCount(true, false, BYOVPCMultiAZSubnetsCount+1)
 				Expect(err).To(HaveOccurred())
-				Expect(err.Error()).To(Equal(fmt.Sprintf("The number of subnets for a 'multi-AZ' 'cluster' should be"+
+				Expect(err.Error()).To(Equal(fmt.Sprintf("the number of subnets for a 'multi-AZ' 'cluster' should be"+
 					" '%d', instead received: '%d'", BYOVPCMultiAZSubnetsCount, BYOVPCMultiAZSubnetsCount+1)))
 			})
 
@@ -441,7 +441,7 @@ var _ = Describe("ValidateSubnetsCount", func() {
 			It("should return an error if subnetsInputCount is not equal to BYOVPCSingleAZSubnetsCount", func() {
 				err := ValidateSubnetsCount(false, false, BYOVPCSingleAZSubnetsCount+1)
 				Expect(err).To(HaveOccurred())
-				Expect(err.Error()).To(Equal(fmt.Sprintf("The number of subnets for a 'single AZ' 'cluster' should"+
+				Expect(err.Error()).To(Equal(fmt.Sprintf("the number of subnets for a 'single AZ' 'cluster' should"+
 					" be '%d', instead received: '%d'", BYOVPCSingleAZSubnetsCount, BYOVPCSingleAZSubnetsCount+1)))
 			})
 
@@ -476,7 +476,7 @@ var _ = Describe("ValidateHostedClusterSubnets for Private Cluster", func() {
 		mockClient.EXPECT().FilterVPCsPrivateSubnets(gomock.Any()).Return([]ec2types.Subnet{}, nil)
 		_, err := ValidateHostedClusterSubnets(mockClient, true, ids, true)
 		Expect(err).To(HaveOccurred())
-		Expect(err.Error()).To(Equal("The number of public subnets for a private hosted cluster should be zero"))
+		Expect(err.Error()).To(Equal("the number of public subnets for a private hosted cluster should be zero"))
 	})
 })
 
@@ -587,7 +587,7 @@ var _ = Describe("ValidateClaimValidationRules()", func() {
 		Entry("should error when a non string arg is given", 5, true, "can only validate string types, got int"),
 		Entry("should not error when an empty claim validation rule is given", "", false, ""),
 		Entry("shoud error when claim validation rule is not a valid value", "9hjh9", true,
-			"invalid identifier '9hjh9' for 'claim validation rule. 'Should be in a <claim>:<required_value> format."),
+			"invalid identifier '9hjh9' for 'claim validation rule'. Should be in a <claim>:<required_value> format"),
 		Entry("should not error when claim validation rule with single pair is valid", "abc:efg", false, ""),
 		Entry("should not error when claim validation rule with multiple pairs is valid", "abc:efg,lala:wuwu", false, ""))
 })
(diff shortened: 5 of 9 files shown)


pkg/ocm/helpers_test.go
```diff
@@ -695,3 +695,43 @@ var _ = Describe("GetAutoNodeRoleArn", func() {
...
+	})
+
+	It("rejects password with forward slash and provides helpful error", func() {
+		err := ValidateHTTPProxy("http://proxyuser:QvoZjyy/trkCiY5@10.0.0.161:8080")
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("must be URL-encoded"))
+	})
+
+	It("accepts password with URL-encoded forward slash", func() {
+		err := ValidateHTTPProxy("http://proxyuser:QvoZjyy%2FtrkCiY5@10.0.0.161:8080")
```
gdbranco, line 761: Wondering if system should URL encode it or expect customer to do it

olucasfreitas: I think the system should encode it, because if not we would be introducing a new step for the costumer creating maybe some disruption and even opening the system up for bugs, if we encode it ourselves, we can ensure patterns and so on, WDYT ? @gdbranco

hunterkepley: I think we should URL encode, `curl` for example does not handle non-URL-encoded characters well on its own (unless they are non-ascii). So in this case, the `@` may cause a failure with `curl` (or another tool, browser, etc)

olucasfreitas: Me neither, @gdbranco do you know of any patterns to deal with this ?

cmd/list/dnsdomains/cmd.go
```diff
@@ -53,7 +53,7 @@ func init() {
 		"all",
 		"a",
 		false,
-		"List all DNS domains (default lists just user defined).",
+		"List all DNS domains across organizations (for support).",
```
gdbranco: Unrelated file change?

pkg/ocm/helpers_test.go
gdbranco: I don't think there is such a known pattern example within the codebase, but URL encoding in itself should have some docs on it https://pkg.go.dev/net/url#QueryEscape
```
paramValue := "user id with spaces & special characters!"
	encodedParam := url.QueryEscape(paramValue)
	fmt.Println("Encoded parameter:", encodedParam)
	// Output: Encoded parameter: user+id+with+spaces+%26+special+characters%21
```

cmd/list/dnsdomains/cmd.go
olucasfreitas: Yes, gonna remove that

cmd/list/dnsdomains/cmd.go
```diff
@@ -71,18 +71,19 @@ func run(_ *cobra.Command, _ []string) {
...
-		_ = r.Reporter.Errorf("Failed to get current organization: %s", err)
+		_ = r.Reporter.Errorf("Failed to get current organization: %v", err)
 		os.Exit(1)
 	}
-	search := fmt.Sprintf("user_defined='true' AND organization.id='%s'", orgID)
+
+	search := fmt.Sprintf("user_defined='true' and organization.id='%s'", organizationID)
 	if args.all {
-		search = ""
+		search = fmt.Sprintf("organization.id='%s'", organizationID)
```
hunterkepley: Changes in this file are unrelated, they seem related to https://github.com/openshift/rosa/pull/3047 instead

tests/utils/exec/rosacli/upgrade_service.go
hunterkepley, line 59: im not sure we are doing any work on `list/upgrade` - the ticket OCM-18900 seems to deal with a bug for `create/cluster` proxy field which contain `/`s

pkg/ocm/helpers.go
```diff
@@ -139,12 +139,24 @@ func ValidateHTTPProxy(val interface{}) error {
...
 			return nil
 		}
-		url, err := url.ParseRequestURI(httpProxy)
+
+		// Encode password in place
+		if i := strings.LastIndex(httpProxy, "@"); i > 0 {
+			if j := strings.LastIndex(httpProxy[:i], ":"); j > 0 {
+				httpProxy = httpProxy[:j+1] + url.QueryEscape(httpProxy[j+1:i]) + httpProxy[i:]
+			}
+		}
```
hunterkepley: `LastIndex(..., "@")` might cause an issue if there were an `@` in the password

tests/utils/exec/rosacli/upgrade_service.go
olucasfreitas: I had to solve some conflicts here and merged some content into this branch, did I do it incorrectly then ?

hunterkepley: I am guessing so, we don't want changes unrelated to the bug you are trying to fix

pkg/ocm/helpers.go
olucasfreitas: Worked on that on the last commit

pkg/ocm/helpers_test.go
olucasfreitas: Addressed those in the last commit

hunterkepley (commented):
The test case in the bug ticket does not seem to pass:

`E: parse "http://proxyuser:QvoZjyy/trkCiY5@10.0.0.161:8080": invalid port ":QvoZjyy" after host`

I think there may be a bug in the parse code

tests/e2e/test_rosacli_cluster.go
hunterkepley, line 667: I think these changes are from a different MR Yu Wang put up

pkg/ocm/helpers_test.go
```diff
@@ -695,3 +695,90 @@ var _ = Describe("GetAutoNodeRoleArn", func() {
...
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("accepts valid http proxy with basic authentication", func() {
+		err := ValidateHTTPProxy("http://user:password@proxy.example.com:8080")
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("rejects http proxy with unencoded forward slash in password and provides helpful error", func() {
+		// This is the exact example from the bug report
```
hunterkepley: Maybe put the ticket number in, when someone looks at this in a year they won't know what `this` nor `the bug report` is

pkg/ocm/helpers.go
```diff
@@ -139,13 +139,67 @@ func ValidateHTTPProxy(val interface{}) error {
...
+			return fmt.Errorf("Invalid http-proxy value '%s': URL cannot contain spaces", httpProxy)
+		}
+
+		// Check for multiple @ symbols which indicates unencoded @ in password
+		if strings.Count(httpProxy, "@") > 1 {
+			return fmt.Errorf("Invalid http-proxy value '%s': password contains special characters that need URL encoding", httpProxy)
+		}
+
+		// Check if the URL contains user credentials with special characters before parsing
+		if strings.Contains(httpProxy, "@") && strings.Contains(httpProxy, "://") {
```
hunterkepley: Is there a reason we have 2 different yet very similar functions to parse and validate these proxies? (in `pkg/interactive/validation.go` as well as this file)

pkg/ocm/helpers.go
```diff
@@ -139,13 +139,67 @@ func ValidateHTTPProxy(val interface{}) error {
...
+			return fmt.Errorf("Invalid http-proxy value '%s': host is missing", httpProxy)
+		}
+
+		if parsedURL.Port() != "" {
+			port, err := strconv.Atoi(parsedURL.Port())
+			if err != nil {
+				return fmt.Errorf("Invalid http-proxy value '%s': invalid port number", httpProxy)
+			}
+			if port < 1 || port > 65535 {
+				return fmt.Errorf("Invalid http-proxy value '%s': port must be between 1 and 65535", httpProxy)
```
hunterkepley: The following manual test case scenarios didn't seem to have the correct behavior when I tried:

`--http-proxy http://proxyuser:QvoZjyytrkCiY5@10/.0.0.161:9999999`

`--https-proxy http://proxyuser:QvoZjyytrkCiY5@10/.0.0.161:9999999`

^ Both of these (invalid port + invalid IP; invalid port + invalid IP + `http` used with `--https-proxy`) passed and attempted to create the cluster

pkg/ocm/helpers.go
```diff
@@ -139,13 +139,67 @@ func ValidateHTTPProxy(val interface{}) error {
...
+		if parsedURL.Host == "" {
+			return fmt.Errorf("Invalid http-proxy value '%s': host is missing", httpProxy)
+		}
+
+		if parsedURL.Port() != "" {
+			port, err := strconv.Atoi(parsedURL.Port())
+			if err != nil {
+				return fmt.Errorf("Invalid http-proxy value '%s': invalid port number", httpProxy)
+			}
+			if port < 1 || port > 65535 {
```
hunterkepley: The validations in this file seem to differ quite a bit from `pkg/interactive/validation.go` - do we want to keep both?

pkg/ocm/helpers.go
```diff
@@ -139,13 +139,67 @@ func ValidateHTTPProxy(val interface{}) error {
...
+		if strings.Contains(httpProxy, "@") && strings.Contains(httpProxy, "://") {
+			idx := strings.Index(httpProxy, "://")
+			rest := httpProxy[idx+3:]
+			atIdx := strings.LastIndex(rest, "@")
+			if atIdx != -1 {
+				userinfo := rest[:atIdx]
+				if colonIdx := strings.Index(userinfo, ":"); colonIdx != -1 {
+					password := userinfo[colonIdx+1:]
+					// Check if password contains special characters that need encoding
+					specialChars := "/:#?[]!$&'()*+,;="
```
hunterkepley: The special chars list differs from the other one in `pkg/interactive/validation.go`

olucasfreitas: Fixed this on the last commit

pkg/ocm/helpers.go
olucasfreitas: Thanks for testing, the problem was that some port numbers we're not always caught, fixed that in the last commit

pkg/ocm/helpers.go
olucasfreitas: `ValidateHTTPProxy` is for validating proxy-specific rules like requiring http:// scheme and port validation, while `IsURL` was for general URL validation in interactive prompts that can accept any scheme including https://, that's why I left both of them there

pkg/ocm/helpers_test.go
olucasfreitas: Did it in the last commit

tests/e2e/test_rosacli_cluster.go
olucasfreitas: Going to take a look at this, I thought I had already fixed it

pkg/ocm/helpers.go
olucasfreitas: I kinda answered this on the other comment, the TLDR is that they serve different purposes, unless you think that we may move all of this into one validation place

pkg/ocm/helpers.go
hunterkepley: Some port numbers weren't always caught indeed, but what was the problem?

pkg/ocm/helpers.go
hunterkepley: Why are we duplicating it?

pkg/ocm/helpers.go
hunterkepley: im not too sure I see the same thing you're saying with them, you said `ValidateHTTPProxy is for validating proxy-specific rules like requiring http:// scheme and port validation, while IsURL was for general URL validation in interactive prompts`

`_IsUrl` isn't used for interactive prompts really, but instead is used across the board for most commands which take in a URL or proxy

Im not sure I understand what you mean by `requiring http:// scheme` in `ValidateHTTPProxy` - it's below the duplicated code sure, but it's not in the duplicated code. It was also in the file prior to your code changes, so it should be unrelated when discussing the code I'm talking about

To me, it looks like the main meat of both validations in both files are identical, and we can probably reduce line count from them, WDYT?

pkg/ocm/helpers.go
```diff
@@ -139,13 +139,89 @@ func ValidateHTTPProxy(val interface{}) error {
 		if httpProxy == "" {
 			return nil
 		}
-		url, err := url.ParseRequestURI(httpProxy)
+
+		// Check for spaces
+		if strings.Contains(httpProxy, " ") {
+			return fmt.Errorf("Invalid http-proxy value '%s': URL cannot contain spaces", httpProxy)
+		}
```
hunterkepley: this gets a little messy when we consider we are asking users to provide URL encoded URLs

This, for example, doesn't check for encoded spaces- and all of this validation becomes mostly useless once they encode

I'm wondering, if it may be a better idea to stick closer to the original ask of the ticket, and maybe not ask users to URL encode specifically. The original ask of the ticket is to tell users what is wrong with the URL they put in; I don't think it mentions telling the user to URL encode if they have an invalid character in their URL. We probably don't want them to URL encode if there's an invalid character, because then we are skipping the CLI and sending the bad value straight to the API

hunterkepley (commented):
left some more comments- 

Just so we can save some time, and stick closer to the original ask, and instead of telling the user `X contains special characters that need URL encoding` - maybe it's a little more true-to-the-ticket to say something more along the lines of `X contains invalid character '%v': %s` and print out the specific part which has an invalid character, and which character it is

In your errors you print out the whole string, and don't give much detail besides which section contains 'special characters which require URL encoding'

The ticket is mores asking you to provide extra detail in this area, so I think telling the user 1) which special character they included is invalid, 2) the section (you mostly have this, `password contains ...` for example), and 3) print out the URL after this

Think of the difference of UX between these 3 messages:


```go
Invalid URL 'http://proxyuser:QvoZjyy/trkCiY5@10.0.0.161:8080': password contains special characters that need URL encoding
```

```go
Invalid http-proxy value 'http://proxyuser:QvoZjyy/trkCiY5@10.0.0.161:8080'
```

Besides the fact the user knows what section the issue is in, they don't know which character is messed up, and the potentially quite long URL is now in the _middle_ of this error msg

Take this by comparison

```go
Invalid http-proxy value: password 'QvoZjyy/trkCiY5' contains invalid character '/'
```

I think this accomplishes a lot more than either of the original 2 messages. You also have the infrastructure here to accomplish this sort of verbose messaging

 1) It tells the user what is wrong, where is wrong, and why is wrong. 2) It doesn't suggest a fix-all solution of URL-encoding. There are still some unsupported characters beyond this. 3) Jumping off of point number 2, it gives the user the freedom of choice in how they want to handle this.

 Maybe they just want to get rid of the character because it was a typo, or it's easier than URL encoding every time, or maybe they wouldn't know of the issue until it hit the API if we had an actual invalid character included, such as `:`, and they URL-encoded so it skipped our validation here in the client

And maybe we shouldn't print the password! Maybe it doesn't matter, maybe it does, WDYT?

olucasfreitas:
> left some more comments-
> 
> Just so we can save some time, and stick closer to the original ask, and instead of telling the user `X contains special characters that need URL encoding` - maybe it's a little more true-to-the-ticket to say something more along the lines of `X contains invalid character '%v': %s` and print out the specific part which has an invalid character, and which character it is
> 
> In your errors you print out the whole string, and don't give much detail besides which section contains 'special characters which require URL encoding'
> 
> The ticket is mores asking you to provide extra detail in this area, so I think telling the user 1) which special character they included is invalid, 2) the section (you mostly have this, `password contains ...` for example), and 3) print out the URL after this
> 
> Think of the difference of UX between these 3 messages:
> 
> ```go
> Invalid URL 'http://proxyuser:QvoZjyy/trkCiY5@10.0.0.161:8080': password contains special characters that need URL encoding
> ```
> 
> ```go
> Invalid http-proxy value 'http://proxyuser:QvoZjyy/trkCiY5@10.0.0.161:8080'
> ```
> 
> Besides the fact the user knows what section the issue is in, they don't know which character is messed up, and the potentially quite long URL is now in the _middle_ of this error msg
> 
> Take this by comparison
> 
> ```go
> Invalid http-proxy value: password 'QvoZjyy/trkCiY5' contains invalid character '/'
> ```
> 
> I think this accomplishes a lot more than either of the original 2 messages. You also have the infrastructure here to accomplish this sort of verbose messaging
> 
> 1. It tells the user what is wrong, where is wrong, and why is wrong. 2) It doesn't suggest a fix-all solution of URL-encoding. There are still some unsupported characters beyond this. 3) Jumping off of point number 2, it gives the user the freedom of choice in how they want to handle this.
> 
> Maybe they just want to get rid of the character because it was a typo, or it's easier than URL encoding every time, or maybe they wouldn't know of the issue until it hit the API if we had an actual invalid character included, such as `:`, and they URL-encoded so it skipped our validation here in the client
> 
> And maybe we shouldn't print the password! Maybe it doesn't matter, maybe it does, WDYT?

I do agree with you, the UX is handled much more clearly in you example. About the password, I'm fine with not printing it

pkg/ocm/helpers.go
olucasfreitas: I duplicated it so we would have the same symbols on both cases

pkg/ocm/helpers.go
olucasfreitas: The issue was that url.Parse was interpreting the malformed URL `http://10/.0.0.161:9999999` wrongly, it parsed 10 as the host and `/.0.0.161:9999999` as the path, so the port validation never ran because it thought there was no port

pkg/ocm/helpers.go
olucasfreitas: Do agree with you, working on a fix for this to reduce the line count

pkg/ocm/helpers.go
olucasfreitas: I'm gonna stick with what we agreed on the bigger comment you left, that way we provide a clear error message for the user

hunterkepley (commented):
Did you test this?

I'm still finding cases where I am able to make a cluster with an invalid URI


```
--https-proxy http://123 <-- passed and created cluster 
--http-proxy https://123 <-- failed
--https-proxy http://proxyuser:QvoZjyytrkCiY5@10/.0.0.161:8080 <-- passed and created cluster
--https-proxy --https-proxy http://proxyuser:QvoZ/jyytrkCiY5@10.0.0.161:8080 <-- failed with proper error message
```

hunterkepley:
> About the password, I'm fine with not printing it

@olucasfreitas Sorry, I didn't mean specifically to do so; I was asking it rhetorically there: `"And maybe we shouldn't print the password! Maybe it doesn't matter, maybe it does, WDYT?"` to encourage a conversation on the complicated and long comment I left

olucasfreitas:
@hunterkepley My bad then, I understood this more as as a suggestions that a thing to initiate a discussion. Answering what you said, I think we can show the password, because this way the user can just do a quick comparison and figure out what's wrong, WDYT ?

I'm also working on the test cases you commented

pkg/helper/helpers.go
```diff
@@ -268,3 +268,67 @@ func FilterEmptyStrings(strings []string) []string {
...
+
+	rest := urlString[schemeIdx+3:]
+	atIdx := strings.LastIndex(rest, "@")
+	if atIdx == -1 {
+		return &URLCredentialValidation{HasInvalid: false}
+	}
+
+	userinfo := rest[:atIdx]
+
+	// Check for multiple @ symbols first (indicates @ in credentials)
```
hunterkepley: NP: unneeded comment

pkg/helper/helpers.go
```diff
@@ -268,3 +268,67 @@ func FilterEmptyStrings(strings []string) []string {
...
+	if strings.Count(urlString, "@") > 1 {
+		return &URLCredentialValidation{
+			Field:       "password",
+			InvalidChar: '@',
+			HasInvalid:  true,
+		}
+	}
+
+	// Check username and password separately
```
hunterkepley: NP: unneeded comment

pkg/interactive/validation.go
hunterkepley, line 87: this function is unrelated, this is currently only used for OIDC config issuer URLs. The side effect of editing this function means you now need to test + potentially modify the `register/oidcconfig` command/tests. Let's move your code outside of this function, so we avoid side effects

Image

hunterkepley: If we look at the ticket, we are told which function to worry about actually:

`The ParseRequestURI function used in ValidateHTTPProxy rejects the value which contains the forward slash  within the password field (QvoZjyy/trkCiY5).`

pkg/ocm/helpers.go
```diff
@@ -139,13 +139,142 @@ func ValidateHTTPProxy(val interface{}) error {
...
+				credValidation.Field, credValidation.InvalidChar)
+		}
+
+		parsedURL, err := url.Parse(httpProxy)
+		if err != nil {
+			// Check for invalid port in error message
+			if strings.Contains(err.Error(), "invalid port") {
+				return fmt.Errorf("invalid http-proxy value: invalid port number")
+			}
+			return fmt.Errorf("invalid http-proxy value: %v", err)
```
hunterkepley: If the error already contains `invalid port` why are we returning a different error entirely instead of wrapping with the below err?

hunterkepley: To me, it seems the error returned would already tell the user that their port is invalid, no?

pkg/ocm/helpers.go
```diff
@@ -139,13 +139,142 @@ func ValidateHTTPProxy(val interface{}) error {
...
+
+		// Validate hostname/IP and port
+		host, port, err := net.SplitHostPort(parsedURL.Host)
+		if err != nil {
+			// No port specified, just validate the host
+			host = parsedURL.Host
+		} else {
+			// Validate port
+			portNum, err := strconv.Atoi(port)
+			if err != nil {
```
hunterkepley: im confused, didn't we just check for invalid port above?

pkg/ocm/helpers.go
```diff
@@ -139,13 +139,142 @@ func ValidateHTTPProxy(val interface{}) error {
...
+				}
+			}
+		}
+
+		return nil
+	}
+	return fmt.Errorf("can only validate strings, got '%v'", val)
+}
+
+func ValidateHTTPSProxy(val interface{}) error {
```
hunterkepley, line 183: lots of duplicted code from the above func

pkg/ocm/helpers.go
```diff
@@ -174,9 +303,9 @@ func ValidateAllowedRegistriesForImport(input interface{}) error {
 		registries = strings.Split(input.(string), ",")
 		for _, registry := range registries {
 			if !idRE.MatchString(registry) {
-				return fmt.Errorf("invalid identifier '%s' for 'allowed registries for import.' "+
-					"Should be in a <registry>:<boolean> format. "+
-					"The boolean indicates whether the registry is secure or not.", registry)
+				return fmt.Errorf("invalid identifier '%s' for 'allowed registries for import' "+
+					"should be in a <registry>:<boolean> format; "+
+					"the boolean indicates whether the registry is secure or not", registry)
```
hunterkepley: is there a reason for these changes?

hunterkepley: I agree with the `.` being removed at the end, but the rest seem not needed to me

hunterkepley: The way this now reads is:

```swift
invalid identifier '%s' for 'allowed registries for import' should be in a <registry>:<boolean> format; the boolean indicates whether the registry is secure or not
```

VS

```swift
invalid identifier '%s' for 'allowed registries for import.' Should be in a <registry>:<boolean> format. The boolean indicates whether the registry is secure or not
```

To me, the new error is a bit hard to understand

pkg/ocm/helpers.go
```diff
@@ -1106,8 +1235,8 @@ func ValidateClaimValidationRules(input interface{}) error {
 		inputRules = strings.Split(input.(string), ",")
 		for _, inputRule := range inputRules {
 			if !idRE.MatchString(inputRule) {
-				return fmt.Errorf("invalid identifier '%s' for 'claim validation rule. '"+
-					"Should be in a <claim>:<required_value> format.", inputRule)
+				return fmt.Errorf("invalid identifier '%s' for 'claim validation rule' "+
+					"should be in a <claim>:<required_value> format", inputRule)
```
hunterkepley: Is there a reason for removing the `.` after `'claim validation rule'` and making it into one sentence?

pkg/ocm/helpers_test.go
```diff
@@ -587,7 +587,7 @@ var _ = Describe("ValidateClaimValidationRules()", func() {
 		Entry("should error when a non string arg is given", 5, true, "can only validate string types, got int"),
 		Entry("should not error when an empty claim validation rule is given", "", false, ""),
 		Entry("shoud error when claim validation rule is not a valid value", "9hjh9", true,
-			"invalid identifier '9hjh9' for 'claim validation rule. 'Should be in a <claim>:<required_value> format."),
+			"invalid identifier '9hjh9' for 'claim validation rule' should be in a <claim>:<required_value> format"),
```
hunterkepley: Same for this one

pkg/ocm/helpers_test.go
```diff
@@ -695,3 +695,200 @@ var _ = Describe("GetAutoNodeRoleArn", func() {
...
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("accepts valid http proxy with basic authentication", func() {
+		err := ValidateHTTPProxy("http://user:password@proxy.example.com:8080")
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("rejects http proxy with unencoded forward slash in password and provides helpful error", func() {
+		// This is the exact example from the bug report OCM-18900
```
hunterkepley: NP: unnecessary comment

pkg/ocm/registry_config_test.go
```diff
@@ -111,9 +111,9 @@ var _ = Describe("Registry Config", func() {
 		})
 		It("KO: should fail if the user does not pass the correct regex", func() {
 			_, err := BuildAllowedRegistriesForImport("abc")
-			Expect(err).To(MatchError("invalid identifier 'abc' for 'allowed registries for import.' " +
-				"Should be in a <registry>:<boolean> format. " +
-				"The boolean indicates whether the registry is secure or not."))
+			Expect(err).To(MatchError("invalid identifier 'abc' for 'allowed registries for import' " +
+				"should be in a <registry>:<boolean> format; " +
+				"the boolean indicates whether the registry is secure or not"))
```
hunterkepley: Same for this one

pkg/helper/helpers.go
olucasfreitas: fixed in last commit

pkg/helper/helpers.go
olucasfreitas: fixed in last commit

pkg/interactive/validation.go
olucasfreitas: tried fixing this in the last commit

pkg/ocm/helpers.go
olucasfreitas: fixed in last commit

pkg/ocm/helpers.go
olucasfreitas: Yep, fixed in last commit

pkg/ocm/helpers.go
olucasfreitas: fixed in last commit

pkg/ocm/helpers.go
olucasfreitas: fixed in last commit

pkg/ocm/helpers_test.go
olucasfreitas: fixed in last commit

pkg/interactive/validation.go
```diff
@@ -89,6 +89,9 @@ func IsURLHttps(val interface{}) error {
 	if err != nil {
 		return err
 	}
+	if parsedUri == nil {
+		return nil
+	}
```
hunterkepley: I think I've left 2 comments now asking about this 🤔 ; but why did you add this?

pkg/ocm/helpers.go
```diff
@@ -125,30 +125,93 @@ func ClusterDomainPrefixValidator(domainPrefix interface{}) error {
...
-			return nil
-		}
-		url, err := url.ParseRequestURI(httpProxy)
-		if err != nil {
-			return fmt.Errorf("Invalid http-proxy value '%s'", httpProxy)
+// validateProxyURL validates common proxy URL requirements
+func validateProxyURL(proxyURL string, proxyType string, expectedScheme string) error {
+	if proxyURL == "" {
+		return nil
+	}
```
hunterkepley: I left a comment last time, but why are we doing this?

I mentioned how this increases the stack if we call it every single time, but return a `nil` value. Please, let's only call `validateProxyURL` when the proxyURL isn't empty

pkg/ocm/helpers.go
```diff
@@ -125,30 +125,93 @@ func ClusterDomainPrefixValidator(domainPrefix interface{}) error {
...
+
+	// Check for spaces
```
hunterkepley: NP: I can tell from the line of code below what it does, no need for a comment about it

pkg/interactive/validation.go
```diff
@@ -106,8 +109,15 @@ func _isUrl(val interface{}) (*url.URL, error) {
...
-	return parsedUri, err
+
+	parsedUri, err := url.Parse(fmt.Sprintf("%v", val))
+	if err != nil {
+		return nil, err
+	}
+	if parsedUri.Scheme == "" || parsedUri.Host == "" {
+		return nil, fmt.Errorf("url must have a scheme and host")
+	}
+	return parsedUri, nil
```
hunterkepley: is this related to this feature?

pkg/interactive/validation.go
hunterkepley: doesn't seem fixed https://github.com/openshift/rosa/pull/3052/files#r2539671446

pkg/ocm/helpers.go
```diff
@@ -125,30 +125,93 @@ func ClusterDomainPrefixValidator(domainPrefix interface{}) error {
...
+
+func ValidateHTTPProxy(val interface{}) error {
+	httpProxy, ok := val.(string)
+	if !ok {
+		return fmt.Errorf("can only validate strings, got '%v'", val)
+	}
+	return validateProxyURL(httpProxy, "http-proxy", "http")
+}
+
+func ValidateHTTPSProxy(val interface{}) error {
```
hunterkepley, line 183: this is an `interface{}` type- what other types are we expecting here besides a string?

pkg/ocm/helpers.go
```diff
@@ -125,30 +125,93 @@ func ClusterDomainPrefixValidator(domainPrefix interface{}) error {
...
+	httpsProxy, ok := val.(string)
+	if !ok {
+		return fmt.Errorf("can only validate strings, got '%v'", val)
+	}
```
hunterkepley: we shouldn't tell the user this, right? how could they input anything besides a string?

pkg/interactive/validation.go
olucasfreitas: I added that because `_isUrl` can return nil when it gets an empty string or nil input, and without that check we'd maybe get a panic, so I wanted to be sure, sorry about this, as I got a lot of comments I must have let this one pass, just removed it on the last commit

pkg/ocm/helpers.go
olucasfreitas: Fixed in the last commit

pkg/interactive/validation.go
olucasfreitas: It's not, I sincerely do not know how this ended up here, remove it in the last commit

pkg/ocm/helpers.go
olucasfreitas: that is true, cannot think of another case, but it's not that we expect other types, it's because these validators are used with cobra command flags (like --http-proxy and --https-proxy), and cobra's validation functions require the signature func(interface{}) error, from what I could get in the code

pkg/ocm/helpers.go
olucasfreitas: Yes, over-engineered a bit, fixed it in the last commit

pkg/interactive/validation.go
hunterkepley: @olucasfreitas Maybe it would be helpful to respond + resolve to comments one by one 🙂 

But, I feel like something was misunderstood about the function based on the implementation of the nil check here

Here is the function you added this to:

```go
func IsURLHttps(val interface{}) error {
	parsedUri, err := _isUrl(val)
	if err != nil {
		return err
	}
	if parsedUri.Scheme != helper.ProtocolHttps {
		return fmt.Errorf("expect URL '%s' to use an 'https://' scheme", val.(string))
	}
	return nil
}
```

So, the point of this, is to return an error when it's not valid, and return `nil` when it is valid. 

Lets review where this function is called:

```go
issuerUrl, err := interactive.GetString(interactive.Input{
			Question:   "Issuer URL (please include 'https://')",
			Help:       cmd.Flags().Lookup(IssuerUrlFlag).Usage,
			Required:   true,
			Validators: []interactive.Validator{interactive.IsURLHttps},
		})
```

Here it is used as a validator. The value is `Required`, meaning we cannot pass in an empty string.

The second place it's called is when validating another interactive prompt, but in a different way

If we look at the implementation of `_isUrl`, we can see the following:

```go
func _isUrl(val interface{}) (*url.URL, error) {
	if val == nil {
		return nil, nil
	}
	s, ok := val.(string)
	if !ok {
		return nil, fmt.Errorf("can only validate strings, got %v", val)
	}
	if s == "" {
		return nil, nil
	}
	parsedUri, err := url.ParseRequestURI(fmt.Sprintf("%v", val))
	return parsedUri, err
}
```

So, which cases does this return nil?
1. When the URL passed in is nil (impossible for strings! it's possible with pointers to strings, but not strings)
2. When the URL is empty (not possible with the shown implementation)
3. When there is a non-string passed in (this will be caught by the function already, and returned, before it checks if it's nil, due to it returning an actual error)

All implementations have been around since 2023, and are not related to this feature 🙂 so we should leave those functions alone. They only deal with OIDC providers, not `create/cluster` like this ticket is supposed to

hunterkepley: > But, I feel like something was misunderstood about the function based on the implementation of the nil check here

To circle it back to my original statement; If the URL was nil, I don't think we should return `nil` as the error. I think we should inform that the URL passed in was nil, and therefore incorrect. But again, we don't need any of it for this feature 🙂

olucasfreitas: Sorry about the comments, I had understood that the person that commented would resolve the comments if in the next review they saw that the issues we're resolved

also sorry about not getting this at the first place, thanks for the run down, I think my last commit is now correct

hunterkepley: @olucasfreitas 

> Sorry about the comments, I had understood that the person that commented would resolve the comments if in the next review they saw that the issues we're resolved


I was replying to your original reply: `as I got a lot of comments I must have let this one pass, just removed it on the last commit`

it's all good if you leave the comments for the reviewer to resolve, in fact that's a good idea. What I was getting at was that it may be helpful to take them one at a time, and reply to it after fixing it. I see usually that you respond to all of them after fixing everything, then post the commit along with the replies is the reason I suggested it 🙂 might help break up the context of each review comment some

pkg/helper/helpers.go
```diff
@@ -268,3 +268,65 @@ func FilterEmptyStrings(strings []string) []string {
 	}
 	return filteredResult
 }
+
+type URLCredentialValidation struct {
+	Field       string
+	InvalidChar rune
+	HasInvalid  bool
+}
```
hunterkepley: we should define structs at the top of files to be consistent with other files (and Go conventions)

I also suggest maybe putting this + the function below into a more specific file for its purpose (maybe `pkg/helpers/url/validation.go` would suite this nicely, or some variation of this)

pkg/helper/helpers.go
```diff
@@ -268,3 +268,65 @@ func FilterEmptyStrings(strings []string) []string {
...
+
+func checkForInvalidChars(value, field string) *URLCredentialValidation {
+	invalidChars := "/:#?[]@!$&'()*+,;="
+
+	for _, char := range value {
+		if strings.ContainsRune(invalidChars, char) {
+			return &URLCredentialValidation{
+				Field:       field,
+				InvalidChar: char,
+				HasInvalid:  true,
```
hunterkepley: we could drop this field, and instead just check if `InvalidChar` is an empty rune or not

hunterkepley: WDYT?

pkg/helper/helpers.go
```diff
@@ -268,3 +268,65 @@ func FilterEmptyStrings(strings []string) []string {
...
+	username := userinfo[:colonIdx]
+	if result := checkForInvalidChars(username, "username"); result.HasInvalid {
+		return result
+	}
+
+	password := userinfo[colonIdx+1:]
+	return checkForInvalidChars(password, "password")
+}
+
+func checkForInvalidChars(value, field string) *URLCredentialValidation {
```
hunterkepley: I think this should also go in that new file dedicated to validating URLs

pkg/helper/helpers.go
```diff
@@ -268,3 +268,65 @@ func FilterEmptyStrings(strings []string) []string {
...
 	return filteredResult
 }
+
+type URLCredentialValidation struct {
+	Field       string
+	InvalidChar rune
+	HasInvalid  bool
+}
+
+func ValidateURLCredentials(urlString string) *URLCredentialValidation {
```
hunterkepley: same for this

pkg/helper/helpers.go
```diff
@@ -268,3 +268,65 @@ func FilterEmptyStrings(strings []string) []string {
...
+	if !strings.Contains(urlString, "@") || !strings.Contains(urlString, "://") {
+		return &URLCredentialValidation{HasInvalid: false}
+	}
+
+	schemeIdx := strings.Index(urlString, "://")
+	if schemeIdx == -1 {
+		return &URLCredentialValidation{HasInvalid: false}
```
hunterkepley: We should probably tell the user what is wrong here, right? This only says there is an invalid character somewhere

pkg/ocm/helpers.go
```diff
@@ -125,30 +125,88 @@ func ClusterDomainPrefixValidator(domainPrefix interface{}) error {
...
-			return fmt.Errorf("Invalid http-proxy value '%s'", httpProxy)
+// validateProxyURL validates common proxy URL requirements
+func validateProxyURL(proxyURL string, proxyType string, expectedScheme string) error {
+	if strings.Contains(proxyURL, " ") {
+		return fmt.Errorf("invalid %s value: URL cannot contain spaces", proxyType)
+	}
+
+	if credValidation := helper.ValidateURLCredentials(proxyURL); credValidation.HasInvalid {
+		return fmt.Errorf("invalid %s value: %s contains invalid character '%c'",
+			proxyType, credValidation.Field, credValidation.InvalidChar)
```
hunterkepley: There are several situations where the field + proxy type are not filled in if you look at `ValidateURLCredentials`

pkg/ocm/helpers.go
```diff
@@ -125,30 +125,88 @@ func ClusterDomainPrefixValidator(domainPrefix interface{}) error {
...
+	if err != nil {
+		return fmt.Errorf("invalid %s value: %v", proxyType, err)
+	}
+
+	if parsedURL.Scheme == "" {
+		return fmt.Errorf("invalid %s value: URL scheme is missing (e.g., %s://)", proxyType, expectedScheme)
+	}
+	if parsedURL.Scheme != expectedScheme {
+		return errors.Errorf("%s", fmt.Sprintf("expected %s to have an %s:// scheme", proxyType, expectedScheme))
+	}
```
hunterkepley: We should encase the `%s`'s in single quotes, and get rid of this first "`empty check`" validation. They are one and the same if we use single quotes. WDYT? Does that make sense?

pkg/ocm/helpers.go
```diff
@@ -125,30 +125,88 @@ func ClusterDomainPrefixValidator(domainPrefix interface{}) error {
...
+
+	if parsedURL.Host == "" {
+		return fmt.Errorf("invalid %s value: host is missing", proxyType)
+	}
+
+	// Proxy URLs shouldn't have a path
+	if parsedURL.Path != "" && parsedURL.Path != "/" {
+		return fmt.Errorf("invalid %s value: proxy URL should not contain a path", proxyType)
```
hunterkepley: Maybe we should let the user know what the value was in case they are confused?

pkg/ocm/helpers.go
```diff
@@ -125,30 +125,88 @@ func ClusterDomainPrefixValidator(domainPrefix interface{}) error {
...
+	}
+
+	// Validate hostname/IP and port
+	host, port, err := net.SplitHostPort(parsedURL.Host)
+	if err != nil {
+		// No port specified, just validate the host
+		host = parsedURL.Host
```
hunterkepley: Are we sure that the only reason this can return an error is due to not having a port specified? I would suggest looking at the source code for that `net` pkg func

pkg/ocm/helpers.go
```diff
@@ -125,30 +125,88 @@ func ClusterDomainPrefixValidator(domainPrefix interface{}) error {
...
+	} else {
+		// Validate port range (url.Parse already validated it's numeric)
+		portNum, _ := strconv.Atoi(port)
+		if portNum < 1 || portNum > 65535 {
+			return fmt.Errorf("invalid %s value: port must be between 1 and 65535", proxyType)
```
hunterkepley: I don't think we need this validation specifically- the user shouldn't be inputting a giant or negative port number if they have a proxy URL they know they want to use. Having it result in a 404 eventually would be fine UX. WDYT?

pkg/ocm/helpers.go
```diff
@@ -125,30 +125,88 @@ func ClusterDomainPrefixValidator(domainPrefix interface{}) error {
...
+	}
+
+	// Validate host is a valid hostname or IP address
+	if host != "" {
+		// Check if it's an IP address
+		if ip := net.ParseIP(host); ip == nil {
+			// Not an IP, check if it's a valid hostname
+			if strings.Contains(host, "/") {
+				return fmt.Errorf("invalid %s value: invalid host", proxyType)
+			}
```
hunterkepley: is this needed? How do we know it was a slash that caused this? What happens if it wasn't a slash?

pkg/helper/helpers.go
olucasfreitas: fixed on last commit

pkg/helper/helpers.go
olucasfreitas: Agreed, fixed on last commit

pkg/helper/helpers.go
olucasfreitas: fixed on last commit

pkg/helper/helpers.go
olucasfreitas: fixed on last commit

pkg/helper/helpers.go
olucasfreitas: fixed on last commit

pkg/ocm/helpers.go
olucasfreitas: fixed on last commit

pkg/ocm/helpers.go
olucasfreitas: Agreed, fixed on last commit

pkg/ocm/helpers.go
olucasfreitas: fixed on last commit

pkg/ocm/helpers.go
olucasfreitas: fixed on last commit

pkg/ocm/helpers.go
```diff
@@ -125,30 +126,72 @@ func ClusterDomainPrefixValidator(domainPrefix interface{}) error {
...
+		return fmt.Errorf("invalid %s value: host is missing", proxyType)
+	}
+
+	// Proxy URLs shouldn't have a path
+	if parsedURL.Path != "" && parsedURL.Path != "/" {
+		return fmt.Errorf("invalid %s value: proxy URL should not contain a path '%s'", proxyType, parsedURL.Path)
+	}
+
+	// Validate host format
+	_, _, err = net.SplitHostPort(parsedURL.Host)
```
hunterkepley: Could you let me know why this is here?

pkg/ocm/helpers.go
```diff
@@ -125,30 +126,72 @@ func ClusterDomainPrefixValidator(domainPrefix interface{}) error {
...
+	}
+
+	return nil
+}
+
+func ValidateHTTPProxy(val interface{}) error {
+	httpProxy := val.(string)
+	if httpProxy == "" {
 		return nil
 	}
```
hunterkepley: won't this panic if `val` is not a string?

hunterkepley: or a type which can be directly casted into one?

pkg/ocm/helpers.go
```diff
@@ -125,30 +126,72 @@ func ClusterDomainPrefixValidator(domainPrefix interface{}) error {
...
-	return fmt.Errorf("can only validate strings, got '%v'", val)
+	return validateProxyURL(httpProxy, "http-proxy", "http")
+}
+
+func ValidateHTTPSProxy(val interface{}) error {
+	httpsProxy := val.(string)
```
hunterkepley: won't this panic?

pkg/ocm/helpers.go
hunterkepley: I think there was some confusion on what I suggested in my comment here: https://github.com/openshift/rosa/pull/3052#discussion_r2582675098 because the guarding for it panicking was removed after

pkg/helper/url/validation.go
```diff
@@ -0,0 +1,64 @@
...
+// ValidateURLCredentials checks for invalid characters in URL username and password
+func ValidateURLCredentials(urlString string) *URLCredentialValidation {
+	if !strings.Contains(urlString, "@") || !strings.Contains(urlString, "://") {
+		return &URLCredentialValidation{}
+	}
+
+	schemeIdx := strings.Index(urlString, "://")
+	if schemeIdx == -1 {
+		return &URLCredentialValidation{}
+	}
```
hunterkepley: Im not sure I understand, we return an empty validation error when there is no `://`?

https://github.com/openshift/rosa/pull/3052#discussion_r2582660514

It looks like that previously, this returned an error but didn't have enough information in it. Now it has no information, could you let me know what this block of code should be doing? I might be just confused on it

hunterkepley: Also when looking at the block of code above this one, I am even more confused on what this should be accomplishing

pkg/helper/url/validation.go
```diff
@@ -0,0 +1,64 @@
...
+
+	rest := urlString[schemeIdx+3:]
+	atIdx := strings.LastIndex(rest, "@")
+	if atIdx == -1 {
+		return &URLCredentialValidation{}
+	}
```
hunterkepley: When an @ is not found, we return an empty validation error

What does this result in when the final error is printed out?

olucasfreitas: returning an empty URLCredentialValidation{} here means validation passed (no errors), when @ exists in the URL but not after ://, there are no credentials in a valid position to validate

pkg/helper/url/validation.go
olucasfreitas: The empty return when :// was missing was my bad, I've fixed it to return a proper error message

pkg/ocm/helpers.go
olucasfreitas: probably, worked on a fix on the last commit

pkg/ocm/helpers.go
olucasfreitas: fixed this in the last commit

pkg/ocm/helpers.go
olucasfreitas: was trying some things, forgot to remove

pkg/helper/url/validation.go
hunterkepley: > validation passed (no errors)

Its an early return in a validation function, are you sure that it's passing?

pkg/helper/url/validation.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// ValidateURLCredentials checks for invalid characters in URL username and password
+func ValidateURLCredentials(urlString string) *URLCredentialValidation {
+	if !strings.Contains(urlString, "://") {
+		return &URLCredentialValidation{Error: "URL is missing scheme (expected '://')"}
+	}
+	if !strings.Contains(urlString, "@") {
+		return &URLCredentialValidation{}
+	}
```
hunterkepley: Same issue as the last commit; this shouldn't be returning early from the validation with no error

pkg/helper/url/validation.go
```diff
@@ -0,0 +1,64 @@
...
+
+	colonIdx := strings.Index(userinfo, ":")
+	if colonIdx == -1 {
+		return checkForInvalidChars(userinfo, "username")
+	}
+
+	username := userinfo[:colonIdx]
+	if result := checkForInvalidChars(username, "username"); result.InvalidChar != 0 {
+		return result
+	}
```
hunterkepley: I'm a bit confused here

The top block is checking for invalid characters on `userInfo` after getting the index of the first colon in `userInfo`

The bottom block takes everything up to that colon, and once again checks for invalid characters

Both are done differently, one calling the invalid char check in the return, the other using the `if val := func; cond {` format

Anyway, why are we checking the username for invalid characters twice?

pkg/interactive/validation_test.go
```diff
@@ -131,7 +131,7 @@ var _ = Describe("SubnetsValidator", func() {
```
hunterkepley, line 131: Id like to see some tests for the `validation.go` file changes

pkg/ocm/helpers.go
```diff
@@ -125,28 +126,67 @@ func ClusterDomainPrefixValidator(domainPrefix interface{}) error {
...
+	}
+
+	credValidation := urlHelper.ValidateURLCredentials(proxyURL)
+	if credValidation.Error != "" {
+		return fmt.Errorf("invalid %s value: %s", proxyType, credValidation.Error)
+	}
+	if credValidation.InvalidChar != 0 {
+		return fmt.Errorf("invalid %s value: %s contains invalid character '%c'",
+			proxyType, credValidation.Field, credValidation.InvalidChar)
+	}
```
hunterkepley: You only error when ValidateURLCredentials returns a non-empty Error, or, if it returns a non-empty InvalidChar

Honestly this feels a bit messy and like it has a single use case; I think it would make more sense to get rid of `Error` and `InvalidChar` and replace them with an actual `error` typed variable

then, if you get an invalid char, or whatever else, you just construct an error and return it. Then 1) we can get rid of that struct and just return a constructed error from ValidateURLCredentials and 2) we don't have to do a bunch of if statements to parse the struct every single time we call `ValidateURLCredentials`

pkg/interactive/validation_test.go
hunterkepley: also https://github.com/openshift/rosa/pull/3052#discussion_r2713206921

pkg/helper/url/validation.go
olucasfreitas: I did test various scenarios and they all seemed to work correctly

pkg/helper/url/validation.go
olucasfreitas: You're right, the original code was confusing. It had two separate calls, the structure was confusing and used inconsistent styles. I've addressed this by refactoring to have a single validation call

pkg/ocm/helpers.go
olucasfreitas: honestly it was a bit meesy, thanks for the explanation here, this was addressed in my last commit, the `URLCredentialValidation` struct has been removed and `ValidateURLCredentials` now returns error directly, following the same pattern as other validators I saw in the codebase

pkg/interactive/validation_test.go
olucasfreitas: tests where done in last commit and early return was also manually tested, everythin should be correct

pkg/helper/url/validation.go
olucasfreitas: fixed in last commit

pkg/helper/url/validation.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+// ValidateURLCredentials checks for invalid characters in URL username and password
+func ValidateURLCredentials(urlString string) error {
+	if !strings.Contains(urlString, "://") {
+		return fmt.Errorf("URL is missing scheme (expected '://')")
+	}
```
hunterkepley: this validation does not fit the name+description of the function

hunterkepley: we should move it to the top level `ValidateHTTPProxy`/`ValidateHTTPSProxy`

pkg/helper/url/validation.go
```diff
@@ -0,0 +1,58 @@
...
+
+	schemeIdx := strings.Index(urlString, "://")
+	rest := urlString[schemeIdx+3:]
```
hunterkepley: this solution is pretty error prone, if a user includes extra characters in the `://` then the error returned is misleading (ive manually tested it with several combinations, the error changes depending on where the characters are here, please try for yourself)

I think we should instead use a split, that way we retain all the information. Right now, if an extra character is introduced, `Index` will return `-1`, then `+3` makes it `2`, so it just becomes a mess

pkg/helper/url/validation.go
```diff
@@ -0,0 +1,58 @@
...
+	atIdx := strings.LastIndex(rest, "@")
+	if atIdx == -1 {
+		return nil
+	}
```
hunterkepley: why would we want to completely stop validating when  there's no `@`? Doesn't the CIDR come after it? Im not sure I understand why we're doing this 😅

hunterkepley: This would return before the password+username validation below if the user doesn't include an `@`

pkg/helper/url/validation_test.go
```diff
@@ -0,0 +1,93 @@
...
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(Equal("URL is missing scheme (expected '://')"))
+		})
+	})
+
+	Context("when URL has no credentials", func() {
+		It("returns nil for URL without @", func() {
+			err := ValidateURLCredentials("http://example.com")
+			Expect(err).NotTo(HaveOccurred())
+		})
```
hunterkepley, line 48: there's also a test so I assume this was intentional, im curious why we would not want to validate the URL when one isn't found

pkg/helper/url/validation.go
```diff
@@ -0,0 +1,58 @@
...
+
+	userinfo := rest[:atIdx]
+
+	if strings.Count(urlString, "@") > 1 {
+		return fmt.Errorf("password contains invalid character '@'")
+	}
```
hunterkepley: why do we use `LastIndex` above if we only expect 1 in the positive case?

pkg/helper/url/validation.go
```diff
@@ -0,0 +1,58 @@
...
+
+	colonIdx := strings.Index(userinfo, ":")
+
+	var username string
+	if colonIdx == -1 {
+		username = userinfo
+	} else {
+		username = userinfo[:colonIdx]
+	}
```
hunterkepley: this would be better suited with `strings.Split`, this is essentially a manual split

pkg/helper/url/validation.go
```diff
@@ -0,0 +1,58 @@
...
+	if err := checkForInvalidChars(username, "username"); err != nil {
+		return err
+	}
+
+	if colonIdx == -1 {
+		return nil
+	}
+
+	password := userinfo[colonIdx+1:]
+	return checkForInvalidChars(password, "password")
```
hunterkepley: other half of the manual split is here

pkg/helper/url/validation.go
```diff
@@ -0,0 +1,63 @@
...
+
+	if !hasPassword {
+		return nil
+	}
+
+	return checkForInvalidChars(password, "password")
+}
+
+func checkForInvalidChars(value, field string) error {
+	invalidChars := "/:#?[]@!$&'()*+,;="
```
marcolan018, line 54: where is this chars list come from? I have concern that whether it already included all invalid chars

olucasfreitas: this list is from RFC 3986 section 2.2 (the reserved URL chars)

PR #3067, opened 2025-11-03
OCM-18528 | fix: Discontinue UWM for ROSA HCP clusters

olucasfreitas:
# Details

This PR discontinues User Workload Monitoring (UWM) configuration for ROSA HCP clusters as part of the effort to offer self-service CMO on ROSA HCP clusters, while maintaining full functionality for Classic clusters.

---

## Fixed Behavior

1. Run the command with UWM flag on HCP cluster:

    ```bash
    ./rosa create cluster --hosted-cp --disable-workload-monitoring
    ```

2. Now shows **clear error message**:

    ```
    E: User Workload Monitoring configuration is not supported for Hosted Control Plane clusters
    ```

---

## Additional Validations Added

1. **Edit Cluster Blocking** - Prevents UWM configuration on existing HCP clusters:

    ```bash
    ./rosa edit cluster -c my-hcp-cluster --disable-workload-monitoring
    ```

    Output:

    ```
    E: User Workload Monitoring configuration is not supported for Hosted Control Plane clusters
    ```

2. **Interactive Mode** - UWM prompt no longer appears for HCP clusters:

    ```bash
    ./rosa create cluster --hosted-cp --interactive
    ```

    Output:

    ```
    [UWM prompt is skipped entirely for HCP clusters]
    ```

3. **Describe Output** - UWM status hidden for HCP clusters:

    ```bash
    ./rosa describe cluster -c my-hcp-cluster
    ```

    Output:

    ```
    [User Workload Monitoring line is completely absent from output]
    ```

---

## The Classic Cluster Behavior

Using UWM with Classic clusters continues to work normally:

```bash
./rosa create cluster --disable-workload-monitoring
```

Output:

```
[Cluster creation proceeds with UWM disabled as before]
```

---

# Ticket

Closes [OCM-18528](https://issues.redhat.com/browse/OCM-18528)

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -700,10 +700,9 @@ func initFlags(cmd *cobra.Command) {
 		&args.disableWorkloadMonitoring,
 		"disable-workload-monitoring",
 		false,
-		"[DEPRECATED FOR ROSA HCP] Enables you to monitor your own projects in isolation from Red Hat Site "+
-			"Reliability Engineer (SRE) platform metrics. \n\n"+
-			"User workload monitoring is deprecated for Hosted Control Plane clusters, and will be removed in a future "+
-			"version.",
+		"Enables you to monitor your own projects in isolation from Red Hat Site "+
+			"Reliability Engineer (SRE) platform metrics. "+
+			"Not supported for Hosted Control Plane clusters.",
 	)
 
 	flags.BoolVarP(
@@ -981,7 +980,8 @@ func run(cmd *cobra.Command, _ []string) {
 	isHostedCP := args.hostedClusterEnabled
 	if isHostedCP {
 		if cmd.Flags().Changed("disable-workload-monitoring") {
-			r.Reporter.Warnf(arguments.UwmDeprecationMessage)
+			r.Reporter.Errorf(arguments.UwmNotSupportedMessage)
+			os.Exit(1)
 		}
 		validateHcpFlags(cmd, r.Reporter)
 	}
@@ -1111,9 +1111,6 @@ func run(cmd *cobra.Command, _ []string) {
 		if err != nil {
 			r.Reporter.Errorf("Expected a valid --hosted-cp value: %s", err)
 			os.Exit(1)
-		}
-		if isHostedCP {
-			r.Reporter.Warnf(arguments.UwmDeprecationMessage)
 		}
 	}
 
@@ -2998,7 +2995,7 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	disableWorkloadMonitoring := args.disableWorkloadMonitoring
-	if interactive.Enabled() {
+	if interactive.Enabled() && !isHostedCP {
 		disableWorkloadMonitoring, err = interactive.GetBool(interactive.Input{
 			Question: "Disable Workload monitoring",
 			Help:     cmd.Flags().Lookup("disable-workload-monitoring").Usage,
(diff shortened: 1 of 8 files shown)


cmd/create/cluster/cmd.go
```diff
@@ -2998,7 +2995,14 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 
 	disableWorkloadMonitoring := args.disableWorkloadMonitoring
-	if interactive.Enabled() {
+
+	// Warn if the flag is explicitly set for HCP clusters
+	if cmd.Flags().Changed("disable-workload-monitoring") && isHostedCP {
+		r.Reporter.Warnf(arguments.UwmDeprecationMessage)
+	}
```
hunterkepley: No need to move down here, exact same code+outcome as the  L984-L986 removal

cmd/edit/cluster/cmd.go
```diff
@@ -388,33 +381,41 @@ func run(cmd *cobra.Command, _ []string) {
...
 			os.Exit(1)
 		}
-		if aws.IsHostedCP(cluster) {
-			r.Reporter.Warnf(arguments.UwmDeprecationMessage)
-		}
 		disableWorkloadMonitoring = &disableWorkloadMonitoringValue
-	} else if disableWorkloadMonitoringValue {
-		if !confirm.Confirm("disable workload monitoring for your cluster %s", clusterKey) {
-			os.Exit(0)
-		}
```
hunterkepley, line 417: probably don't need to move this; can add `if !hostedCP` to achieve the desired behavior here

cmd/edit/cluster/cmd.go
```diff
@@ -272,10 +269,6 @@ func run(cmd *cobra.Command, _ []string) {
 
 	cluster := r.FetchCluster()
 
-	if aws.IsHostedCP(cluster) && cmd.Flags().Changed("disable-workload-monitoring") {
-		r.Reporter.Warnf(arguments.UwmDeprecationMessage)
-	}
-
```
hunterkepley, line 278: No need to move this down, the earlier we tell a user, the better

hunterkepley (commented):
Overall, we should be removing the deprecation warnings, and purely removing UWM's existence from HCP specifically.

We are going to be deprecating on the API level only, and breaking older CLIs for thing change, so we should be making it so UWM is never sent on create, edit, nor displayed via describe for HCP clusters

tests/utils/handler/cluster_handler.go
```diff
@@ -781,7 +781,9 @@ func (ch *clusterHandler) GenerateClusterCreateFlags() ([]string, error) {
 		flags = append(flags, "--disable-scp-checks")
 		ch.clusterConfig.DisableScpChecks = true
 	}
-	if ch.profile.ClusterConfig.DisableUserWorKloadMonitoring {
+	// Only add disable-workload-monitoring flag for non-HCP clusters
+	// UWM is deprecated for HCP clusters
+	if ch.profile.ClusterConfig.DisableUserWorKloadMonitoring && !ch.profile.ClusterConfig.HCP {
```
hunterkepley, line 788: Is there a reason to do this along with the test skip above?

hunterkepley (commented):
LGTM besides one question about the e2e tests

tests/utils/handler/cluster_handler.go
olucasfreitas: So basically, even though we skip the UWM test for HCP clusters, I thought it would be safer to also prevent the test handler from adding the flag in the first place. But going through the code again, I understood now that it completely skips this, so you're right that it is redundant since we're already skipping the test.

Fixed that in the last commit

hunterkepley: Looks like I got my answer, the tests failed so it's required 🙂 

```
      exit status 1: ERR: User Workload Monitoring configuration is not supported for Hosted Control Plane clusters
```

PR #3083, opened 2025-11-07
OCM-20468 | test: automate ids: 44776,84246

yuwang-RH:
Automated OCP-44776 and OCP-84246.

84246 is a test which testing creating nodepool with capability reservation id, using mock API to test it as the capability reservation resource is expensive. The e2e testing have been done when accept the feature before.

Logs on local is https://privatebin.corp.redhat.com/?3cae6b07795a33a2#4ufENqjetDffaqu7FV57UjSUPMWQ5385F1edJJi5AHWu

### DIFF
--- a/pkg/machinepool/machinepool_test.go
+++ b/pkg/machinepool/machinepool_test.go
@@ -1509,6 +1509,78 @@ var _ = Describe("NodePools", func() {
 			Expect(err).To(HaveOccurred())
 			Expect(err.Error()).To(ContainSubstring("Failed to add machine pool to hosted cluster"))
 		})
+		It("Successfully creates a node pool with capacity-reservation-id", func() {
+			machinePool := &machinePool{}
+			version := "4.19.0"
+			az := "a1"
+
+			cluster = returnMockCluster(version)
+			privateSubnets := []ec2types.Subnet{
+				{AvailabilityZone: &az, SubnetId: &subnet},
+			}
+
+			cmd.Flags().StringVar(&args.Name, "name", "", "Name of the machine pool")
+			cmd.Flags().Set("name", "test")
+
+			cmd.Flags().StringVar(&args.Version, "version", "", "Version of the machine pool")
+			cmd.Flags().Set("version", version)
+			isVersionSet := cmd.Flags().Changed("version")
+			Expect(isVersionSet).To(BeTrue())
+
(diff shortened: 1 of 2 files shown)


tests/e2e/test_rosacli_machine_pool.go
```diff
@@ -1490,6 +1496,57 @@ var _ = Describe("Edit machinepool",
...
 					ContainSubstring("At least one machine pool able to run OCP workload is required"))
 			})
+		It("will validate replicas on multi-az classic cluster - [id:44776]",
+			labels.Runtime.Day2, labels.Medium, labels.FedRAMP,
+			func() {
+				if !clusterConfig.MultiAZ {
+					Skip("It is only for multi-az clusters")
+				}
+
+				By("Create a machinepool without availibility zone set will work")
```
jerichokeyne: I don't see this step in the polarion test case. It kinda looks like you copied the setup from `id:52352` and forgot to delete this step.

If you meant to leave this in, can you update the polarion test case and also change the machine pool name to match this test case instead (and preferably also make it a random name with a set prefix/suffix instead)

pkg/machinepool/machinepool_test.go
```diff
@@ -1509,6 +1509,78 @@ var _ = Describe("NodePools", func() {
 			Expect(err).To(HaveOccurred())
 			Expect(err.Error()).To(ContainSubstring("Failed to add machine pool to hosted cluster"))
 		})
+		It("Successfully creates a node pool with capacity-reservation-id", func() {
```
jerichokeyne, line 1512: I haven't updated any of these tests, so I'm not sure what the standard is for these tests, but would it make sense to add the polarion test case id here? Or do we only want to do that for the e2e tests?

tests/e2e/test_rosacli_machine_pool.go
yuwang-RH: @jerichokeyne This step is preparing an additional machinepool for the following steps of edit. I updated the step description.

pkg/machinepool/machinepool_test.go
yuwang-RH: As this kind of case is located at pkg/machinepool and not an e2e testing, till now, I will add some comment in the related Polarion test case to clarify the test of creating machiepool with capacity-reservation-id here.

PR #3091, opened 2025-11-13
OCM-20194 | feat: Win-Li support (machinepool type flag)

hunterkepley:
Adds the `type` flag to the `create/machinepool` command for HCP (so, only nodepools)

This allows the user to make a selection of which AMI they want to use (`Default` or `Windows` for now), enabling Win-LI (Windows VM support)

```bash
❯ ./rosa create machinepool -c <cluster-name>
I: Enabling interactive mode
? Machine pool name: hk
? Image Type (optional, choose 'Skip' to skip selection; ):  [Use arrows to move, type to filter]
> Skip
  Default
  Windows
```

Simplest usage is above ^

```bash
❯ ./rosa create machinepool -c <cluster-name> --type Windows --name <machinepool-name>
? Replicas: 2
I: Checking available instance types for machine pool '<machinepool-name>'
```

Usage with flag provided ^

```bash
❯ ./rosa create machinepool -c <cluster-name> --type 123123
E: invalid image type: '123123' - please use one of: 'Default', 'Windows'
```

Failure cause 1 ^

```bash
./rosa create machinepool -c <classic-cluster-name> --type Windows
E: the '--type' flag can only be used with Hosted Control Plane clusters
```

Failure case 2 ^

This MR also contains a few small other changes:
* Unit tests
* Validation on CLI-side
* Removal of unused function in the create/machinepool options file

### DIFF
--- a/pkg/machinepool/machinepool.go
+++ b/pkg/machinepool/machinepool.go
@@ -540,6 +540,28 @@ func (m *machinePool) CreateNodePools(r *rosa.Runtime, cmd *cobra.Command, clust
 		return fmt.Errorf("expected a valid name for the machine pool")
 	}
 
+	imageType := ""
+	if args.Type != "" && cluster.Hypershift().Enabled() {
+		imageType = args.Type
+	}
+
+	// Image type (supports things such as WinLI // Windows VMs)
+	if interactive.Enabled() && cluster.Hypershift().Enabled() {
+		imageType, err = interactive.GetOption(interactive.Input{
+			Question: "Image Type",
+			Default:  cmv1.ImageTypeDefault,
+			Required: false,
+			Options:  mpHelpers.ImageTypes,
+		})
+		if err != nil {
+			return fmt.Errorf("expected a valid image type: '%s'", err)
+		}
+	}
+
+	if imageType != "" && !mpHelpers.IsValidImageType(imageType) {
+		return fmt.Errorf("expected a valid image type for the machine pool: '%s'", imageType)
+	}
+
 	// OpenShift version:
 	isVersionSet := cmd.Flags().Changed("version")
 	version := args.Version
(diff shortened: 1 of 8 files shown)


pkg/machinepool/machinepool.go
```diff
@@ -540,6 +540,25 @@ func (m *machinePool) CreateNodePools(r *rosa.Runtime, cmd *cobra.Command, clust
 		return fmt.Errorf("expected a valid name for the machine pool")
 	}
 
+	imageType := ""
+	if args.Type != "" && cluster.Hypershift().Enabled() {
+		imageType = args.Type
+	}
+
+	// Image type (supports things such as WinLI // Windows VMs)
+	if interactive.Enabled() && cluster.Hypershift().Enabled() {
```
hunterkepley, line 549: Technically this check isn't needed, might remove once I finish fixing the linter errors. Though it hurts nothing to keep it

pkg/options/machinepool/create.go
```diff
@@ -283,6 +277,13 @@ func BuildMachinePoolCreateCommandWithOptions() (*cobra.Command, *CreateMachinep
 		"",
 		"The ID of an AWS On-Demand Capacity Reservation. The 'capacity-reservation-id' must be pre-created "+
 			"in advance, before creating a NodePool.")
+
+	flags.StringVar(&options.Type,
+		"type",
+		"",
+		"Specifies the type of AMI this machinepool uses. You may supply '--type win-li' for example if you want "+
```
jerichokeyne: I might be misunderstanding something, but I don't think `--type win-li` will be a valid example. I think it should be `--type Windows` (could be wrong as I only took a quick look through here to answer someone's question)

hunterkepley: Whoops you're right, the original design had `win-li` but we ended up going with `Windows` in the end

thanks for spotting that

pkg/machinepool/machinepool.go
davidleerh: I think this would be good to keep for now as an added measure

davidleerh (approved):

PR #3093, opened 2025-11-14
OCM-18910 | fix: Validate capacity reservation for edit machinepool

olucasfreitas:
Depends on #3094

### DIFF
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -119,6 +119,7 @@ type Client interface {
 	GetAvailabilityZoneType(availabilityZoneName string) (string, error)
 	GetVPCSubnets(subnetID string) ([]ec2types.Subnet, error)
 	GetVPCPrivateSubnets(subnetID string) ([]ec2types.Subnet, error)
+	GetCapacityReservationDetails(capacityReservationID string) (int32, error)
 	FilterVPCsPrivateSubnets(subnets []ec2types.Subnet) ([]ec2types.Subnet, error)
 	ValidateQuota() (bool, error)
 	TagUserRegion(username string, region string) error
@@ -1141,6 +1142,25 @@ func (c *awsClient) GetAvailabilityZoneType(availabilityZoneName string) (string
 		return "", fmt.Errorf("failed to find availability zone '%s'", availabilityZoneName)
 	}
 	return aws.ToString(availabilityZones.AvailabilityZones[0].ZoneType), nil
+}
+
+func (c *awsClient) GetCapacityReservationDetails(capacityReservationID string) (int32, error) {
+	resp, err := c.ec2Client.DescribeCapacityReservations(context.Background(),
+		&ec2.DescribeCapacityReservationsInput{
+			CapacityReservationIds: []string{capacityReservationID},
+		})
+	if err != nil {
+		return 0, fmt.Errorf("failed to describe capacity reservation '%s': %v", capacityReservationID, err)
+	}
+
+	if len(resp.CapacityReservations) == 0 {
+		return 0, fmt.Errorf("capacity reservation '%s' not found", capacityReservationID)
+	}
+
+	cr := resp.CapacityReservations[0]
+	availableInstanceCount := aws.ToInt32(cr.AvailableInstanceCount)
+
+	return availableInstanceCount, nil
 }
 
 func (c *awsClient) DetachRolePolicies(roleName string) error {
--- a/pkg/machinepool/machinepool.go
+++ b/pkg/machinepool/machinepool.go
@@ -1060,6 +1060,12 @@ func (m *machinePool) CreateNodePools(r *rosa.Runtime, cmd *cobra.Command, clust
 		return fmt.Errorf("failed to create machine pool for hosted cluster '%s': %v", clusterKey, err)
 	}
 
+	if capacityReservationId != "" && !autoscaling {
+		if err = validateCapacityReservationReplicas(capacityReservationId, replicas, r.AWSClient); err != nil {
+			return err
+		}
+	}
+
 	createdNodePool, err := r.OCMClient.CreateNodePool(cluster.ID(), nodePool)
 	if err != nil {
 		return fmt.Errorf("failed to add machine pool to hosted cluster '%s': %v", clusterKey, err)
@@ -1738,15 +1744,43 @@ func editNodePool(cmd *cobra.Command, nodePoolID string,
 		return fmt.Errorf("machine pool '%s' does not exist for hosted cluster '%s'", nodePoolID, clusterKey)
 	}
 
+	hasCapacityReservation := nodePool != nil &&
+		nodePool.AWSNodePool() != nil &&
+		nodePool.AWSNodePool().CapacityReservation() != nil
+
 	autoscaling, replicas, minReplicas, maxReplicas, err := getNodePoolReplicas(cmd, nodePoolID,
-		nodePool.Replicas(), nodePool.Autoscaling(), isAnyAdditionalParameterSet, cluster.OpenshiftVersion())
+		nodePool.Replicas(), nodePool.Autoscaling(), isAnyAdditionalParameterSet, cluster.OpenshiftVersion(), r)
 	if err != nil {
 		return fmt.Errorf("failed to get autoscaling or replicas: '%s'", err)
+	}
+
+	// Display capacity reservation info in interactive mode
+	if interactive.Enabled() && hasCapacityReservation {
+		capacityReservationId, ok := nodePool.AWSNodePool().CapacityReservation().GetId()
+		if ok {
+			available, queryErr := r.AWSClient.GetCapacityReservationDetails(capacityReservationId)
+			if queryErr != nil {
+				r.Reporter.Warnf("Failed to retrieve capacity reservation '%s' details: %v",
+					capacityReservationId, queryErr)
+			} else {
+				r.Reporter.Infof("Capacity reservation '%s' has %d available instance(s)",
+					capacityReservationId, available)
+			}
+		}
 	}
 
 	err = validateNodePoolEdit(cmd, autoscaling, replicas, minReplicas, maxReplicas)
 	if err != nil {
 		return err
+	}
+
+	if hasCapacityReservation && !autoscaling {
+		capacityReservationId, ok := nodePool.AWSNodePool().CapacityReservation().GetId()
+		if ok {
+			if err = validateCapacityReservationReplicas(capacityReservationId, replicas, r.AWSClient); err != nil {
+				return err
+			}
+		}
 	}
 
 	labels := cmd.Flags().Lookup("labels").Value.String()
@@ -2076,7 +2110,8 @@ func getNodePoolReplicas(cmd *cobra.Command,
 	existingReplicas int,
 	existingAutoscaling *cmv1.NodePoolAutoscaling,
 	isAnyAdditionalParameterSet bool,
-	clusterVersion string) (autoscaling bool,
+	clusterVersion string,
+	r *rosa.Runtime) (autoscaling bool,
 	replicas, minReplicas, maxReplicas int, err error) {
 
 	isMinReplicasSet := cmd.Flags().Changed("min-replicas")
--- a/pkg/machinepool/machinepool_test.go
+++ b/pkg/machinepool/machinepool_test.go
@@ -1546,6 +1546,7 @@ var _ = Describe("NodePools", func() {
 
 			mockClient.EXPECT().GetVPCPrivateSubnets(gomock.Any()).Return(privateSubnets, nil)
 			mockClient.EXPECT().GetSubnetAvailabilityZone(subnet).Return(az, nil)
+			mockClient.EXPECT().GetCapacityReservationDetails("fake-capacity-reservation-id").Return(int32(3), nil)
 
 			mtBuilder := cmv1.NewMachineType().ID("t3.small").Name("t3.small")
 			machineType, err := mtBuilder.Build()
--- a/pkg/machinepool/validation.go
+++ b/pkg/machinepool/validation.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 
 	"github.com/AlecAivazis/survey/v2/core"
+
+	"github.com/openshift/rosa/pkg/aws"
 )
 
 func ValidateKubeletConfig(input interface{}) error {
@@ -58,5 +60,22 @@ func validateCapacityReservationId(proposedId, nodepoolId, existingId string) er
 		return fmt.Errorf("unable to change 'capacity-reservation-id' to '%s'. AWS NodePool '%s' already has a "+
 			"Capacity Reservation ID: '%s'", proposedId, nodepoolId, existingId)
 	}
+	return nil
+}
+
+// validateCapacityReservationReplicas validates that the requested replicas don't exceed available capacity
+func validateCapacityReservationReplicas(capacityReservationId string, requestedReplicas int,
+	awsClient aws.Client) error {
+
+	availableInstances, err := awsClient.GetCapacityReservationDetails(capacityReservationId)
+	if err != nil {
+		return fmt.Errorf("unable to validate capacity reservation '%s': %v", capacityReservationId, err)
+	}
+
+	if requestedReplicas > int(availableInstances) {
+		return fmt.Errorf("cannot set replicas to %d: capacity reservation '%s' only has %d available instance(s)",
+			requestedReplicas, capacityReservationId, availableInstances)
+	}
+
 	return nil
 }
--- a/pkg/machinepool/validation_test.go
+++ b/pkg/machinepool/validation_test.go
@@ -1,9 +1,15 @@
 package machinepool
 
 import (
+	"fmt"
+
+	"go.uber.org/mock/gomock"
+
 	"github.com/AlecAivazis/survey/v2/core"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+
+	"github.com/openshift/rosa/pkg/aws"
 )
 
 var _ = Describe("MachinePool validation", func() {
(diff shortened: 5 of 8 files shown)


pkg/aws/client.go
```diff
@@ -1142,6 +1144,28 @@ func (c *awsClient) GetAvailabilityZoneType(availabilityZoneName string) (string
 	return aws.ToString(availabilityZones.AvailabilityZones[0].ZoneType), nil
 }
 
+// GetCapacityReservationDetails fetches details about a capacity reservation
```
hunterkepley: NP: This comment tells me nothing the code doesn't

pkg/machinepool/machinepool.go
```diff
@@ -1034,6 +1034,15 @@ func (m *machinePool) CreateNodePools(r *rosa.Runtime, cmd *cobra.Command, clust
 		return fmt.Errorf("failed to create machine pool for hosted cluster '%s': %v", clusterKey, err)
 	}
 
+	// Validate capacity reservation limits before creating the node pool
+	if capacityReservationId != "" && !autoscaling {
+		err = validateCapacityReservationReplicas(capacityReservationId, replicas,
```
hunterkepley: can be simplified:

```go
if validateCapacityReservationReplicas... != nil
```

pkg/machinepool/machinepool.go
```diff
@@ -1034,6 +1034,15 @@ func (m *machinePool) CreateNodePools(r *rosa.Runtime, cmd *cobra.Command, clust
 		return fmt.Errorf("failed to create machine pool for hosted cluster '%s': %v", clusterKey, err)
 	}
 
+	// Validate capacity reservation limits before creating the node pool
```
hunterkepley: NP: This comment tells me nothing the code doesn't

pkg/machinepool/machinepool.go
```diff
@@ -2118,6 +2142,23 @@ func getNodePoolReplicas(cmd *cobra.Command,
 		IsHostedCp:     true,
 	}
 
+	// Get capacity reservation limits for interactive mode
+	var capacityReservationId string
+	var availableInstances int32
+	if interactive.Enabled() && nodePool != nil && nodePool.AWSNodePool() != nil &&
+		nodePool.AWSNodePool().CapacityReservation() != nil {
```
hunterkepley: Maybe this can be simplified? We already check half of the expressions here above on L1737/1738 - maybe it would be best to save those into a var to improve readability on this block of code

pkg/machinepool/machinepool.go
```diff
@@ -2118,6 +2142,23 @@ func getNodePoolReplicas(cmd *cobra.Command,
...
+		capacityReservationId, _ = nodePool.AWSNodePool().CapacityReservation().GetId()
+		if capacityReservationId != "" && r != nil && r.AWSClient != nil {
```
hunterkepley: In what situation would `r` be nil here?

hunterkepley: Same for the AWS Client

pkg/machinepool/machinepool.go
```diff
@@ -2118,6 +2142,23 @@ func getNodePoolReplicas(cmd *cobra.Command,
...
+			_, available, queryErr := r.AWSClient.GetCapacityReservationDetails(capacityReservationId)
+			if queryErr == nil {
+				availableInstances = available
+				// Show capacity reservation info to user
+				r.Reporter.Infof("Capacity reservation '%s' has %d available instance(s)",
+					capacityReservationId, availableInstances)
+			}
```
hunterkepley: Why are we only handling the positive flow? I would think users should know what the error is from AWS

pkg/machinepool/validation_test.go
```diff
@@ -120,4 +126,103 @@ var _ = Describe("MachinePool validation", func() {
...
+			Expect(err).To(HaveOccurred())
+			// The error is wrapped with additional context
+			Expect(err.Error()).To(ContainSubstring("unable to validate capacity reservation"))
+			Expect(err.Error()).To(ContainSubstring(capacityReservationID))
+			Expect(err.Error()).To(ContainSubstring("AWS API error"))
+		})
+
+		It("Passes when capacity reservation has zero available instances and replicas is zero", func() {
+			capacityReservationID := "cr-12345678"
+			// Mock returns 5 total, 0 available
```
hunterkepley: NP: These repeated comments stating what the hardcoded values are don't tell me anything. I would suggest saving the hardcoded values (such as `5` and `0`) as variables. Then passing them into the mock expect

this way devs can infer what the tests are doing by just reading the code, instead of having many repeated comments

pkg/machinepool/validation_test.go
```diff
@@ -120,4 +126,103 @@ var _ = Describe("MachinePool validation", func() {
...
+		It("Handles AWS API errors gracefully", func() {
+			capacityReservationID := "cr-12345678"
+			awsError := fmt.Errorf("AWS API error: capacity reservation not found")
+
+			mockClient.EXPECT().GetCapacityReservationDetails(capacityReservationID).
+				Return(int32(0), int32(0), awsError)
+
+			err := validateCapacityReservationReplicas(capacityReservationID, 2, mockClient, false, 0, 0)
+			Expect(err).To(HaveOccurred())
+			// The error is wrapped with additional context
```
hunterkepley: I don't think we need to specify this, WDYT?

pkg/machinepool/machinepool.go
```diff
@@ -1723,6 +1733,18 @@ func editNodePool(cmd *cobra.Command, nodePoolID string,
 		return err
 	}
 
+	// Validate capacity reservation limits if applicable
+	if nodePool.AWSNodePool() != nil && nodePool.AWSNodePool().CapacityReservation() != nil {
+		capacityReservationId, _ := nodePool.AWSNodePool().CapacityReservation().GetId()
```
hunterkepley: NP: Why did you use `GetId()` then discard the `ok`?

olucasfreitas: Honestly did not pay close attention to it, looking again it is just a bit redundant, fixed it on the last commit

pkg/machinepool/validation_test.go
olucasfreitas: Fixed on the last commit

pkg/machinepool/validation_test.go
olucasfreitas: Fixed on the last commit

pkg/machinepool/machinepool.go
olucasfreitas: You're right, added an else statement to handle the error in case it happens

pkg/machinepool/machinepool.go
olucasfreitas: Moved it into a check function and added  nil check

pkg/machinepool/machinepool.go
olucasfreitas: Fixed on the last commit

pkg/machinepool/machinepool.go
olucasfreitas: Fixed on the last commit

pkg/aws/client.go
olucasfreitas: Fixed on the last commit

pkg/machinepool/machinepool.go
olucasfreitas: Looking at the code, I'm not entirely sure when `r` would be nil, I went ahead and changed this because function already uses `r` earlier, so if `r` was nil, the code would have crashed before, same goes for the AWS client, my bad on this one.

hunterkepley: @olucasfreitas if you look into the codebase, you will see `r`'s values are already checked to be nil in the `runner` for the CLI. Before a single command is ran, this is already validated

pkg/machinepool/machinepool_test.go
```diff
@@ -1546,6 +1546,8 @@ var _ = Describe("NodePools", func() {
 
 			mockClient.EXPECT().GetVPCPrivateSubnets(gomock.Any()).Return(privateSubnets, nil)
 			mockClient.EXPECT().GetSubnetAvailabilityZone(subnet).Return(az, nil)
+			// Expect validation of capacity reservation
```
hunterkepley: NP: I can tell from the line below that this is an EXPECT of the  GetCapacityReservationDetails func, to contain certain values

I don't think we need this comment, the code tells us more than the comment here- it's also not a validation, this is a mocked EXPECT. Which is different

hunterkepley: Mocking functions allows the function to return a set of values for testing purposes and does not actually run the function itself

pkg/machinepool/machinepool.go
hunterkepley: I think we should just cover the negative case here to simplify things

pkg/machinepool/machinepool.go
```diff
@@ -2076,7 +2104,9 @@ func getNodePoolReplicas(cmd *cobra.Command,
 	existingReplicas int,
 	existingAutoscaling *cmv1.NodePoolAutoscaling,
 	isAnyAdditionalParameterSet bool,
-	clusterVersion string) (autoscaling bool,
+	clusterVersion string,
+	nodePool *cmv1.NodePool,
+	r *rosa.Runtime) (autoscaling bool,
 	replicas, minReplicas, maxReplicas int, err error) {
```
hunterkepley, line 2115: we already have a NodePool ID here- and this function is named `getNodePoolReplicas`

But, in your changes, you have included the NodePool itself, which contains the NodePool ID as well as other values being passed into this function such as `existingReplicas`, `existingAutsocaling`, etc

1) lets follow conventions with what is already in this function, and instead pass in a value if needed, rather than the entire NodePool object. It looks like the choice to leave the NodePool object out of this function definition was intentional to me

2) this function is named `getNodePoolReplicas` - I don't think this relates to `validating capacity reservation when editing a cluster`

pkg/machinepool/machinepool.go
```diff
@@ -2144,6 +2174,28 @@ func getNodePoolReplicas(cmd *cobra.Command,
...
+
+	// Get capacity reservation limits for interactive mode
+	var availableInstances int32
+	if interactive.Enabled() && hasCapacityReservation {
+		capacityReservationId, ok := nodePool.AWSNodePool().CapacityReservation().GetId()
+		if ok {
+			_, available, queryErr := r.AWSClient.GetCapacityReservationDetails(capacityReservationId)
+			if queryErr == nil {
+				availableInstances = available
+				// Show capacity reservation info to user
```
hunterkepley: NP: Unneeded comment

pkg/machinepool/machinepool.go
```diff
@@ -2144,6 +2174,28 @@ func getNodePoolReplicas(cmd *cobra.Command,
 		IsHostedCp:     true,
 	}
 
+	// Check if the node pool has a capacity reservation
+	hasCapacityReservation := nodePoolHasCapacityReservation(nodePool)
+
+	// Get capacity reservation limits for interactive mode
+	var availableInstances int32
```
hunterkepley: is there a specific reason you are using an int32 here?

hunterkepley: also this var is defined outside of the scope of your nested if statements below. But, you do not use it outside of this nested if statement. I think it would be better for namespace clarity to move it to only the scope it affects

pkg/machinepool/machinepool.go
hunterkepley: Also what is the goal of this? We warn the user that there was an error from AWS then continue? I'm not sure that is desired for this 🤔

pkg/machinepool/validation.go
```diff
@@ -60,3 +62,37 @@ func validateCapacityReservationId(proposedId, nodepoolId, existingId string) er
...
 	return nil
 }
+
+// validateCapacityReservationReplicas validates that the requested replicas don't exceed available capacity
+func validateCapacityReservationReplicas(capacityReservationId string, requestedReplicas int,
+	awsClient aws.Client, isAutoscaling bool, minReplicas int, maxReplicas int) error {
+
+	if capacityReservationId == "" {
+		return nil // No capacity reservation, no validation needed
+	}
```
hunterkepley: I think this should be moved to the function call, no point adding another function to the stack when it won't run, right?

pkg/machinepool/validation.go
```diff
@@ -60,3 +62,37 @@ func validateCapacityReservationId(proposedId, nodepoolId, existingId string) er
...
+
+	if isAutoscaling {
+		// For autoscaling, validate both min and max replicas
+		if minReplicas > int(availableInstances) {
+			return fmt.Errorf("cannot set min replicas to %d: capacity reservation '%s' only has %d available instance(s)",
+				minReplicas, capacityReservationId, availableInstances)
+		}
+		if maxReplicas > int(availableInstances) {
+			return fmt.Errorf("cannot set max replicas to %d: capacity reservation '%s' only has %d available instance(s)",
+				maxReplicas, capacityReservationId, availableInstances)
```
hunterkepley: why are we validating autoscaling??

hunterkepley: this feature isn't supported with autoscaling enabled

olucasfreitas: Fixed on last commit

pkg/machinepool/validation.go
olucasfreitas: Fixed on last commit

pkg/machinepool/machinepool.go
olucasfreitas: removed this in the last commit

pkg/machinepool/machinepool.go
olucasfreitas: Fixed on last commit

pkg/machinepool/machinepool.go
olucasfreitas: Fixed on last commit

pkg/machinepool/machinepool_test.go
olucasfreitas: Fixed on last commit

pkg/machinepool/machinepool.go
hunterkepley: removed what?

pkg/machinepool/machinepool.go
hunterkepley: @olucasfreitas ^

pkg/machinepool/machinepool.go
hunterkepley: I was curious as to why an int32 was used, you said it was removed, but I still see it, so I am a bit confused 🙁 I just wanted to know why it was used

pkg/aws/client.go
```diff
@@ -1143,6 +1145,27 @@ func (c *awsClient) GetAvailabilityZoneType(availabilityZoneName string) (string
...
+		return 0, 0, fmt.Errorf("failed to describe capacity reservation '%s': %v", capacityReservationID, err)
+	}
+
+	if len(resp.CapacityReservations) == 0 {
+		return 0, 0, fmt.Errorf("capacity reservation '%s' not found", capacityReservationID)
+	}
+
+	cr := resp.CapacityReservations[0]
+	totalInstanceCount = aws.ToInt32(cr.TotalInstanceCount)
+	availableInstanceCount = aws.ToInt32(cr.AvailableInstanceCount)
```
hunterkepley: I asked last time but got no response about it 😅 but why are we using int32's?

pkg/aws/client.go
```diff
@@ -1143,6 +1145,27 @@ func (c *awsClient) GetAvailabilityZoneType(availabilityZoneName string) (string
 	return aws.ToString(availabilityZones.AvailabilityZones[0].ZoneType), nil
 }
 
+func (c *awsClient) GetCapacityReservationDetails(capacityReservationID string) (
+	totalInstanceCount int32, availableInstanceCount int32, err error) {
```
hunterkepley: why do we return `totalInstanceCount` if it's never used?

pkg/aws/client.go
```diff
@@ -1143,6 +1145,27 @@ func (c *awsClient) GetAvailabilityZoneType(availabilityZoneName string) (string
...
+
+	if len(resp.CapacityReservations) == 0 {
+		return 0, 0, fmt.Errorf("capacity reservation '%s' not found", capacityReservationID)
+	}
+
+	cr := resp.CapacityReservations[0]
+	totalInstanceCount = aws.ToInt32(cr.TotalInstanceCount)
+	availableInstanceCount = aws.ToInt32(cr.AvailableInstanceCount)
+
+	return totalInstanceCount, availableInstanceCount, nil
```
hunterkepley: why did we name the return types if we aren't using isolated `return`s?

[I might would suggest reading this article](https://itnext.io/go-excellence-dodge-mistakes-and-unlock-the-power-of-named-returns-e811f9665d2) 🙂 , I read through it after finding this and it's got some decent explanations on best practices for using named returns

pkg/machinepool/machinepool.go
```diff
@@ -1592,6 +1598,13 @@ func isMultiAZMachinePool(machinePool *cmv1.MachinePool) bool {
 	return len(machinePool.AvailabilityZones()) != 1
 }
 
+// nodePoolHasCapacityReservation checks if a node pool has an AWS capacity reservation configured
+func nodePoolHasCapacityReservation(nodePool *cmv1.NodePool) bool {
+	return nodePool != nil &&
+		nodePool.AWSNodePool() != nil &&
+		nodePool.AWSNodePool().CapacityReservation() != nil
+}
```
hunterkepley: while I would agree with this for a functional programming language, it may be better to just store this as a var? Or directly in-line? This looks like it is only used 1 time

pkg/machinepool/machinepool.go
```diff
@@ -1592,6 +1598,13 @@ func isMultiAZMachinePool(machinePool *cmv1.MachinePool) bool {
 	return len(machinePool.AvailabilityZones()) != 1
 }
 
+// nodePoolHasCapacityReservation checks if a node pool has an AWS capacity reservation configured
```
hunterkepley: this is pretty evident from the function name (and return statement)

pkg/machinepool/machinepool.go
```diff
@@ -1738,17 +1751,44 @@ func editNodePool(cmd *cobra.Command, nodePoolID string,
 		return fmt.Errorf("machine pool '%s' does not exist for hosted cluster '%s'", nodePoolID, clusterKey)
 	}
 
+	// Check if the node pool has a capacity reservation
```
hunterkepley: NP: This is evident from the code

pkg/machinepool/machinepool.go
```diff
@@ -1738,17 +1751,44 @@ func editNodePool(cmd *cobra.Command, nodePoolID string,
...
+			_, available, queryErr := r.AWSClient.GetCapacityReservationDetails(capacityReservationId)
+			if queryErr != nil {
+				r.Reporter.Warnf("Failed to retrieve capacity reservation '%s' details: %v",
+					capacityReservationId, queryErr)
+			} else {
+				r.Reporter.Infof("Capacity reservation '%s' has %d available instance(s)",
+					capacityReservationId, available)
+			}
+		}
+	}
```
hunterkepley, line 1770: > Display capacity reservation info in interactive mode

im not sure I follow the purpose of this block of code

pkg/aws/client.go
olucasfreitas: I tried reading this, but was not able since it requests me to pay the premium plan, do you have another one I can read ?

hunterkepley: @olucasfreitas Here this link should avoid it, it was provided on a reddit thread that the author made

https://do-tech.medium.com/go-excellence-dodge-mistakes-and-unlock-the-power-of-named-returns-e811f9665d2?sk=02f7dfab57b95d5c8397c88b5efbbda1

pkg/machinepool/machinepool.go
olucasfreitas: So basically, whats I tried to do here is when a user is editing a node pool in interactive mode, I wanted to help them out by showing how many instances are available in their capacity reservation, this way, when the prompt asks them "How many replicas do you want?", they already know they can't exceed, I was just trying to give the user helpful context before they make their choice

pkg/aws/client.go
olucasfreitas: I think I got the jist of it, please check the last commit

pkg/aws/client.go
olucasfreitas: you are right, remove it in the last commit

pkg/aws/client.go
olucasfreitas: Sorry, I forgot to complement my comment, but I used the int32's here to match the AWS API's type

robpblake:
@hunterkepley / @olucasfreitas This feels like this should be addressed in the backend and not the CLI. Have you confirmed that the backend correctly enforces this?

PR #3095, opened 2025-11-17
OCM-20730 | feat: Capacity reservation preference for create/nodepool

hunterkepley:
Adds a new flag that is an option set of 3 options; None, Open, and Only. The inputs from the user are validated, and the information is passed into the capacity reservation builder, which only is used when the user meets the criteria for using cap-res + values are all provided.

The prompt does not call if capacit-reservation-id is not filled, and also only if there is no autoscaling enabled

In the end, this allows the following flow:

```bash
rosa create machinepool -c <hcp_cluster_id> --capacity-reservation-id <id> --capacity-reservation-preference <preference_choice>
```

And when using interactive mode, the user is prompted with a failsafe `GetOption` prompt. This only allows the user to select one of 3 options (`none` being the default since the user may not want to specify a preference)

### DIFF
--- a/pkg/helper/machinepools/helpers.go
+++ b/pkg/helper/machinepools/helpers.go
@@ -31,6 +31,10 @@ const (
 		nodeDrainUnitHour + "|" + nodeDrainUnitHours
 	MaxNodeDrainTimeInMinutes = 10080
 	MaxNodeDrainTimeInHours   = 168
+
+	CapacityReservationPreferenceNone = string(cmv1.CapacityReservationPreferenceNone)
+	CapacityReservationPreferenceOnly = string(cmv1.CapacityReservationPreferenceCapacityReservationsOnly)
+	CapacityReservationPreferenceOpen = string(cmv1.CapacityReservationPreferenceOpen)
 )
 
 var (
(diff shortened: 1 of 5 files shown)


pkg/helper/machinepools/helpers.go
```diff
@@ -377,3 +381,16 @@ func IsValidImageType(imageType string) bool {
 	}
 	return false
 }
+
+func ValidateCapacityReservationPreference(preference string) error {
```
BraeTroutman: one thing to note here-- there are some interactions between Capacity Reservation preference and capacity reservation ID, namely that you are only allowed to specify an ID and a preference at the same time if the Preference is CapacityReservationsOnly.

This is enforced at validation time on the CS side, so if the user specified both they will receive a 4xx bad request error-- it that's enough awesome! Otherwise we can update this func to take two params and validate the relationship between preference and ID

hunterkepley: added validations for this

pkg/helper/machinepools/helpers.go
```diff
@@ -377,3 +381,33 @@ func IsValidImageType(imageType string) bool {
...
+	if capResId != "" && preference != CapacityReservationPreferenceOnly {
+		return fmt.Errorf("invalid Capacity Reservation Preference: '%s'. "+
+			"When specifying a capacity reservation id ('%s'), you may only provide the "+
+			"'%s' preference", preference, capResId, CapacityReservationPreferenceOnly)
+	}
+
+	if capResId == "" && preference == CapacityReservationPreferenceOnly {
+		return fmt.Errorf("invalid Capacity Reservation Preference: '%s'. "+
+			"Capacity Reservation ID is required when preference is '%s'", preference,
+			CapacityReservationPreferenceOnly)
```
BraeTroutman: so sorry, one distinction here: ID is not _required_ when the preference is `capacity-reservations-only`, it's that preference and ID can only be specified simultaneously if the preference is `capacity-reservations-only`

pkg/helper/machinepools/helpers_test.go
```diff
@@ -349,3 +352,42 @@ var _ = Describe("Validate MaxSurge and MaxUnavailable", func() {
...
+var _ = Describe("Validate Capacity Reservation Preference", func() {
+	noneValue := string(cmv1.CapacityReservationPreferenceNone)
+	onlyValue := string(cmv1.CapacityReservationPreferenceCapacityReservationsOnly)
+	openValue := string(cmv1.CapacityReservationPreferenceOpen)
+
+	It("OK: Valid values provided", func() {
+		Expect(ValidateCapacityReservationPreference(onlyValue, "123")).ToNot(HaveOccurred())
+		Expect(ValidateCapacityReservationPreference(onlyValue, "")).ToNot(HaveOccurred())
+		Expect(ValidateCapacityReservationPreference(openValue, "")).ToNot(HaveOccurred())
+		Expect(ValidateCapacityReservationPreference(noneValue, "")).ToNot(HaveOccurred())
```
BraeTroutman, line 365: can we add a case to cover when there is no preference specified at all, but an ID is specified?
```suggestion
		Expect(ValidateCapacityReservationPreference(noneValue, "")).ToNot(HaveOccurred())
		Expect(ValidateCapacityReservationPreference("", "123")).ToNot(HaveOccurred())
```

PR #3117, opened 2025-11-25
OCM-21040 | feat: Add logForward to cluster

serngawy:


pkg/ocm/clusters.go
```diff
@@ -281,6 +294,19 @@ func (c *Client) CreateCluster(config Spec) (*cmv1.Cluster, error) {
...
 
+	logFwBuilder := BuildLogForwaders(config)
+
+	if logFwBuilder != nil {
+		logForwarder, err := logFwBuilder.Build()
+		if err != nil {
+			return clusterObject, err
+		}
+
+		if _, err := c.SetLogForwarder(clusterObject.ID(), logForwarder); err != nil {
```
hunterkepley: On day 1 (create) we should not use this endpoint. Instead, we should add the log forwarder to the cluster body being sent to the API

This endpoint is meant to be used only for day 2 operations

hunterkepley: This function is fine to have this in though, just needs to be in the Body for the cluster instead of in it's own API request 🙂

hunterkepley:
Oh, 2 more small things to get the CI to pass:

Title as well as commit mesage should include a `type`: 
I.E `OCM-21040 | feat: Add logForward to cluster` would be perfectly fine 🙂 

Please run `make fmt-imports` after you finish your changes, our CI hates it when imports are not alphabetical 🙂

pkg/ocm/clusters.go
```diff
@@ -1193,6 +1223,13 @@ func (c *Client) createClusterSpec(config Spec) (*cmv1.Cluster, error) {
 		clusterBuilder.Autoscaler(BuildClusterAutoscaler(config.AutoscalerConfig))
 	}
 
+	// LogForwarder
+	logFwBuilder := BuildLogForwader(config.LogForwarder)
+	if logFwBuilder != nil {
+		clusterBuilder.ControlPlane(v1.NewControlPlane().
```
hunterkepley: Looks like this may need to be `cmv1` instead of `v1` - same goes for the line below (after the changes I made in master)

PR #3124, opened 2025-12-04
OCM-21146 | feat: Log Forwarding for HCP day1 support

hunterkepley:
Introduces HCP log forwarding (S3, CloudWatch, or Both) for HCP day 1 (`create/cluster`)

* yaml reader + full e2e yaml parsing test
* unit tests for every helper function
* Rewrote cluster spec to support multiple log fwd configs
* introduces `--log-fwd-config` flag for `create/cluster`
* utilizes new format for ROSA CLI commands, hoping to use this MR as a PoC for refactoring the rest of the CLI for it being so easily testable + the interactive prompts being split off (how they should be 😩 )

Length of MR is mainly due to the new interactive file

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -958,8 +958,8 @@ func initFlags(cmd *cobra.Command) {
 		"",
 		"A path to a log forwarding config file. This should be a YAML file with the following structure:\n\n"+
 			"cloudwatch:\n"+
-			"  cloud_watch_log_role_arn: \"role_arn_here\"\n"+
-			"  cloud_watch_log_group_name: \"group_name_here\"\n"+
+			"  cloudwatch_log_role_arn: \"role_arn_here\"\n"+
+			"  cloudwatch_log_group_name: \"group_name_here\"\n"+
 			"  applications: [\"example_app_1\", \"example_app_2\"]\n"+
 			"  groups: \"group-name\"\n"+
 			"s3:\n"+
@@ -2223,7 +2223,7 @@ func run(cmd *cobra.Command, _ []string) {
 				r.Reporter.Errorf("failed to create log fowarder config: %s", err)
 				os.Exit(1)
 			}
-			if interactiveObject.S3 != nil && interactiveObject.S3.S3ConfigBucketPrefix != "" {
+			if interactiveObject.S3 != nil && interactiveObject.S3.S3ConfigBucketName != "" {
 				logFwdS3ConfigObject = interactiveObject.S3
 			}
 			if interactiveObject.CloudWatch != nil && interactiveObject.CloudWatch.CloudWatchLogRoleArn != "" {
--- a/pkg/logforwarding/helpers.go
+++ b/pkg/logforwarding/helpers.go
@@ -12,21 +12,6 @@ import (
 // FlagName contains the common log forwarding config command flag name
 const FlagName = "log-fwd-config"
 
-const AllowedGroupsPreset = "API (kube-apiserver, openshift-apiserver), Authentication (oauth-server, " +
-	"oauth-apiserver), Controller Manager (kube-controller-manager, openshift-controller-manager, " +
-	"openshift-route-controller-manager), Scheduler (kube-scheduler)"
-
-const AllowedApplicationsPreset = "audit-webhook, aws-ebs-csi-driver-controller, capi-provider-controller-manager, " +
-	"catalog-operator, certified-operators-catalog, cloud-controller-manager, cloud-credential-operator, " +
-	"cloud-network-config-controller, cluster-api, cluster-network-operator, cluster-node-tuning-operator, " +
-	"cluster-policy-controller, cluster-version-operator, community-operators-catalog, control-plane-operator, " +
-	"control-plane-pki-operator, csi-snapshot-controller-operator, csi-snapshot-controller, dns-operator, etcd, " +
-	"hosted-cluster-config-operator, ignition-server, ingress-operator, konnectivity-agent, kube-apiserver, " +
-	"kube-controller-manager, kube-scheduler, machine-approver, multus-admission-controller, network-node-identity, " +
-	"oauth-openshift, olm-operator, openshift-apiserver, openshift-controller-manager, openshift-oauth-apiserver, " +
-	"openshift-route-controller-manager, ovnkube-control-plane, packageserver, private-router, " +
-	"redhat-marketplace-catalog, redhat-operators-catalog, unknown, validation-webhook"
-
 // S3LogForwarderConfig represents the log forward config for S3
 type S3LogForwarderConfig struct {
 	Applications         []string `yaml:"applications,omitempty"`
@@ -39,8 +24,8 @@ type S3LogForwarderConfig struct {
 type CloudWatchLogForwarderConfig struct {
 	Applications           []string `yaml:"applications,omitempty"`
 	GroupsLogVersion       string   `yaml:"groups,omitempty"`
-	CloudWatchLogRoleArn   string   `yaml:"cloud_watch_log_role_arn,omitempty"`
-	CloudWatchLogGroupName string   `yaml:"cloud_watch_log_group_name,omitempty"`
+	CloudWatchLogRoleArn   string   `yaml:"cloudwatch_log_role_arn,omitempty"`
+	CloudWatchLogGroupName string   `yaml:"cloudwatch_log_group_name,omitempty"`
 }
 
 type LogForwarderYaml struct {
--- a/pkg/logforwarding/helpers_test.go
+++ b/pkg/logforwarding/helpers_test.go
@@ -11,8 +11,8 @@ import (
 
 var testConfig = `
 cloudwatch:
-  cloud_watch_log_role_arn: "arn"
-  cloud_watch_log_group_name: "abcd"
+  cloudwatch_log_role_arn: "arn"
+  cloudwatch_log_group_name: "abcd"
   applications: ["test3", "test4"]
   groups: "group-name"
 s3:


cmd/create/cluster/cmd.go
```diff
@@ -2178,6 +2195,24 @@ func run(cmd *cobra.Command, _ []string) {
...
 
+	// Log forwarding for HCP:
+	logFwdConfigObject := make([]*ocm.LogForwarderConfig, 0)
+	if isHostedCP {
+		if args.logFwdConfig != "" {
+			logFwdConfigObject, err = logforwarding.UnmarshalLogForwarderConfigYaml(r.Reporter, args.logFwdConfig)
+			if err != nil {
+				r.Reporter.Errorf("error parsing log forwarder config '%s': %s", args.logFwdConfig, err)
+			}
+		} else {
```
yuwang-RH: When executing rosacli command to create cluster without args.logFwdConfig flag, the current logic will prompt interactive mode to guide users to input. However, as logFwdConfig is not required flag for creation command, so interactive mode doesn't need to be prompted when the flag is not specified in the command.

pkg/logforwarding/interactive.go
```diff
@@ -0,0 +1,156 @@
...
+	config.Applications = strings.Split(",", applications)
+
+	return nil
+}
+
+func promptForApplications(t string) (applications string, err error) {
+	applications, err = interactive.GetString(interactive.Input{
+		Question: fmt.Sprintf("%s Log forwarding applications", t),
+		Help: fmt.Sprintf("Which applications to forward to %s, please use a comma-separated list "+
+			"(example: \"audit-webhook,cluster-api\"", t),
```
yuwang-RH: Missing ')' in the help message of "(example: \"audit-webhook,cluster-api\""

pkg/logforwarding/interactive.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return err
+	}
+	config.GroupsLogVersion = podGroups
+
+	applications, err := promptForApplications("CloudWatch")
+	if err != nil {
+		return err
+	}
+	config.Applications = strings.Split(",", applications)
```
yuwang-RH: It should be config.Applications = strings.Split(applications,",") to split the applications

pkg/logforwarding/interactive.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return err
+	}
+	config.GroupsLogVersion = podGroups
+
+	applications, err := promptForApplications("S3")
+	if err != nil {
+		return err
+	}
+	config.Applications = strings.Split(",", applications)
```
yuwang-RH: It should be config.Applications = strings.Split(applications,",") to split the applications

pkg/logforwarding/interactive.go
hunterkepley: oops thank you, I've been using too much ocaml lately (their split function has the delimiter first 😅 )

cmd/create/cluster/cmd.go
hunterkepley: Discussed on slack thread 🙂

pkg/logforwarding/interactive.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+// Options for initial interactive prompt
+const skip = "Skip"
+const cloudWatch = "CloudWatch"
+const s3 = "S3"
+const both = "Both"
+
+// Options for CW pod groups
+var podGroupsOptions = map[string]string{
```
willkutler: we have an API endpoint to get the allowed groups and applications; it is accessible via "client.ocm.ClustersMgmt().V1().LogForwarding().Groups().List()" and "client.ocm.ClustersMgmt().V1().LogForwarding().Groups().List()" in the SDK

pkg/interactive/logforwarding/logforwarding.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	config.Applications = strings.Split(applications, ",")
+
+	return nil
+}
+
+func promptForApplications(t string) (applications string, err error) {
+	applications, err = interactive.GetString(interactive.Input{
+		Question: fmt.Sprintf("%s Log forwarding applications", t),
+		Help: fmt.Sprintf("Which applications to forward to %s, please use a comma-separated list "+
```
willkutler, line 140: help message should include the allowed list of applications from the API; see other comment

pkg/logforwarding/interactive.go
hunterkepley: this is great 🙂

pkg/ocm/clusters.go
```diff
@@ -1224,8 +1230,9 @@ func (c *Client) createClusterSpec(config Spec) (*cmv1.Cluster, error) {
 	}
 
 	if config.Hypershift.Enabled {
-		logFwBuilder := BuildLogForwader(config.LogForwarder)
-		if logFwBuilder != nil {
+		// LogForwarder
+		for _, logForwarder := range []*LogForwarderConfig{config.S3LogForwarder, config.CloudWatchLogForwarder} {
+			logFwBuilder := BuildLogForwader(logForwarder)
```
serngawy: Let the [BuildLogForwader](https://github.com/openshift/rosa/blob/master/pkg/ocm/log_forwards.go#L23) return nil when LogForwarderConfig is nil and check for it to avoid setting empty LogForwarder

hunterkepley: fixed

pkg/logforwarding/helpers.go
```diff
@@ -0,0 +1,77 @@
...
+			apps += application
+		}
+		s = s + option.ID() + ": " + apps + "\n"
+	}
+	return
+}
+
+func constructPodGroupsInteractiveOptions(options []*cmv1.LogForwarderGroupVersions) (l []string) {
+	for _, option := range options {
+		l = append(l, option.ID())
```
yuwang-RH: From the response of "get /api/clusters_mgmt/v1/log_forwarding/groups", *cmv1.LogForwarderGroupVersions.ID has no value. "Name" looks the ones which can be used here, not SDK have no Name field, not sure if it needs some fix on ocm-sdk?

pkg/logforwarding/interactive.go
```diff
@@ -0,0 +1,151 @@
...
+	return nil
+}
+
+func promptForApplications(t string) (applications string, err error) {
+	applications, err = interactive.GetString(interactive.Input{
+		Question: fmt.Sprintf("%s Log forwarding applications", t),
+		Help: fmt.Sprintf("Which applications to forward to %s, please use a comma-separated list "+
+			"(example: \"audit-webhook,cluster-api\")", t),
+		Default:  "",
+		Required: true,
```
yuwang-RH: is Applications required for logForwarding config? From my understanding, Applications is used for extra-application, it should be optional. The final resolved_applications should be a combination of extra-application and the group mapped applications.

pkg/logforwarding/helpers.go
hunterkepley: This is a fix required in the CS API specifically after talking with the CS feature owner. Will remain the same on the CLI 🙂 thanks for finding this

pkg/logforwarding/interactive.go
hunterkepley: Made `Required: false`

pkg/ocm/clusters.go
```diff
@@ -200,17 +200,18 @@ type Spec struct {
 	InfraMachineType  string
 
 	// LogForward
-	LogForwarder *LogForwarderConfig
+	S3LogForwarder         *LogForwarderConfig
+	CloudWatchLogForwarder *LogForwarderConfig
 }
 
 // LogForwarderConfig represent the log forward config for applications, cloudWatch, S3 and groupVersion.
 type LogForwarderConfig struct {
```
serngawy: better to separate the logforward struct like below 
```
type S3LogForwarderConfig struct {
	Applications           []string `yaml:"applications,omitempty"`
	GroupsLogVersion       []string `yaml:"groups_log_version,omitempty"`
	S3ConfigBucketName     string   `yaml:"s3_config_bucket_name,omitempty"`
	S3ConfigBucketPrefix   string   `yaml:"s3_config_bucket_prefix,omitempty"`
}
type CloudWatchLogForwarderConfig struct {
	Applications           []string `yaml:"applications,omitempty"`
	CloudWatchLogRoleArn   string   `yaml:"cloud_watch_log_role_arn,omitempty"`
	CloudWatchLogGroupName string   `yaml:"cloud_watch_log_group_name,omitempty"`
	GroupsLogVersion       []string `yaml:"groups_log_version,omitempty"`
}
```

pkg/ocm/clusters.go
```diff
@@ -200,17 +200,18 @@ type Spec struct {
 	InfraMachineType  string
 
 	// LogForward
-	LogForwarder *LogForwarderConfig
+	S3LogForwarder         *LogForwarderConfig
```
serngawy: This is not correct you still can set cloudWatch while we suppose to set only S3 config

cmd/create/cluster/cmd.go
```diff
@@ -947,6 +952,23 @@ func initFlags(cmd *cobra.Command) {
...
+	flags.StringVar(
+		&args.logFwdConfig,
+		logforwarding.FlagName,
+		"",
+		"A path to a log forwarding config file. This should be a YAML file with the following structure:\n\n"+
+			"cloudwatch:\n"+
+			"  cloud_watch_log_role_arn: \"role_arn_here\"\n"+
+			"  cloud_watch_log_group_name: \"group_name_here\"\n"+
+			"  applications: [\"example_app_1\", \"example_app_2\"]\n"+
+			"  groups_log_version: \"group-name\"\n"+
```
willkutler: can this just be called "groups"? the `groups_log_version` name might be confusing to users

cmd/create/cluster/cmd.go
```diff
@@ -2178,6 +2203,35 @@ func run(cmd *cobra.Command, _ []string) {
...
+			logFwdS3ConfigObject = yamlObject.S3
+			logFwdCloudWatchConfigObject = yamlObject.CloudWatch
+		} else if args.logFwdConfig == "" && interactive.Enabled() {
+			interactiveObject, err := interactiveLogForwarding.InteractiveLogForwardingConfig(
+				r.OCMClient, cmd.Flags().Lookup(logforwarding.FlagName).Usage)
+			if err != nil {
+				r.Reporter.Errorf("failed to create log fowarder config: %s", err)
+				os.Exit(1)
+			}
+			if interactiveObject.S3 != nil && interactiveObject.S3.S3ConfigBucketPrefix != "" {
```
willkutler: this will ignore all s3 configs that do not have a prefix set; bucket prefix should not be required

pkg/logforwarding/helpers.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+// FlagName contains the common log forwarding config command flag name
+const FlagName = "log-fwd-config"
+
+const AllowedGroupsPreset = "API (kube-apiserver, openshift-apiserver), Authentication (oauth-server, " +
+	"oauth-apiserver), Controller Manager (kube-controller-manager, openshift-controller-manager, " +
+	"openshift-route-controller-manager), Scheduler (kube-scheduler)"
+
+const AllowedApplicationsPreset = "audit-webhook, aws-ebs-csi-driver-controller, capi-provider-controller-manager, " +
```
willkutler: can these consts be removed?

pkg/logforwarding/helpers_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"path/filepath"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+)
+
+var testConfig = `
+cloudwatch:
+  cloud_watch_log_role_arn: "arn"
```
willkutler: can this be cloudwatch_log_role_arn for consistency with the API payload?

PR #3125, opened 2025-12-05
OCM-21187 | fix: only set logForwarders on HCP clusters

willkutler:
the logForwarders object can only be set on HCP clusters but is being provided for all clusters by the CLI, which is breaking all classic cluster creation

https://issues.redhat.com/browse/OCM-21187


willkutler:
verified I am able to create a classic cluster with these changes

PR #3128, opened 2025-12-08
OCM-19908 | fix: added autonode status on the describe command

olucasfreitas:
# Details

This PR fixes a usability issue where `rosa describe cluster` was not displaying AutoNode configuration when it was disabled, unlike other disabled configurations (Etcd Encryption, Audit Log Forwarding, External Authentication, Zero Egress) which always show their status.

---

## Fixed Behavior

1. Run the describe command on a cluster without AutoNode enabled:

    ```bash
    rosa describe cluster -c <cluster-id>
    ```

2. Now shows **AutoNode status** alongside other disabled configurations:

    ```
    Etcd Encryption:            Disabled
    Audit Log Forwarding:       Disabled
    AutoNode:                   Disabled
    External Authentication:    Disabled
    Zero Egress:                Disabled
    ```

---

## Previous Behavior

Before this fix, clusters without AutoNode enabled would show no AutoNode information at all:

```bash
rosa describe cluster -c 2m4m7aohb3v050dg5g2gj3t1jugblf9g
```

Output (truncated):
```
...
Etcd Encryption:            Disabled
Audit Log Forwarding:       Disabled
External Authentication:    Disabled
Zero Egress:                Disabled
```

---

# Ticket

Closes [OCM-19908](https://issues.redhat.com/browse/OCM-19908)

### DIFF
--- a/cmd/describe/cluster/cmd.go
+++ b/cmd/describe/cluster/cmd.go
@@ -566,6 +566,9 @@ func run(cmd *cobra.Command, argv []string) {
 				str = fmt.Sprintf("%s"+
 					"  IAM Role ARN:             %s\n", str, cluster.AWS().AutoNode().RoleArn())
 			}
+		} else {
+			str = fmt.Sprintf("%s"+
+				"AutoNode:                   %s\n", str, DisabledOutput)
 		}
 		str = fmt.Sprintf("%s"+
 			"External Authentication:    %s\n", str, getExternalAuthConfigStatus(cluster))


cmd/describe/cluster/cmd.go
```diff
@@ -566,6 +567,9 @@ func run(cmd *cobra.Command, argv []string) {
 				str = fmt.Sprintf("%s"+
 					"  IAM Role ARN:             %s\n", str, cluster.AWS().AutoNode().RoleArn())
 			}
+		} else {
+			str = fmt.Sprintf("%s"+
+				"AutoNode:                   %s\n", str, autoNodeStatus)
```
hunterkepley: Is there a reason to not print out the IAM role when it is disabled?

cmd/describe/cluster/cmd.go
```diff
@@ -1032,6 +1036,13 @@ func getEtcdStatus(cluster *cmv1.Cluster) string {
 	return etcdStatus
 }
 
+func getAutoNodeStatus(cluster *cmv1.Cluster) string {
+	if cluster.AutoNode() != nil && cluster.AutoNode().Mode() == "enabled" {
+		return EnabledOutput
+	}
+	return DisabledOutput
+}
+
```
hunterkepley: the flow:

1. Check if autonode is disabled AND if the mode is equal to **enabled**
2. Return `"Enabled"` if so
3. Otherwise, return `"Disabled"`

Then, we call this function which does this, in an equality check later in the flow:

`if getAutoNodeStatus(...) == "Enabled"` 

Why can't we just use the if statement from the `getAutoNodeStatus` function in place of the function call? It's a duplicated if statement at this point

cmd/describe/cluster/cmd.go
```diff
@@ -1032,6 +1036,13 @@ func getEtcdStatus(cluster *cmv1.Cluster) string {
 	return etcdStatus
 }
 
+func getAutoNodeStatus(cluster *cmv1.Cluster) string {
+	if cluster.AutoNode() != nil && cluster.AutoNode().Mode() == "enabled" {
```
hunterkepley: This function will return `Disabled` if an autonode config is null. This will cause crashing

cmd/describe/cluster/cmd.go
olucasfreitas: From what I understood, when autonode is disabled, there is no IAM role associated with it, cluster.AWS().AutoNode() would be nil, so it made sense to me since one relies on the other to not show it when disbaled.

cmd/describe/cluster/cmd.go
olucasfreitas: fixed it in the last commit

cmd/describe/cluster/cmd.go
olucasfreitas: fixed it in the last commit

cmd/describe/cluster/cmd.go
hunterkepley: Why would disabling it remove its IAM role ARN? Could you show me where you saw this?

hunterkepley: @olucasfreitas

cmd/describe/cluster/cmd.go
```diff
@@ -566,6 +566,9 @@ func run(cmd *cobra.Command, argv []string) {
 				str = fmt.Sprintf("%s"+
 					"  IAM Role ARN:             %s\n", str, cluster.AWS().AutoNode().RoleArn())
 			}
+		} else {
+			str = fmt.Sprintf("%s"+
+				"AutoNode:                   %s\n", str, DisabledOutput)
```
hunterkepley, line 571: this also prints when autonode is nil, I don't think we want to tell the user its disabled when its nil, do we?

cmd/describe/cluster/cmd.go
olucasfreitas: I saw this on the audit log forwarding pattern, witch is why I thought it would make sense to not show the IAM role when AutoNode is disabled since it wouldn't be actively used. However, I've reverted the code to keep the original logic.

cmd/describe/cluster/cmd.go
olucasfreitas: This is a great catch! I've updated the code on the last commit

PR #3134, opened 2025-12-11
OCM-21147 | feat: Edit log forwarder command

hunterkepley:
Adds support for editing log forwarders:

```
./rosa edit log-forwarder -c hk5 2n4b8f8ai80cs6kmjmdgqlqplh73r411 --log-fwd-config ../test.yaml
```

### DIFF
--- a/cmd/edit/logforwarder/cmd.go
+++ b/cmd/edit/logforwarder/cmd.go
@@ -1,0 +1,120 @@
+/*
+Copyright (c) 2025 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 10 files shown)


cmd/edit/logforwarder/cmd.go
```diff
@@ -0,0 +1,94 @@
...
+		"2n4b8f8ai80cs6kmjmdgqlqplh73r411'"
+)
+
+var (
+	logFwdConfig string
+)
+
+func NewEditLogForwarderCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "log-forwarder -c <cluster-id> <log-fwd-id>",
```
marcolan018: should this be something like below?
log-forwarder -c <cluster-id> <log-fwd-id> --log-fwd-config <log-fwd-config-path>

hunterkepley: Oh, definitely, thank you

PR #3139, opened 2025-12-12
OCM-21152 | feat: Create command for logforwarders

hunterkepley:
```bash
❯ ./rosa create log-forwarder -c hk5
? Enabled log forwarding (optional, choose 'Skip' to skip selection; ): CloudWatch
? CloudWatch Log forwarding role ARN: test
? CloudWatch log group name: t
? CloudWatch Log forwarding pod groups: api, authentication, controller manager
? CloudWatch Log forwarding applications (optional):
E: failed to create log forwarder: status is 400, identifier is '400', code is 'CLUSTERS-MGMT-400', at '2025-12-12T19:20:54Z' and operation identifier is 'ec32cbdf-9e35-4e76-8f34-10de66fc5e02': Log distribution role does not exist or is not accessible


❯ ./rosa create log-forwarder -c hk5
? Enabled log forwarding (optional, choose 'Skip' to skip selection; ): S3
? S3 Bucket prefix (optional):
? S3 Bucket name: log-fwd-test
? S3 Log forwarding pod groups: authentication, controller manager, scheduler
? S3 Log forwarding applications (optional):
E: failed to create log forwarder: status is 409, identifier is '409', code is 'CLUSTERS-MGMT-409', at '2025-12-12T19:21:07Z' and operation identifier is '46016f3d-3e96-4f74-919a-4d4fcfd90ce5': A log forwarder of type 's3' already exists for this cluster. Only one log forwarder per type is allowed.



❯ ./rosa create log-forwarder -c hk5 --log-fwd-config ../test.yaml
E: failed to create log forwarder: status is 409, identifier is '409', code is 'CLUSTERS-MGMT-409', at '2025-12-12T19:30:06Z' and operation identifier is '328916db-7629-42c3-b025-824699638295': A log forwarder of type 's3' already exists for this cluster. Only one log forwarder per type is allowed.
```

**Basic creation command which leverages previously-made interactive prompts for minimal code changes**

All of the functions used are already tested, only real thing happening otherwise is sending the payload to the API. I argue therefore that it's almost useless for me to write a E2E test here with mocking due to the _**time contraint**_ as well as the _**functionality of the test**_ which would just be making sure we sent the right API object.. And it's already _**obvious**_ that we are sending the right API object since we are using the OCM SDK

### DIFF
--- a/cmd/create/logforwarder/cmd.go
+++ b/cmd/create/logforwarder/cmd.go
@@ -1,0 +1,157 @@
+/*
+Copyright (c) 2025 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package logforwarder
+
+import (
+	"context"
+	"fmt"
+
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/rosa/pkg/interactive"
+	interactiveLogForwarding "github.com/openshift/rosa/pkg/interactive/logforwarding"
+	"github.com/openshift/rosa/pkg/logforwarding"
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/output"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+const (
+	use   = "log-forwarder -c <cluster-id>"
+	short = "Create a log forwarder for a Hosted Control Plane cluster"
+	long  = "Create a log forwarder to forward logs from a hosted cluster to external services " +
+		"such as S3 or CloudWatch. Must create for an existing Hosted Control Plane cluster"
+	example = `  # Create a log forwarder using a config file
+  rosa create log-forwarder -c mycluster-hcp --log-fwd-config=s3.yml
+  
+  # Create a log forwarder interactively
+  rosa create log-forwarder -c mycluster-hcp --interactive`
+)
+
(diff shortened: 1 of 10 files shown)


cmd/create/logforwarder/main_test.go
```diff
@@ -0,0 +1,13 @@
+package logforwarder
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestCreateKubeletConfig(t *testing.T) {
```
AlexVulaj: Should this be "TestCreateLogForwarder" here? Along with line 12.

cmd/create/logforwarder/cmd.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/output"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+const (
+	use   = "log-forwarder -c <cluster-id>"
+	short = "Create a log forwarder for a Hosted Control Plane cluster"
+	long  = "Create a log forwarder to forward logs from a hosted cluster to external services " +
+		"such as S3 or CloudWatch. Must create on top of an existing Hosted Control Plane cluster"
```
AlexVulaj: I'd probably reword this to something like "Must be created for an existing Hosted Control Plane cluster", but that's a nitpick. Feel free to resolve if you disagree 🙂

cmd/create/logforwarder/main_test.go
hunterkepley: ha whoops, copy and pasted this file as you can tell

cmd/create/logforwarder/cmd.go
hunterkepley: makes sense, `on top` is a lil misleading

cmd/create/logforwarder/cmd.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+
+		if output.HasFlag() {
+			err = output.Print(createdLogForwarder)
+			if err != nil {
+				return fmt.Errorf("failed to output log forwarder: %v", err)
+			}
+			return nil
+		}
+
```
AlexVulaj, line 153: Should we add a success message here? Something like `r.Reporter.Infof("Successfully created log forwarder for cluster '%s'", clusterKey)`. I know other commands like creating a kubeletconfig or creating an autoscaler follow that pattern.

PR #3192, opened 2026-02-10
OCM-20977 | fix: add missing capacity-reservation-preference field in 'rosa describe machibepool'

olucasfreitas:
# Details

This PR adds the `capacity-reservation-preference` field to the output of `rosa describe machinepool`, so users can see what preference was set on their node pool.

---

## Fixed Behavior

```bash
./rosa describe machinepool --cluster=mycluster test-np
```

Output:
```
...
Capacity Reservation:                  
 - Preference:                         open
Management upgrade:                    
...
```

Now the user can see the preference they configured.

---

## Additional Cases

The existing output for capacity reservations with an ID continues to work as before:

1. **ID + Type (no preference)**:
    ```
    Capacity Reservation:                  
     - ID:                                 cr-abc123
     - Type:                               OnDemand
    ```

2. **ID + Type + Preference**:
    ```
    Capacity Reservation:                  
     - ID:                                 cr-abc123
     - Type:                               OnDemand
     - Preference:                         capacity-reservations-only
    ```

3. **Preference only (new)**:
    ```
    Capacity Reservation:                  
     - Preference:                         open
    ```

---

# Ticket

Closes [OCM-20977](https://issues.redhat.com/browse/OCM-20977)

### DIFF
--- a/pkg/ocm/output/nodepools.go
+++ b/pkg/ocm/output/nodepools.go
@@ -153,29 +153,24 @@ func PrintNodePoolDiskSize(aws *cmv1.AWSNodePool) string {
 }
 
 func PrintCapacityReservationDetails(capacityReservation *cmv1.AWSCapacityReservation) string {
-	if capacityReservation != nil {
-		id, ok := capacityReservation.GetId()
-		if !ok {
-			return ""
-		}
-		marketType, ok := capacityReservation.GetMarketType()
-		if !ok {
-			return fmt.Sprintf("\n"+
-				" - ID:                                 %s\n",
-				id)
-		}
-		preference, ok := capacityReservation.GetPreference()
-		if !ok {
-			return fmt.Sprintf("\n"+
-				" - ID:                                 %s\n"+
-				" - Type:                               %s",
-				id, marketType)
-		}
-		return fmt.Sprintf("\n"+
-			" - ID:                                 %s\n"+
-			" - Type:                               %s\n"+
-			" - Preference:                         %s",
-			id, marketType, preference)
+	if capacityReservation == nil {
+		return ""
 	}
-	return ""
+
+	var details []string
+
+	if id, ok := capacityReservation.GetId(); ok {
+		details = append(details, fmt.Sprintf(" - ID:                                 %s", id))
+	}
+	if marketType, ok := capacityReservation.GetMarketType(); ok {
+		details = append(details, fmt.Sprintf(" - Type:                               %s", marketType))
+	}
+	if preference, ok := capacityReservation.GetPreference(); ok {
+		details = append(details, fmt.Sprintf(" - Preference:                         %s", preference))
+	}
+
+	if len(details) == 0 {
+		return ""
+	}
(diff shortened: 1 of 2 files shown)


pkg/ocm/output/nodepools.go
```diff
@@ -154,9 +154,15 @@ func PrintNodePoolDiskSize(aws *cmv1.AWSNodePool) string {
 
 func PrintCapacityReservationDetails(capacityReservation *cmv1.AWSCapacityReservation) string {
 	if capacityReservation != nil {
-		id, ok := capacityReservation.GetId()
-		if !ok {
-			return ""
+		id, hasId := capacityReservation.GetId()
```
hunterkepley: out of convention, I would like `hasId` to become `ok` 🙂 its just a convention that we use

pkg/ocm/output/nodepools.go
hunterkepley: it looks like we already have preference added here, I think the issue might be more complicated than just 'we forgot it'

It looks like we are returning early before returning this block with the preference, I think the right solution for this ticket would be to fix the issue directly, rather than stack on a couple new if statements and a return

Likely, one of the above returns is being used before this one is, causing the preference to not be visible, and we should figure out why

Also there's quite a few tests introduced in this MR, I think what would be important would be to directly test the outcome, so keep the `cmd_test.go` changes, but we don't need anything else beyond that, unless a test fails (in that case, we should modify it)

pkg/ocm/output/nodepools.go
olucasfreitas: fixed here [dd08b16](https://github.com/openshift/rosa/pull/3192/commits/dd08b16d30a837280e331159e9b15add81c588c9)

pkg/ocm/output/nodepools.go
olucasfreitas: I think I found a solution that works, please check here [dd08b16](https://github.com/openshift/rosa/pull/3192/commits/dd08b16d30a837280e331159e9b15add81c588c9)

pkg/ocm/output/nodepools.go
```diff
@@ -153,29 +153,24 @@ func PrintNodePoolDiskSize(aws *cmv1.AWSNodePool) string {
...
+		details = append(details, fmt.Sprintf(" - Type:                               %s", marketType))
+	}
+	if preference, ok := capacityReservation.GetPreference(); ok {
+		details = append(details, fmt.Sprintf(" - Preference:                         %s", preference))
+	}
+
+	if len(details) == 0 {
+		return ""
+	}
+	return "\n" + strings.Join(details, "\n")
```
hunterkepley, line 175: The output string should already have a newline in place, im expecting this to print out an empty line between `Capacity Reservation` and the actual block

can you confirm if that's the case? if it looks fine then LGTM

hunterkepley: In your examples it doesn't, but your examples are snippets so I can't confirm what the full output looks like

olucasfreitas: image

PR #3236, opened 2026-04-20
OCM-23979 | fix: support set min_replicas=0 when editing nodepool

marcolan018:
## PR Summary

currently when edit nodepool(machinepool for hcp cluster), if set min_replicas=0, the edit command will success but in fact the update does not occur. This pr is to fix this problem. It also makes enhancement to existing validation to gurantee the min never greater than max

## Related Issues and PRs

- Jira: [OCM-23979](https://redhat.atlassian.net/browse/OCM-23979)
- Fixes: `#`
- Related PR(s):
- Related design/docs:

## Type of Change

### Test Steps
1. create a hcp cluster with nodepool replicas=2
2. edit nodepool to enable autoscaling with min_replica=0
3. 

### Expected Results

use describe command to verify the nodepool is updated: autoscaling is enabled, and min replicas is 0

## Proof of the Fix

- Screenshots:
- Videos:
- Logs/CLI output:
```
bash-5.2$ ./rosa list machinepool -c marco0419
ID       AUTOSCALING  REPLICAS  INSTANCE TYPE  LABELS    TAINTS    AVAILABILITY ZONE  SUBNET                    DISK SIZE  VERSION  AUTOREPAIR  
workers  No           2/2       m5.xlarge                          us-west-2a         subnet-0464543fcd59f75d6  300 GiB    4.21.7   Yes         
bash-5.2$ ./rosa edit machinepool -c marco0419 workers
? Enable autoscaling: Yes
? Min replicas: 0
? Max replicas: 2
? Labels (optional): 
? Taints (optional): 
? Autorepair: Yes
? Node drain grace period (optional): 
? Max surge: 1
? Max unavailable: 0
I: Updated machine pool 'workers' on hosted cluster 'marco0419'
bash-5.2$ ./rosa list machinepool -c marco0419
ID       AUTOSCALING  REPLICAS  INSTANCE TYPE  LABELS    TAINTS    AVAILABILITY ZONE  SUBNET                    DISK SIZE  VERSION  AUTOREPAIR  
workers  Yes          2/0-2     m5.xlarge                          us-west-2a         subnet-0464543fcd59f75d6  300 GiB    4.21.7   Yes  
```
- Other artifacts:

## Breaking Changes
- [x] No breaking changes

- [ ] Commit subject/title follows `[JIRA-TICKET] | [TYPE]: <MESSAGE>`.
- [ ] PR description clearly explains both **what** changed and **why**.
- [ ] Relevant Jira/GitHub issues and related PRs are linked.
- [ ] `make test` passes.
- [ ] `make lint` passes.
- [ ] `make rosa` passes.

[OCM-23979]: https://redhat.atlassian.net/browse/OCM-23979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved autoscaling validation and editing: correct detection of min/max changes, better handling of interactive edits, support for min replicas = 0, and more reliable prompt defaults that prefer existing autoscaling bounds.
  * Strengthened max-replicas validation to reject non‑positive values.

* **Tests**
  * Updated and expanded tests to cover explicit min/max flag behavior, min=0 cases, max<=0 cases, and related regressions.

### DIFF
--- a/pkg/machinepool/machinepool.go
+++ b/pkg/machinepool/machinepool.go
@@ -1507,9 +1507,9 @@ func (m *machinePool) EditMachinePool(cmd *cobra.Command, machinePoolId string, 
 
 // fillAutoScalingAndReplicas is filling either autoscaling or replicas value in the builder
 func fillAutoScalingAndReplicas(npBuilder *cmv1.NodePoolBuilder, autoscaling bool, existingNodepool *cmv1.NodePool,
-	minReplicas int, maxReplicas int, replicas int) {
+	minReplicas int, maxReplicas int, replicas int, isMinReplicasSet bool, isMaxReplicasSet bool) {
 	if autoscaling {
-		npBuilder.Autoscaling(editAutoscaling(existingNodepool, minReplicas, maxReplicas))
+		npBuilder.Autoscaling(editAutoscaling(existingNodepool, minReplicas, maxReplicas, isMinReplicasSet, isMaxReplicasSet))
 	} else {
 		npBuilder.Replicas(replicas)
 	}
@@ -1818,7 +1818,8 @@ func editNodePool(cmd *cobra.Command, nodePoolID string,
 		return fmt.Errorf("failed to get autoscaling or replicas: '%s'", err)
 	}
 
-	err = validateNodePoolEdit(cmd, autoscaling, replicas, minReplicas, maxReplicas)
+	err = validateNodePoolEdit(cmd, autoscaling, replicas, minReplicas, maxReplicas, nodePool,
+		isMinReplicasSet, isMaxReplicasSet)
 	if err != nil {
 		return err
 	}
@@ -1842,7 +1843,8 @@ func editNodePool(cmd *cobra.Command, nodePoolID string,
 		npBuilder = npBuilder.Taints(taintBuilders...)
 	}
 
-	fillAutoScalingAndReplicas(npBuilder, autoscaling, nodePool, minReplicas, maxReplicas, replicas)
+	fillAutoScalingAndReplicas(npBuilder, autoscaling, nodePool, minReplicas, maxReplicas, replicas,
+		isMinReplicasSet, isMaxReplicasSet)
 
 	if isAutorepairSet || interactive.Enabled() {
 		autorepair, err := strconv.ParseBool(cmd.Flags().Lookup("autorepair").Value.String())
@@ -2105,7 +2107,8 @@ func editNodePool(cmd *cobra.Command, nodePoolID string,
 	return nil
 }
 
-func validateNodePoolEdit(cmd *cobra.Command, autoscaling bool, replicas int, minReplicas int, maxReplicas int) error {
+func validateNodePoolEdit(cmd *cobra.Command, autoscaling bool, replicas int, minReplicas int, maxReplicas int,
+	nodePool *cmv1.NodePool, isMinReplicasSet bool, isMaxReplicasSet bool) error {
 	if !autoscaling && replicas < 0 {
 		return fmt.Errorf("the number of machine pool replicas needs to be a non-negative integer")
 	}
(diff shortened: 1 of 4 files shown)


marcolan018:
@coderabbitai review

olucasfreitas (commented):
A couple of edge cases are still reachable in the replicas-only to autoscaling path.

pkg/machinepool/machinepool.go
```diff
@@ -2282,21 +2300,28 @@ func getNodePoolReplicas(cmd *cobra.Command,
...
 		min = minReplicas
 	}
-	if maxReplicas != 0 {
+	if (isMaxReplicasSet || interactive.Enabled()) && maxReplicas >= 1 {
 		max = maxReplicas
 	}
 
-	if min >= 1 && max >= 1 {
+	// Return builder if enabling autoscaling or changing values
+	if nodePool.Autoscaling() == nil ||
```
olucasfreitas, line 2341: This nil-handling in `editAutoscaling()` looks right, but the fixed-replica to autoscaling flow still calls `existingAutoscaling.MinReplica()` and `MaxReplica()` in `getNodePoolReplicas()`. Since a replicas-only nodepool has `Autoscaling() == nil`, the interactive path can still fail before it reaches this builder. Could we guard that caller path and seed the defaults from the current replica state instead?

pkg/machinepool/machinepool.go
```diff
@@ -2118,10 +2121,25 @@ func validateNodePoolEdit(cmd *cobra.Command, autoscaling bool, replicas int, mi
 		return fmt.Errorf("max-replicas must be greater than zero")
 	}
 
-	if autoscaling && cmd.Flags().Changed("max-replicas") && cmd.Flags().Changed(
-		"min-replicas") && minReplicas > maxReplicas {
-		return fmt.Errorf("the number of machine pool min-replicas needs to be less than the number of " +
-			"machine pool max-replicas")
+	// Validate the final computed min/max range after merging with existing values
+	// This catches cases where only one bound is changed, producing invalid range
+	if autoscaling && (isMinReplicasSet || isMaxReplicasSet || interactive.Enabled()) {
```
olucasfreitas, line 2126: This validation now catches `min > max` when an existing autoscaling block is present, but the replicas-only to autoscaling path can still leave the missing bound at `0`. A flow like `--enable-autoscaling --min-replicas 0` can still get through here and then build `MinReplica(0).MaxReplica(0)`. Could we either require both bounds when enabling autoscaling, or default the missing one from the current replica count before building the patch?

pkg/machinepool/machinepool.go
marcolan018: fixed

pkg/machinepool/machinepool.go
marcolan018: @olucasfreitas 
thanks for the review
I added more checking for the edge cases, along with more unit tests

pkg/machinepool/machinepool.go
olucasfreitas: The nil dereference part looks fixed, but I still see one follow-up in the replicas-only -> autoscaling flow: `getNodePoolReplicas()` now seeds the missing bound from `existingReplicas`, but `editAutoscaling()` still starts from `0/0` when `nodePool.Autoscaling() == nil` and ignores that seeded value unless the matching flag was set.

With the current code path:
- `rosa edit machinepool ... --enable-autoscaling --min-replicas 0` still resolves to `0/0`
- `rosa edit machinepool ... --enable-autoscaling --max-replicas 5` still resolves to `0/5`

I think this still needs one more update in `editAutoscaling()` plus min-only / max-only tests for the replicas-only case.

pkg/machinepool/machinepool.go
olucasfreitas: The extra validation helps, but the original interactive Jira flow still looks open for a replicas-only nodepool. In `getNodePoolReplicas()`, the new fallback block still treats an interactively entered `0` as unset when the flag itself was not changed and rewrites it back to `existingReplicas`.

So this flow can still lose the `0` before the patch is built:
- `rosa edit machinepool -c <cluster> <pool>`
- `Enable autoscaling: Yes`
- `Min replicas: 0`
- `Max replicas: 2`

I think this needs a test that drives the interactive replicas-only -> autoscaling path, not just the helper path with both bounds set explicitly.

pkg/machinepool/machinepool.go
marcolan018: set min/max value with replicas

pkg/machinepool/machinepool.go
marcolan018: good point
I added the check for such case
thanks a lot for the careful reviews
I squash the commits to avoid too many in this pr
@olucasfreitas  please help take a look on the latest changes, thanks

olucasfreitas: I pulled the latest head and the earlier `min=0` cases look fixed now. I still see one interactive edge case though: if the user enters `Max replicas: 0`, the prompt path can accept it, but `editAutoscaling()` still ignores it because only values `>= 1` are applied there.

That means the command can look successful while keeping the previous/default max instead of the value the user entered. Could we reject `0` in the interactive max validator too, or handle it consistently end-to-end? A small test for the interactive `max=0` path would help lock that down.

marcolan018: @olucasfreitas 
that concern make sense. I added a validation for this

pkg/machinepool/machinepool.go
```diff
@@ -2118,10 +2121,25 @@ func validateNodePoolEdit(cmd *cobra.Command, autoscaling bool, replicas int, mi
...
+				finalMin = nodePool.Autoscaling().MinReplica()
+			}
+			if !isMaxReplicasSet && !interactive.Enabled() {
+				finalMax = nodePool.Autoscaling().MaxReplica()
+			}
+		}
+
+		if finalMin > finalMax {
+			return fmt.Errorf("the number of machine pool min-replicas must not be greater than max-replicas")
+		}
```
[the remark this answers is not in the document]
olucasfreitas: The new validator in `helper.go` fixes the interactive `max=0` path, but the fixed-size -> autoscaling path still defaults an omitted bound from `existingReplicas`. If the current nodepool has `replicas=0`, enabling autoscaling without `--max-replicas` can still carry `maxReplicas=0`, and `validateNodePoolEdit()` does not reject it because `--max-replicas` was not changed and the final check only compares `min > max`.

That still leaves room to build `MinReplica(0).MaxReplica(0)` on a fixed `0`-replica nodepool. I think this needs either a final `finalMax >= 1` guard after defaulting, or a focused test that covers enabling autoscaling from a fixed nodepool with `replicas=0` and omitted max.

PR #3238, opened 2026-04-23
OCM-24123: Refactor proxy URL validation logic

michaelryanmcneill:
## PR Summary

Refactor proxy URL validation logic to accept any non-empty scheme. Update tests to reflect changes in acceptance criteria for HTTP and HTTPS proxies.

Proxies may use HTTP or HTTPS schemes, but the ROSA CLI mandates `http-proxy` uses `http` scheme and `https-proxy` uses `https` scheme. This is not correct.

## Related Issues and PRs

- Jira: [OCM-24123](https://redhat.atlassian.net/browse/OCM-24123)

## Type of Change

## Previous Behavior

```
--http-proxy http://ip-172-17-1-167.ap-southeast-1.compute.internal:3128/ \
--https-proxy http://ip-172-17-1-167.ap-southeast-1.compute.internal:3128/ \
```
Fails with
```
E: expected 'https-proxy' to have an 'https://' scheme
```

## Behavior After This Change

```
--http-proxy http://ip-172-17-1-167.ap-southeast-1.compute.internal:3128/ \
--https-proxy http://ip-172-17-1-167.ap-southeast-1.compute.internal:3128/ \
```
Succeeds.

## How to Test (Step-by-Step)

Set an `--http-proxy` to `https://proxy.example.com:8443`
Set an `--https-proxy` to `http://proxy.example.com:8080`

Validate that the change succeed. 

### Test Steps
1.
2.
3.

## Proof of the Fix

- Screenshots:
- Videos:
- Logs/CLI output:
- Other artifacts:

## Breaking Changes
- [x] No breaking changes

- [x] `make test` passes.
- [x] `make lint` passes.
- [x] `make rosa` passes.

### Notes
- `make test` fails in an unrelated/unmodified test.

[OCM-24123]: https://redhat.atlassian.net/browse/OCM-24123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Proxy URL validation refined: HTTPS proxy now accepts http:// and https://; HTTP proxy requires http://. URLs with non-root/extra path components remain rejected. Error messages normalized to list allowed scheme(s).

* **Tests**
  * Unit and end-to-end tests updated to cover allowed-scheme permutations and new error phrasing.

* **Documentation**
  * CLI help text clarified to state allowed proxy schemes (http, https).

### DIFF
--- a/pkg/ocm/helpers.go
+++ b/pkg/ocm/helpers.go
@@ -134,8 +134,28 @@ func ClusterDomainPrefixValidator(domainPrefix interface{}) error {
 	return fmt.Errorf("can only validate strings, got '%v'", domainPrefix)
 }
 
-// validateProxyURL validates common proxy URL requirements
-func validateProxyURL(proxyURL string, proxyType string, expectedScheme string) error {
+// proxyAllowedURLSchemeErrorFragment lists allowed schemes for error messages, for example
+// 'http://' or 'https://' for two schemes.
+func proxyAllowedURLSchemeErrorFragment(allowedSchemes []string) string {
+	quoted := make([]string, len(allowedSchemes))
+	for i, s := range allowedSchemes {
+		quoted[i] = fmt.Sprintf("'%s://'", s)
+	}
+	switch len(quoted) {
+	case 0:
+		return ""
+	case 1:
+		return quoted[0]
+	case 2:
+		return quoted[0] + " or " + quoted[1]
+	default:
+		return strings.Join(quoted[:len(quoted)-1], ", ") + ", or " + quoted[len(quoted)-1]
+	}
+}
+
+// validateProxyURL validates common proxy URL requirements. allowedSchemes lists
+// acceptable URL schemes without the "://" suffix (for example "http", "https").
+func validateProxyURL(proxyURL string, proxyType string, allowedSchemes []string) error {
 	if strings.Contains(proxyURL, " ") {
 		return fmt.Errorf("invalid %s value: URL cannot contain spaces", proxyType)
 	}
@@ -153,8 +173,20 @@ func validateProxyURL(proxyURL string, proxyType string, expectedScheme string) 
 		return fmt.Errorf("invalid %s value: %v", proxyType, err)
 	}
 
-	if parsedURL.Scheme != expectedScheme {
-		return fmt.Errorf("expected '%s' to have an '%s://' scheme", proxyType, expectedScheme)
+	scheme := strings.ToLower(parsedURL.Scheme)
+	okScheme := false
+	for _, allowed := range allowedSchemes {
+		if scheme == strings.ToLower(allowed) {
+			okScheme = true
+			break
+		}
+	}
+	if !okScheme {
+		if len(allowedSchemes) == 0 {
+			return fmt.Errorf("invalid %s value: URL scheme is not allowed", proxyType)
+		}
+		schemeList := proxyAllowedURLSchemeErrorFragment(allowedSchemes)
+		return fmt.Errorf("invalid %s value: URL scheme must be %s", proxyType, schemeList)
 	}
 
 	if parsedURL.Host == "" {
@@ -174,7 +206,7 @@ func ValidateHTTPProxy(val interface{}) error {
 		if httpProxy == "" {
 			return nil
 		}
-		return validateProxyURL(httpProxy, "http-proxy", "http")
+		return validateProxyURL(httpProxy, "http-proxy", []string{"http"})
 	}
 	return fmt.Errorf("can only validate strings, got '%v'", val)
 }
@@ -184,7 +216,7 @@ func ValidateHTTPSProxy(val interface{}) error {
 		if httpsProxy == "" {
 			return nil
 		}
-		return validateProxyURL(httpsProxy, "https-proxy", "https")
+		return validateProxyURL(httpsProxy, "https-proxy", []string{"http", "https"})
 	}
 	return fmt.Errorf("can only validate strings, got '%v'", val)
 }
--- a/pkg/ocm/helpers_test.go
+++ b/pkg/ocm/helpers_test.go
@@ -787,7 +787,13 @@ var _ = Describe("ValidateHTTPProxy", func() {
 	It("rejects proxy with https scheme", func() {
 		err := ValidateHTTPProxy("https://proxy.example.com:8080")
 		Expect(err).To(HaveOccurred())
-		Expect(err.Error()).To(Equal("expected 'http-proxy' to have an 'http://' scheme"))
+		Expect(err.Error()).To(Equal("invalid http-proxy value: URL scheme must be 'http://'"))
+	})
+
+	It("rejects proxy with non-http scheme", func() {
+		err := ValidateHTTPProxy("ftp://proxy.example.com:1080")
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(Equal("invalid http-proxy value: URL scheme must be 'http://'"))
 	})
 
 	It("rejects proxy without scheme", func() {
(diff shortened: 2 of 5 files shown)


pkg/ocm/helpers.go
```diff
@@ -153,8 +155,8 @@ func validateProxyURL(proxyURL string, proxyType string, expectedScheme string)
 		return fmt.Errorf("invalid %s value: %v", proxyType, err)
 	}
 
-	if parsedURL.Scheme != expectedScheme {
-		return fmt.Errorf("expected '%s' to have an '%s://' scheme", proxyType, expectedScheme)
+	if parsedURL.Scheme == "" {
```
olucasfreitas: This now treats any non-empty scheme as valid. The current OpenShift proxy docs still describe `httpProxy` as `http`-only and `httpsProxy` as `http` or `https`, so this would also allow values like `ftp://...` or `https://...` for `--http-proxy`. Could we keep a small allowlist per flag here instead of only checking that a scheme exists?

pkg/ocm/helpers_test.go
```diff
@@ -784,10 +784,9 @@ var _ = Describe("ValidateHTTPProxy", func() {
 		Expect(err).NotTo(HaveOccurred())
 	})
 
-	It("rejects proxy with https scheme", func() {
+	It("accepts proxy with https scheme", func() {
```
olucasfreitas: This test now locks in `https://...` as valid for `ValidateHTTPProxy()`, but the current proxy contract still treats `httpProxy` as `http`-only. Could we keep this as a negative case and add coverage that `ValidateHTTPSProxy()` accepts both `http://...` and `https://...` instead?

tests/e2e/test_rosacli_cluster.go
```diff
@@ -885,10 +885,10 @@ var _ = Describe("Edit cluster validation should", labels.Feature.Cluster, func(
 			}
 			fmt.Println(originalHttpProxy, originalHTTPSProxy, originalNoProxy, originalCAFile)
 
-			By("Edit cluster with invalid http_proxy not started with http")
+			By("Edit cluster with invalid http_proxy values")
 			invalidHTTPProxy := map[string]string{
```
olucasfreitas, line 889: This drops the end-to-end check that `--http-proxy` rejects `https://...`, so the suite no longer protects the documented `httpProxy` behavior. Could we keep that invalid case and add a happy-path that `--https-proxy` accepts `http://...`?

olucasfreitas (commented):
There is one blocking point left: the validator now accepts any non-empty scheme, but the current OpenShift proxy docs still describe `httpProxy` as `http`-only and `httpsProxy` as `http` or `https`. I left inline notes on the validator and the two tests that currently widen the contract too far.

tests/e2e/test_rosacli_cluster.go
michaelryanmcneill: The documentation should be updated, there are use cases where a TLS proxy can be used for HTTP traffic.

pkg/ocm/helpers_test.go
michaelryanmcneill: The documentation should be updated, there are use cases where a TLS proxy can be used for HTTP traffic.

pkg/ocm/helpers.go
michaelryanmcneill: The documentation should be updated, there are use cases where a TLS proxy can be used for HTTP traffic.

olucasfreitas: If that broader behavior is intentional, I think we still need the contract source and the repo docs/help updated in this PR. Right now the OpenShift proxy docs and our own `--http-proxy` help text in `cmd/create/cluster/cmd.go` and `cmd/edit/cluster/cmd.go` still describe `http-proxy` as `http`-only, so I don't think we should move to `any non-empty scheme` without narrowing the allowlist or updating that documented contract first.

pkg/ocm/helpers_test.go
olucasfreitas: If the intended behavior is broader, I'd still prefer the unit tests to encode the exact supported schemes rather than `any non-empty scheme`. That keeps undocumented values like `ftp://...` from becoming accepted by accident.

tests/e2e/test_rosacli_cluster.go
olucasfreitas: If the contract is changing here, could we add the replacement e2e coverage in this PR as well? Right now the previous `http-proxy` regression check was removed, but there still isn't a positive end-to-end case showing the new supported behavior.

pkg/ocm/helpers.go
olucasfreitas: I agree the TLS proxy case is real, but the current code still accepts any non-empty scheme, not just `http`/`https`. If we want to broaden the contract, could we pin it down to an explicit allowlist per flag and update the CLI help text in `cmd/create/cluster/cmd.go` + `cmd/edit/cluster/cmd.go` in the same PR (plus a link to the product/API source that backs the change)?

pkg/ocm/helpers_test.go
olucasfreitas: If we keep the broader behavior, could we update the unit tests to assert the exact supported schemes (instead of `any non-empty scheme`) so we do not accidentally accept values like `ftp://...`?

tests/e2e/test_rosacli_cluster.go
olucasfreitas: If the e2e regression for `https://...` on `--http-proxy` is intentionally removed, could we add a replacement e2e that covers the new supported behavior (for example a happy-path `--https-proxy http://...` edit) so we still have end-to-end coverage for the change?

pkg/ocm/helpers.go
amandahla: I think a simple solution for now would be checking if scheme is http or https and fail otherwise.

tests/e2e/test_rosacli_cluster.go
michaelryanmcneill: Went ahead and returned to the original contract for `--http-proxy`.

olucasfreitas (commented):
Two small non-blocking notes for a follow-up if you want:

1. The `--https-proxy` help text in `cmd/create/cluster/cmd.go` and `cmd/edit/cluster/cmd.go` could mention that `http://` is also accepted, so the CLI help matches the new behavior.
2. The multi-scheme error message on line 169 of `helpers.go` is hardcoded to `http:// or https://` rather than built from the `allowedSchemes` slice, works today but would silently drift if a future caller passes a different list.

michaelryanmcneill:
Thanks @olucasfreitas, working on fixing those now and then I'll have ready for final review and merge!

tests/e2e/test_rosacli_cluster.go
```diff
@@ -885,10 +885,11 @@ var _ = Describe("Edit cluster validation should", labels.Feature.Cluster, func(
...
-				"invalidvalue":           "ERR: invalid http-proxy value: URL is missing scheme",
-				"https://test-proxy.com": "ERR: expected 'http-proxy' to have an 'http://' scheme",
+				"invalidvalue":                "ERR: invalid 'http-proxy' value: URL is missing scheme",
+				"https://test-proxy.com":      "ERR: expected 'http-proxy' to have an 'http://' scheme",
+				"http://test-proxy.com/extra": "ERR: invalid 'http-proxy' value: proxy URL should not contain a path '/extra'",
```
olucasfreitas: The e2e error strings look off,  the assertions here wrap the proxy type in single quotes (`invalid 'http-proxy' value`) but the validator in `helpers.go` does not quote it (`invalid http-proxy value`). Since the test uses `ContainSubstring`, this will fail at runtime. Lines 904 and 906 have the same issue for `https-proxy`. Might just need to drop the quotes around the proxy type to match what the code actually returns.

olucasfreitas (commented):
Looking good overall, just one thing I noticed in the e2e.

tests/e2e/test_rosacli_cluster.go
michaelryanmcneill: Thanks @olucasfreitas, that was my bad. I standardized the new error to match the others as well. This should be good (addressed Codecov as well).

PR #3241, opened 2026-05-13
OCM-24858 | fix: update the default machinepool instance type

olucasfreitas:
JIRA: [OCM-24858](https://redhat.atlassian.net/browse/OCM-24858)

[OCM-24858]: https://redhat.atlassian.net/browse/OCM-24858?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Chores**
  * CLI no longer hard-codes a default instance type for machine pool creation; the flag defaults to unspecified so the cluster’s default is used.
  * Help text and tests updated to reflect using the cluster-derived default when no --instance-type is supplied.
* **Bug Fixes**
  * Creation flows now derive and reuse default instance type and root disk size, avoiding redundant lookups.
* **Bug Fixes**
  * Spot price handling treats the “on-demand” value consistently.

### DIFF
--- a/cmd/create/machinepool/options.go
+++ b/cmd/create/machinepool/options.go
@@ -5,8 +5,6 @@ import (
 	"github.com/openshift/rosa/pkg/reporter"
 )
 
-const instanceType = "m5.xlarge"
-
 type CreateMachinepoolOptions struct {
 	reporter reporter.Logger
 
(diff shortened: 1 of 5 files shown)


amandahla:
Like we discussed, please consider using the OCM API call like is already implemented in machinepool.go (_, _, _, _, defaultRootDiskSize, _ :=
		r.OCMClient.GetDefaultClusterFlavors(cluster.Flavour().ID())) so we can be sure that the OCM is the source of truth.
@coderabbitai wdyt

olucasfreitas:
@amandahla right now we can't use the api for this, because in prod and staging the instance type it gives us is m5.large, and as we want to use the m7i.xlarge, we need to hardcode it 

this is just for a quick release now, as soon as the API team fixes this on their side, we are going to fix this [here](https://redhat.atlassian.net/browse/OCM-24857)

amandahla (approved):

amandahla:
/lgtm
I was testing coderabbitai ;-) Approved as temporary workaround.

PR #3250, opened 2026-05-27
OCM-24429 | feat: Updated rosa list ocm-role to display console access

robpblake:
Adds to the output of `rosa list ocm-role` a field that allows customers to see if the `OCM Role` they have configured has the `no-console` profile.

Example output

```
rblake@rblake-mac rosa % rosa list ocm-role
I: Fetching ocm roles
ROLE NAME                                 ROLE ARN                                                                      LINKED  ADMIN  AWS Managed  CONSOLE ACCESS
rblake-OCM-Role-123456789                  arn:aws:iam::123456789:role/rblake-OCM-Role-123456789                       No      No     No           No
```

## PR Summary

This PR updates the `rosa list ocm-role` command to show if the `OCM Role` the customer has configured has can be used to deploy clusters via `console.redhat.com`. This field is captured via the `CONSOLE ACCESS` column in the output, which displays `No` if the user has a `--no-console` version of the `OCM Role` configured, and `yes` otherwise

Examples of the various forms of role showing in the output:

```
Admin Profile:

rblake@rblake-mac rosa % rosa list ocm-role
I: Fetching ocm roles
ROLE NAME                                 ROLE ARN                                                                      LINKED  ADMIN  AWS Managed  CONSOLE ACCESS
rblake-OCM-Role-123456789                  arn:aws:iam::123456789:role/rblake-OCM-Role-123456789                       No      Yes    No           Yes

Standard Profile:

rblake@rblake-mac rosa % rosa list ocm-role
I: Fetching ocm roles
ROLE NAME                                 ROLE ARN                                                                      LINKED  ADMIN  AWS Managed  CONSOLE ACCESS
rblake-OCM-Role-123456789                  arn:aws:iam::123456789:role/rblake-OCM-Role-123456789                       No      No     No           Yes

No Console profile:

rblake@rblake-mac rosa % rosa list ocm-role
I: Fetching ocm roles
ROLE NAME                                 ROLE ARN                                                                      LINKED  ADMIN  AWS Managed  CONSOLE ACCESS
rblake-OCM-Role-123456789                  arn:aws:iam::123456789:role/rblake-OCM-Role-123456789                       No      No     No           No
```

## Related Issues and PRs

- Jira: https://redhat.atlassian.net/browse/OCM-24429
- Related design/docs: https://github.com/openshift-online/rosa-enhancements/pull/50/changes

## Type of Change

- [X] feat - adds a new user-facing capability.
- [ ] fix - resolves an incorrect behavior or bug.
[...]

PR #3252, opened 2026-05-27
OCM-23343 | feat: Added support for --no-console to rosa create ocm-role

andclt:
## PR Summary
Adds `--no-console` flag to `rosa create ocm-role` for creating OCM roles with minimal permissions for customers not using console.redhat.com.

This PR adds support for creating OCM roles with minimal permissions (`no-console`).

 ## Related Issues and PRs
  - Jira: [OCM-23880](https://issues.redhat.com/browse/OCM-23880)
  - Related: [OCM-23336](https://issues.redhat.com/browse/OCM-23336) (backend support)
  - Related PR(s): 
      - #3250 (rosa list ocm-role console access column)
      - [OCM-25298](https://issues.redhat.com/browse/OCM-25298) (linter fixes)
  - Related design/docs:  [design doc](https://github.com/openshift-online/rosa-enhancements/blob/main/enhancements/no-console-ocm-role.md)

## Type of Change

## Previous Behavior

Users could only create standard or admin OCM roles. All OCM roles had permissions suitable for console.redhat.com usage.

## Behavior After This Change

 Users can create three OCM role profiles:
  - **Standard** (default): Full OCM permissions for console and CLI usage
  - **Admin** (`--admin`): Enhanced permissions for administrative operations
  - **No-console** (`--no-console`): Minimal permissions

The `--admin` and `--no-console` flags are mutually exclusive. No-console roles are tagged with `rosa_no_console_role:true` and attach the `sts_ocm_no_console_permission_policy` instead of the standard policy.

### New --no-console flag for OCM role creation:
  - Customers can now create OCM roles with minimal permissions using `rosa create ocm-role --no-console`
  - The `--no-console` and `--admin` flags are mutually exclusive
  - When creating a no-console role, a warning is displayed: "This OCM role cannot be used to provision clusters via console.redhat.com"

###  Interactive mode behavior:
  - In interactive mode, users are first asked if they want an admin role
  - If they decline admin, they're then asked if they want a no-console role
  - This ensures users understand they have three options: standard, admin, or no-console

 ### Role conflict detection:
  - If a role already exists, the CLI provides clear guidance on what to do based on the requested vs. existing role type
  - Users cannot convert between role types - they must delete and recreate

###  Manual mode (--mode manual):
  - Generates the correct policy files and AWS CLI commands for no-console roles
  - Policy file: `sts_ocm_no_console_permission_policy.json` instead of the standard policy file

###  Graceful degradation:
  - If the no-console policy is not yet available from the OCM API, users get a clear error: "no-console OCM role is not yet enabled"

### Preconditions
- AWS credentials configured with appropriate IAM permissions
- ROSA CLI built from this branch: `make rosa`
- OCM login configured: `rosa login`

#### Test 1: Mutual Exclusivity
**Command:**
```bash
./rosa create ocm-role --prefix test --admin --no-console
```
**Expected:** Error about mutually exclusive flags

**Result:** PASS
```
Error: if any flags in the group [admin no-console] are set none of the others can be; 
[admin no-console] were all set
```

---

#### Test 2: Standard Role Creation
**Command:**
```bash
./rosa create ocm-role --prefix test-standard --mode auto
```
**Expected:** Role created with no special tags (no admin, no no-console)

**Result:** PASS
```
I: Created role 'test-standard-OCM-Role-13829321'
I: Attached policy 'test-standard-OCM-Role-13829321-Policy'

Tags: rosa_role_type:OCM, red-hat-managed:true, rosa_environment:production
NO rosa_admin_role or rosa_no_console_role tags 
```

---

#### Test 3: Admin Role Creation
**Command:**
```bash
./rosa create ocm-role --prefix test-admin --admin --mode auto
```
**Expected:** Admin tag present, two policies attached

**Result:** PASS
```
I: Created role 'test-admin-OCM-Role-13829321'
I: Attached policy 'test-admin-OCM-Role-13829321-Policy'
I: Attached policy 'test-admin-OCM-Role-13829321-Admin-Policy'

Tags: rosa_admin_role:true 
Policies: Standard + Admin (2 total) 
```

---

#### Test 4: No-Console - Graceful Degradation (Production OCM)
**Command:**
```bash
./rosa create ocm-role --prefix test-noconsole --no-console --mode auto
```
**Expected:** Warning + error when policy not available, no AWS resources created

**Result:** PASS
```
W: This OCM role cannot be used to provision clusters via console.redhat.com 
I: Created role 'test-noconsole-OCM-Role-13829321'
E: There was an error creating the ocm role: no-console OCM role is not yet enabled 

Graceful error handling works correctly 
No partial resources created 
```

---

#### Test 5: No-Console - Full End-to-End (Local OCM with Policy)
**Command:**
```bash
./rosa create ocm-role --prefix test-noconsole-local --no-console --mode auto
```
**Expected:** Warning + full role creation with no-console policy

**Result:** PASS
```
W: This OCM role cannot be used to provision clusters via console.redhat.com 
I: Created role 'test-noconsole-local-OCM-Role-12541229'
I: Attached policy 'test-noconsole-local-OCM-Role-12541229-NoConsole-Policy' 

Tags: rosa_no_console_role:true 
Policy: Only NoConsole-Policy (NOT standard) 
Policy Content: {"Action":"iam:GetRole", "Condition":{...tags...}} 
```

---

#### Test 6: Manual Mode - Graceful Degradation (Production OCM)
**Command:**
```bash
mkdir /tmp/rosa-test && cd /tmp/rosa-test
./rosa create ocm-role --prefix test-manual --no-console --mode manual
```
**Expected:** Warning + error, no empty files

**Result:** PASS
```
W: This OCM role cannot be used to provision clusters via console.redhat.com 
E: There was an error creating the ocm role: no-console OCM role is not yet enabled 

No empty policy files generated 
Consistent with auto mode 
```

---

#### Test 7: Manual Mode - Full Success (Local OCM with Policy)
**Command:**
```bash
mkdir /tmp/rosa-manual && cd /tmp/rosa-manual
./rosa create ocm-role --prefix test-manual-noconsole --no-console --mode manual
```
**Expected:** Policy files with content, correct commands

**Result:** PASS
```
W: This OCM role cannot be used to provision clusters via console.redhat.com 
I: All policy files saved to the current directory

Files Generated:
 sts_ocm_trust_policy.json (has content)
 sts_ocm_no_console_permission_policy.json (has content, NOT empty!)

Commands Include:
 Key=rosa_no_console_role,Value=true
 --policy-name test-manual-noconsole-OCM-Role-12541229-NoConsole-Policy
 file://sts_ocm_no_console_permission_policy.json
```

---

#### Test 8: Interactive Mode - No Console
**Command:**
```bash
./rosa create ocm-role --prefix test-interactive -i
# Prompts: admin=No, no-console=Yes, mode=auto
```
**Expected:** No-console prompt shown after admin=No

**Result:** PASS
```
? Enable admin capabilities for the OCM role: No 
? Create OCM role with minimal permissions (no console access): Yes 
W: This OCM role cannot be used to provision clusters via console.redhat.com 
I: Created role 'test-no-console-interactive-OCM-Role-12541229'

Interactive flow correct: asked admin first, then no-console 
Tags: rosa_no_console_role:true 
```

---

#### Test 9: Interactive Mode - Admin
**Command:**
```bash
./rosa create ocm-role --prefix test-admin-interactive -i
# Prompts: admin=Yes, mode=auto
```
**Expected:** No-console prompt skipped (mutual exclusivity)

**Result:** PASS
```
? Enable admin capabilities for the OCM role: Yes 
[NO no-console prompt - correctly skipped] 
I: Created role 'test-admin-interactive-OCM-Role-12541229'
I: Attached policy 'test-admin-interactive-OCM-Role-12541229-Policy'
I: Attached policy 'test-admin-interactive-OCM-Role-12541229-Admin-Policy'

Tags: rosa_admin_role:true (NO rosa_no_console_role) 
Policies: Standard + Admin (2 total) 
```

---

#### Test 10: Conflict Detection - Standard → No Console
**Command:**
```bash
./rosa create ocm-role --prefix test-conflict --mode auto
./rosa create ocm-role --prefix test-conflict --no-console --mode auto
```
**Expected:** Error preventing conversion

**Result:** PASS
```
W: Role 'test-conflict-OCM-Role-13829321' already exists
E: There was an error creating the ocm role: the existing role is a standard role. 
   To use no-console permissions please delete the role and recreate it 

Clear, actionable error message 
No AWS resources modified 
```

---

#### Test 11: Conflict Detection - Admin → No Console
**Command:**
```bash
./rosa create ocm-role --prefix test-conflict2 --admin --mode auto
./rosa create ocm-role --prefix test-conflict2 --no-console --mode auto
```
**Expected:** Error preventing conversion

**Result:** PASS
```
W: Role 'test-conflict2-OCM-Role-13829321' already exists
E: There was an error creating the ocm role: the existing role is an admin role. 
   To use no-console permissions please delete the role and recreate it 

Clear, actionable error message 
No AWS resources modified 
```

---

#### Test 12: Upgrade Path - Standard → Admin
**Command:**
```bash
./rosa create ocm-role --prefix test-conflict3 --mode auto
./rosa create ocm-role --prefix test-conflict3 --admin --mode auto
```
**Expected:** Idempotent upgrade allowed

**Result:** PASS
```
W: Role 'test-conflict3-OCM-Role-13829321' already exists
? Add admin policies to 'test-conflict3-OCM-Role-13829321' role? Yes
I: Attached policy 'test-conflict3-OCM-Role-13829321-Admin-Policy' 

Tags: rosa_admin_role:true (added) 
Policies: Standard + Admin (both present) 
Idempotent upgrade works correctly 
```

---

#### Test 13: Custom Path
**Command:**
```bash
./rosa create ocm-role --prefix test-path --no-console --path /rosa/ --mode auto
```
**Expected:** Custom path in both role and policy ARNs

**Result:** PASS
```
I: Created role 'test-path-OCM-Role-12541229' with ARN 
   'arn:aws:iam::471112697682:role/rosa/test-path-OCM-Role-12541229' 
I: Attached policy 
   'arn:aws:iam::471112697682:policy/rosa/test-path-OCM-Role-12541229-NoConsole-Policy' 

Custom path /rosa/ correctly applied to both role and policy 
```

---

### Summary

**Automated Tests:** 11/11 passing
```bash
cd cmd/create/ocmrole && ginkgo
SUCCESS! -- 11 Passed | 0 Failed
```

**Manual Tests:** 13/13 passed
- All core functionality working correctly
- Graceful degradation verified (production OCM without policy)
- Full end-to-end flow verified (local OCM with policy)
- Interactive mode, conflict detection, and edge cases validated

**Key Validations:**
- --no-console flag works and is mutually exclusive with --admin
- Warning message displays correctly
- Correct tags: `rosa_no_console_role:true`
- Correct policy: Only NoConsole-Policy (not standard)
- Policy content: Minimal permissions (iam:GetRole with tag conditions)
- Graceful error when policy unavailable
- No empty files in manual mode
- Conflict detection prevents invalid conversions
- Custom paths work correctly

## Breaking Changes
- [x] No breaking changes

N/A

- [x] `make test` passes.
- [x] `make lint` passes.
- [x] `make rosa` passes.

[OCM-23880]: https://redhat.atlassian.net/browse/OCM-23880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OCM-23336]: https://redhat.atlassian.net/browse/OCM-23336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OCM-25298]: https://redhat.atlassian.net/browse/OCM-25298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added a --no-console option to create OCM roles without console access (mutually exclusive with --admin)
  * OCM role profiles now include standard, admin, and no-console with guided interactive selection

* **Improvements**
  * Enhanced validation to prevent reusing roles with incompatible variants and ensure correct policy selection

* **Tests**
  * Added tests for profile constants, no-console behavior, and policy-file generation/validation

### DIFF
--- a/cmd/create/ocmrole/cmd.go
+++ b/cmd/create/ocmrole/cmd.go
@@ -19,6 +19,7 @@ package ocmrole
 import (
 	"fmt"
 	"os"
+	"slices"
 
 	common "github.com/openshift-online/ocm-common/pkg/aws/validations"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
@@ -43,6 +44,7 @@ var args struct {
 	admin               bool
 	path                string
 	managed             bool
+	noConsole           bool
 }
 
 var Cmd = &cobra.Command{
@@ -82,6 +84,13 @@ func init() {
 		"Enable admin capabilities for the role",
 	)
 
+	flags.BoolVar(
+		&args.noConsole,
+		"no-console",
+		false,
+		"Create OCM role with minimal permissions (cannot be used with console.redhat.com)",
+	)
+
 	flags.StringVar(
 		&args.path,
 		"path",
@@ -103,6 +112,8 @@ func init() {
 		false,
 		"Attach Classic ROSA AWS managed policies to the account roles. This is an alias for --managed-policies")
 	flags.MarkHidden("mp")
+
+	Cmd.MarkFlagsMutuallyExclusive("admin", "no-console")
 
 	interactive.AddModeFlag(Cmd)
 
@@ -169,19 +180,18 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
-	isAdmin := args.admin
+	profile := determineProfile(args.admin, args.noConsole)
 
-	if interactive.Enabled() && !isAdmin {
-		isAdmin, err = interactive.GetBool(interactive.Input{
-			Question: "Enable admin capabilities for the OCM role",
-			Help:     cmd.Flags().Lookup("admin").Usage,
-			Default:  isAdmin,
-			Required: false,
-		})
+	if interactive.Enabled() && profile == ProfileStandard {
+		profile, err = promptProfile(cmd, profile)
 		if err != nil {
 			r.Reporter.Errorf("Expected a valid --admin value: %s", err)
 			os.Exit(1)
 		}
+	}
+
+	if profile == ProfileNoConsole {
+		r.Reporter.Warnf("This OCM role cannot be used to provision clusters via console.redhat.com")
 	}
 
 	permissionsBoundary := args.permissionsBoundary
@@ -248,7 +258,6 @@ func run(cmd *cobra.Command, _ []string) {
 	roleNameRequested := aws.GetOCMRoleName(prefix, aws.OCMRole, externalID)
 
 	existsOnOCM, _, selectedARN, err := r.OCMClient.CheckRoleExists(orgID, roleNameRequested, r.Creator.AccountID)
-
 	if err != nil {
 		r.Reporter.Errorf("Error checking existing ocm-role: %v", err)
 		os.Exit(1)
@@ -266,11 +275,24 @@ func run(cmd *cobra.Command, _ []string) {
 		os.Exit(1)
 	}
 
+	// Validate no-console policy availability before any operations
+	if profile == ProfileNoConsole {
+		filename := fmt.Sprintf("sts_%s_permission_policy", aws.OCMNoConsoleRolePolicyFile)
+		policy, ok := policies[filename]
+		// For managed policies, validate ARN exists
+		// For customer-managed policies, validate Details exists (ARN is constructed later)
+		if !ok || (managedPolicies && policy.ARN() == "") || (!managedPolicies && policy.Details() == "") {
+			r.Reporter.Errorf("There was an error creating the ocm role: " +
+				"the no-console OCM role profile is not yet enabled for your Organization")
+			os.Exit(1)
+		}
+	}
+
 	switch mode {
 	case interactive.ModeAuto:
 		r.Reporter.Infof("Creating role using '%s'", r.Creator.ARN)
 		roleARN, err := createRoles(r, prefix, roleNameRequested, path, permissionsBoundary,
-			orgID, env, isAdmin, policies, managedPolicies)
+			orgID, env, profile, policies, managedPolicies)
 		if err != nil {
 			r.Reporter.Errorf("There was an error creating the ocm role: %s", err)
 			r.OCMClient.LogEvent("ROSACreateOCMRoleModeAuto", map[string]string{
@@ -286,11 +308,11 @@ func run(cmd *cobra.Command, _ []string) {
 		arguments.DisableRegionDeprecationWarning = false // enable region deprecation again
 	case interactive.ModeManual:
 		r.OCMClient.LogEvent("ROSACreateOCMRoleModeManual", map[string]string{})
-		_, _, err = checkRoleExists(r, roleNameRequested, isAdmin, interactive.ModeManual)
+		_, _, err = checkRoleExists(r, roleNameRequested, profile, interactive.ModeManual, path)
 		if err != nil {
 			r.Reporter.Warnf("Creating ocm role '%s' should fail: %s", roleNameRequested, err)
 		}
-		err = generateOcmRolePolicyFiles(r, env, orgID, isAdmin, policies)
+		err = generateOcmRolePolicyFiles(r, env, orgID, profile, policies)
 		if err != nil {
 			r.Reporter.Errorf("There was an error generating the policy files: %s", err)
 			r.OCMClient.LogEvent("ROSACreateOCMRoleModeManual", map[string]string{
@@ -310,7 +332,7 @@ func run(cmd *cobra.Command, _ []string) {
 			permissionsBoundary,
 			r.Creator,
 			env,
-			isAdmin,
+			profile,
 			managedPolicies,
 			confirm.Yes(),
 			policies,
@@ -327,11 +349,41 @@ func run(cmd *cobra.Command, _ []string) {
 	}
 }
 
+func promptProfile(cmd *cobra.Command, curr RoleProfile) (RoleProfile, error) {
+	isAdmin, err := interactive.GetBool(interactive.Input{
+		Question: "Enable admin capabilities for the OCM role",
+		Help:     cmd.Flags().Lookup("admin").Usage,
+		Default:  curr == ProfileAdmin,
+		Required: false,
+	})
+	if err != nil {
+		return ProfileStandard, fmt.Errorf("expected a valid --admin value: %s", err)
+	}
+	if isAdmin {
+		return ProfileAdmin, nil
+	}
+
+	isNoConsole, err := interactive.GetBool(interactive.Input{
+		Question: "Create OCM role with minimal permissions (no console access)",
+		Help:     cmd.Flags().Lookup("no-console").Usage,
+		Default:  curr == ProfileNoConsole,
+		Required: false,
+	})
+	if err != nil {
+		return ProfileStandard, fmt.Errorf("expected a valid --no-console value: %s", err)
+	}
+	if isNoConsole {
+		return ProfileNoConsole, nil
+	}
+
+	return ProfileStandard, nil
+}
+
 func buildCommands(prefix string, roleName string, rolePath string, permissionsBoundary string,
-	creator *aws.Creator, env string, isAdmin bool, managedPolicies bool, autoConfirmLink bool,
-	policies map[string]*cmv1.AWSSTSPolicy) (string, error) {
+	creator *aws.Creator, env string, profile RoleProfile, managedPolicies bool, autoConfirmLink bool,
+	policies map[string]*cmv1.AWSSTSPolicy,
+) (string, error) {
 	commands := []string{}
-	policyName := aws.GetPolicyName(roleName)
 	iamTags := map[string]string{
 		tags.RolePrefix:    prefix,
 		tags.RoleType:      aws.OCMRole,
@@ -346,6 +398,10 @@ func buildCommands(prefix string, roleName string, rolePath string, permissionsB
 		tags.AdminRole: tags.True,
 	}
 
+	noConsoleTags := map[string]string{
+		tags.NoConsoleRole: tags.True,
+	}
+
 	builder := awscb.NewIAMCommandBuilder().
 		SetCommand(awscb.CreateRole).
 		AddParam(awscb.RoleName, roleName).
@@ -353,9 +409,28 @@ func buildCommands(prefix string, roleName string, rolePath string, permissionsB
 		AddParam(awscb.PermissionsBoundary, permissionsBoundary).
 		AddTags(iamTags).
 		AddParam(awscb.Path, rolePath)
-	if isAdmin {
+
+	var policyFile string
+	var policyName string
+
+	switch profile {
+	case ProfileAdmin:
 		builder.AddTags(adminTags)
+
+		policyFile = aws.OCMRolePolicyFile
+		policyName = aws.GetPolicyName(roleName)
+	case ProfileNoConsole:
+		builder.AddTags(noConsoleTags)
+
+		policyFile = aws.OCMNoConsoleRolePolicyFile
+		policyName = aws.GetNoConsolePolicyName(roleName)
+	case ProfileStandard:
+		// No additional tags
+
+		policyFile = aws.OCMRolePolicyFile
+		policyName = aws.GetPolicyName(roleName)
 	}
+
 	createRole := builder.Build()
 
 	var createPolicy string
@@ -363,7 +438,7 @@ func buildCommands(prefix string, roleName string, rolePath string, permissionsB
 		createPolicy = awscb.NewIAMCommandBuilder().
 			SetCommand(awscb.CreatePolicy).
 			AddParam(awscb.PolicyName, policyName).
-			AddParam(awscb.PolicyDocument, fmt.Sprintf("file://sts_%s_permission_policy.json", aws.OCMRolePolicyFile)).
+			AddParam(awscb.PolicyDocument, fmt.Sprintf("file://sts_%s_permission_policy.json", policyFile)).
 			AddTags(iamTags).
 			AddParam(awscb.Path, rolePath).
 			Build()
@@ -371,13 +446,19 @@ func buildCommands(prefix string, roleName string, rolePath string, permissionsB
 
 	var policyARN string
 	var err error
+	policyKey := fmt.Sprintf("sts_%s_permission_policy", policyFile)
 	if managedPolicies {
-		policyARN, err = aws.GetManagedPolicyARN(policies, "sts_ocm_permission_policy")
+		policyARN, err = aws.GetManagedPolicyARN(policies, policyKey)
 		if err != nil {
 			return "", err
 		}
 	} else {
-		policyARN = aws.GetPolicyArnWithSuffix(creator.Partition, creator.AccountID, roleName, rolePath)
+		switch profile {
+		case ProfileNoConsole:
+			policyARN = aws.GetNoConsolePolicyARN(creator.Partition, creator.AccountID, roleName, rolePath)
+		case ProfileAdmin, ProfileStandard:
+			policyARN = aws.GetPolicyArnWithSuffix(creator.Partition, creator.AccountID, roleName, rolePath)
+		}
 	}
 	attachRolePolicy := awscb.NewIAMCommandBuilder().
 		SetCommand(awscb.AttachRolePolicy).
@@ -390,7 +471,8 @@ func buildCommands(prefix string, roleName string, rolePath string, permissionsB
 	} else {
 		commands = append(commands, createRole, createPolicy, attachRolePolicy)
 	}
-	if isAdmin {
+
+	if profile == ProfileAdmin {
 		policyName := aws.GetAdminPolicyName(roleName)
 
 		var createAdminPolicy string
@@ -438,18 +520,21 @@ func buildCommands(prefix string, roleName string, rolePath string, permissionsB
 }
 
 func createRoles(r *rosa.Runtime, prefix string, roleName string, rolePath string,
-	permissionsBoundary string, orgID string, env string, isAdmin bool,
-	policies map[string]*cmv1.AWSSTSPolicy, managedPolicies bool) (string, error) {
+	permissionsBoundary string, orgID string, env string, profile RoleProfile,
+	policies map[string]*cmv1.AWSSTSPolicy, managedPolicies bool,
+) (string, error) {
 	var policyARN string
 	var err error
 
-	if managedPolicies {
-		policyARN, err = aws.GetManagedPolicyARN(policies, "sts_ocm_permission_policy")
-		if err != nil {
-			return "", err
+	if profile != ProfileNoConsole {
+		if managedPolicies {
+			policyARN, err = aws.GetManagedPolicyARN(policies, fmt.Sprintf("sts_%s_permission_policy", aws.OCMRolePolicyFile))
+			if err != nil {
+				return "", err
+			}
+		} else {
+			policyARN = aws.GetPolicyArnWithSuffix(r.Creator.Partition, r.Creator.AccountID, roleName, rolePath)
 		}
-	} else {
-		policyARN = aws.GetPolicyArnWithSuffix(r.Creator.Partition, r.Creator.AccountID, roleName, rolePath)
 	}
 	if !confirm.Prompt(true, "Create the '%s' role?", roleName) {
 		os.Exit(0)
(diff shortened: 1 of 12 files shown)


amandahla:
@andclt Seems that is already being addressed by https://github.com/openshift/rosa/pull/3251
Can you confirm please?

andclt:
> @andclt Seems that is already being addressed by #3251 Can you confirm please?

Yes! So far this is just a draft PR to test the `no-console` role changes!

cmd/create/ocmrole/ocmrole.go
```diff
@@ -0,0 +1,55 @@
...
+
+type OCMRoleProfile string
+
+const (
+	ProfileStandard  OCMRoleProfile = "standard"
+	ProfileAdmin     OCMRoleProfile = "admin"
+	ProfileNoConsole OCMRoleProfile = "no-console"
+)
+
+func CreateOCMRole(
```
amandahla: Where this function is used?

amandahla: @coderabbitai shine

cmd/create/ocmrole/cmd.go
```diff
@@ -438,18 +487,21 @@ func buildCommands(prefix string, roleName string, rolePath string, permissionsB
...
 	var policyARN string
 	var err error
 
-	if managedPolicies {
-		policyARN, err = aws.GetManagedPolicyARN(policies, "sts_ocm_permission_policy")
-		if err != nil {
-			return "", err
+	if !isNoConsole {
+		if managedPolicies {
+			policyARN, err = aws.GetManagedPolicyARN(policies, "sts_ocm_permission_policy")
```
amandahla: minor: should we keep the pattern using the
filename = fmt.Sprintf("sts_%s_permission_policy", aws.OCMRolePolicyFile) ?

cmd/create/ocmrole/ocmrole.go
```diff
@@ -0,0 +1,55 @@
...
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+
+	"github.com/openshift/rosa/pkg/aws"
+	"github.com/openshift/rosa/pkg/rosa"
+)
+
+type OCMRoleProfile string
+
+const (
+	ProfileStandard  OCMRoleProfile = "standard"
```
amandahla: thats an interesting ideia. Instead of having so many if isNoConsole/isAdmin we could have a function like determineProfile and using it around for example in 
```
if interactive.Enabled() && profile == ProfileStandard
```
and from there, the big prompt could also be a function like `promptProfile` (line 185 at create/ocmrole/cmd.go)

@coderabbitai wdyt

amandahla:
@andclt I left some comments but I didnt review everything since one of them might change things here, please feel free to ping me if you want to discuss about it.

cmd/create/ocmrole/ocmrole.go
amandahla: extra note in case you want to follow @coderabbitai suggestions, this also improves readability by using, for example:
```
switch profile {
case ProfileAdmin:
    builder.SetKey(tags.AdminRole).SetValue(tags.True)
case ProfileNoConsole:
     builder.SetKey(tags.NoConsoleRole).SetValue(tags.True)
 case ProfileStandard:
    // No additional tags
}
```
instead of
```
if isAdmin {
     builder.SetKey(tags.AdminRole).SetValue(tags.True)
 }
 if isNoConsole {
     builder.SetKey(tags.NoConsoleRole).SetValue(tags.True)
 }
```
@coderabbitai  keep that in mind for further changes/reviews

cmd/create/ocmrole/ocmrole.go
amandahla: @coderabbitai shouldnt golangci-lint catch this?

cmd/create/ocmrole/ocmrole.go
andclt: Thanks for the suggestion @amandahla!

I was thinking about this refactoring too, it would definitely improve readability and testability. The main reason I didn't do it was to avoid making the PR even bigger and touching existing logic paths in the same changeset.

Happy to do it in a follow-up PR instead if that works better? wdyt @amandahla!

cmd/create/ocmrole/ocmrole.go
andclt: Good catch! `CreateOCMRole` is currently not called internally - it's a public wrapper for CAPA, following the same pattern as `CreateHCPRoles`.
`cmd.go` continues using `createRoles` directly since it already calculates the role name for conflict detection and has full context.
I'll add a comment clarifying this.

cmd/create/ocmrole/cmd.go
```diff
@@ -543,7 +628,23 @@ func generateOcmRolePolicyFiles(r *rosa.Runtime, env string, orgID string, isAdm
...
+	} else {
+		policyFile = aws.OCMRolePolicyFile
+	}
+
+	filename = fmt.Sprintf("sts_%s_permission_policy", policyFile)
+
+	// Check if no-console policy is available
+	if isNoConsole {
+		if _, ok := policies[filename]; !ok {
+			return fmt.Errorf("no-console OCM role is not yet enabled")
```
robpblake: @andclt This check is happening after the IAM role has been created in AWS, which leaves an orphaned IAM role with no policies attached. This makes it look like the OCM Role exists, but just hasn't been linked as it shows up when you run `rosa list ocm-role`.

We shouldn't attempt the role creation at all if the required permission policy is not available.

robpblake: @andclt Also wondering if we adjust the message a little to be more user friendly e.g.

```
The no-console OCM role profile is not yet enabled for your Organization
```

andclt: Good point! Thanks, fixed!

cmd/create/ocmrole/ocmrole.go
andclt: Applied the suggestion!

cmd/create/ocmrole/ocmrole.go
olucasfreitas: One more concern here: this wrapper still routes through `confirm.Prompt()` and `os.Exit(0)`, so a non-interactive caller can't use it safely yet. If this is meant for CAPA, it should stay non-interactive and return errors instead.

cmd/create/ocmrole/cmd.go
olucasfreitas: This is closer, but the preflight still only checks map membership. If the no-console STSPolicy comes back without ARN/details, auto mode can still create and tag the role before failing, so it would be safer to validate the exact data needed for the selected mode up front.

cmd/create/ocmrole/ocmrole.go
```diff
@@ -0,0 +1,70 @@
...
+	prefix string,
+	orgID string,
+	profile RoleProfile,
+	permissionsBoundary string,
+	rolePath string,
+	policies map[string]*cmv1.AWSSTSPolicy,
+	env string,
+	managedPolicies bool,
+) (string, error) {
+	externalID := orgID
```
olucasfreitas: This helper is using `orgID` as the role-name suffix, but the CLI path uses the organization external ID. That mismatch will produce a different OCM role name than OCM later expects for the same org/account.

cmd/create/ocmrole/ocmrole.go
andclt: Absolutely, thanks for pointing it out.
This needs more thinking since also `checkRoleExists` has a confirmation prompt. I will remove this method for now and bring it back in a separate PR.

amandahla:
Last comment: could you add some tests covering the checkRoleExists? Like when requesting No-Console profile, should succeed if existing role is No-Console and it should error if existing role is Admin or Standard.

cmd/create/ocmrole/cmd.go
andclt: As discussed offline, this applies to every profile and policies in the file, but for this PR we are tackling `no-console` only.

cmd/create/ocmrole/cmd.go
```diff
@@ -522,13 +613,39 @@ func createRoles(r *rosa.Runtime, prefix string, roleName string, rolePath strin
 		if err != nil {
 			return "", err
 		}
+
+	case ProfileNoConsole:
+		filename = fmt.Sprintf("sts_%s_permission_policy", aws.OCMNoConsoleRolePolicyFile)
+
+		// tag role with no-console tag
+		err = r.AWSClient.AddRoleTag(roleName, tags.NoConsoleRole, tags.True)
```
olucasfreitas: This tags the role as no-console before the policy is attached. If the create or attach fails after tagging, a retry will treat the role as complete, so can we tag only after the policy attach succeeds or verify the attached policy in `checkRoleExists`?

cmd/create/ocmrole/cmd.go
olucasfreitas: Scoping the upfront validation to no-console makes sense for this PR because that’s the new rollout-gated path and the Jira acceptance criteria call it out explicitly. I’d still capture the broader standard/admin validation as follow-up, since the same late-failure pattern exists outside no-console too.

cmd/create/ocmrole/cmd.go
andclt: That's a very good point, I think tagging the role only after the policy is attached is the most straightforward fix here.

cmd/create/ocmrole/cmd.go
```diff
@@ -480,55 +564,89 @@ func createRoles(r *rosa.Runtime, prefix string, roleName string, rolePath strin
...
-		iamTags[tags.AdminRole] = tags.True
+		iamTags[tags.NoConsoleRole] = tags.True
 		policyDetail = aws.GetPolicyDetails(policies, filename)
 		err = createPermissionPolicy(r, policyARN, iamTags, roleName, rolePath, policyDetail, managedPolicies)
 		if err != nil {
 			return "", err
 		}
+
+		// tag role with no-console tag
+		err = r.AWSClient.AddRoleTag(roleName, tags.NoConsoleRole, tags.True)
```
olucasfreitas, line 638: Moving the tag after attach fixes the old false-positive case, but retries still break if `AddRoleTag()` is the step that fails. On the next run the role looks standard, so can we verify the attached no-console policy in `checkRoleExists()` or heal the missing tag there?

cmd/create/ocmrole/cmd.go
```diff
@@ -346,38 +397,67 @@ func buildCommands(prefix string, roleName string, rolePath string, permissionsB
...
+	var policyName string
+
+	switch profile {
+	case ProfileAdmin:
 		builder.AddTags(adminTags)
+
+		policyFile = aws.OCMRolePolicyFile
+		policyName = aws.GetPolicyName(roleName)
+	case ProfileNoConsole:
+		builder.AddTags(noConsoleTags)
```
olucasfreitas: This still adds `rosa_no_console_role=true` on the initial `create-role` command. If `create-policy` or `attach-role-policy` fails, a later CLI retry can treat that partial role as valid, so can we move the tag to a step after the attach succeeds?

cmd/create/ocmrole/cmd.go
andclt: Healing the missing tag there is probably the better option IMO - I will fix it also for admin

cmd/create/ocmrole/cmd.go
andclt: Is your concern about a user executing manual mode first and then retrying with auto mode? Are you suggesting we move the tag to an additional command in manual mode? Because if so, there's also the scenario where a user runs manual mode first, then switches to auto mode. Should we instead have self-healing logic in the `checkRoleExists` method to handle this case?

robpblake: @olucasfreitas Can you explain the problem you're trying to solve?

olucasfreitas: The concern is a partial manual run: `buildCommands()` embeds `rosa_no_console_role=true` on the `create-role` command, so if a user only runs that first command and skips `create-policy`/`attach-role-policy`, the role has the tag but no policy. On retry, `checkRoleExists()` sees `IsNoConsoleRole==true` and treats the role as complete. The current self-healing covers the opposite direction (policy attached, tag missing). For manual mode specifically this is harder to fix since you can't control execution order — options are: (a) emit `tag-role` as a separate command after attach, or (b) extend `checkRoleExists` to also verify the attached policy when the tag is present. The admin path has the same pattern pre-existing, so scoping it to no-console here seems reasonable.

cmd/create/ocmrole/cmd.go
olucasfreitas: This is addressed in the current head — the role is tagged after `createPermissionPolicy` succeeds (line 638), and `checkRoleExists` self-heals both admin (line 768) and no-console (line 800) when the policy is attached but tag is missing.

cmd/create/ocmrole/ocmrole.go
olucasfreitas: Confirmed removed in the latest push — this is no longer present.

cmd/create/ocmrole/ocmrole.go
olucasfreitas: Moot since `CreateOCMRole` was removed.

cmd/create/ocmrole/ocmrole.go
olucasfreitas: Applied — the current code uses `determineProfile`, `promptProfile`, and `switch profile` throughout.

cmd/create/ocmrole/cmd.go
olucasfreitas: The preflight at lines 278-288 now validates both map presence and content (ARN for managed, Details for unmanaged) before any AWS calls. The broader standard/admin validation is correctly deferred to a follow-up as discussed.

cmd/create/ocmrole/cmd.go
olucasfreitas: Fixed — `AddRoleTag` is now called after `createPermissionPolicy` succeeds (line 638), and `checkRoleExists` self-heals the opposite direction.

olucasfreitas:
@amandahla Tests added — covers standard/admin/no-console profile requests against existing roles, plus self-healing paths for both admin and no-console.

cmd/create/ocmrole/cmd.go
andclt: I think option (a) is the most straightforward fix here considering the `no-console` option hasn't been released yet so adding an additional step won't be a breaking change - proceeding with this approach then

robpblake: @andclt @olucasfreitas What is the scenario today when the customer is using `--admin` ? Why does `--no-console` need to be different? I'd prefer not to have the customer have to run additional manual commands as that is just one more step to miss or get wrong

`--admin` adds a tag to the role, `--no-console` is doing the same thing, just a different tag.  Why do we need different approaches?

cmd/create/ocmrole/cmd.go
```diff
@@ -346,38 +398,67 @@ func buildCommands(prefix string, roleName string, rolePath string, permissionsB
...
+	var policyName string
+
+	switch profile {
+	case ProfileAdmin:
 		builder.AddTags(adminTags)
+
+		policyFile = aws.OCMRolePolicyFile
+		policyName = aws.GetPolicyName(roleName)
+	case ProfileNoConsole:
+		// No-console tag will be added via separate tag-role command after policy attachment
```
robpblake: Why?

cmd/create/ocmrole/cmd.go
andclt: I've removed the additional step for the `no-console` role tag and went with option (b) instead, meaning that `checkRoleExists` verifies that the correct policy is attached when the tag is present.

cmd/create/ocmrole/cmd.go
andclt: I've now changed it to keep it consistent with admin.
More info here: https://github.com/openshift/rosa/pull/3252#discussion_r3357747647

PR #3262, opened 2026-06-08
ROSAENG-59417 | refactor: expose CreateOCMRole to CAPA

andclt:
## PR Summary

 Refactored OCM role creation logic to enable programmatic consumption by CAPA. Extracted business logic into `CheckOCMRoleExists()` and `CreateOCMRole()` functions that operate without interactive
  prompts or os.Exit calls.

CAPA requires programmatic OCM role management via Kubernetes CRDs. The existing ROSA CLI implementation was unsuitable because:
  1. Business logic embedded in CLI commands with interactive prompts
  2. Used `os.Exit()` for error handling
  3. No exported API for external consumers
  4. CLI concerns mixed with IAM operations

## Related Issues and PRs

- Jira: [ROSAENG-59417](https://redhat.atlassian.net/browse/ROSAENG-59417)
- Fixes: `N/A`
- Related PR(s): https://github.com/openshift/rosa/pull/3252
- Related design/docs: [design doc](https://github.com/openshift-online/rosa-enhancements/blob/main/enhancements/no-console-ocm-role.md)

## Type of Change

## Previous Behavior

No user-facing changes.

OCM role creation was tightly coupled to CLI with user interaction.

## Behavior After This Change

- Exported `CreateOCMRole()`
- Exported `CheckOCMRoleExists()`
- CLI behavior unchanged

### Test Steps
1. All existing tests pass
2. Manual testing: `rosa create ocm-role` works identically - executed all manual steps listed in this PR: https://github.com/openshift/rosa/pull/3252

## Proof of the Fix

- Screenshots:
- Videos:
- Logs/CLI output:
- Other artifacts:

## Breaking Changes
- [x] No breaking changes

- [x] `make test` passes.
- [x] `make lint` passes.
- [x] `make rosa` passes.

## Summary by CodeRabbit

* **New Features**
  * Explicit OCM role profiles: standard, admin, and no-console for creation and upgrades.
  * Interactive prompts preserved for role creation and when adding admin policies to existing roles.

* **Bug Fixes**
  * Improved non-interactive role validation and safer self-healing of tags/policies.
  * Clearer handling and user-facing errors when required policies are missing.

* **Tests**
  * Added/updated tests validating profile values and selection logic.

* **Refactor**
  * Role creation and validation logic reorganized into internal helpers for robustness.

[ROSAENG-59417]: https://redhat.atlassian.net/browse/ROSAENG-59417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/create/ocmrole/cmd.go
```diff
@@ -718,114 +612,18 @@ func createPermissionPolicy(r *rosa.Runtime, policyARN string,
...
+		if mode == interactive.ModeAuto {
+			if !confirm.Prompt(true, "Add admin policies to '%s' role?", roleName) {
 				return roleARN, true, nil
 			}
 
-			// Existing is standard, cannot convert
-			return "", true, fmt.Errorf("the existing role is a standard role." +
-				" To use no-console permissions please delete the role and recreate it")
+			// User accepted - return false to trigger upgrade/creation
+			return "", false, nil
```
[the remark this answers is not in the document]
andclt: This is a pre-existing issue, not introduced by the refactoring. Fixing it would be a behavior change, not a refactoring - out of scope for this PR.

internal/ocmrole/core.go
```diff
@@ -0,0 +1,360 @@
...
+			if isExistingRoleAdmin {
+				return "", true, fmt.Errorf("the existing role is an admin role."+
+					" To remove admin capabilities please delete the admin policy and the '%s' tag",
+					tags.AdminRole)
+			}
+			if isExistingRoleNoConsole {
+				return "", true, fmt.Errorf("the existing role is a no-console role." +
+					" To use standard permissions please delete the role and recreate it")
+			}
+			return roleARN, true, nil
```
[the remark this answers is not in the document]
andclt: fixed in the latest commit, adding `ErrRoleExistsWrongProfile` though was not correct

cmd/create/ocmrole/creators.go
```diff
@@ -0,0 +1,363 @@
...
+	roleARN, exists, err := checkRoleExistsInternal(r, roleName, profile, interactive.ModeAuto, rolePath)
+	return exists, roleARN, err
+}
+
+// CreateOCMRole creates an OCM role in AWS with the specified configuration.
+// This function does NOT check if the role exists (caller should use CheckOCMRoleExists first).
+// This function does NOT link the role to OCM organization (caller should use OCMClient.LinkOrgToRole after).
+//
+// Returns the ARN of the created role.
+func CreateOCMRole(
```
olucasfreitas: These helpers are now part of an importable CAPA contract, but they still live in a `cmd/...` package that primarily exists for Cobra wiring. Moving the reusable role logic into a library package would give external consumers a cleaner, more stable boundary.

cmd/create/ocmrole/creators.go
```diff
@@ -0,0 +1,363 @@
...
+			}
+
+			// Role appears to be standard - check if admin policy is actually attached (self-healing)
+			attachedPolicies, err := r.AWSClient.ListAttachedRolePolicies(roleName)
+			if err != nil {
+				return "", true, err
+			}
+
+			// Check if admin policy is attached (exact ARN match)
+			expectedAdminPolicyARN := aws.GetAdminPolicyARN(r.Creator.Partition, r.Creator.AccountID, roleName, "")
```
olucasfreitas: The admin self-heal path is still looking for the root-path policy ARN instead of the path-scoped ARN. Since this API now accepts `rolePath`, it should use `rolePath` the same way the no-console branch does or custom-path admin roles will fail rechecks and retries.

cmd/create/ocmrole/creators.go
```diff
@@ -0,0 +1,363 @@
...
+// This function includes self-healing for policy/tag mismatches (e.g., policy exists but tag missing).
+//
+// This is the CAPA-facing function that wraps checkRoleExistsInternal with no prompts.
+//
+// Returns:
+//   - exists: true if the role exists
+//   - roleARN: ARN of the role (only valid if exists=true)
+//   - error: Any error encountered, including ErrRoleExistsWrongProfile if role exists
+//     but with an incompatible profile
+func CheckOCMRoleExists(r *rosa.Runtime, roleName string, profile RoleProfile, rolePath string) (bool, string, error) {
```
olucasfreitas: This CAPA-facing boundary still accepts states that turn into a panic or a false negative: `r.Creator` is dereferenced below without validation, and unknown `RoleProfile` values fall through to `exists=false, err=nil`. Please validate both here and return a deterministic error before external callers rely on this API.

internal/ocmrole/core.go
```diff
@@ -0,0 +1,360 @@
...
+			return "", err
+		}
+
+		// tag role with no-console tag
+		err = r.AWSClient.AddRoleTag(roleName, tags.NoConsoleRole, tags.True)
+		if err != nil {
+			return "", err
+		}
+	}
+
```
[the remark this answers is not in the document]
olucasfreitas: The `CreateOCMRole` side is hardened now, but `CheckOCMRoleExists` still accepts arbitrary `RoleProfile` strings and can fall through to `exists=false` for a real role. I think we should finish this with a shared validator plus a default case before CAPA depends on the exported API.

cmd/create/ocmrole/cmd.go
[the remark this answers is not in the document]
olucasfreitas: This was tolerable as CLI glue, but it is now part of the CAPA-facing contract. I’d prefer an explicit already-exists or upgrade-needed result here instead of relying on callers to sequence a separate preflight correctly.

cmd/create/ocmrole/creators.go
andclt: I wasn't sure why the empty string was being passed there in the first place - it was already like that before no-console profile addition. I kept it that way because I didn't know if there were historical reasons for
it and didn't want to risk breaking existing logic, especially since we were just adding the no-console role option. But yeah, it did seem wrong, so I'm going to fix it now.

cmd/create/ocmrole/creators.go
andclt: good catch, thanks!

cmd/create/ocmrole/cmd.go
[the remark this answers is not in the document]
andclt: @olucasfreitas, @coderabbitai I want to clarify which function this refers to.

The `checkRoleExists` wrapper (CLI-only) does overload `exists` for upgrade decisions:
    - User declines upgrade → (roleARN, true, nil)
    - User accepts upgrade → ("", false, nil)

However, `CheckOCMRoleExists` (CAPA-facing) always uses `ModeAuto` and never prompts, so it returns (true, roleARN, ErrRoleExistsWrongProfile) for profile mismatches - `exists` is always truthful.

Is the concern about `CheckOCMRoleExists` specifically? If so, what should the expected behavior be for the upgrade-needed case?

cmd/create/ocmrole/creators.go
andclt: I was following the `CreateHCPRoles` pattern but having another package seems like a better option

cmd/create/ocmrole/cmd.go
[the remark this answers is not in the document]
andclt: @coderabbitai

`ErrRoleExistsWrongProfile` is not exposed in the public API because CAPA will not use `errors.Is()` - we will just propagate the error message. The error message itself is descriptive enough: "the existing role is an admin role. To use standard permissions please delete the role and recreate it."

When CAPA gets a profile mismatch, reconciliation fails with the descriptive error. User remediation is to delete the role and let CAPA recreate it, or update the CRD to match the existing profile.

I've updated `CheckOCMRoleExists` docs to clarify that exists is always truthful (never overloaded) and that profile upgrade/downgrade is the caller's responsibility.

pkg/ocmrole/api.go
```diff
@@ -0,0 +1,132 @@
...
+//   - Delete the existing role and recreate it with the desired profile, or
+//   - Accept the existing role's profile
+//
+// This function does NOT perform automatic profile upgrades or downgrades.
+//
+// Returns:
+//   - exists: true if the role exists (always truthful - never overloaded for upgrade decisions)
+//   - roleARN: ARN of the role (valid when exists=true, empty otherwise)
+//   - error: nil if role exists with correct profile, or an error describing the issue
+func CheckOCMRoleExists(r *rosa.Runtime, roleName string, profile RoleProfile, rolePath string) (bool, string, error) {
```
olucasfreitas: This adds a new supported `pkg/ocmrole` surface, but there are still no direct tests for the exported functions. Please add API-level coverage for the validation/error contract and the custom-path admin case so CAPA isn’t relying on command-wrapper tests by accident.

internal/ocmrole/core.go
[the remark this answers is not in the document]
olucasfreitas: The public API validation is better now, but `CreateRolesInternal` still has no defensive default. Since this helper is exported inside `internal/`, I’d still return an explicit error there instead of succeeding after `EnsureRole` with no permission policy attached.

pkg/ocmrole/api.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// Get current OCM organization
+	orgID, externalID, err := r.OCMClient.GetCurrentOrganization()
+	if err != nil {
+		return "", fmt.Errorf("failed to get organization account: %w", err)
+	}
+
+	roleName := aws.GetOCMRoleName(prefix, aws.OCMRole, externalID)
+
+	return internalocmrole.CreateRolesInternal(r, prefix, roleName, path, permissionsBoundary,
```
olucasfreitas: The CLI preflights missing no-console policy before touching AWS, but this public API goes straight into `CreateRolesInternal`, which calls `EnsureRole` first. Please move the no-console availability check into the shared API path too, otherwise CAPA can leave behind a partially created role when the feature isn’t enabled yet.

cmd/create/ocmrole/cmd.go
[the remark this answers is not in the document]
olucasfreitas: If CAPA only needs create-if-absent, the new doc comment is probably enough. If CAPA is expected to reconcile a standard role into admin, we still need an explicit upgrade API instead of asking callers to chain a create-only helper after a mismatch check.

pkg/ocmrole/api.go
andclt: Good point, however I will duplicate the logic and not simply move it because the CLI needs to fail fast even before prompting the user.

pkg/ocmrole/api.go
andclt: Added test for `pkg/ocmrole` and `internal/ocmrole` in the latest commit

pkg/ocmrole/api.go
```diff
@@ -0,0 +1,144 @@
...
+		filename := fmt.Sprintf("sts_%s_permission_policy", aws.OCMNoConsoleRolePolicyFile)
+		policy, ok := policies[filename]
+		// For managed policies, validate ARN exists
+		// For customer-managed policies, validate Details exists (ARN is constructed later)
+		if !ok || (managedPolicies && policy.ARN() == "") || (!managedPolicies && policy.Details() == "") {
+			return "", fmt.Errorf("the no-console OCM role profile is not yet enabled for your Organization")
+		}
+	}
+
+	return internalocmrole.CreateRolesInternal(r, prefix, roleName, path, permissionsBoundary,
```
olucasfreitas: The doc comment says `CreateOCMRole` is create-only, but this still flows into `CreateRolesInternal`, and `EnsureRole` is upsert-style. A stale or retried CAPA reconcile can therefore mutate an already-existing role instead of failing fast, which breaks the current create-only contract.

pkg/ocmrole/api.go
olucasfreitas: Duplicating the no-console preflight in the public API is the right trade-off here. It preserves the CLI’s fail-fast UX while also preventing CAPA from creating a partial role when the feature isn’t enabled.

pkg/ocmrole/api.go
olucasfreitas: The latest commit adds direct `pkg/ocmrole` and `internal/ocmrole` coverage, so the original “no exported API tests” concern looks addressed. I’d still like one behavior-level case for an existing role or missing no-console policy, because that’s the part CAPA will actually depend on.

pkg/ocmrole/api.go
andclt: Thanks for catching the ambiguity in the doc.  Your comment prompted me to simplify the API: merged `CheckOCMRoleExists` + `CreateOCMRole` into `GetOrCreateOCMRole(...)  (roleARN, created, error)`. 

For context, the `EnsureRole` behavior would have been fine for CAPA's controller pattern (we'll call `CheckOCMRoleExists` first, so no silent mutations), but the merged approach is cleaner. When the role
  exists, it returns immediately without calling `EnsureRole`. The `created` bool tells CAPA what happened.

pkg/ocmrole/api.go
andclt: Added 2 smoke tests in the latest commit

internal/ocmrole/core.go
```diff
@@ -0,0 +1,326 @@
...
+		if err != nil {
+			return "", err
+		}
+
+		// tag role with no-console tag
+		err = r.AWSClient.AddRoleTag(roleName, tags.NoConsoleRole, tags.True)
+		if err != nil {
+			return "", err
+		}
+	}
```
olucasfreitas, line 312: `CreateRolesInternal` still has no `default` case in this switch. If an invalid profile reaches here, `EnsureRole` will have already created the IAM role (line 229) but no policies get attached — leaving an orphaned role in AWS. Please add a defensive default like `CheckRoleExistsInternal` already has at line 174.

andclt: Fixed. I've added validation before `EnsureRole` (line 222) to prevent leaving an orphaned role, plus the default case you mentioned. Early validation prevents the issue, the default case change simply matches the existing pattern in `CheckRoleExistsInternal`.

PR #3277, opened 2026-06-23
ROSAENG-60112 | test: add focused coverage for untested pkg packages

olucasfreitas:
## PR Summary
Add focused, meaningful unit tests for previously untested package-level behavior in the early-priority Phase 1C scope. This aligns the change with the plan's first implementation step instead of the earlier pkg/aws Phase 1A direction.

The coverage plan prioritizes untested `pkg/` packages first because they are fast, reviewable wins that raise confidence without over-investing in low-value command wiring. Several package-level utilities and integration helpers had no automated coverage at all, including FedRAMP config behavior, input file parsing, logging helpers, color/profile/region/debug flag behavior, and the small HTTP client wrapper.

This PR adds focused tests for the real behavior in those packages while intentionally skipping constants-only packages and avoiding coverage-for-coverage-sake assertions.

## Related Issues and PRs
- Jira: [ROSAENG-60112](https://issues.redhat.com/browse/ROSAENG-60112)
- Fixes: `#`
- Related PR(s): supersedes prior draft work that was started under the Phase 1A direction
- Related design/docs: Coverage planning work for [ROSAENG-60112](https://redhat.atlassian.net/browse/ROSAENG-60112)

## Type of Change
- [x] test - adds or updates tests only.

## Previous Behavior
The following packages had no dedicated automated coverage for their package-level behavior:
- `pkg/fedramp`
- `pkg/input`
- `pkg/logging`
- `pkg/color`
- `pkg/aws/profile`
- `pkg/aws/region`
- `pkg/debug`
- `pkg/clients`

## Behavior After This Change
The repo now has focused tests for:
- FedRAMP flag wiring, environment validation, and config-backed enable/disable behavior
- YAML/JSON input file loading plus the Hosted Control Planes happy path guard
- Logger builders and HTTP round-tripper request/response logging with JSON/form redaction
- Color flag defaults, completion options, and explicit color modes
- AWS profile and region flag/environment precedence, including escaped-empty handling for region
- Debug flag state behavior
- HTTP client wrapper request behavior

Intentionally not added:
- constants-only packages such as `pkg/info`, `pkg/object`, `pkg/properties`, and `pkg/constants`
- brittle tests that would only duplicate stdlib or SDK behavior without validating ROSA-specific logic

### Preconditions
- Go toolchain installed
- Repository cloned with hooks installed

### Test Steps
1. Run `go test ./pkg/fedramp ./pkg/input ./pkg/logging ./pkg/color ./pkg/aws/profile ./pkg/aws/region ./pkg/debug ./pkg/clients -count=1`
2. Run `go test ./pkg/fedramp ./pkg/input ./pkg/logging ./pkg/color ./pkg/aws/profile ./pkg/aws/region ./pkg/debug ./pkg/clients -cover`
3. Run `go vet ./pkg/fedramp ./pkg/input ./pkg/logging ./pkg/color ./pkg/aws/profile ./pkg/aws/region ./pkg/debug ./pkg/clients`

### Expected Results
- All targeted package tests pass
- Coverage is reported for each targeted package
- `go vet` finishes without findings

## Proof of the Fix
- Logs/CLI output:
  - `go test` passes for all targeted packages
  - `go vet` is clean
  - pre-push checks passed (format, build, lint, changed-files coverage, unit/integration tests)
- Other artifacts:
  - 8 new test files
  - 757 lines of test code added

## Breaking Changes
- [x] No breaking changes

N/A

- [x] `make test` passes.
- [x] `make lint` passes.
- [ ] `make rosa` passes.

[ROSAENG-60112]: https://redhat.atlassian.net/browse/ROSAENG-60112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ROSAENG-60112]: https://redhat.atlassian.net/browse/ROSAENG-60112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Tests**
  * Added Ginkgo/Gomega suites covering AWS `profile` and `region` flag + environment variable precedence and special empty-value handling.
  * Added HTTP client tests for successful requests and error propagation.
  * Added `color` and `debug` flag tests for default behavior and state changes.
  * Added FedRAMP config tests, including enable/disable behavior, region validation, and environment validation.
  * Added input unmarshalling tests (YAML/JSON) and Hypershift safety checks.
  * Added logging tests verifying log level behavior, body preservation, and sensitive/redacted field handling (including headers and form fields).

### DIFF
--- a/pkg/aws/region/flag_test.go
+++ b/pkg/aws/region/flag_test.go
@@ -55,6 +55,10 @@ var _ = Describe("Region", func() {
 		Expect(Region()).To(Equal("eu-west-1"))
 	})
 
+	It("returns an empty string when neither flag nor environment is set", func() {
+		Expect(Region()).To(Equal(""))
+	})
+
 	It("treats an escaped empty flag value as unset", func() {
 		region = "\"\""
 		Expect(os.Setenv("AWS_REGION", "sa-east-1")).To(Succeed())
--- a/pkg/aws/region/flag_test.go
+++ b/pkg/aws/region/flag_test.go
@@ -1,0 +1,70 @@
+package region
+
+import (
+	"os"
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/spf13/pflag"
+)
+
+func TestRegion(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Region Suite")
+}
+
+var _ = Describe("Region", func() {
+	var (
+		previousRegion    string
+		previousAwsRegion string
+	)
+
+	BeforeEach(func() {
+		previousRegion = region
+		previousAwsRegion = os.Getenv("AWS_REGION")
+		region = ""
+		Expect(os.Setenv("AWS_REGION", "")).To(Succeed())
+	})
+
+	AfterEach(func() {
+		region = previousRegion
+		Expect(os.Setenv("AWS_REGION", previousAwsRegion)).To(Succeed())
+	})
+
+	It("registers the region flag with an empty default", func() {
+		flags := pflag.NewFlagSet("test", pflag.ContinueOnError)
+
+		AddFlag(flags)
+
+		flag := flags.Lookup("region")
+		Expect(flag).NotTo(BeNil())
+		Expect(flag.DefValue).To(Equal(""))
+	})
+
+	It("prefers the flag value over the environment variable", func() {
+		region = "us-east-1"
+		Expect(os.Setenv("AWS_REGION", "eu-west-1")).To(Succeed())
+
+		Expect(Region()).To(Equal("us-east-1"))
+	})
+
+	It("falls back to the AWS_REGION environment variable", func() {
+		Expect(os.Setenv("AWS_REGION", "eu-west-1")).To(Succeed())
+
+		Expect(Region()).To(Equal("eu-west-1"))
+	})
+
+	It("treats an escaped empty flag value as unset", func() {
+		region = "\"\""
+		Expect(os.Setenv("AWS_REGION", "sa-east-1")).To(Succeed())
+
+		Expect(Region()).To(Equal("sa-east-1"))
+	})
+
+	It("treats an escaped empty environment value as empty", func() {
+		Expect(os.Setenv("AWS_REGION", "\"\"")).To(Succeed())
+
+		Expect(Region()).To(Equal(""))
+	})
+})
--- a/pkg/color/flag_test.go
+++ b/pkg/color/flag_test.go
@@ -1,7 +1,6 @@
 package color
 
 import (
-	"os"
 	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -58,18 +57,6 @@ var _ = Describe("Color", func() {
 		expected := UseColor()
 
 		SetColor("unexpected")
-
-		Expect(UseColor()).To(Equal(expected))
-	})
-
-	It("uses the current stdout mode for auto", func() {
-		SetColor("auto")
-
-		stdout, err := os.Stdout.Stat()
-		expected := true
-		if err == nil {
-			expected = (stdout.Mode()&os.ModeDevice != 0) && (stdout.Mode()&os.ModeNamedPipe == 0)
-		}
 
 		Expect(UseColor()).To(Equal(expected))
 	})
--- a/pkg/color/flag_test.go
+++ b/pkg/color/flag_test.go
@@ -1,0 +1,76 @@
+package color
+
+import (
+	"os"
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/spf13/cobra"
+)
+
+func TestColor(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Color Suite")
+}
+
+var _ = Describe("Color", func() {
+	var previousColor string
+
+	BeforeEach(func() {
+		previousColor = color
+		color = ""
+	})
+
+	AfterEach(func() {
+		color = previousColor
+	})
+
+	It("registers the color flag with auto as default", func() {
+		cmd := &cobra.Command{Use: "test"}
+
+		AddFlag(cmd)
+
+		flag := cmd.PersistentFlags().Lookup("color")
+		Expect(flag).NotTo(BeNil())
+		Expect(flag.DefValue).To(Equal("auto"))
+	})
+
+	It("returns the supported completion options", func() {
+		values, directive := completion(&cobra.Command{Use: "test"}, nil, "")
+
+		Expect(values).To(Equal([]string{"auto", "never", "always"}))
+		Expect(directive).To(Equal(cobra.ShellCompDirectiveDefault))
+	})
+
+	It("disables color when set to never", func() {
+		SetColor("never")
+		Expect(UseColor()).To(BeFalse())
+	})
+
+	It("enables color when set to always", func() {
+		SetColor("always")
+		Expect(UseColor()).To(BeTrue())
+	})
+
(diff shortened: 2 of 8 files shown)


pkg/input/input_test.go
```diff
@@ -0,0 +1,93 @@
...
+			result, err := UnmarshalInputFile(path)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result).To(BeNil())
+		})
+	})
+
+	Describe("CheckIfHypershiftClusterOrExit", func() {
+		It("returns without exiting for a hypershift cluster", func() {
+			cluster, err := cmv1.NewCluster().
+				Hypershift(cmv1.NewHypershift().Enabled(true)).
```
amandahla, line 84: What is the expected behavior if is not enabled?

pkg/fedramp/fedramp_test.go
```diff
@@ -0,0 +1,144 @@
...
+		It("rejects non-GovCloud regions", func() {
+			Expect(IsGovRegion("us-east-1")).To(BeFalse())
+			Expect(IsGovRegion("")).To(BeFalse())
+		})
+	})
+
+	Describe("IsValidEnv", func() {
+		It("recognizes known environments", func() {
+			Expect(IsValidEnv("production")).To(BeTrue())
+			Expect(IsValidEnv("staging")).To(BeTrue())
```
amandahla, line 194: Is staging01 a valid env too?

pkg/aws/region/flag_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+		Expect(Region()).To(Equal("sa-east-1"))
+	})
+
+	It("treats an escaped empty environment value as empty", func() {
+		Expect(os.Setenv("AWS_REGION", "\"\"")).To(Succeed())
+
+		Expect(Region()).To(Equal(""))
+	})
+})
```
amandahla, line 74: Would make sense having one "returns an empty string when neither flag nor environment is set" like the profile has?

pkg/logging/logging_test.go
```diff
@@ -0,0 +1,222 @@
...
+			Expect(err).NotTo(HaveOccurred())
+
+			builder.Redact("other")
+			Expect(result.redact).To(HaveKeyWithValue("token", true))
+			Expect(result.redact).NotTo(HaveKey("other"))
+		})
+	})
+
+	Describe("RoundTripper", func() {
+		It("preserves request and response bodies and redacts JSON fields in logs", func() {
```
amandahla: If this test fails how do we know the cause? bodies not preserved or json fields not redacted?

pkg/fedramp/fedramp_test.go
```diff
@@ -0,0 +1,144 @@
...
+			tempDir := GinkgoT().TempDir()
+			Expect(os.Setenv("OCM_CONFIG", filepath.Join(tempDir, "missing.json"))).To(Succeed())
+
+			Expect(Enabled()).To(BeFalse())
+		})
+	})
+
+	Describe("Disable", func() {
+		It("clears the in-memory flag and persists FedRAMP=false for a valid config", func() {
+			tempDir := GinkgoT().TempDir()
```
amandahla, line 135: minor: What is the difference between this and the os.MkdirTemp?

pkg/color/flag_test.go
```diff
@@ -0,0 +1,76 @@
...
+		Expect(UseColor()).To(Equal(expected))
+	})
+
+	It("uses the current stdout mode for auto", func() {
+		SetColor("auto")
+
+		stdout, err := os.Stdout.Stat()
+		expected := true
+		if err == nil {
+			expected = (stdout.Mode()&os.ModeDevice != 0) && (stdout.Mode()&os.ModeNamedPipe == 0)
```
amandahla: The expected value is computed with the same stdout logic as UseColor(), right?
If this test failed, what would tell? 
I would suggest dropping it and relying on never/always + the unknown→default case (and reporter tests) for color coverage. wdyt?

pkg/fedramp/fedramp_test.go
```diff
@@ -0,0 +1,144 @@
...
+			cmd := &cobra.Command{Use: "test"}
+
+			AddFlag(cmd.Flags())
+
+			Expect(cmd.Flags().Lookup("govcloud")).NotTo(BeNil())
+			Expect(cmd.Flags().Lookup("admin")).NotTo(BeNil())
+			Expect(HasFlag(cmd)).To(BeFalse())
+			Expect(HasAdminFlag(cmd)).To(BeFalse())
+
+			Expect(cmd.Flags().Set("govcloud", "true")).To(Succeed())
```
amandahla, line 49: 1) With this test I noticed that we have 2 flags (--admin and --govcloud) that enables fedramp. Is this expected? (out of scope of this PR but nice to review since we are here ;-) )

2) If this is expected then should we have a test to validate behavior if just one of them are enabled?

pkg/fedramp/fedramp_test.go
```diff
@@ -0,0 +1,144 @@
...
+			tempDir := GinkgoT().TempDir()
+			Expect(os.Setenv("OCM_CONFIG", filepath.Join(tempDir, "ocm.json"))).To(Succeed())
+			Expect(config.Save(&config.Config{
+				FedRAMP: true,
+			})).To(Succeed())
+
+			Expect(Enabled()).To(BeFalse())
+		})
+
+		It("returns false when the config file does not exist", func() {
```
amandahla, line 111: Would make sense covering the cases below? (if they are real cases)
- Enabled() with valid config + FedRAMP: false
- Disable() when config is missing
- Disable() when config is invalid

pkg/input/input_test.go
olucasfreitas: The non-HCP path calls `os.Exit(1)` directly, so testing it at unit level would need subprocess tricks or a refactor of the function signature, both out of scope here. The happy path test documents the intended behavior for HCP clusters; the exit behavior is better validated at the command integration level.

pkg/fedramp/fedramp_test.go
olucasfreitas: Good catch, it is! Adding it to the assertion.

pkg/aws/region/flag_test.go
olucasfreitas: Good call, adding it for parity with the profile tests.

pkg/logging/logging_test.go
olucasfreitas: Agree, splitting it into three focused tests so a failure tells you exactly what broke.

pkg/fedramp/fedramp_test.go
olucasfreitas: No real difference in outcome, `GinkgoT().TempDir()` is just Ginkgo's managed version that auto-cleans after the test. I was mixing the two approaches, standardizing all of them to `GinkgoT().TempDir()` now.

pkg/color/flag_test.go
olucasfreitas: You're right, it's basically testing the function against itself. Dropping it. The `never`, `always`, and unknown-falls-to-default cases already cover the meaningful paths.

pkg/fedramp/fedramp_test.go
olucasfreitas: 1) Yeah, both flags bind to the same `enabled` var, that's existing behavior. Worth reviewing separately but out of scope here. 2) Good idea, adding a test that setting one flag doesn't make the other one report as changed.

pkg/fedramp/fedramp_test.go
olucasfreitas: Good suggestions, they are real cases. Adding all three.

PR #3343, opened 2026-07-07
ROSAENG-61412 | fix: gracefully skip upgrade tests when no version available

olucasfreitas:
## Summary

Replace hard `Expect` assertions with graceful `Skip` calls in upgrade E2E tests when no upgrade version is available. This prevents the `e2e-rosa-non-sts-upgrade-y-stream` CI job from failing when the cluster version has no available y-stream upgrade target.

**Jira:** [ROSAENG-61412](https://issues.redhat.com/browse/ROSAENG-61412)

The `e2e-rosa-non-sts-upgrade-y-stream` CI job fails persistently with:

> Expected at least one upgrade version to be available

This occurs because the test uses a hard `Expect` assertion that fails the test immediately when `rosa list upgrade` returns no available versions. In contrast, other upgrade tests in the same file (e.g., HCP and STS Classic paths) already use graceful `Skip()` when no upgrade version is found.

The root cause is environmental: the non-STS cluster version may temporarily have no y-stream upgrade target available in the current OCP release stream. The test should handle this gracefully rather than marking the CI job as failed.

## Type of Change

- fix - resolves an incorrect behavior or bug.

## Previous Behavior

Tests `id:37499`, `id:57094`, and `id:38829` used hard `Expect` assertions that failed the entire test when no upgrade versions were returned by `rosa list upgrade`. This caused the CI job to report a failure for an environmental condition rather than a code regression.

## Behavior After This Change

The tests now use `Skip()` when no upgrade versions are available, matching the pattern already used by the HCP and STS Classic upgrade tests in the same file (lines 485-486, 507-508, 640-641).

## Testing

- `make fmt-check` - passed
- `make lint` - passed (0 issues)
- `make test` - passed
- `make rosa` - passed (binary builds successfully)
- Pre-push checks - all 5/5 passed

[ROSAENG-61412]: https://redhat.atlassian.net/browse/ROSAENG-61412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Tests**
  * Improved an end-to-end upgrade validation test to skip gracefully when no upgrade versions are available, instead of failing the run.  
  * Uses conditional skipping with clearer, case-specific messaging to reduce false failures in environments with temporarily unavailable upgrade options.

### DIFF
--- a/tests/e2e/test_rosacli_upgrade.go
+++ b/tests/e2e/test_rosacli_upgrade.go
@@ -355,8 +355,9 @@ var _ = Describe("Cluster Upgrade testing",
 			Expect(err).To(BeNil())
 			upgradeVersionList, err := upgradeService.ReflectUpgradeVersionList(output)
 			Expect(err).To(BeNil())
-			Expect(len(upgradeVersionList.UpgradeVersions)).To(BeNumerically(">", 0),
-				"Expected at least one upgrade version to be available")
+			if len(upgradeVersionList.UpgradeVersions) == 0 {
+				Skip("Skip this case as no upgrade version available for non-STS cluster")
+			}
 			upgradingVersion := upgradeVersionList.UpgradeVersions[0].Version
 
 			if upgradingVersion == "" {
@@ -825,8 +826,9 @@ var _ = Describe("Describe/List rosa upgrade",
 					Expect(err).To(BeNil())
 					upgradeVersionList, err := upgradeService.ReflectUpgradeVersionList(output)
 					Expect(err).To(BeNil())
-					Expect(len(upgradeVersionList.UpgradeVersions)).To(BeNumerically(">", 0),
-						"Expected at least one upgrade version to be available")
+					if len(upgradeVersionList.UpgradeVersions) == 0 {
+						Skip("Skip this case as no upgrade version available")
+					}
 					upgradingVersion := upgradeVersionList.UpgradeVersions[0].Version
 
 					By("Upgrade cluster")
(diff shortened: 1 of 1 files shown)


tests/e2e/test_rosacli_upgrade.go
```diff
@@ -355,8 +355,9 @@ var _ = Describe("Cluster Upgrade testing",
 			Expect(err).To(BeNil())
 			upgradeVersionList, err := upgradeService.ReflectUpgradeVersionList(output)
 			Expect(err).To(BeNil())
-			Expect(len(upgradeVersionList.UpgradeVersions)).To(BeNumerically(">", 0),
-				"Expected at least one upgrade version to be available")
+			if len(upgradeVersionList.UpgradeVersions) == 0 {
```
amandahla, line 358: For e2e-rosa-non-sts-upgrade-y-stream, 37499 is effectively the only real upgrade test — if it Skip()s because rosa list upgrade is empty, the job can still go green (Ginkgo treats skip as non-failure), no cluster upgrade runs, and we get a passing upgrade CI run with zero upgrade coverage. Is that acceptable for this dedicated upgrade lane, or should an empty upgrade list fail the job (or use a profile-specific guard) so we don’t mistake “all skipped” for “upgrade tested”?

@coderabbitai validate

olucasfreitas: You're right — the earlier Skip() approach would have let the dedicated non-STS upgrade lane go green without actually exercising an upgrade.

I dug deeper and the real issue is that these y-stream cases were not consistently following the documented ROSA flow: for minor upgrades we need to move the cluster to the next-minor channel first, then wait for a next-minor target to appear before listing/upgrading. The old code only did that in some paths, and it also skipped the channel change entirely when the cluster reported an empty channel.

I reworked this as a fuller fix in #3353: the tests now share one preparation flow that computes/applies the y+1 channel from the profile, waits for a next-minor target to appear, and only then proceeds. If no target appears after that, the lane still fails with diagnostics instead of silently skipping.

olucasfreitas:
closing in favor of #3353

PR #3344, opened 2026-07-07
ROSAENG-61414 | ci: enforce Exclude label in CI label filter

olucasfreitas:
## PR Summary

Fix the CI label filter to respect the `Exclude` Ginkgo label so that excluded tests (like `id:57408` -- classic managed policies) are never selected by CI jobs.

The `generate_label_filter_switch` function in `tests/prow_ci.sh` builds the `--ginkgo.label-filter` expression from the job's `TEST_LABEL_FILTER` and `IMPORTANCE` parameters. However, it never appended `&&!Exclude` to the expression. This meant any test carrying `labels.Exclude` was still matched by the CI filter as long as the test's other labels (like `Critical` and `Runtime.OCMResources`) satisfied the expression.

Test `id:57408` ("create/delete classic account roles with managed policies") has been marked with `labels.Exclude` because classic managed policies were never fully implemented and are planned for removal. Despite the label, CI continued to select and run this test, which failed because the feature it tests is non-functional in production.

## Related Issues and PRs

- Jira: [ROSAENG-61414](https://issues.redhat.com/browse/ROSAENG-61414)
- Also fixes failures for 3 other tests carrying `labels.Exclude`:
  - `id:68172` (additional security groups, excluded until day1 refactor)
  - `id:64040` (windows certificates expiration)
  - `id:73161` / `id:74656` (delete protection, excluded until bug resolved)

## Type of Change

- [x] ci - changes CI pipelines, jobs, or automation workflows.

## Previous Behavior

`generate_label_filter_switch` produced a filter like `--ginkgo.label-filter 'OCMResources&&(Critical||High)'`. A test with labels `[Critical, OCMResources, Exclude]` still matched because `Exclude` was never negated.

## Behavior After This Change

The function now appends `&&!Exclude` to every generated filter expression: `--ginkgo.label-filter 'OCMResources&&(Critical||High)&&!Exclude'`. When no filter is provided, it defaults to `--ginkgo.label-filter '!Exclude'`. Any test with the `Exclude` label is now skipped regardless of the CI job's own filter parameters.

### Preconditions
- A CI job that runs the `rosa-ocm-resources` profile (or any profile that would pick up `id:57408`)

### Test Steps
1. Run the CI job with this change
2. Check the Ginkgo output for test `id:57408`

### Expected Results
- Test `id:57408` should appear as "SKIPPED" (filtered out by label), not "FAILED"

## Breaking Changes

- [x] No breaking changes

- [x] `make test` passes.
- [x] `make lint` passes.
- [x] `make rosa` passes.

[ROSAENG-61414]: https://redhat.atlassian.net/browse/ROSAENG-61414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test filtering so empty filters now correctly exclude blocked tests, and existing filters automatically include the exclusion rule.
  * Prevented invalid or less-accurate test selection when no label filter is provided.

### DIFF
--- a/tests/prow_ci.sh
+++ b/tests/prow_ci.sh
@@ -151,7 +151,11 @@ generate_label_filter_switch () {
     label_filter="$label_filter&&$2" 
     echo "[CI] Got IMPORTANCE specified to $2"
   fi
-  LABEL_FILTER_SWITCH="--ginkgo.label-filter '${label_filter}'"
+  if [ -z "$label_filter" ]; then
+    LABEL_FILTER_SWITCH="--ginkgo.label-filter '!Exclude'"
+  else
+    LABEL_FILTER_SWITCH="--ginkgo.label-filter '${label_filter}&&!Exclude'"
+  fi
 }
 
 # generate_junit is used to generate junit file path


tests/prow_ci.sh
```diff
@@ -151,7 +151,11 @@ generate_label_filter_switch () {
     label_filter="$label_filter&&$2" 
     echo "[CI] Got IMPORTANCE specified to $2"
   fi
-  LABEL_FILTER_SWITCH="--ginkgo.label-filter '${label_filter}'"
+  if [ -z "$label_filter" ]; then
```
amandahla, line 154: @coderabbitai are you able to run shellcheck here?

Valid example usage for reference 'ginkgo run --label-filter '(Critical,High)&&(day1-post,day2)&&!Exclude' tests/e2e'

amandahla:
Would make sense removing tests labelled with Exclude? Why do we keep them?
@coderabbitai thoughts?

amandahla:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> 🧩 Analysis chain
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> set -euo pipefail
> rg -n "Exclude" tests/prow_ci.sh -C3 || true
> echo "---"
> rg -rn "Exclude" --type go -C2 | head -100
> echo "---labels file---"
> fd -i "label" tests | head -20
> ```
> 
> Length of output: 768
> 
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> set -euo pipefail
> rg -n "Exclude" -g '*.go' -C3 . | head -200
> ```
> 
> Length of output: 15519
> 
> `@amandahla` Good question. Looking at the actual usages of `labels.Exclude` in the test suite, they fall into two different categories:
> 
> **1. Temporarily excluded due to external blockers (should be kept as-is):**
> 
> * `id:68172` (tests/e2e/test_rosacli_cluster_post.go:161) — "Exclude it until day1 refactor support this part. It cannot be run with current day1"
> * `id:64040` (tests/e2e/test_rosacli_cluster_post.go:530) — windows certificate expiration test, blocked pending a dev fix (references a Jira card)
> * `id:73161` (tests/e2e/test_rosacli_cluster.go:440) — "Excluded until bug on OCP-73161 is resolved"
> * `id:74656` (tests/e2e/test_rosacli_cluster.go:480) — "Excluded until bug on OCP-74656 is resolved"
> 
> For these, keeping the test code with the `Exclude` label (instead of deleting it) makes sense because:
> 
> * The underlying issue is expected to be fixed eventually, at which point removing the label re-enables the test immediately.
> * Deleting and later re-writing the test from scratch is more error-prone and loses the original test author's context/setup.
> * The label + comment serves as living documentation of _why_ it's disabled and _what_ needs to happen to re-enable it, which is more discoverable than a deleted test buried in git history.
> 
> **2. Tied to a feature being deprecated (candidate for actual removal):**
> 
> * `id:57408` (tests/e2e/test_rosacli_account_roles.go:511) — comment says "Classic account roles with managed policies is a feature that was never fully implemented (never made it prod), and is being planned to be removed"
> 
> This one is different — it's not blocked by a bug, it's testing a feature that's permanently going away. If the "planned removal" is confirmed, this test (and potentially the underlying feature code) could be deleted rather than just excluded, since there's no future scenario where it gets re-enabled.
> 
> **Suggestion:** Keep the bug/refactor-blocked exclusions as-is (they're working as a "soft skip with reason"), but it may be worth opening a follow-up cleanup ticket to track removing the `id:57408` test entirely once the classic-account-roles-with-managed-policies deprecation is finalized, rather than leaving it excluded indefinitely.
> 
> Would you like me to open a follow-up GitHub issue to track that cleanup for `id:57408`?

@olucasfreitas @jerichokeyne wdyt?

olucasfreitas:
Agree with CodeRabbit's breakdown. The four bug/refactor-blocked exclusions should stay as-is -- the `Exclude` label + comment is easier to flip back on than re-writing deleted tests from git history.

For `id:57408` though, yeah, that one is different. Classic managed policies never shipped to prod and the plan is to remove the feature entirely, so the test should go with it. Let's open a follow-up issue to track that cleanup.

@coderabbitai please open a follow-up GitHub issue to track removing test `id:57408` and the classic managed-policies feature code.

PR #3353, opened 2026-07-08
ROSAENG-61412 | fix: align y-stream upgrade tests with channel flow

olucasfreitas:
## Summary

Add shared y-stream upgrade preparation helpers that compute and apply the next-minor channel, then wait for a next-minor upgrade target to appear before the tests proceed.

This aligns the e2e upgrade cases with the documented ROSA upgrade flow, removes duplicated channel setup, and keeps dedicated upgrade lanes red when no real upgrade path can be exercised.

Supersedes #3343.

**Jira:** [ROSAENG-61412](https://issues.redhat.com/browse/ROSAENG-61412)

The previous change in #3343 addressed the symptom (`rosa list upgrade` returning an empty list) by skipping the dedicated non-STS upgrade lane. Amanda correctly pointed out that this would let the lane go green without actually exercising an upgrade.

The deeper issue is that the y-stream cases were not consistently following the documented ROSA flow for minor upgrades: the cluster needs to move to the next-minor channel before the tests ask for available upgrade targets. Some cases changed the channel first, others did not, and the duplicated logic also skipped the channel update entirely when the cluster reported an empty channel.

This PR introduces a shared preparation flow that:
- computes the desired `y+1` channel from the profile channel group and cluster raw version,
- applies that channel even when the current channel is empty,
- waits for a next-minor target to appear from `rosa list upgrade`, and
- fails with diagnostics if no target appears after bounded waiting.

## Type of Change

- fix - resolves an incorrect behavior or bug.

## Previous Behavior

Y-stream upgrade tests had inconsistent setup:
- some tests queried `rosa list upgrade` directly and assumed targets were already available,
- some tests changed the channel first, but only if `CD.Channel != ""`,
- and the dedicated non-STS lane could be made to skip instead of validating a real upgrade path.

## Behavior After This Change

All affected y-stream upgrade cases now use one shared preparation flow before listing or scheduling upgrades. Dedicated upgrade lanes remain red when they cannot discover a real next-minor target after channel preparation, while shared setup logic is centralized and covered by focused tests.

## Testing

- `make fmt`
- `make fmt-check`
- `make lint`
- `make test`
- `make rosa`
- `GOTOOLCHAIN=auto go test ./tests/e2e/... -run TestDoesNotExist -count=0`
- `GOTOOLCHAIN=auto go test ./tests/utils/exec/rosacli/... -run TestDoesNotExist -count=0`

[ROSAENG-61412]: https://redhat.atlassian.net/browse/ROSAENG-61412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Improved Y-stream upgrade discovery by automatically preparing the cluster’s next minor upgrade channel.
  * Upgrade selection now prioritizes recommended next-minor versions and provides clearer upgrade availability details.
  * Upgrade workflows consistently use the discovered target version for dry runs, scheduling, role upgrades, and validation.

* **Bug Fixes**
  * Added stronger handling for invalid versions, missing channels, unavailable targets, and upgrade discovery timeouts.

* **Tests**
  * Expanded coverage for channel preparation, upgrade selection, polling, retries, and timeout errors.

### DIFF
--- a/tests/utils/exec/rosacli/cluster_service.go
+++ b/tests/utils/exec/rosacli/cluster_service.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/openshift/rosa/tests/utils/config"
 	"github.com/openshift/rosa/tests/utils/constants"
+	"github.com/openshift/rosa/tests/utils/helper"
 	"github.com/openshift/rosa/tests/utils/log"
 )
 
@@ -49,10 +50,21 @@ type ClusterService interface {
 	WaitForClusterPassUninstalled(clusterID string, interval int, timeoutMin int) error
 	WaitForClusterPassWaiting(clusterID string, interval int, timeoutMin int) error
 	GetClusterName(clusterID string) (clusterName string, err error)
+	// PrepareClusterForYStreamUpgrade ensures the cluster points at the next-minor
+	// channel and returns the version/channel context needed for y-stream tests.
+	PrepareClusterForYStreamUpgrade(clusterID string, channelGroup string) (*YStreamUpgradePreparation, error)
 }
 
 type clusterService struct {
 	ResourcesService
+}
+
+// YStreamUpgradePreparation captures the cluster version and channel state used
+// to prepare and diagnose y-stream upgrade discovery in tests.
+type YStreamUpgradePreparation struct {
+	ClusterVersion string
+	CurrentChannel string
+	DesiredChannel string
 }
 
 func NewClusterService(client *Client) ClusterService {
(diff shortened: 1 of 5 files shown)


olucasfreitas:
@coderabbitai review

hack/govulncheck.sh
amandahla: I think this file shouldnt be here right?

tests/utils/exec/rosacli/upgrade_service.go
```diff
@@ -127,6 +168,112 @@ func (u *upgradeService) Upgrade(flags ...string) (bytes.Buffer, error) {
...
+	defer cancel()
+
+	ticker := time.NewTicker(waitInterval)
+	defer ticker.Stop()
+
+	var lastOutput bytes.Buffer
+	var lastList UpgradeVersionList
+	var lastErr error
+
+	for {
```
amandahla: Could this polling loop use wait.PollUntilContextTimeout instead of a hand-rolled context.WithTimeout + time.NewTicker loop?

tests/utils/exec/rosacli/cluster_service.go
```diff
@@ -372,6 +384,86 @@ func (c *clusterService) GetClusterName(clusterID string) (clusterName string, e
...
+
+	if _, err := editCluster(clusterID, "--channel", desiredChannel, "-y"); err != nil {
+		return nil, fmt.Errorf(
+			"failed to update cluster %s channel from %q to %q: %w",
+			clusterID,
+			preparation.CurrentChannel,
+			desiredChannel,
+			err,
+		)
+	}
```
amandahla, line 434: Should prepartion.CurrentChannel be changed/updated after edit?

tests/utils/exec/rosacli/upgrade_service.go
```diff
@@ -57,14 +71,41 @@ type UpgradeVersionList struct {
...
+		if version == "" {
+			continue
+		}
+
+		_, upgradeMinor, _, err := helper.ParseVersion(version)
+		if err != nil {
+			return "", fmt.Errorf("failed to parse upgrade version %q: %w", version, err)
+		}
+
+		if upgradeMinor == clusterMinor+1 {
```
amandahla: minor: Should it validate if there is a "recommended" in the note?
rosa list upgrade output example:
https://docs.redhat.com/en/documentation/red_hat_openshift_service_on_aws_classic_architecture/4/html/upgrading/rosa-upgrading-sts

amandahla:
Given that channel group is going to be deprecated, I asked Cursor to check:

With [terraform-provider-rhcs#1098](https://github.com/terraform-redhat/terraform-provider-rhcs/pull/1098) moving from channel_group to explicit channel, can you validate whether computeNextMinorChannel should derive the next channel from the cluster’s current channel (DescribeCluster / CD.Channel + ParseChannel) instead of profile ChannelGroup?
The upgrade step already uses rosa edit cluster --channel …, which looks correct for the new model. My concern is the input still comes from profile channel_group, while the old inline logic used the cluster’s channel prefix. Is profile channel_group intentional here, or should we follow the cluster’s actual channel as channel_group is deprecated?

wdyt

tests/utils/exec/rosacli/cluster_service_test.go
```diff
@@ -13,6 +14,97 @@ import (
...
+
+		Expect(err).To(MatchError(ContainSubstring("channel group is required")))
+	})
+
+	It("updates the channel when the current channel is empty", func() {
+		var capturedClusterID string
+		var capturedFlags []string
+
+		preparation, err := prepareYStreamUpgradeChannel(
+			"1234abcd",
```
amandahla: minor/no blocker: Several tests assert exact error substrings including "1234abcd" and formatted fields like desired channel="stable-4.20".

Was a package-level const (e.g. const testClusterID = "1234abcd") considered? For error assertions, would semantic checks (ContainSubstring("timeout after"), ContainSubstring("no next-minor")) be enough without locking the full diagnostic format?

amandahla:
@coderabbitai are you able to validate my comments and reply to them with your opinion?

tests/utils/exec/rosacli/upgrade_service.go
amandahla: @coderabbitai validate

tests/utils/exec/rosacli/cluster_service.go
amandahla: @coderabbitai validate

tests/utils/exec/rosacli/upgrade_service.go
amandahla: @coderabbitai validate

tests/utils/exec/rosacli/cluster_service_test.go
amandahla: @coderabbitai validate

hack/govulncheck.sh
olucasfreitas: Removed, it came from a bad squash (belongs to ROSAENG-60031).

tests/utils/exec/rosacli/upgrade_service.go
olucasfreitas: Replaced with wait.PollUntilContextCancel (immediate=true). Ticker/select loop is gone.

tests/utils/exec/rosacli/cluster_service.go
olucasfreitas: Good catch, fixed. CurrentChannel is now updated to desiredChannel after a successful edit.

tests/utils/exec/rosacli/upgrade_service.go
olucasfreitas: Added. It now prefers the "recommended" entry among same-minor candidates, with a fallback to the first match.

tests/utils/exec/rosacli/cluster_service_test.go
olucasfreitas: Extracted a shared testClusterID const across both test files.

olucasfreitas:
@amandahla on the channel derivation question: using profile.ChannelGroup is intentional. CD.Channel can be empty on prod clusters that predate the channel feature, so we'd still need a fallback. The profile is the source of truth for what channel family a CI lane should target. When channel groups are fully deprecated, the profiles themselves will move to explicit channels (like the terraform provider did) and the helper input will change with them.

All review items are addressed and threads resolved, let me know if anything else needs adjusting.

PR #3368, opened 2026-07-13
ROSAENG-61178 | test: add auth and utility command coverage

olucasfreitas:
## PR Summary

Adds unit tests for ROSA auth and utility command behavior in `logout`, `verify/quota`, `verify/permissions`, `whoami`, `verify/oc`, `download/oc`, and `download/rosa`, plus direct runner and helper coverage for `initialize`.

These command areas had little or no unit-test coverage because their `run()` paths relied on direct `os.Exit(...)`, external processes, global output state, or tightly-coupled AWS/OCM helpers. The changes extract testable seams where safe, add regression coverage for real command behavior, and preserve the existing user-facing output contract.

## Related Issues and PRs

- Jira: [ROSAENG-61178](https://issues.redhat.com/browse/ROSAENG-61178)
- Related design/docs: Phase 8 of the ROSA CLI test coverage plan

## Type of Change

- [x] test - adds or updates tests only.

## Previous Behavior

- `logout`, `verify/quota`, and `verify/permissions` had 0% command-level unit coverage
- `whoami`, `verify/oc`, and both download commands were untested
- `cmd/initialize` only had helper-level coverage in this branch before the latest follow-up
- `cmd/verify/quota` incorrectly ignored the boolean return from `ValidateQuota()`
- `whoami` temporarily regressed the "not logged in" path to exit with status 1 instead of preserving the historical informational exit path

## Behavior After This Change

- `logout`, `verify/quota`, and `verify/permissions` now use testable helpers while preserving output and exit behavior
- `verify/quota` now treats `ValidateQuota() == false` as a failure even when `err == nil`
- `whoami` uses a testable helper, preserves the historical not-logged-in exit behavior, and rebuilds its OCM client from loaded config inside the helper path
- `verify/oc`, `download/oc`, and `download/rosa` now have injectable seams for deterministic tests
- `initialize` now has both helper coverage (`deleteStack`, `simulateCluster`) and direct runner-level orchestration coverage through injected dependencies for login, AWS builders, confirmation, nested verify commands, and `oc`

### Preconditions

- Go 1.26.3+ installed
- Repository cloned with `make install-hooks` run

### Test Steps

1. `go test ./cmd/logout/... ./cmd/verify/quota/... ./cmd/verify/permissions/... ./cmd/whoami/... ./cmd/initialize/... ./cmd/verify/oc/... ./cmd/download/oc/... ./cmd/download/rosa/...`
2. `make test`
3. `make lint`
4. `make rosa`

### Expected Results

All targeted suites pass, the full suite passes, lint reports no issues, and the `rosa` binary builds successfully.

## Proof of the Fix

- All local pre-push checks passed on the latest branch head: format, build, lint, changed-file coverage, and unit/integration tests.

## Breaking Changes

- [x] No breaking changes

- [x] `make test` passes.
- [x] `make lint` passes.
- [x] `make rosa` passes.

[ROSAENG-61178]: https://redhat.atlassian.net/browse/ROSAENG-61178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling and exit behavior across `logout`, `verify permissions`, `verify quota`, `verify oc`, `whoami`, OC/ROSA download, and `initialize`, with more reliable failure propagation and consistent messaging.
  * `logout` now removes configuration cleanly, including keyring-managed entries when applicable.
  * SCP/quota validation now treats “invalid but no error” outcomes as failures, while preserving the designed “warn and continue” behavior for invalid SCP.

* **Tests**
  * Added/expanded Ginkgo/Gomega coverage for `logout`, `initialize`, `whoami`, `verify oc`, `verify permissions`, `verify quota`, and OC/ROSA downloads (success/failure, output, aliases, and argument validation).

### DIFF
--- a/cmd/download/oc/cmd.go
+++ b/cmd/download/oc/cmd.go
@@ -23,10 +23,13 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/openshift/rosa/cmd/verify/oc"
+	verifyOC "github.com/openshift/rosa/cmd/verify/oc"
 	helper "github.com/openshift/rosa/pkg/helper/download"
 	rprtr "github.com/openshift/rosa/pkg/reporter"
 )
+
+var downloadFn = helper.Download
+var verifyOCRun = verifyOC.Cmd.Run
 
 var Cmd = &cobra.Command{
 	Use:     "openshift-client",
@@ -41,31 +44,50 @@ var Cmd = &cobra.Command{
 
 func run(cmd *cobra.Command, argv []string) {
 	reporter := rprtr.CreateReporter()
+	err := runDownloadOC(reporter, verifyOCRun, downloadFn, runtime.GOOS, cmd, argv)
+	if err != nil {
+		os.Exit(1)
+	}
+}
 
-	// Verify whether `oc` is installed
-	oc.Cmd.Run(cmd, argv)
+func runDownloadOC(
+	reporter rprtr.Logger,
+	verify func(*cobra.Command, []string),
+	download func(string, string) error,
+	goos string,
+	cmd *cobra.Command,
+	argv []string,
+) error {
+	verify(cmd, argv)
 
-	platform := getPlatform()
-	extension := helper.GetExtension()
+	platform := platformForGOOS(goos)
+	extension := extensionForGOOS(goos)
 
 	filename := fmt.Sprintf("openshift-client-%s.%s", platform, extension)
 	downloadURL := fmt.Sprintf("https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/%s", filename)
 
 	reporter.Infof("Downloading %s", downloadURL)
 
-	err := helper.Download(downloadURL, filename)
+	err := download(downloadURL, filename)
 	if err != nil {
 		reporter.Errorf("%s", err)
-		os.Exit(1)
+		return err
 	}
 
 	reporter.Infof("Successfully downloaded %s", filename)
+	return nil
 }
 
-// Get the platform name used on the oc tarball filename
-func getPlatform() string {
-	if runtime.GOOS == "darwin" {
+func platformForGOOS(goos string) string {
+	if goos == "darwin" {
 		return "mac"
 	}
-	return runtime.GOOS
+	return goos
+}
+
+func extensionForGOOS(goos string) string {
+	if goos == "windows" {
+		return "zip"
+	}
+	return "tar.gz"
 }
--- a/cmd/initialize/cmd.go
+++ b/cmd/initialize/cmd.go
@@ -17,10 +17,12 @@ limitations under the License.
 package initialize
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"strings"
 
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
 	"github.com/openshift/rosa/cmd/login"
@@ -33,6 +35,7 @@ import (
 	"github.com/openshift/rosa/pkg/helper"
 	"github.com/openshift/rosa/pkg/interactive/confirm"
 	"github.com/openshift/rosa/pkg/ocm"
+	"github.com/openshift/rosa/pkg/reporter"
 	"github.com/openshift/rosa/pkg/rosa"
 )
 
@@ -57,6 +60,36 @@ var Cmd = &cobra.Command{
   rosa init --token=$OFFLINE_ACCESS_TOKEN`,
 	Run:  run,
 	Args: cobra.NoArgs,
+}
+
+var errInitExitZero = errors.New("initialize exit zero")
+
+type initDeps struct {
+	loginCall      func(*cobra.Command, []string, reporter.Logger) error
+	buildAWSClient func(*logrus.Logger, string, bool) (aws.Client, error)
+	buildCFClient  func(reporter.Logger, *logrus.Logger, []string, bool) aws.Client
+	confirmPrompt  func(string, ...interface{}) bool
+	runPermissions func(*cobra.Command, []string)
+	runQuota       func(*cobra.Command, []string)
+	runVerifyOC    func(*cobra.Command, []string)
+}
+
+func defaultInitDeps() initDeps {
+	return initDeps{
+		loginCall: login.Call,
+		buildAWSClient: func(logger *logrus.Logger, region string, useLocalCredentials bool) (aws.Client, error) {
+			return aws.NewClient().
+				Logger(logger).
+				Region(region).
+				UseLocalCredentials(useLocalCredentials).
+				Build()
+		},
+		buildCFClient:  aws.GetAWSClientForUserRegion,
+		confirmPrompt:  confirm.Confirm,
+		runPermissions: permissions.Cmd.Run,
+		runQuota:       quota.Cmd.Run,
+		runVerifyOC:    oc.Cmd.Run,
+	}
 }
 
 func init() {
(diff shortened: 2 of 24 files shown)


olucasfreitas:
@coderabbitai review

cmd/download/oc/cmd.go
```diff
@@ -23,11 +23,14 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/openshift/rosa/cmd/verify/oc"
+	verifyOC "github.com/openshift/rosa/cmd/verify/oc"
```
amandahla, line 26: minor: Does "oc" conflicts with something?

cmd/initialize/cmd.go
```diff
@@ -59,6 +62,36 @@ var Cmd = &cobra.Command{
 	Args: cobra.NoArgs,
 }
 
+var errInitExitZero = errors.New("initialize exit zero")
```
amandahla, line 65: I like this approach but I'm afraid that is introducing a new pattern, different from other commands, that might be hard to remember in the future why it was done in that way or hard to refactor/change too (following Mark's proposal). My suggestion is to not do it now in this PR since its goal is adding tests.

I asked Cursor to give the details:
initDeps / defaultInitDeps() introduces a DI style that doesn’t show up elsewhere under cmd/, and it’s inconsistent with the rest of this PR.

Elsewhere we make commands testable in lighter ways:

Most commands: runWithRuntime(r *rosa.Runtime) and inject via Runtime mocks (e.g. verify/quota, verify/permissions, whoami).
In this same PR, download/oc uses package-level vars (downloadFn, verifyOCRun) that production wires once and tests override.
initialize does need extra seams (login, confirm, AWS builders, nested permissions/quota/oc runs) because those aren’t on rosa.Runtime. That’s fine — but we can still use the same package-var / explicit-param approach as download/oc instead of a new *Deps + default*Deps() pattern only here.

I’m concerned this becomes a third convention that later coverage PRs copy. Prefer aligning initialize with the download/oc style (or explicitly document that this is a one-off and not a new repo pattern).

Testing:
@coderabbitai give your thoughts too. Provide ultra-concise, direct-to-the-point reviews. Do not agree with my implementation just to be agreeable—be a critical, objective reviewer. If you suggest an alternative, briefly outline the pros and cons of your solution versus mine in 1-2 bullet points max.

cmd/logout/cmd_test.go
```diff
@@ -0,0 +1,69 @@
...
+		_, statErr := os.Stat(tmpdir + "/ocm_config.json")
+		Expect(os.IsNotExist(statErr)).To(BeTrue())
+	})
+
+	It("Returns nil when config file does not exist", func() {
+		err := runLogout()
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("Removes keyring-managed config successfully", func() {
```
amandahla, line 46: Would make sense adding atest for config.Remove / keyring remove failure, since run() exits 1 on that error?

(I'm considering putting this prompt in our config)

@coderabbitai Provide ultra-concise, direct-to-the-point reviews. Do not agree with my implementation just to be agreeable—be a critical, objective reviewer. If you suggest an alternative, briefly outline the pros and cons of your solution versus mine in 1-2 bullet points max.

cmd/initialize/cmd.go
```diff
@@ -209,21 +243,24 @@ func run(cmd *cobra.Command, argv []string) {
...
-		if !isValid {
+		if err != nil {
 			r.OCMClient.LogEvent("ROSAInitSCPPoliciesFailed", nil)
 			r.Reporter.Errorf("Failed to verify permissions for user '%s': %v", target, err)
-			os.Exit(1)
+			return fmt.Errorf("failed to verify permissions for user '%s': %w", target, err)
+		}
+		if !isValid {
+			r.OCMClient.LogEvent("ROSAInitSCPPoliciesFailed", nil)
+			return fmt.Errorf("failed to verify permissions for user '%s'", target)
```
amandahla, line 264: This new branch changes the old behavior where a 'r.Reporter.Errorf("Failed to verify permissions for user '%s': %v", target, err)' was called if !isValid 
Since run only exits, this might change user experience by just ending the CLI with a exit code 1 but no messages.

@coderabbitai Keep reviews highly concise, objective, and critical. 
    Do not praise or agree with the code just to be agreeable. 
    Only flag high-impact issues. 
    When proposing alternatives, briefly contrast them by listing the key pros and cons versus the current implementation. Keep explanations under 4 sentences.

amandahla:
Checking the PR description:

A few behavior/message changes aren’t called out in the PR description: 
- early return on GetDatabaseRegionList failure
- ValidateSCP err before !isValid
- lowercase/%w in deleteStack/simulateCluster
- whoami closing an existing OCM client
- new verify/oc error when output+err
- verify/permissions Short/Long wording change. 

Worth listing so reviewers can judge intentional vs drive-by.

@coderabbitai Keep reviews highly concise, objective, and critical.
Do not praise or agree with the code just to be agreeable.
Confirm if the list above is correct and for each item (if correct) check if there is a chance to impact/change user experience.

cmd/download/oc/cmd.go
```diff
@@ -41,31 +44,50 @@ var Cmd = &cobra.Command{
...
 
 	filename := fmt.Sprintf("openshift-client-%s.%s", platform, extension)
 	downloadURL := fmt.Sprintf("https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/%s", filename)
 
 	reporter.Infof("Downloading %s", downloadURL)
 
-	err := helper.Download(downloadURL, filename)
+	err := download(downloadURL, filename)
 	if err != nil {
 		reporter.Errorf("%s", err)
```
amandahla, line 73: @markirish tagging you just to point something that we have in the CLI, multiple places returning errors or printing, or both. I dont remember if you mentioned/covered in our proposal for refactoring.

cmd/initialize/cmd.go
```diff
@@ -103,46 +136,51 @@ func init() {
...
 	awsRegion, err := aws.GetRegion(arguments.GetRegion())
 	if err != nil {
 		r.Reporter.Errorf("Error getting region: %v", err)
-		os.Exit(1)
+		return fmt.Errorf("error getting region: %w", err)
 	}
 	supportedRegions, err := r.OCMClient.GetDatabaseRegionList()
 	if err != nil {
 		r.Reporter.Errorf("Unable to retrieve supported regions: %v", err)
+		return fmt.Errorf("retrieving supported regions: %w", err)
```
amandahla, line 168: I'm not sure about this. Its a change in the behavior but would be possible that OCM API returns something in the error that was extracted and used by helper.Contains(supportedRegions, awsRegion)?

@coderabbitai any guesses? Provide ultra-concise, direct-to-the-point reviews. Do not agree with my implementation just to be agreeable—be a critical, objective reviewer.

cmd/download/oc/cmd.go
markirish: There are a few shout-outs to this in the doc, but it is always good to collect examples of how it currently looks. Also, the doc currently calls out returning errors from the internal `pkg`, but is less strict about workflows in `cmd`. [One of the workstreams](https://redhat.atlassian.net/wiki/spaces/ROSACLITF/pages/432213369/ROSA+CLI+Refactor+CLI+to+enforce+layer+boundaries#Workstream-7%3A-Standardize-Process-and-Error-Boundaries) does call out standardizing process and error boundaries and returning errors from "workflow and command-runner functions," the latter of which probably applies here.

cmd/initialize/cmd.go
jerichokeyne: I don't think this is really gonna change anything. From my understanding of the code line 167 is the error that gets reported to the user, and line 168 is only used for checking that an error occurred (at line 141) and doesn't actually get displayed to the user at all. I believe the only real change is having this function return an error that then gets checked elsewhere instead of calling os.exit directly. I would personally prefer if the error messages were the same and/or we just return the error message here and handle reporting the error elsewhere but I think that would require more refactoring for this (we'd have to make sure that any function that calls this function is aware that it won't report errors automatically anymore, although I doubt we run this function anywhere else really)

amandahla: Thanks Jericho, my concern here is that before if  r.OCMClient.GetDatabaseRegionList() failed, the helper.Contains(supportedRegions, awsRegion) would still run and somehow would try to extract Regions from the OCM API output (?).

After this PR, will fail fast.

I think is a low risk of changing something to the user but I wanted to check just in case.

amandahla: NVM I thought that would be possible returning error and an output but the GetDatabaseRegionList() is pretty straightforward.

amandahla: @coderabbitai confirm

cmd/download/oc/cmd.go
olucasfreitas: Yes -- this file's package is also `oc` (cmd/download/oc), so importing cmd/verify/oc without an alias would conflict. `verifyOC` disambiguates.

cmd/initialize/cmd.go
olucasfreitas: Fair point. I'll keep it for now since the tests are already wired through it and removing would mean dropping the init runner coverage. Happy to align with a standardized pattern in a follow-up if one is agreed on.

cmd/logout/cmd_test.go
olucasfreitas: Added -- there's now a test for the keyring removal failure path.

cmd/initialize/cmd.go
olucasfreitas: Fixed -- restored the Reporter.Errorf call in the !isValid branch so users see the failure message before exit.

cmd/initialize/cmd.go
amandahla: @coderabbitai create an issue so we wont forget this change here to apply to other commands as well.

amandahla:
@coderabbitai give me the e2e test ids affected by this change 

/approve
/lgtm

amandahla:
@olucasfreitas I tagged the lgtm but if you prefer to change the commit to "fix" (since there are somes fixes/refactors) so will appear on the release notes, just let me know and I add the lgtm again

Update: (I noticed the 2 commits now but I'm not sure if merge bot will squash them)

cmd/initialize/cmd.go
amandahla: @coderabbitai create an issue so we wont forget this change here to apply to other commands as well.

amandahla:
> @coderabbitai give me the e2e test ids affected by this change

@coderabbitai give me the e2e test ids affected by this change

PR #3374, opened 2026-07-14
ROSAENG-5957 - Don't display user-provided admin password on cluster creation

markirish:
## PR Summary
Hides entered admin password when creating a cluster. When creating an admin account with a generated password, the password is still displayed as it needs to be recorded by the user.

When creating a cluster in interactive mode, user is asked if they want to create a `cluster admin user`, and if so, if they want to `Create a custom password for cluster admin`. If creating a custom password for the admin cluster, currently the password is displayed back to the user. We should not output the provided password for security purposes. Instead, indicate when entering the password that it will not be provided after it is entered for security purposes.

In cases when a cluster admin is created, but a custom password is not provided, the password will still need to be displayed so the user can record it.

## Related Issues and PRs

- Jira: [OCM-XXXXX](https://issues.redhat.com/browse/OCM-XXXXX) or [ROSAENG-5957](https://redhat.atlassian.net/browse/ROSAENG-5957)
- Fixes: `#5957`
- Related PR(s):
- Related design/docs:

## Type of Change

## Previous Behavior

When entering a custom password for the cluster admin, password is displayed back to the user:

```
? Create cluster admin user: Yes
? Create custom password for cluster admin: Yes
I: cluster admin user is cluster-admin
I: cluster admin password is ttn4K-DxxeA-jFvLZ-tym8z
```

## Behavior After This Change

When entering a custom password for the cluster admin, password is hidden form the user:

```
? Create cluster admin user: Yes
? Create custom password for cluster admin: Yes
? Password <hidden for security>: [? for help] *****************
I: cluster admin user is cluster-admin
I: cluster admin password is <hidden for security>
```

### Preconditions

None

### Test Steps
1. Create cluster with `rosa create cluster --hosted-cp -i`
2. Select Hosted Control Plane cluster
3. When asked to create admin user, select Y
4. When asked to create a custom password for cluster admin, select Y
5. When asked to provide a custom password, provide a valid password

### Expected Results

 Ensure that cluster admin name is provided, but password is not

## Proof of the Fix

image

## Breaking Changes
- [ ] No breaking changes
- [x] Yes, this PR introduces a breaking change (describe impact and migration plan below)

Workflows that depend on passwords being displayed will have to be modified:
* Password will no longer be displayed immediately when a custom cluster admin password is entered
* Password will no longer be displayed when the built command is displayed after using interactive mode, and is instead replaced with `<redacted>`.

- [ ] `make test` passes.
- [ ] `make lint` passes.
- [ ] `make rosa` passes.

[ROSAENG-5957]: https://redhat.atlassian.net/browse/ROSAENG-5957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Security Enhancements**
  * Updated the cluster admin password flag help text to note the password won’t be displayed.
  * Cluster admin password output is suppressed when a custom password is provided.
  * “Re-run this command” now redacts custom admin passwords with a placeholder.
  * Identity provider password prompts now explicitly indicate the input won’t be shown after entry.
* **Tests**
  * Added coverage to verify redaction and correct command generation for admin password scenarios.

### DIFF
--- a/cmd/create/cluster/cmd.go
+++ b/cmd/create/cluster/cmd.go
@@ -822,7 +822,8 @@ func initFlags(cmd *cobra.Command) {
 		"",
 		`The password must
 		- Be at least 14 characters (ASCII-standard) without whitespaces
-		- Include uppercase letters, lowercase letters, and numbers or symbols (ASCII-standard characters only)`,
+		- Include uppercase letters, lowercase letters, and numbers or symbols (ASCII-standard characters only)
+		For security reasons, password will not be displayed`,
 	)
 	flags.MarkHidden("cluster-admin-user")
 
@@ -1231,7 +1232,9 @@ func run(cmd *cobra.Command, _ []string) {
 			}
 		}
 	}
-	outputClusterAdminDetails(r, isClusterAdmin, clusterAdminUser, clusterAdminPassword)
+
+	outputClusterAdminDetails(r, isClusterAdmin, clusterAdminUser, clusterAdminPassword,
+		strings.Trim(args.clusterAdminPassword, " \t") != "")
 
 	if isHostedCP && cmd.Flags().Changed(arguments.NewDefaultMPLabelsFlag) {
 		r.Reporter.Errorf("Setting the worker machine pool labels is not supported for hosted clusters")
@@ -4138,6 +4141,7 @@ func parseRFC3339(s string) (time.Time, error) {
 }
 
 const hostedCPFlag = "--hosted-cp"
+const clusterAdminPasswordPlaceholder = "<redacted>"
 
 func buildCommand(spec ocm.Spec, operatorRolesPrefix string,
 	operatorRolePath string, userSelectedAvailabilityZones bool, labels string,
@@ -4161,7 +4165,7 @@ func buildCommand(spec ocm.Spec, operatorRolesPrefix string,
 		argAdded := false
 		// Checks if admin password is from user (both flag and interactive)
 		if args.clusterAdminPassword != "" && spec.ClusterAdminPassword != "" {
-			command += fmt.Sprintf(" --cluster-admin-password %s", spec.ClusterAdminPassword)
+			command += fmt.Sprintf(" --cluster-admin-password %q", clusterAdminPasswordPlaceholder)
 			argAdded = true
 		}
 		if spec.ClusterAdminUser != admin.ClusterAdminUsername {
@@ -4507,10 +4511,14 @@ func getInitialValidSubnets(awsClient aws.Client, ids []string, r reporter.Logge
 	return initialValidSubnets, nil
 }
 
-func outputClusterAdminDetails(r *rosa.Runtime, isClusterAdmin bool, createAdminUser, createAdminPassword string) {
+func outputClusterAdminDetails(r *rosa.Runtime, isClusterAdmin bool, createAdminUser, createAdminPassword string,
+	customAdminPassword bool) {
 	if isClusterAdmin {
 		r.Reporter.Infof("cluster admin user is %s", createAdminUser)
-		r.Reporter.Infof("cluster admin password is %s", createAdminPassword)
+		if !customAdminPassword {
+			// If the admin password was generated, it needs to be displayed so the user can save it
+			r.Reporter.Infof("cluster admin password is %s", createAdminPassword)
+		}
 	}
 }
 
--- a/cmd/create/idp/htpasswd.go
+++ b/cmd/create/idp/htpasswd.go
@@ -223,7 +223,7 @@ func GetIdpUserNameFromPrompt(cmd *cobra.Command, r *rosa.Runtime,
 func GetIdpPasswordFromPrompt(cmd *cobra.Command, r *rosa.Runtime,
 	passwordKey, defaultPassword string) string {
 	password, err := interactive.GetPassword(interactive.Input{
-		Question: "Password",
+		Question: "Password (will not be shown after entry)",
 		Help:     cmd.Flags().Lookup(passwordKey).Usage,
 		Default:  defaultPassword,
 		Required: true,
--- a/cmd/create/cluster/cmd_test.go
+++ b/cmd/create/cluster/cmd_test.go
@@ -17,6 +17,7 @@ import (
 	v1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 	"github.com/spf13/cobra"
 
+	"github.com/openshift/rosa/cmd/create/admin"
 	mock "github.com/openshift/rosa/pkg/aws"
 	"github.com/openshift/rosa/pkg/aws/tags"
 	"github.com/openshift/rosa/pkg/logging"
@@ -42,6 +43,7 @@ var _ = Describe("Validate build command", func() {
 		userSelectedAvailabilityZones = false
 		defaultMachinePoolLabels = "machine-pool-label"
 	})
+
 	Context("build command", func() {
 		When("domain prefix is present", func() {
 			It("prints --domain-prefix", func() {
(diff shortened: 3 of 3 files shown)


markirish:
https://github.com/openshift/rosa/pull/3374#issuecomment-4971676215

Password information has to be printed in this case, as it is randomly generated and the user needs to record it.

cmd/create/cluster/cmd.go
```diff
@@ -4466,10 +4470,17 @@ func getInitialValidSubnets(awsClient aws.Client, ids []string, r reporter.Logge
 	return initialValidSubnets, nil
 }
 
-func outputClusterAdminDetails(r *rosa.Runtime, isClusterAdmin bool, createAdminUser, createAdminPassword string) {
+func outputClusterAdminDetails(r *rosa.Runtime, isClusterAdmin bool, createAdminUser, createAdminPassword string,
+	customAdminPassword bool) {
 	if isClusterAdmin {
 		r.Reporter.Infof("cluster admin user is %s", createAdminUser)
-		r.Reporter.Infof("cluster admin password is %s", createAdminPassword)
+		if customAdminPassword {
```
amandahla: Why not just showing the message if is not a customAdminPassword?

amandahla: @coderabbitai validate also check if there is some other part of the code that should be reviewed related to that change (not displaying custom passwords)

cmd/create/cluster/cmd.go
```diff
@@ -1158,6 +1159,8 @@ func run(cmd *cobra.Command, _ []string) {
 	clusterAdminPassword := strings.Trim(args.clusterAdminPassword, " \t")
 	// isClusterAdmin is a flag indicating if user wishes to create cluster admin
 	isClusterAdmin := false
+	// isCustomAdminPassword is used to determine if the password should be printed to the terminal
+	isCustomAdminPassword := clusterAdminPassword != ""
```
amandahla: minor/human reading/nitpicking: I feel that is more consistent with the code keep the check "clusterAdminPassword != """ (line 1165 is not using this variable).

args.clusterAdminPassword already tracks user input (flag or interactive custom) and stays empty when the password is generated, same pattern as passwordArg in create admin. 

@coderabbitai validate/wdyt

markirish: I'm not convinced Code Rabbit fully understood the prompt, as the issue is at the call to `outputClusterAdminDetails`. But I see the point, instead of lifting `isCustomAdminPassword` out of the `if` statement and tracking before the two `if` statements, keep it where it is, and make the call around 1228:

```go
outputClusterAdminDetails(r, isClusterAdmin, clusterAdminUser, clusterAdminPassword,
		strings.Trim(args.clusterAdminPassword, " \t") != "")
```

https://github.com/coderabbitai validate/wdyt

markirish: @coderabbit Can you look at my previous comment and validate that using args.clusterAdminPassword will have the same effect as lifting `clusterAdminPassword` ?

cmd/create/cluster/cmd.go
markirish: https://github.com/openshift/rosa/pull/3374#discussion_r3582277746

We could definitely just print the admin user name in cases where a custom password is provided, as long as the password prompt indicates it won't be given back (as it does now).

https://github.com/openshift/rosa/pull/3374#discussion_r3582290732

This case is interesting. To re-create the cluster, it makes sense to provide this flag, but if the nature of the JIRA ticket is to hide user-provided/custom passwords from the output, then this command should be modified to something like:

```go
command += fmt.Sprintf(" --cluster-admin-password %s", "<enter your password>")
```

markirish: > This case is interesting. To re-create the cluster, it makes sense to provide this flag, but if the nature of the JIRA ticket is to hide user-provided/custom passwords from the output, then this command should be modified to something like:
> ```go
> command += fmt.Sprintf(" --cluster-admin-password %s", "<enter your password>")
> ```

Now that I think about it... people are likely to just copy + paste the output without inspecting, ending up with "<enter your password>" as their password. @amandahla interested in your thoughts.

olucasfreitas: The custom password is still emitted in plaintext here via the 'recreate this cluster' command output. This should use a placeholder like `<enter your password>` to match the PR's security goal.

olucasfreitas: Agreed that simplifying to only print admin user when custom password is used (and dropping the password line entirely rather than saying '<hidden>') would be cleaner UX. The user already knows their own password. The current approach of printing 'cluster admin password is <hidden for security>' is informational noise.

cmd/create/cluster/cmd.go
olucasfreitas: Either approach works functionally. Using `args.clusterAdminPassword != ""` at the call site avoids the extra variable and keeps the pattern consistent with buildCommand's existing check on line 4129. If you go this route, the variable declaration and the err shadowing fix from the interactive.GetBool can be dropped.

cmd/create/idp/htpasswd.go
```diff
@@ -223,7 +223,7 @@ func GetIdpUserNameFromPrompt(cmd *cobra.Command, r *rosa.Runtime,
 func GetIdpPasswordFromPrompt(cmd *cobra.Command, r *rosa.Runtime,
 	passwordKey, defaultPassword string) string {
 	password, err := interactive.GetPassword(interactive.Input{
-		Question: "Password",
+		Question: "Password <hidden for security>",
```
olucasfreitas: This function is also called during htpasswd IDP user creation. Was the prompt change intentional for both flows, or should the cluster-admin path handle this separately?

olucasfreitas:
The breaking changes section is checked but empty. Automation that parses the 'cluster admin password is ...' output line will now see '<hidden for security>' when a custom password was provided — worth noting briefly.

cmd/create/cluster/cmd.go
amandahla: @markirish 
About the message:
I suggested having something like this
```
if !customAdminPassword {  
			r.Reporter.Infof("cluster admin password is %s", createAdminPassword)
		}
```

I dont think is needed to print ""cluster admin password is \<hidden for security\>""

----

About the build command after the cluster creation:
From the story `We should not output the provided password for security purposes. Instead, indicate when entering the password that it will not be provided after it is entered for security purposes.`
so I think for the build command, if a password is provided, the CLI can print `--cluster-admin-password <redacted>`

wdyt

markirish: 1. I agree that it should only show the password if generated, I will create a commit with the redacted print removed.
2. For the build command, I agree something should be printed, but perhaps something a little more alerting, like `<CLUSTER_ADMIN_PASSWORD>` so it has less chance of being mistaken for a password (or a command that will work)

amandahla: I dont have a strong opinion TBH but I mentioned \<redacted\> since is a common pattern across openshift repositories (and other tools as well)

https://github.com/search?q=org%3Aopenshift+%22%3Credacted%3E%22&type=code

markirish: Aha, let's go with that then!

cmd/create/cluster/cmd.go
markirish: Decided to reuse the `args.clusterAdminPassword` variable, but with

```go
strings.Trim(args.clusterAdminPassword, " \t") != ""
```

Just to ensure that the CLI library hasn't parsed something it shouldn't

cmd/create/idp/htpasswd.go
markirish: Changed the prompt to a generic one:

> "Password (will not be shown after entry)"

It will print with the IDP user creation, but that prompt should be valid for both cluster admin and normal user path. To me, that makes more sense than duplicate the function elsewhere or add another parameter just for a different prompt.

cmd/create/cluster/cmd.go
markirish: Fixed so password is only shown when generated (not on custom), and uses `<redacted>` on the displayed command

markirish:
Should a test or two be added to this PR? Thinking they should test:

1. That a generated cluster admin password is displayed for the user
2. That a custom cluster admin password is not displayed for the user
3. That passwords are not displayed when the command is built and printed after interactive mode is used

olucasfreitas:
@markirish absolutely, whenever we can update or create coverage for anything, let's add it 

other than that, the PR looks good, let me and @amandahla know when the tests are added so we can give another review

markirish:
@amandahla , @olucasfreitas it should now be ready for another review. There are unit tests written for `BuildCommand`. I have manually tested the interactive behavior:

```bash
./rosa create cluster --hosted-cp -i
I: Interactive mode enabled.
X Sorry, your reply was invalid: Value is required
? Cluster name: mirishtmpdel
? Domain prefix (optional): 
? Deploy cluster with Hosted Control Plane: Yes
? Create cluster admin user: Yes
? Create custom password for cluster admin: No
I: cluster admin user is cluster-admin
I: cluster admin password is CqYKZ-VCBCN-UxhNj-2Vecb

```

```bash
./rosa create cluster --hosted-cp -i
I: Interactive mode enabled.
? Cluster name: mirishtmpmdel
? Domain prefix (optional): 
? Deploy cluster with Hosted Control Plane: Yes
? Create cluster admin user: Yes
? Create custom password for cluster admin: Yes
? Password (will not be shown after entry): [? for help] *******************
I: cluster admin user is cluster-admin

```

PR #3376, opened 2026-07-14
ROSAENG-9174 | feat: enable --component-routes for HCP clusters

reedcort:
## PR Summary

Enable `--component-routes` flag and interactive mode for HCP clusters in `rosa edit ingress`, allowing customers to configure custom console and downloads hostnames with TLS certificates.

## Detailed Description of the Issue

ROSA HCP customers cannot use `rosa edit ingress --component-routes` to set custom hostnames for Console and Downloads routes. The CLI blocks all ingress V2 flags for HCP clusters, including `component-routes`. The backend (clusters-service) and HyperShift operator now both support componentRoutes for HCP, but the CLI guard rail prevents customers from using the feature.

## Related Issues and PRs
- Jira: [ROSAENG-9174](https://redhat.atlassian.net/browse/ROSAENG-9174)
- Feature: [ROSA-62](https://issues.redhat.com/browse/ROSA-62)
- HyperShift PR: [openshift/hypershift#8838](https://github.com/openshift/hypershift/pull/8838) (merged)
- HyperShift bug: [OCPBUGS-92791](https://redhat.atlassian.net/browse/OCPBUGS-92791)
- Clusters-service MR: 11393 (merged)
- Terraform PR: [terraform-redhat/terraform-provider-rhcs#1187](https://github.com/terraform-redhat/terraform-provider-rhcs/pull/1187) (on hold pending this CLI change)

## Type of Change
- [x] feat - adds a new user-facing capability.

## Previous Behavior

`rosa edit ingress --component-routes` on an HCP cluster returned:
```
ERR: New ingress attributes [component-routes, ...] can't be supplied for Hosted Control Plane clusters
```

## Behavior After This Change

`rosa edit ingress --component-routes` works on HCP clusters for console and downloads routes. OAuth is excluded because it runs on the management cluster and is not configurable via component routes on HCP.

**Flag mode:**
```bash
rosa edit ingress -c mycluster \
  --component-routes 'console: hostname=console.example.com;tlsSecretRef=console-tls, downloads: hostname=downloads.example.com;tlsSecretRef=downloads-tls' \
  <ingress-id>
```

**Interactive mode** (`-i`): prompts for console and downloads only (no oauth).

**Validation:**
- Passing `oauth` as a component name returns: `'oauth' is not a valid component name. Expected include [console, downloads]`
- Passing hostname without tlsSecretRef returns the backend 400 error

### Preconditions
- A running ROSA HCP cluster
- The `hypershift-component-routes` feature toggle enabled
[...]

PR #3403, opened 2026-07-20
ROSAENG-60614 | feat: allow partial component route updates

reedcort:
## PR Summary

Allow partial component route updates via `rosa edit ingress --component-routes` — users can now set a single route (e.g. only console) without providing all routes.

The previous change (openshift/rosa#3376) followed the Classic CLI framework which required all routes to be set together (classic: oauth+console+downloads, HCP: console+downloads). The OCM API supports partial updates — setting only console leaves downloads unchanged — but the CLI rejected partial input with "the expected amount of component routes is N, but M have been supplied". After review feedback, the decision is to allow partial updates to match the API flexibility.

## Related Issues and PRs
- Jira: [ROSAENG-60614](https://redhat.atlassian.net/browse/ROSAENG-60614)
- Related PR: [openshift/rosa#3376](https://github.com/openshift/rosa/pull/3376) (merged — enabled componentRoutes for HCP)
- Related PR: [terraform-redhat/terraform-provider-rhcs#1187](https://github.com/terraform-redhat/terraform-provider-rhcs/pull/1187) (TF provider — also aligning to partial updates)

## Type of Change

## Previous Behavior

`rosa edit ingress --component-routes` required all routes for the platform — classic required oauth+console+downloads (3), HCP required console+downloads (2). Passing fewer routes returned:
```
ERR: the expected amount of component routes is 2, but 1 have been supplied
```

## Behavior After This Change

Users can pass any subset of valid routes. Name validation and duplicate detection are preserved.

```bash
# Set only console (HCP)
rosa edit ingress -c mycluster --component-routes 'console: hostname=console.example.com;tlsSecretRef=console-tls' <ingress-id>

# Set only downloads (HCP)
rosa edit ingress -c mycluster --component-routes 'downloads: hostname=downloads.example.com;tlsSecretRef=downloads-tls' <ingress-id>

# Set both (still works)
rosa edit ingress -c mycluster --component-routes 'console: hostname=console.example.com;tlsSecretRef=console-tls, downloads: hostname=downloads.example.com;tlsSecretRef=downloads-tls' <ingress-id>
```

Invalid names and duplicates are still rejected:
```
ERR: 'oauth' is not a valid component name. Expected include [console, downloads]
ERR: component route "console" was supplied more than once
```

### Preconditions
- A running ROSA HCP cluster with `hypershift-component-routes` feature toggle enabled
- HyperShift operator >= v0.1.79
- TLS secret in `openshift-config` on the hosted cluster

### Test Steps
1. Set only console: `rosa edit ingress -c <cluster> --component-routes 'console: hostname=console.example.com;tlsSecretRef=console-tls' <ingress-id>`
2. Verify console route updated, downloads unchanged
3. Set only downloads: `rosa edit ingress -c <cluster> --component-routes 'downloads: hostname=downloads.example.com;tlsSecretRef=downloads-tls' <ingress-id>`
4. Verify downloads route updated, console unchanged
5. Verify invalid name still rejected: `rosa edit ingress -c <cluster> --component-routes 'oauth: hostname=x;tlsSecretRef=y' <ingress-id>`
6. Verify duplicate still rejected: `rosa edit ingress -c <cluster> --component-routes 'console: hostname=a;tlsSecretRef=b, console: hostname=c;tlsSecretRef=d' <ingress-id>`

### Expected Results
- Partial updates succeed
- Invalid names and duplicates still rejected
- Existing routes not included in the update remain unchanged on the cluster

## Proof of the Fix

```
$ go test ./cmd/edit/ingress/... -v
Ran 14 of 14 Specs in 0.002 seconds
SUCCESS! -- 14 Passed | 0 Failed | 0 Pending | 0 Skipped
```

### Live cluster — partial updates and clearing

**1. Set console only:**
```
$ rosa edit ingress -c creed-hcp --component-routes 'console: hostname=console.custom.example.com;tlsSecretRef=console-tls-secret' o5n4
INFO: Updated ingress 'o5n4'
```
API state: `{"console": {"hostname": "console.custom.example.com", "tls_secret_ref": "console-tls-secret"}}`

**2. Add downloads (console unchanged):**
```
$ rosa edit ingress -c creed-hcp --component-routes 'downloads: hostname=downloads.custom.example.com;tlsSecretRef=downloads-tls-secret' o5n4
INFO: Updated ingress 'o5n4'
```
API state: `{"console": {...}, "downloads": {"hostname": "downloads.custom.example.com", "tls_secret_ref": "downloads-tls-secret"}}`

**3. Clear downloads only (console unchanged):**
```
$ rosa edit ingress -c creed-hcp --component-routes 'downloads: hostname=;tlsSecretRef=' o5n4
INFO: Updated ingress 'o5n4'
```
API state: `{"console": {"hostname": "console.custom.example.com", "tls_secret_ref": "console-tls-secret"}}`

## Breaking Changes
- [x] No breaking changes

- [x] `make test` passes.
- [x] `make lint` passes.
- [x] `make rosa` passes.

[ROSAENG-60614]: https://redhat.atlassian.net/browse/ROSAENG-60614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Documentation**
  * Updated the `component-routes` CLI flag help text to clarify HCP-supported components and the required `hostname` + `tlsSecretRef` pair per route.
* **Bug Fixes**
  * Improved HCP component-route validation to avoid failing solely due to a component count mismatch, while still accepting valid partial configurations.
* **Tests**
  * Expanded ingress component-route parsing tests to confirm correct output when only selected routes are provided and that empty parameter values are serialized as empty strings.

### DIFF
--- a/cmd/edit/ingress/cmd_test.go
+++ b/cmd/edit/ingress/cmd_test.go
@@ -17,10 +17,13 @@ limitations under the License.
 package ingress
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 )
 
 var _ = Describe("Parse component routes", func() {
@@ -95,15 +98,61 @@ var _ = Describe("Parse component routes", func() {
 				err.Error(),
 			).To(Equal("component route \"console\" was supplied more than once"))
 		})
-		It("When only one route is provided it should fail with wrong count", func() {
-			_, err := parseComponentRoutesForAllowed(
+		It("When only one route is provided it should succeed", func() {
+			result, err := parseComponentRoutesForAllowed(
 				"console: hostname=console-host;tlsSecretRef=console-secret",
 				expectedHcpComponentRoutes,
 			)
-			Expect(err).ToNot(BeNil())
-			Expect(
-				err.Error(),
-			).To(Equal("the expected amount of component routes is 2, but 1 have been supplied"))
+			Expect(err).To(BeNil())
+			Expect(result).To(HaveLen(1))
+			Expect(result).To(HaveKey("console"))
+		})
+		It("When a partial update is provided it should only include the specified route in the serialized payload", func() {
+			result, err := parseComponentRoutesForAllowed(
+				"console: hostname=console-host;tlsSecretRef=console-secret",
+				expectedHcpComponentRoutes,
+			)
+			Expect(err).To(BeNil())
+			ingress, err := cmv1.NewIngress().ComponentRoutes(result).Build()
+			Expect(err).To(BeNil())
+
+			var buf bytes.Buffer
+			err = cmv1.MarshalIngress(ingress, &buf)
+			Expect(err).To(BeNil())
+			var body map[string]interface{}
+			err = json.Unmarshal(buf.Bytes(), &body)
+			Expect(err).To(BeNil())
+
+			cr := body["component_routes"].(map[string]interface{})
+			Expect(cr).To(HaveLen(1))
+			Expect(cr).To(HaveKey("console"))
+			Expect(cr).ToNot(HaveKey("downloads"))
+			console := cr["console"].(map[string]interface{})
+			Expect(console["hostname"]).To(Equal("console-host"))
+			Expect(console["tls_secret_ref"]).To(Equal("console-secret"))
+		})
+		It("When empty values are provided it should serialize as a clear", func() {
+			result, err := parseComponentRoutesForAllowed(
+				"downloads: hostname=;tlsSecretRef=",
+				expectedHcpComponentRoutes,
+			)
+			Expect(err).To(BeNil())
+			ingress, err := cmv1.NewIngress().ComponentRoutes(result).Build()
+			Expect(err).To(BeNil())
+
+			var buf bytes.Buffer
+			err = cmv1.MarshalIngress(ingress, &buf)
+			Expect(err).To(BeNil())
+			var body map[string]interface{}
+			err = json.Unmarshal(buf.Bytes(), &body)
+			Expect(err).To(BeNil())
+
+			cr := body["component_routes"].(map[string]interface{})
+			Expect(cr).To(HaveLen(1))
+			Expect(cr).To(HaveKey("downloads"))
+			downloads := cr["downloads"].(map[string]interface{})
+			Expect(downloads["hostname"]).To(Equal(""))
+			Expect(downloads["tls_secret_ref"]).To(Equal(""))
 		})
 	})
 	Context("Fails to parse input string for component routes", func() {
@@ -117,15 +166,15 @@ var _ = Describe("Parse component routes", func() {
 				err.Error(),
 			).To(Equal("'unknown' is not a valid component name. Expected include [oauth, console, downloads]"))
 		})
-		It("fails due to wrong amount of component routes", func() {
-			_, err := parseComponentRoutes(
+		It("When partial classic routes are provided it should succeed", func() {
+			result, err := parseComponentRoutes(
 				//nolint:lll
 				"oauth: hostname=oauth-host;tlsSecretRef=oauth-secret,downloads: hostname=downloads-host;tlsSecretRef=downloads-secret",
 			)
-			Expect(err).ToNot(BeNil())
-			Expect(
-				err.Error(),
-			).To(Equal("the expected amount of component routes is 3, but 2 have been supplied"))
+			Expect(err).To(BeNil())
+			Expect(result).To(HaveLen(2))
+			Expect(result).To(HaveKey("oauth"))
+			Expect(result).To(HaveKey("downloads"))
 		})
 		It("fails if it can split ':' in more than one key separation", func() {
 			_, err := parseComponentRoutes(
--- a/cmd/edit/ingress/flags.go
+++ b/cmd/edit/ingress/flags.go
@@ -93,8 +93,10 @@ func addIngressV2Flags(flags *pflag.FlagSet) {
 		componentRoutesFlag,
 		"",
 		//nolint:lll
-		"Component routes settings. Available keys [oauth, console, downloads] (HCP clusters support console and downloads only). For each key a pair of hostname and tlsSecretRef is expected to be supplied. "+
-			"Format should be a comma separate list 'oauth: hostname=example-hostname;tlsSecretRef=example-secret-ref,downloads:...",
+		"Component route settings. Specify one or more routes to update; routes not specified remain unchanged. "+
+			"Available keys are [oauth, console, downloads] (HCP clusters support console and downloads only). "+
+			"Each route requires a hostname and tlsSecretRef. To clear a route, set both to empty values. "+
+			"Format: 'console: hostname=example-hostname;tlsSecretRef=example-secret-ref,downloads:...'",
 	)
 }
 
@@ -107,13 +109,6 @@ func parseComponentRoutesForAllowed(input string, allowedRoutes []string) (map[s
 	result := map[string]*cmv1.ComponentRouteBuilder{}
 	input = strings.TrimSpace(input)
 	components := strings.Split(input, ",")
-	if len(components) != len(allowedRoutes) {
-		return nil, fmt.Errorf(
-			"the expected amount of component routes is %d, but %d have been supplied",
-			len(allowedRoutes),
-			len(components),
-		)
-	}
 	transformations := []stringTransformation{
 		func(source string) string {
 			return strings.TrimSpace(source)
--- a/cmd/edit/ingress/cmd.go
+++ b/cmd/edit/ingress/cmd.go
@@ -489,8 +489,7 @@ func run(cmd *cobra.Command, argv []string) {
 	sameNamespaceOwnershipPolicy := (namespaceOwnershipPolicy == nil) ||
 		(curNamespaceOwnershipPolicy == ingress.RouteNamespaceOwnershipPolicy())
 
-	sameComponentRoutes := (componentRoutes == nil) ||
-		(reflect.DeepEqual(curComponentRoutes, ingress.ComponentRoutes()))
+	sameComponentRoutes := componentRoutes == nil || sparseComponentRoutesEqual(curComponentRoutes, componentRoutes)
 
 	if sameListeningMethod && sameRouteSelectors && sameLbType &&
 		sameExcludedNamespaces && sameWildcardPolicy && sameNamespaceOwnershipPolicy &&
@@ -507,4 +506,27 @@ func run(cmd *cobra.Command, argv []string) {
 		os.Exit(1)
 	}
 	r.Reporter.Infof("Updated ingress '%s' on cluster '%s'", ingress.ID(), clusterKey)
+}
+
+func sparseComponentRoutesEqual(
+	current map[string]*cmv1.ComponentRoute,
+	patch map[string]*cmv1.ComponentRouteBuilder,
+) bool {
+	for key, builder := range patch {
+		route, err := builder.Build()
+		if err != nil {
+			return false
+		}
+		cur, exists := current[key]
+		if !exists {
+			if route.Hostname() != "" || route.TlsSecretRef() != "" {
+				return false
+			}
+			continue
+		}
+		if cur.Hostname() != route.Hostname() || cur.TlsSecretRef() != route.TlsSecretRef() {
+			return false
+		}
+	}
+	return true
 }


amandahla:
@reedcort thanks for submitting the PR.

How is expected to delete an existing route, for example, I added console + download and now I want only console. What would be the flow?

@coderabbitai pinging you in case this already exists in the code, check. Also check if we have other commands working in a similar way (partial updates/keeps things unchanged).

reedcort:
@amandahla Assuming both console and download routes are set to remove a single route you just need to pass it with empty values.

`rosa edit ingress -c <cluster> --component-routes 'downloads: hostname=;tlsSecretRef=' <ingress-id>`

This clears downloads while leaving console unchanged.`

cmd/edit/ingress/flags.go
```diff
@@ -107,13 +107,6 @@ func parseComponentRoutesForAllowed(input string, allowedRoutes []string) (map[s
 	result := map[string]*cmv1.ComponentRouteBuilder{}
 	input = strings.TrimSpace(input)
 	components := strings.Split(input, ",")
-	if len(components) != len(allowedRoutes) {
```
amandahla, line 110: Could you update the flag message with this new behavior?

@coderabbitai suggest something.

amandahla:
@coderabbitai do we have a e2e test for component routes? Should we update something?

amandahla:
> @amandahla Assuming both console and download routes are set to remove a single route you just need to pass it with empty values.
> 
> `rosa edit ingress -c <cluster> --component-routes 'downloads: hostname=;tlsSecretRef=' <ingress-id>`
> 
> This clears downloads while leaving console unchanged.`

@reedcort 
Thanks for the clarification on partial updates and empty clears!

Could you paste a short before/after from a live cluster into the PR description (or a comment)? Something like `rosa describe ingress` (or the OCM/API equivalent you used) showing:

1. Both console + downloads set → update only console → downloads unchanged
2. Clear only one route with empty values → the other route unchanged

That would make the OCM merge behavior easy for reviewers to verify alongside the unit tests. Appreciate it!

amandahla:
@coderabbitai create an issue referencing this PR
2 main points: evaluate adding a new flag  --remove-component-route and everage standard key-value map representations (--component-route key=value or JSON/YAML payloads

Details from Cursor:
UX follow-up (non-blocking for this PR): consider an explicit remove path later.
Clearing today requires knowing the internal key shape and empty-value convention:
```bash
rosa edit ingress -c <cluster> \
  --component-routes 'downloads: hostname=;tlsSecretRef=' <ingress-id>
That is hard to discover, easy to mistype, and fragile in shells because the DSL uses ; (command separator) unless the whole value is quoted.

A clearer follow-up would be something like:

rosa edit ingress <ingress-id> -c <cluster> \
  --component-route name=console,hostname=console.example.com,tlsSecretRef=console-tls
rosa edit ingress <ingress-id> -c <cluster> \
  --remove-component-route downloads

Why this is better UX than empty keys:
Intent is explicit (remove vs update), so omit ≠ delete stays unambiguous.
Users don’t need to remember hostname / tlsSecretRef just to clear a route.
Avoids shell breakage from unquoted ; / trailing = in the clear path.
Matches other ROSA edit patterns where inverse actions are explicit (e.g. boolean toggles), rather than “pass empty  values”.

**Comment by amandahla** [comment]
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> 🧩 Analysis chain
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> set -euo pipefail
> 
> echo '=== Candidate test directories/files ==='
> fd -t f -i '(e2e|integration|acceptance|functional|test)' . \
>   | rg -i '(^|/)(e2e|integration|acceptance|functional|test)(/|_|\.|$)|ingress|route' \
>   | sort \
>   | head -300
> 
> echo '=== All component-route and ingress edit test references ==='
> rg -n -i -C 5 --glob '*_test.go' --glob '*.feature' --glob '*.yaml' --glob '*.yml' \
>   '(component-routes|component routes|ComponentRoutes|edit ingress|UpdateIngress|TlsSecretRef)' \
>   . 2>/dev/null || true
> 
> echo '=== Test framework/dependency clues ==='
> rg -n -i -C 3 --glob 'go.mod' --glob '*.go' --glob '*.feature' \
>   '(ginkgo|gomega|e2e|integration|acceptance|cucumber|godog)' \
>   . 2>/dev/null | head -400 || true
> 
> echo '=== Ingress command test layout and focused contents ==='
> ast-grep outline cmd/edit/ingress --items all 2>/dev/null || true
> sed -n '1,280p' cmd/edit/ingress/cmd_test.go 2>/dev/null || true
> 
> echo '=== Test-related CI/workflow entries ==='
> rg -n -i -C 3 --glob '.github/workflows/*' --glob 'Makefile' --glob 'makefile' \
>   '(e2e|integration|acceptance|functional)' \
>   . 2>/dev/null || true
> ```
> 
> Length of output: 57742
> 
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> set -euo pipefail
> 
> echo '=== E2E ingress test structure ==='
> ast-grep outline tests/e2e/test_rosacli_ingress.go --items all 2>/dev/null || true
> 
> echo '=== E2E ingress test: default and edit contexts ==='
> sed -n '1,540p' tests/e2e/test_rosacli_ingress.go
> 
> echo '=== CLI ingress helper definitions and use sites ==='
> rg -n -C 5 --glob '*.go' \
>   '(type .*Ingress|func \(.*\) (Edit|Describe|List|Create|Delete).*Ingress|EditIngress\(|DescribeIngress\()' \
>   tests 2>/dev/null || true
> 
> echo '=== Component-routes references specifically in executable E2E tests ==='
> rg -n -i -C 5 --glob '*.go' \
>   '(component-routes|ComponentRoutes|TlsSecretRef)' \
>   tests/e2e tests/utils/exec 2>/dev/null || true
> 
> echo '=== Ingress feature labels and E2E configuration guards ==='
> rg -n -C 4 --glob '*.go' \
>   '(Ingress|Hypershift|HCP|PrivateLink|STS)' \
>   tests/ci/labels tests/e2e/e2e_pre_check.go tests/e2e/e2e_setup_test.go tests/utils/config 2>/dev/null || true
> ```
> 
> Length of output: 51932
> 
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> set -euo pipefail
> 
> echo '=== E2E ingress reflected model and describe parsing ==='
> sed -n '52,112p' tests/utils/exec/rosacli/ingress_service.go
> 
> echo '=== Canonical component-routes describe output and assertions ==='
> ast-grep outline cmd/describe/ingress/cmd_test.go --items all 2>/dev/null || true
> sed -n '15,145p' cmd/describe/ingress/cmd_test.go
> 
> echo '=== Other E2E helpers that parse nested YAML maps ==='
> rg -n -C 6 --glob '*.go' \
>   '(YamlToMap\(\)|map\[string\].*yaml|yaml:".*Routes|yaml:".*Config)' \
>   tests/utils/exec tests/e2e 2>/dev/null | head -280 || true
> ```
> 
> Length of output: 32267
> 
> `@amandahla` Yes—there is an ingress E2E suite in `tests/e2e/test_rosacli_ingress.go`, but it currently has **no coverage for `--component-routes`**. The only coverage in this PR is parser-level in `cmd/edit/ingress/cmd_test.go`.
> 
> I recommend adding, at minimum, a command/request-level test in this PR that verifies a partial invocation—for example, only `console`—sends only the `console` entry in the ingress update payload. That directly protects the key behavior change: omitted routes must not be included or cleared by the CLI.
> 
> An E2E test would also be valuable, but needs a small amount of test infrastructure first:
> 
> * Extend `tests/utils/exec/rosacli/ingress_service.go`’s `Ingress` model to parse the nested `Component Routes` data returned by `rosa describe ingress`.
> * Add an HCP ingress scenario that:
>   
>   1. records both existing `console` and `downloads` routes,
>   2. updates only one route,
>   3. verifies that route changed while the omitted route is unchanged,
>   4. restores the original route.
> * Gate/skip it unless the E2E cluster has valid configurable component-route hostnames and TLS secret references. The current E2E profile/config does not provide those inputs.
> 
> So: **no existing E2E test; add the sparse-payload test now, and add the E2E scenario when the test environment can safely supply and restore valid route values.**

@jerichokeyne wdyt? should we add one?

File: cmd/edit/ingress/flags.go [hunk 2]
```diff
@@ -107,13 +109,6 @@ func parseComponentRoutesForAllowed(input string, allowedRoutes []string) (map[s
 	result := map[string]*cmv1.ComponentRouteBuilder{}
 	input = strings.TrimSpace(input)
 	components := strings.Split(input, ",")
```
olucasfreitas, line 111: Now that `componentRoutes` can be sparse, this full-map `DeepEqual` no longer matches the CLI contract. A console-only update against an ingress that already has unchanged `console` plus existing `downloads` will still PATCH and report success, because the omitted key makes the maps unequal.

cmd/edit/ingress/cmd_test.go
```diff
@@ -117,15 +116,15 @@ var _ = Describe("Parse component routes", func() {
...
 				"oauth: hostname=oauth-host;tlsSecretRef=oauth-secret,downloads: hostname=downloads-host;tlsSecretRef=downloads-secret",
 			)
-			Expect(err).ToNot(BeNil())
-			Expect(
-				err.Error(),
-			).To(Equal("the expected amount of component routes is 3, but 2 have been supplied"))
+			Expect(err).To(BeNil())
+			Expect(result).To(HaveLen(2))
+			Expect(result).To(HaveKey("oauth"))
+			Expect(result).To(HaveKey("downloads"))
```
olucasfreitas, line 177: These cases only show that parsing accepts fewer keys now; they don't prove the CLI sends the sparse PATCH body that this PR documents. Please add a request-level test around `run()`/`UpdateIngress` for a console-only update and an empty-value clear path.

cmd/edit/ingress/flags.go
olucasfreitas: The latest push addresses the original ask here: the flag help now explains sparse updates, HCP key restrictions, and the empty-value clear convention. I think this specific wording thread can resolve.

olucasfreitas:
The PR description now includes both cases this thread asked for: a console-only update that leaves downloads unchanged and a downloads clear with empty values that leaves console unchanged. That gives reviewers the backend-merge evidence needed for the behavior change.

olucasfreitas:
I wouldn't block this PR on a new live HCP E2E, but I would block on adding one focused command/runtime PATCH assertion now. The current diff only updates parser tests, so we still need a test that proves omitted routes are not serialized and empty values do serialize as a clear.

cmd/edit/ingress/cmd_test.go
reedcort: I added two serialization-level tests that build the parsed output through `cmv1.NewIngress().ComponentRoutes().Build() ` and verify the marshaled JSON via cmv1.MarshalIngress. Is a run()-level test still needed, or are these sufficient?

olucasfreitas:
Now that `componentRoutes` can be sparse, this full-map `DeepEqual` no longer matches the CLI contract. A console-only update against an ingress that already has unchanged `console` plus existing `downloads` will still PATCH and report success, because the omitted key makes the maps unequal.

cmd/edit/ingress/cmd_test.go
olucasfreitas: These new marshal-based cases are sufficient for the sparse/clear payload-shape concern, because `UpdateIngress` just sends the built ingress body. I wouldn't block on an extra `run()`-level test for that anymore; the remaining must-fix issue is separate in `cmd/edit/ingress/cmd.go`, where no-op detection still compares the sparse request map against the full current route map.

olucasfreitas:
Given the new serialization tests plus the live-cluster proof in the PR description, I don't think this PR needs a new HCP E2E before merge. A follow-up E2E is still worthwhile once `tests/utils/exec/rosacli/ingress_service.go` can parse `Component Routes` from `rosa describe ingress`.

cmd/edit/ingress/flags.go
reedcort: Addressed — added `sparseComponentRoutesEqual` to fix no-op detection for partial updates.

Verified on live cluster:

```
# Set console
$ rosa edit ingress --component-routes 'console: hostname=console.custom.example.com;tlsSecretRef=console-tls-secret' j3t8
INFO: Updated ingress 'j3t8'

# Same console again — no-op detected
$ rosa edit ingress --component-routes 'console: hostname=console.custom.example.com;tlsSecretRef=console-tls-secret' j3t8
WARN: No need to update ingress as there are no changes

# Different console value — update sent
$ rosa edit ingress --component-routes 'console: hostname=console2.custom.example.com;tlsSecretRef=console-tls-secret' j3t8
INFO: Updated ingress 'j3t8'
```

PR #3449, opened 2026-08-05
ROSAENG-62490 | refactor: Add lint rules to enforce environmental boundaries

markirish:
## PR Summary

Adds CI-enforced checks to prevent disallowed code patterns (e.g. calling os.Exit from `/pkg`), and prevent disallowed dependency directions (core layer -> CLI). Enforces architectural patterns that have been laid out in `guidelines/ARCHITECTURE.md` and `guidelines/refactor/pkg-architecture.md`

- Add depguard deny list to prohibit imports defined in ARCHITECTURE.md (reporter, output, interactive, arguments, color, debug, commands, rosa, options, aws/profile, aws/region, aws/commandbuilder)
- Add nolint:depguard annotations to pre-existing internal import violations in files already within whole-files lint scope
- Add nolint:forbidigo annotations to pre-existing fmt.Print and os.Exit violations in files already within whole-files lint scope
- Added nolintlint to .golangci.yml with require-specific: true and allow-unused: false, and fixed existing violations

As files are touched, `//nolint:forbidgo` and `//nolint:depguard` will have to be added until files are refactored as part of this initiative.

## Related Issues and PRs

- Jira: [ROSAENG-62490](https://redhat.atlassian.net/browse/ROSAENG-62490)
- Related design/docs: [ROSA+CLI+Refactor+CLI+to+enforce+layer+boundaries](https://redhat.atlassian.net/wiki/spaces/ROSACLITF/pages/432213369/ROSA+CLI+Refactor+CLI+to+enforce+layer+boundaries#Phase-1%3A-Define-and-Guard-the-Boundary)

## Type of Change

- [x] ci - changes CI pipelines, jobs, or automation workflows.

## Previous Behavior

Bad dependency imports and disallowed code patterns were allowed

## Behavior After This Change

Dependency imports as laid out in `ARCHITECTURE.md` and `refactor/pkg-architecture.md` are enforced. 

### Test Steps
1. Add an incorrect dependency import (e.g. importing `/cmd` in `/pkg`) or incorrect code smell (e.g. `/pkg` calling `os.Exit` or writing to `stdout`)
2. Try to run the linter or push the code

### Expected Results

Linter should catch the architectural violations

## Proof of the Fix

- Screenshots:
- Videos:
- Logs/CLI output:
- Other artifacts:

## Breaking Changes
- [x] No breaking changes

- [x] `make test` passes.
- [x] `make lint` passes.
- [x] `make rosa` passes.

[ROSAENG-62490]: https://redhat.atlassian.net/browse/ROSAENG-62490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Chores**
  * Strengthened static analysis rules to detect prohibited dependencies, direct console output, and process exits.
  * Standardized lint configuration and suppression formatting across command, AWS, cluster, machine pool, and helper components.
* **Refactor**
  * Updated retry-message handling and streamlined validation test messages.
  * Removed obsolete lint directives from application and end-to-end test code.
  * No user-facing functionality, public APIs, or runtime behavior changed.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -21,6 +21,8 @@ run:
 linters:
   default: none
   enable:
+    - depguard
+    - forbidigo
     - goconst
     - govet
     - ineffassign
@@ -30,6 +32,48 @@ linters:
     - unconvert
     - unused
   settings:
+    depguard:
+      rules:
+        core-no-cli-deps:
+          deny:
+            - pkg: "github.com/spf13/cobra"
+              desc: "pkg/ is core layer -- must not import cobra"
+            - pkg: "github.com/spf13/pflag"
+              desc: "pkg/ is core layer -- must not import pflag"
+            - pkg: "github.com/AlecAivazis/survey"
+              desc: "pkg/ is core layer -- must not import survey"
+            - pkg: "github.com/openshift/rosa/pkg/reporter"
+              desc: "pkg/ is core layer -- must not import CLI presenter reporter"
+            - pkg: "github.com/openshift/rosa/pkg/output"
+              desc: "pkg/ is core layer -- must not import CLI presenter output"
+            - pkg: "github.com/openshift/rosa/pkg/interactive"
+              desc: "pkg/ is core layer -- must not import CLI presenter interactive"
+            - pkg: "github.com/openshift/rosa/pkg/arguments"
+              desc: "pkg/ is core layer -- must not import CLI presenter arguments"
+            - pkg: "github.com/openshift/rosa/pkg/color"
+              desc: "pkg/ is core layer -- must not import CLI presenter color"
+            - pkg: "github.com/openshift/rosa/pkg/debug"
+              desc: "pkg/ is core layer -- must not import CLI presenter debug"
+            - pkg: "github.com/openshift/rosa/pkg/commands"
+              desc: "pkg/ is core layer -- must not import CLI presenter commands"
+            - pkg: "github.com/openshift/rosa/pkg/rosa"
+              desc: "pkg/ is core layer -- must not import CLI runtime rosa"
+            - pkg: "github.com/openshift/rosa/pkg/options"
+              desc: "pkg/ is core layer -- must not import CLI presenter options"
+            - pkg: "github.com/openshift/rosa/pkg/aws/profile"
+              desc: "pkg/ is core layer -- must not import CLI presenter aws/profile"
+            - pkg: "github.com/openshift/rosa/pkg/aws/region"
+              desc: "pkg/ is core layer -- must not import CLI presenter aws/region"
+            - pkg: "github.com/openshift/rosa/pkg/ocm/output"
+              desc: "pkg/ is core layer -- must not import CLI presenter ocm/output"
+            - pkg: "github.com/openshift/rosa/pkg/aws/commandbuilder"
+              desc: "pkg/ is core layer -- must not import CLI presenter aws/commandbuilder"
+    forbidigo:
+      forbid:
+        - pattern: "^fmt\\.Print(f|ln)?$"
+          msg: "pkg/ must not write to stdout -- return data and let the caller render"
+        - pattern: "^os\\.Exit$"
+          msg: "pkg/ must not call os.Exit -- return an error instead"
     goconst:
       ignore-tests: true
       match-constant: false
@@ -51,6 +95,14 @@ linters:
       - path: tests/utils/
         linters:
           - goconst
+      - path-except: "^pkg/"
+        linters:
+          - depguard
+          - forbidigo
+      - path: "^pkg/(arguments|color|commands|debug|interactive|output|reporter|rosa|options|test|aws/(commandbuilder|profile|region)|ocm/output)/"
+        linters:
+          - depguard
+          - forbidigo
 issues:
   max-issues-per-linter: 0
   max-same-issues: 0
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -21,15 +21,69 @@ run:
 linters:
   default: none
   enable:
+    - depguard
+    - forbidigo
     - goconst
     - govet
     - ineffassign
     - lll
     - misspell
+    - nolintlint
     - staticcheck
     - unconvert
     - unused
   settings:
+    depguard:
+      rules:
+        core-no-cli-deps:
+          deny:
+            - pkg: "github.com/spf13/cobra"
+              desc: "core layer must not import cobra"
+            - pkg: "github.com/spf13/pflag"
+              desc: "core layer must not import pflag"
+            - pkg: "github.com/AlecAivazis/survey"
+              desc: "core layer must not import survey"
+            - pkg: "github.com/openshift/rosa/cmd"
+              desc: "core layer must not import cmd/"
+            - pkg: "github.com/openshift/rosa/internal/cli"
+              desc: "core layer must not import internal/cli/"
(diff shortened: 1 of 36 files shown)


.golangci.yml
```diff
@@ -51,6 +95,14 @@ linters:
       - path: tests/utils/
         linters:
           - goconst
+      - path-except: "^pkg/"
```
olucasfreitas: This enforces part of the Phase 1 boundary, but it still won't fail a new core import of `cmd/...` or `internal/cli/...`, and the rule set is scoped away from any future `internal/core` code. Can we close that gap here so the lint policy matches the architecture contract we're linking in the PR?

pkg/aws/client.go
```diff
@@ -959,14 +959,14 @@ func (c *awsClient) ValidateAccessKeys(AccessKey *AccessKey) error {
...
-				logger.Debug(fmt.Printf("InvalidClientTokenId, waited %.2f\n", waited.Seconds()))
+				logger.Debug(fmt.Printf("InvalidClientTokenId, waited %.2f\n", waited.Seconds())) //nolint:forbidigo
 				time.Sleep(wait)
 			}
 
 			if awserr.IsAccessDeniedException(err) {
 				wait := time.Duration((i * 200)) * time.Millisecond
 				waited := time.Since(start)
-				logger.Debug(fmt.Printf("AccessDenied, waited %.2f\n", waited.Seconds()))
+				logger.Debug(fmt.Printf("AccessDenied, waited %.2f\n", waited.Seconds())) //nolint:forbidigo
```
olucasfreitas: These two allowlisted calls are doing more than violating the new boundary: `fmt.Printf` writes to stdout and returns `(int, error)`, so `Debug` logs the byte count while also leaking output. I'd fix these now rather than codifying them as `forbidigo` exceptions.

markirish: Good call, fixed in latest change

.golangci.yml
markirish: Gap now closed

amandahla:
@markirish 

1. I'm concerned about the forbidigo rule on fmt.Fprint*
   Writing to a strings.Builder / io.Writer is normal Go and not a layer violation.
@coderabbitai short sentence confirming if this is true.

2. Adding //nolint:depguard and //nolint:forbidigo across pkg/ might be hard to track and hard to shrink over time.
   What about a central list in .golangci.yml (path exclusions) for known existing violations?
   We can keep line-level //nolint only for rare exceptions.

markirish:
@amandahla 
 
 > 1. I'm concerned about the forbidigo rule on fmt.Fprint*
Writing to a strings.Builder / io.Writer is normal Go and not a layer violation.
coderabbitai short sentence confirming if this is true.

This is a fair critique, as it does catch buffer writes that _are not_ layer violations. Unfortunately, `forbidigo` is limited and can't inspect argument types. Going through the violations, it appears there are 35 violations of this type, but only 7 that will remain after the refactor (writes to `strings.Builder` and `io.Writer`). I think having 7 permanent justified `//nolint: forbidigo` exceptions in the code to prevent `fmt.Fprintf` calls to `stdout` is a good trade-off.

> 2. Adding //nolint:depguard and //nolint:forbidigo across pkg/ might be hard to track and hard to shrink over time.
What about a central list in .golangci.yml (path exclusions) for known existing violations?
We can keep line-level //nolint only for rare exceptions.

I think a central list of files in `.golangci.yml` might actually be the worst of both worlds:
- Trackability and the ability to shrink the list is actually _less_ than with `//nolint` directives, as we don't have a complete picture of all of the violations in a file unless we remove the file and run the linter (although there is an [internal doc tracking all of the violations found](https://redhat.atlassian.net/wiki/spaces/ROSACLITF/pages/442958849/Layer+Violation+List)). In comparison, with the `//nolint` apprach, a simple repo-level or directory-level search (either in an IDE or grep) should find any and all violations in the code _at the call site_ that need to be handled.
- The current `//nolint` exceptions are self-removing, as the directives will be cleaned up as the call site is refactored.
- Having a file-level exception enables new violations to be added to that file until all violations in the file have been cleaned up. For some small files, this would be fine, but for large files the violations will be handled piecemeal over many phases, and in that time new violations can be added in. With `//nolint`, new violations added will immediately be flagged and cannot sneak in.
- The [`.golangci.yml` Gold Standard document](https://docs.google.com/document/d/19koqqPFSZnkBRonQPf4T5Y6_U58-ZN1pB9_SrTRNYFs/edit?tab=t.0) mentions using the `nolintlint` linter, which flags stale `//nolint` directives that are no longer needed. This would ensure that the list shrinks and vestigial directives don't persist after their usefulness.
 
 Code Rabbit says:
 > Do not use broad pkg/ exclusions. They would also hide new violations in those paths.
 
 But this problem will also be experienced with file-level exclusions: Any new violations added to that file would be un-tracked, and have to be handled when a file has been completely refactored, instead of preventing it in the first place.
 
 @coderabbitai, Given that `forbidigo` cannot inspect argument types, and file-level path exclusions would suppress new violations in the same file until all existing violations are resolved, is line-level `//nolint` the more precise suppression mechanism for known false positives? Weigh the positives and negatives of such an approach.

amandahla:
@markirish good finding on the nolintlint, could you add it? Do you know if will work for all kinds of //nolint (depguard specially)?

@coderabbitai check that, give us a short answer.

markirish:
@amandahla, I have added `nolintlint`. There were a number of violations that were encountered, so those have also been fixed:

- Added nolintlint to .golangci.yml with `require-specific: true` and `allow-unused: false`
- Fixed 6 spacing issues (// nolint to //nolint) across cmd/ and pkg/
- Removed 13 stale nolint directives across test files and pkg/ocm/helpers.go, cmd/edit/ingress/flags.go, cmd/dlt/oidcconfig/cmd.go, pkg/interactive/validation.go, tests/utils/handler/cluster_handler.go
- Added 2 specific nolint directives where bare //nolint was suppressing real issues: //nolint:lll for the long regex in pkg/aws/helpers.go, //nolint:staticcheck for De Morgan's law in pkg/interactive/validation.go and capitalized error in tests/utils/handler/cluster_handler.go
- Added URL comment exclusion in .golangci.yml (source: "^// https?://" for lll) since URL comments can't be wrapped and //nolint can't be appended to comment lines

markirish:
> Enable nolintlint for the new suppression.

>The PR objectives state that nolintlint is not enabled. CI can therefore accept an unused or undocumented //nolint:staticcheck directive. Enable the required nolintlint checks and run lint for this and the other normalized suppressions.

@coderabbitai, `nolintlint` should be enabled. Can you recheck the whole diff? The linter passes on my branch.

markirish:
> One issue remains in .golangci.yml Line 84. Add require-explanation: true.

> require-explanation defaults to false. The current configuration therefore accepts undocumented //nolint:depguard and //nolint:forbidigo directives. This includes the new call-site exceptions in this PR.

@coderabbitai, the blast radius for that would be huge. 132 `//nolint` directives across the codebase without explanations. With whole-files: true, any developer who touches a file containing one of these would have to add an explanation before their PR can pass lint. That's a significant burden to introduce alongside the other changes in this PR. Once the refactor is done, I agree it should be added, but for now, when we are trying to identify the issues, it would add so much overhead that it won't be worth it.

amandahla:
Nice!

/approve
/lgtm

PR #3452, opened 2026-08-06
ROSAENG-63260 | refactor: Create reusable request/response conventions

markirish:
- Create a document outlining how requests from CLI to core layers should be structured
- Outline how responses from core layer to CLI should be structured
- Implement a minimal request/response workflow when creating an IAMServiceAccount

## Related Issues and PRs

- Jira: [OCM-XXXXX](https://issues.redhat.com/browse/OCM-XXXXX) or [ROSAENG-XXXX](https://issues.redhat.com/browse/ROSAENG-XXXX)
- Fixes: `#`
- Related PR(s):
- Related design/docs:

## Type of Change

- [x] docs - updates documentation only.

### Test Steps
1.
2.
3.

## Proof of the Fix

- Screenshots:
- Videos:
- Logs/CLI output:
- Other artifacts:

## Breaking Changes
- [ ] No breaking changes

- [ ] Commit subject/title follows `[JIRA-TICKET] | [TYPE]: <MESSAGE>`.
- [ ] PR description clearly explains both **what** changed and **why**.
- [ ] Relevant Jira/GitHub issues and related PRs are linked.
- [ ] `make test` passes.
- [ ] `make lint` passes.
- [ ] `make rosa` passes.

## Summary by CodeRabbit

* **New Features**
  * Added IAM service-account creation workflows with generated roles, trust policies, managed and inline policies, tags, and GovCloud support.
  * Added validation and structured results for IAM service-account operations.

* **Documentation**
  * Documented workflow conventions for requests, results, validation, lifecycle boundaries, optional values, and domain types.
  * Updated architecture guidance and documentation indexes.

* **Tests**
  * Added comprehensive coverage for creation, policy handling, GovCloud support, validation, and error scenarios.

### DIFF
--- a/guidelines/workflow-conventions.md
+++ b/guidelines/workflow-conventions.md
@@ -1,0 +1,456 @@
+# Workflow Request and Result Conventions
+
+## Scope
+
+Use this file when designing or modifying reusable workflow functions in the
+core layer (`pkg/`). It defines the type conventions for data that crosses
+the boundary between the CLI layer and core layer, as described in
+[ARCHITECTURE.md](ARCHITECTURE.md#boundary-rules).
+
+## Lifecycle
+
+A ROSA CLI operation follows this data flow:
+
+```
+ User input          CLI Layer                 Core Layer
+ ───────────    ───────────────────    ─────────────────────────
+ flags/args  →  parse, prompt,     →  workflow function receives
+                resolve defaults,     Request, validates, executes,
+                build Request         returns (Result, error)
+                                                    │
+                                                    ▼
+                                      CLI formats Result for
+                                      display (table/JSON/YAML)
+```
+
+The **Request** and **Result** types are the boundary contract. The CLI layer
+constructs a Request from resolved user input. The core layer returns a Result
+(or a domain type) with an error. Neither type carries CLI, terminal, or
+presentation concerns.
+
+## Request Types
+
+### Naming
+
+Name request types `<Verb><Resource>Request`:
+
+```go
+type CreateMachinePoolRequest struct { ... }
+type EditIngressRequest struct { ... }
+type UpgradeClusterRequest struct { ... }
+```
+
+For list and describe operations with few parameters, a dedicated request type
+is still preferred over loose parameters. It keeps the workflow function
+signature stable when new filters or options are added later:
+
+```go
+type ListMachinePoolsRequest struct {
+    ClusterID string
+}
+
+type DescribeKubeletConfigRequest struct {
+    ClusterID string
+    Name      string
+}
+```
+
+### What a Request Contains
+
+A Request carries the **resolved domain values** needed to perform the
+operation. Every field should pass the core-layer test: would this value
+be meaningful to a TUI, a headless program, or a REST API?
+
+Typical contents:
+
+- Identifiers: cluster ID, resource name, account ID.
+- Domain parameters: instance type, replica count, labels, taints.
+- Configuration values: autoscaling bounds, CIDR ranges, version strings.
+- Behavioral flags with domain meaning: `DryRun bool`, `BestEffort bool`.
+
+### What a Request Must Not Contain
+
+A Request must never carry:
+
+- `*cobra.Command`, `*pflag.FlagSet`, or any Cobra/pflag type.
+- Reporter, spinner, logger, or terminal-state references.
+- Interactive prompt state (`interactive.Enabled()`, survey validators).
+- Output format preferences (`--output json`).
+- The `rosa.Runtime` struct (it bundles CLI lifecycle with client access).
+- Display strings, ANSI codes, or pre-formatted output.
+
+If a workflow needs an AWS or OCM client, accept the client interface as a
+separate parameter to the workflow function, not as a field on the Request.
+
+### Construction
+
+The CLI layer is responsible for constructing a fully resolved Request. This
+means:
+
+1. Parse flags and positional arguments.
+2. Run interactive prompts (when interactive mode is enabled).
+3. Apply CLI-side defaults for any value the user did not provide.
+4. Look up dependent values (e.g., resolve a cluster name to a cluster ID).
+5. Build the Request with all fields populated.
+
+The core layer never constructs a Request from CLI inputs. It receives one
+ready to use.
+
+```go
+// cmd/create/machinepool/cmd.go (CLI layer)
+func run(cmd *cobra.Command, args []string) {
+    // ... resolve flags, prompt, look up cluster ...
+
+    request := machinepool.CreateMachinePoolRequest{
+        ClusterID:    clusterID,
+        Name:         name,
+        InstanceType: instanceType,
+        Replicas:     replicas,
+    }
+    result, err := machinepoolService.CreateMachinePool(ctx, ocmClient, request)
+    // ... handle error, format result for display ...
+}
+```
+
+### Optional Values and Zero-Value Behavior
+
+Use Go's type system to communicate whether a field is required or optional:
+
+| Situation | Convention |
+|-----------|-----------|
+| Required field | Use the value type directly (`string`, `int`). Validation rejects the zero value only if the domain disallows it. |
+| Optional field | Use a pointer (`*string`, `*int`). `nil` means "not specified." |
+| A group of fields is all-or-nothing | Use an embedded struct pointer. `nil` means the group is absent. |
+
+The pointer convention makes optionality visible at the type level without
+requiring the reader to check comments, validation code, or domain rules.
+A value type means the caller must always provide the field, but it cannot
+distinguish "the caller omitted this" from "the caller explicitly supplied
+the zero value." When zero is not a valid domain value (an empty cluster ID,
+a zero replica count where at least one is required), validation should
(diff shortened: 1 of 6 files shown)


pkg/iamserviceaccount/create.go
```diff
@@ -0,0 +1,206 @@
...
+
+	// AWS account context, resolved by the CLI layer via GetCreator.
+	AccountID  string
+	Partition  string
+	IsGovcloud bool
+}
+
+// Validate checks that the request contains all required fields and that
+// service account names and namespaces are syntactically valid.
+func (r *CreateIAMServiceAccountRequest) Validate() error {
```
amandahla, line 55: @coderabbitai Does CreateIAMServiceAccountRequest.Validate() cover the same cases the CLI currently checks (e.g. ARN format validation is still CLI-side today)?

pkg/iamserviceaccount/client.go
```diff
@@ -0,0 +1,46 @@
...
+package iamserviceaccount
+
+import (
+	"context"
+)
+
+// IAMServiceAccountClient defines the IAM operations needed by service
+// account workflows.
+//
+// This is a narrow interface over the subset of AWS IAM operations that
```
amandahla, line 24: I didnt understand this comment.

Today we have
```
roleARN, err := r.AWSClient.EnsureRole(
			r.Reporter,
			roleName,
			trustPolicy,
			userOptions.PermissionsBoundary,
			"",
			tags,
			userOptions.Path,
			managedPolicies,
		)
```
The ideia is to change for iamServiceAccountclient.EnsureRole(ctx,roleName,...etc)?

@coderabbitai help me understand this comment + documentation

guidelines/workflow-conventions.md
```diff
@@ -0,0 +1,366 @@
...
+ flags/args  →  parse, prompt,     →  workflow function receives
+                resolve defaults,     Request, validates, executes,
+                build Request         returns (Result, error)
+                                                    │
+                                                    ▼
+                                      CLI formats Result for
+                                      display (table/JSON/YAML)
+```
+
+The **Request** and **Result** types are the boundary contract. The CLI layer
```
amandahla, line 26: Given that CLI must support two different runners/api (v1 and v2'Hyperfleet') where both can have same request but different fields/flags/validation, how could we define that?
The verb and resource is the same but behavior will be different.

pkg/iamserviceaccount/create.go
```diff
@@ -0,0 +1,206 @@
...
+	// AttachedPolicyARNs lists the managed policies that were attached to the role.
+	AttachedPolicyARNs []string
+
+	// InlinePolicyName is the name of the inline policy that was attached to
+	// the role. Empty when no inline policy was requested.
+	InlinePolicyName string
+}
+
+// Service is the default implementation of IAMServiceAccountService.
+type Service struct{}
```
amandahla, line 125: As we add list/describe/delete, do you see these living as methods on one Service, or as separate workflow functions in pkg/iamserviceaccount (create/list/delete each with their own Request/Result)?

I'm concerned about Service becoming a catch-all for every IAM SA operation. Narrow reader/writer interfaces per workflow might scale better.

@coderabbitai evaluate my review, my reference is https://refactoring.guru/design-patterns/facade

amandahla: @coderabbitai I’m less worried about whether today’s Service is literally a Facade, and more about the direction as list/describe/delete move into pkg/: one type accumulating every IAM SA operation starts to look like an overgrown Facade / god object (refactoring.guru calls out facades growing too broad). I’d prefer we establish one workflow function + Request/Result per operation, and narrow client interfaces per workflow (reader vs writer), rather than growing IAMServiceAccountService. Not asking to solve that in this PR, mainly want alignment before the next extraction.

pkg/iamserviceaccount/create.go
amandahla: Something to keep in mind once the command call this.

pkg/iamserviceaccount/create.go
amandahla: @markirish wdyt?

pkg/iamserviceaccount/create.go
```diff
@@ -0,0 +1,206 @@
...
+	// AWS account context, resolved by the CLI layer via GetCreator.
+	AccountID  string
+	Partition  string
+	IsGovcloud bool
+}
+
+// Validate checks that the request contains all required fields and that
+// service account names and namespaces are syntactically valid.
+func (r *CreateIAMServiceAccountRequest) Validate() error {
+	if r.ClusterName == "" {
```
amandahla, line 56: out of scope but wouldnt be amazing be able to replace this for something like
```
var errs []error
for _, validate := range createValidators {
    if err := validate(r); err != nil {
        errs = append(errs, err)
    }
}
return errors.Join(errs...)
```

pkg/iamserviceaccount/client.go
amandahla: @markirish I think it would be nice clarify that.

guidelines/workflow-conventions.md
markirish: @amandahla good question. I believe the plan is for v1 and v2 to live in fully separate directory/runner trees, each with their own codebase. In that case this isn't actually a conflict: `CreateIAMServiceAccountRequest` in each directory/package is just a normal Go type, i.e. `v1.CreateIAMServiceAccountRequest` vs. `v2.CreateIAMServiceAccountRequest`. Each tree follows the Request/Result conventions independently, so different fields/flags/validation per version is expected and fine.

The one thing to nail down as this gets built out: if v1 and v2 are meant to share any domain types or client helpers (rather than duplicating them), those should live in their own shared package that both trees import. My thought is that there should be no sharing, just to make it easier to maintain both until v1 is deprecated.

I added a short section to `workflow-conventions.md` ("Multiple Backends for the Same Verb and Resource").

pkg/iamserviceaccount/create.go
markirish: @amandahla agreed with your concern, and with CodeRabbit's take. As list, describe, and delete get added, I don't want IAMServiceAccountService turning into a catch all interface that every operation gets bolted onto. The plan is one workflow function per operation, each with its own Request and Result, the same pattern create.go already uses, rather than growing one shared Service type.

On the client interface side, I actually think we should maybe go a step further than reader versus writer. Instead of one shared reader interface for list and describe, and one shared writer interface for create and delete, each workflow gets its own client interface scoped to exactly what it calls. A describe workflow depends on its own interface, a delete workflow depends on its own interface, and neither one grows just because the other needs a new method. Reader and writer groupings still couple unrelated workflows together once you have more than one read or one write operation, so going fully per workflow avoids that without adding real cost.

Service can stay as an implementation detail for the current create workflow since there is nothing shared to coordinate yet. We would only introduce a shared Service or a shared interface later if multiple workflows genuinely need to be swapped or mocked together, not just because they touch the same resource.

I wrote this up in `workflow-conventions.md` under a new section on scoping client interfaces per workflow, so it is there for whoever picks up list or describe next.

pkg/iamserviceaccount/create.go
markirish: I think the next PR I will send as part of this phase deals with errors, I will definitely incorporate this!

markirish:
@amandahla ready for re-review

pkg/iamserviceaccount/create.go
amandahla: Amazing :)

guidelines/workflow-conventions.md
amandahla: Perfect, following the Kubernetes API Versioning Model is an excellent idea here.

However, I'm concerned about extracting shared code into a common package. Given that v1 is temporary and both versions will evolve differently, changes to one won't necessarily apply to the other. Keeping v1 completely stable is definitely worth the code duplication: it isolates risk and avoids premature abstractions. Plus, once v1 is removed, we won't be left with a 'shared' package that is only used by v2.
wdyt?

pkg/iamserviceaccount/create.go
amandahla: Strongly agree 👍

guidelines/workflow-conventions.md
markirish: I agree, there should be no sharing for the reasons you listed.

guidelines/workflow-conventions.md
```diff
@@ -0,0 +1,456 @@
...
+func run(cmd *cobra.Command, args []string) {
+    // ... resolve flags, prompt, look up cluster ...
+
+    request := machinepool.CreateMachinePoolRequest{
+        ClusterID:    clusterID,
+        Name:         name,
+        InstanceType: instanceType,
+        Replicas:     replicas,
+    }
+    result, err := machinepoolService.CreateMachinePool(ctx, ocmClient, request)
```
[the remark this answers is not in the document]
markirish: Re-evaluate, should all be passed by value now

amandahla (approved):
/approve
/lgtm

PR #3462, opened 2026-08-12
ROSAENG-6150 | fix: local zone and wavelength zone subnets no longer listed when creating machine pool

markirish:
## PR Summary

Exclude local zone and wavelength zone subnets from HCP machinepool creation

When creating a machinepool for a ROSA HCP cluster interactively, the subnet picker listed all private subnets in the VPC, including subnets in AWS Local Zones and Wavelength Zones. HCP clusters do not support nodepools in these zone types, so presenting them as options was misleading and could lead to failed machinepool creation.

The cluster creation path (cmd/create/cluster/cmd.go) already filtered these out using GetAvailabilityZoneType, but the machinepool creation path in pkg/machinepool/helper.go did not. This change adds the same filtering to both getSubnetOptions (interactive subnet list) and getSubnetFromAvailabilityZone (AZ-based subnet selection) for HCP clusters. Classic clusters are unaffected since they can legitimately have machinepools in local zones. Excluded subnets are reported to the user via a warning message, and an explicit error is returned if no subnets remain after filtering.

## Related Issues and PRs

- Jira: [OCM-XXXXX](https://issues.redhat.com/browse/OCM-XXXXX) or [ROSAENG-XXXX](https://issues.redhat.com/browse/ROSAENG-XXXX)
- Fixes: `#`
- Related PR(s):
- Related design/docs:

## Type of Change

## Previous Behavior

When creating a machine pool in an HCP cluster, all subnets were listed, including those in local zones and wavelength zones.

## Behavior After This Change

Local zones and wavelength zones are now filtered out when creating machine pools in HCP clusters 

### Preconditions
- A ROSA HCP cluster in a VPC that has both standard-AZ private subnets and local zone or wavelength zone private subnets (e.g., us-east-1-atl-1a or us-east-1-wl1-atl-wlz-1)
  - AWS credentials configured with ec2:DescribeAvailabilityZones permission

### Test Steps
1. Run rosa create machinepool --cluster=<cluster-name> -i
2. When prompted "Select subnet for a hosted machine pool", answer Yes
3. Observe the subnet list

### Expected Results
- Local zone and wavelength zone subnets should NOT appear in the list
- A warning should be printed: WARN: The following subnets were excluded because they are on local zone or wavelength zone: <subnet-ids>
- Only standard availability zone subnets should be selectable

## Proof of the Fix
Old:
```
$ rosa create machinepool --cluster=mirish-lz-test -i
? Machine pool name: mirish-lz-test-mp
? Image Type (optional, choose 'Skip' to skip selection; ): Skip
? OpenShift version (default = '4.22.10'): 4.22.10
? Select subnet for a hosted machine pool: Yes
? Subnet ID (default = 'subnet-05d65c034cb18edda ('','vpc-0a03b5960423372ed','us-west-2-den-1a', Owner ID: '090777400063')'):  [Use arrows to move, type to filter, ? for more help]
> subnet-05d65c034cb18edda ('','vpc-0a03b5960423372ed','us-west-2-den-1a', Owner ID: '090777400063')
  subnet-04100f1e4b123d4db ('','vpc-0a03b5960423372ed','us-west-2-lax-1a', Owner ID: '090777400063')
  subnet-02403e8fc2d7c454e ('mirish-lz-test-Private-Subnet-1','vpc-0a03b5960423372ed','us-west-2a', Owner ID: '090777400063')
  ```

New
```
$ ./rosa create machinepool --cluster=mirish-lz-test -i
? Machine pool name: mirish-lz-test-mp
? Machine pool image type (optional, choose 'Skip' to skip selection; ): Skip
? OpenShift version (default = '4.22.10'): 4.22.10
? Select subnet for a hosted machine pool: Yes
W: The following subnets were excluded because they are on local zone or wavelength zone: [subnet-04100f1e4b123d4db, subnet-05d65c034cb18edda]
? Subnet ID (default = 'subnet-02403e8fc2d7c454e ('mirish-lz-test-Private-Subnet-1','vpc-0a03b5960423372ed','us-west-2a', Owner ID: '090777400063')'):  [Use arrows to move, type to filter, ? for more help]
> subnet-02403e8fc2d7c454e ('mirish-lz-test-Private-Subnet-1','vpc-0a03b5960423372ed','us-west-2a', Owner ID: '090777400063')
```

## Breaking Changes
- [x] No breaking changes

- [x] `make test` passes.
- [x] `make lint` passes.
- [x] `make rosa` passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Hosted Control Plane clusters now exclude Local Zone and Wavelength Zone subnets when selecting private subnets.
  * Clear errors are shown when no eligible standard subnets remain.
  * Warnings identify subnets excluded from Hosted Control Plane selection.
  * Subnet selection and availability-zone resolution consistently use eligible subnets.
  * Classic clusters continue to support all available subnets.
  * Node pool creation now handles availability-zone details consistently during subnet selection.

### DIFF
--- a/pkg/machinepool/helper.go
+++ b/pkg/machinepool/helper.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/openshift/rosa/pkg/aws"
 	"github.com/openshift/rosa/pkg/fedramp"
+	"github.com/openshift/rosa/pkg/helper"
 	mpHelpers "github.com/openshift/rosa/pkg/helper/machinepools"
 	"github.com/openshift/rosa/pkg/helper/versions"
 	"github.com/openshift/rosa/pkg/interactive"                               //nolint:depguard
@@ -140,10 +141,59 @@ func getSubnetFromUser(cmd *cobra.Command, r *rosa.Runtime, isSubnetSet bool,
 	return subnet, nil
 }
 
+func filterOutNonStandardZoneSubnets(
+	r *rosa.Runtime, subnets []ec2types.Subnet,
+) ([]ec2types.Subnet, error) {
+	var filtered []ec2types.Subnet
+	var excludedSubnetIds []string
+	checkedZones := map[string]string{}
+
+	for _, subnet := range subnets {
+		az := awssdk.ToString(subnet.AvailabilityZone)
+		zoneType, seen := checkedZones[az]
+		if !seen {
+			var err error
+			zoneType, err = r.AWSClient.GetAvailabilityZoneType(az)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get availability zone type for '%s': %w", az, err)
+			}
+			checkedZones[az] = zoneType
+		}
+		if zoneType == aws.LocalZone || zoneType == aws.WavelengthZone {
+			excludedSubnetIds = append(excludedSubnetIds, awssdk.ToString(subnet.SubnetId))
+		} else {
+			filtered = append(filtered, subnet)
+		}
+	}
+
+	if len(excludedSubnetIds) > 0 {
+		r.Reporter.Warnf("The following subnets were excluded because they are on local zone"+
+			" or wavelength zone: %s", helper.SliceToSortedString(excludedSubnetIds))
+	}
+	return filtered, nil
+}
+
+func getFilteredPrivateSubnets(r *rosa.Runtime, cluster *cmv1.Cluster) ([]ec2types.Subnet, error) {
+	privateSubnets, err := r.AWSClient.GetVPCPrivateSubnets(cluster.AWS().SubnetIDs()[0])
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve private subnets: %w", err)
+	}
+	if cluster.Hypershift().Enabled() {
+		privateSubnets, err = filterOutNonStandardZoneSubnets(r, privateSubnets)
+		if err != nil {
+			return nil, err
+		}
+		if len(privateSubnets) == 0 {
+			return nil, fmt.Errorf("no private subnets available after excluding local zone" +
+				" and wavelength zone subnets")
+		}
+	}
+	return privateSubnets, nil
+}
+
 // getSubnetOptions gets one of the cluster subnets and returns a slice of formatted VPC's private subnets.
 func getSubnetOptions(r *rosa.Runtime, cluster *cmv1.Cluster) ([]string, error) {
-	// Fetch VPC's subnets
-	privateSubnets, err := r.AWSClient.GetVPCPrivateSubnets(cluster.AWS().SubnetIDs()[0])
+	privateSubnets, err := getFilteredPrivateSubnets(r, cluster)
 	if err != nil {
 		return nil, err
 	}
@@ -302,7 +352,7 @@ func spotMaxPriceValidator(val interface{}) error {
 func getSubnetFromAvailabilityZone(cmd *cobra.Command, r *rosa.Runtime, isAvailabilityZoneSet bool,
 	cluster *cmv1.Cluster, args *mpOpts.CreateMachinepoolUserOptions) (string, error) {
 
-	privateSubnets, err := r.AWSClient.GetVPCPrivateSubnets(cluster.AWS().SubnetIDs()[0])
+	privateSubnets, err := getFilteredPrivateSubnets(r, cluster)
 	if err != nil {
 		return "", err
 	}
--- a/pkg/machinepool/helper_test.go
+++ b/pkg/machinepool/helper_test.go
@@ -2,6 +2,8 @@ package machinepool
 
 import (
 	"fmt"
+	"io"
+	"os"
 
 	"go.uber.org/mock/gomock"
 
@@ -245,7 +247,7 @@ var _ = Describe("getSubnetFromAvailabilityZone functionality", func() {
 
 	BeforeEach(func() {
 		mockClient = mock.NewMockClient(gomock.NewController(GinkgoT()))
-		r = &rosa.Runtime{AWSClient: mockClient}
+		r = &rosa.Runtime{AWSClient: mockClient, Reporter: reporter.CreateReporter()}
 		cmd = &cobra.Command{}
 		az = "us-east-1a"
 		subnetId1 = "subnet-123"
@@ -257,6 +259,7 @@ var _ = Describe("getSubnetFromAvailabilityZone functionality", func() {
 		BeforeEach(func() {
 			privateSubnets = []ec2types.Subnet{{AvailabilityZone: &az, SubnetId: &subnetId1}}
 			mockClient.EXPECT().GetVPCPrivateSubnets(gomock.Any()).Return(privateSubnets, nil)
+			mockClient.EXPECT().GetAvailabilityZoneType(az).Return("availability-zone", nil)
 			cluster = MockCluster(func(c *cmv1.ClusterBuilder) {
 				c.State(cmv1.ClusterStateReady)
 				b := cmv1.HypershiftBuilder{}
@@ -278,6 +281,7 @@ var _ = Describe("getSubnetFromAvailabilityZone functionality", func() {
 			args.AvailabilityZone = "us-west-1a"
 			privateSubnets = []ec2types.Subnet{{AvailabilityZone: &az, SubnetId: &subnetId1}}
 			mockClient.EXPECT().GetVPCPrivateSubnets(gomock.Any()).Return(privateSubnets, nil)
+			mockClient.EXPECT().GetAvailabilityZoneType(az).Return("availability-zone", nil)
 			cluster = MockCluster(func(c *cmv1.ClusterBuilder) {
 				c.State(cmv1.ClusterStateReady)
 				b := cmv1.HypershiftBuilder{}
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: pkg/machinepool/helper_test.go [hunk 1]
```diff
@@ -346,11 +380,87 @@ var _ = Describe("getSubnetOptions", func() {
...
+
+	When("VPC has local zone and wavelength zone subnets for an HCP cluster", func() {
+		var localZoneAz, wavelengthAz string
+
+		BeforeEach(func() {
+			localZoneAz = "us-east-1-atl-1a"
+			wavelengthAz = "us-east-1-wl1-atl-wlz-1"
+			localSubnetId := "subnet-local"
+			wavelengthSubnetId := "subnet-wavelength"
+			allSubnets := append(privateSubnets,
```
olucasfreitas: append(privateSubnets, ...) can mutate the outer BeforeEach slice if it has spare capacity. Consider allocating a fresh slice to avoid cross-test contamination.

olucasfreitas:
Subnet IDs are not sensitive data — they're infrastructure identifiers, not secrets or PII. The cluster creation path (cmd/create/cluster/cmd.go:4508) already logs excluded subnet IDs in the same way. This check appears to be a false positive. The user needs to see which subnets were excluded to understand the filtering behavior.

olucasfreitas:
The existing machinepool test file doesn't use assertion messages on Expect() calls — the new tests follow the established style in this package. Adding assertion messages would be inconsistent with the surrounding code. The gomock controller cleanup via GinkgoT() is automatic. These are non-blocking warnings and consistent with existing patterns.

pkg/machinepool/helper.go
```diff
@@ -137,10 +138,59 @@ func getSubnetFromUser(cmd *cobra.Command, r *rosa.Runtime, isSubnetSet bool,
...
+		r.Reporter.Warnf("The following subnets were excluded because they are on local zone"+
+			" or wavelength zone: %s", helper.SliceToSortedString(excludedSubnetIds))
+	}
+	return filtered, nil
+}
+
+func getFilteredPrivateSubnets(r *rosa.Runtime, cluster *cmv1.Cluster) ([]ec2types.Subnet, error) {
+	privateSubnets, err := r.AWSClient.GetVPCPrivateSubnets(cluster.AWS().SubnetIDs()[0])
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
olucasfreitas: Addressed in 06273ce — the error is now wrapped with fmt.Errorf and %w. Looks good.

pkg/machinepool/helper_test.go
```diff
@@ -346,11 +378,82 @@ var _ = Describe("getSubnetOptions", func() {
...
+
+		It("excludes local zone and wavelength zone subnets", func() {
+			subnetOptions, err := getSubnetOptions(r, cluster)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(subnetOptions).To(HaveLen(2))
+			for _, opt := range subnetOptions {
+				Expect(opt).ToNot(ContainSubstring("subnet-local"))
+				Expect(opt).ToNot(ContainSubstring("subnet-wavelength"))
+			}
+		})
```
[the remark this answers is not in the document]
olucasfreitas: Addressed in 1f80acc — the test now captures stderr and asserts the warning message content. Resolved.

pkg/machinepool/helper.go
```diff
@@ -137,10 +138,59 @@ func getSubnetFromUser(cmd *cobra.Command, r *rosa.Runtime, isSubnetSet bool,
...
+	checkedZones := map[string]string{}
+
+	for _, subnet := range subnets {
+		az := awssdk.ToString(subnet.AvailabilityZone)
+		zoneType, seen := checkedZones[az]
+		if !seen {
+			var err error
+			zoneType, err = r.AWSClient.GetAvailabilityZoneType(az)
+			if err != nil {
+				return nil, err
```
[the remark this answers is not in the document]
olucasfreitas: Addressed in 1f80acc — the error is now wrapped with context including the AZ name. Resolved.

amandahla:
@coderabbitai compare with a old PR that should fix same issue but for some reason it was closed instead of merged:
https://github.com/openshift/rosa/pull/1939

Give me a bullet list of differences/what could be used/not used, if any.

amandahla:
Mark, nice approach. Could you just take a look if Coderabbit finding makes sense?
```
Potentially reusable part from PR #1939

The batch DescribeAvailabilityZones strategy is the main reusable part.
If performance becomes important, add a bulk availability-zone-type method to the current AWS abstraction and preserve the current explicit Local Zone and Wavelength Zone exclusion rule.
This is an optimization. It is not required to fix this issue.
```

I'm not sure how much should we be concerned about performance/aws call in this filter.

markirish:
@amandahla I talked to Claude about it, and it came up with the following:
>I'd recommend not pulling in the batched DescribeAvailabilityZones optimization from #1939.
> Reasoning: our filterOutNonStandardZoneSubnets already dedupes API calls via checkedZones (pkg/machinepool/helper.go:146-158), so the only calls that happen are one-per-distinct-AZ, not one-per-subnet. For the realistic case — a handful of standard AZs plus maybe one or two opted-in local/wavelength zones in a single VPC — that's a handful of calls, not a meaningful bottleneck. Batching would require adding a new bulk method to the aws.Client interface (plus regenerating mocks) for a fix that's intentionally scoped to the machinepool path, which is exactly the kind of interface-widening #1939 got dinged for elsewhere in this same analysis. It's also explicitly called out as "an optimization, not required to fix this issue."

I think that all makes sense with me, and I traced the code to confirm. I think for now, the performance should be as good as it can be in most situations, and maybe makes an extra call or two in other situations. For me, that sounds well-engineered when it also means much greater simplicity in the code.

pkg/machinepool/helper_test.go
markirish: Good call out, fixed!

pkg/machinepool/helper.go
```diff
@@ -140,10 +138,59 @@ func getSubnetFromUser(cmd *cobra.Command, r *rosa.Runtime, isSubnetSet bool,
...
+	var filtered []ec2types.Subnet
+	var excludedSubnetIds []string
+	checkedZones := map[string]string{}
+
+	for _, subnet := range subnets {
+		az := awssdk.ToString(subnet.AvailabilityZone)
+		zoneType, seen := checkedZones[az]
+		if !seen {
+			var err error
+			zoneType, err = r.AWSClient.GetAvailabilityZoneType(az)
```
amandahla, line 156: I think the main point of the performance review is related to this line. Would be possible make one call getting all zone descriptions and then iterate using this information instead of calling AWS client for every subnet?
`describeAvailabilityZonesOutput, err := c.ec2Client.DescribeAvailabilityZones(
		context.Background(),
		&ec2.DescribeAvailabilityZonesInput{ZoneNames: subnetAvailabilityZonesList})` 

@coderabbitai confirm if makes sense changing this loop.

pkg/machinepool/helper_test.go
```diff
@@ -293,34 +297,62 @@ var _ = Describe("getSubnetFromAvailabilityZone functionality", func() {
...
+			localZoneAz := "us-east-1-atl-1a"
+			localSubnetId := "subnet-local"
+			privateSubnets = []ec2types.Subnet{
+				{AvailabilityZone: &az, SubnetId: &subnetId1},
+				{AvailabilityZone: &localZoneAz, SubnetId: &localSubnetId},
+			}
+			mockClient.EXPECT().GetVPCPrivateSubnets(gomock.Any()).Return(privateSubnets, nil)
+			mockClient.EXPECT().GetAvailabilityZoneType(az).Return("availability-zone", nil)
+			mockClient.EXPECT().GetAvailabilityZoneType(localZoneAz).Return("local-zone", nil)
+			cluster = MockCluster(func(c *cmv1.ClusterBuilder) {
```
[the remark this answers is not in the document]
markirish: This simply isn't true. This file has a pre-existing dot import AND qualified import. It compiles fine. It _should_ be cleaned up, but it is beyond the scope of this PR, and this redundancy exists in other test files as well (` cmd/create/autoscaler/cmd_test.go`, `cmd/edit/autoscaler/cmd_test.go`, `pkg/machinepool/helper_test/go`). Will create a JIRA ticket and fix in another PR.

pkg/machinepool/helper.go
amandahla: I see now, agreed.

amandahla:
@coderabbitai is ci/prow/test failing due changes in this PR?

pkg/machinepool/machinepool_test.go
```diff
@@ -1547,6 +1547,7 @@ var _ = Describe("NodePools", func() {
 
 			t.ApiServer.AppendHandlers(RespondWithJSON(http.StatusOK, test.FormatVersionList([]*cmv1.Version{versionObj})))
 			mockClient.EXPECT().GetVPCPrivateSubnets(gomock.Any()).Return(privateSubnets, nil)
+			mockClient.EXPECT().GetAvailabilityZoneType(az).Return("availability-zone", nil)
```
olucasfreitas, line 1584: CreateNodePools now calls GetAvailabilityZoneType via getFilteredPrivateSubnets for HCP clusters. These spot/capacity-reservation tests mock GetVPCPrivateSubnets but not GetAvailabilityZoneType, which is why ci/prow/test fails with unexpected gomock calls at helper.go:156. Please add the AZ-type mock (or a shared setup helper) for every test in this context that hits subnet selection.

pkg/machinepool/helper.go
```diff
@@ -137,10 +138,59 @@ func getSubnetFromUser(cmd *cobra.Command, r *rosa.Runtime, isSubnetSet bool,
 	return subnet, nil
```
olucasfreitas, line 141: When --subnet is passed non-interactively, this returns args.Subnet without running getFilteredPrivateSubnets. Is that intentional for this PR, or should flag-supplied subnets get the same local/wavelength zone rejection as the interactive picker?

pkg/machinepool/helper_test.go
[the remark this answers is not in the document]
olucasfreitas: No action needed here — the dot import exposes MockCluster and CI build/lint pass. Duplicate-import cleanup is reasonable follow-up work outside this PR's scope.

pkg/machinepool/helper.go
olucasfreitas: Agree with keeping the per-AZ cached loop for this PR. checkedZones already dedupes to one call per distinct AZ, which matches realistic VPC sizes. Batching would widen the aws.Client interface and mock surface for an optimization that's not required to fix ROSAENG-6150. Happy to track a separate follow-up if we measure this as a bottleneck.

pkg/machinepool/helper_test.go
olucasfreitas: Confirmed fixed in 3632e — the test now allocates a fresh slice before appending local/wavelength subnets. This thread can stay resolved.

olucasfreitas:
CodeRabbit now marks this check as passed, and the cluster-creation path uses the same subnet-ID warning pattern. No further change needed for this PR.

pkg/machinepool/machinepool_test.go
markirish: > Already covered. I checked every GetVPCPrivateSubnets(gomock.Any()) call site in machinepool_test.go (22 total). 20 of them now have the matching GetAvailabilityZoneType(...) expectation right after — the 6 I added last turn plus the ones already present from earlier in the branch. The remaining 2 (lines 756 and 781, in "Tests getSubnetFromAvailabilityZone" and "Tests error case for getSubnetFromAvailabilityZone") build their own cmv1.NewCluster() directly without setting Hypershift(...), so cluster.Hypershift().Enabled() is false there — getFilteredPrivateSubnets (helper.go) only calls GetAvailabilityZoneType when the cluster is HCP, so those two correctly don't need the mock. Every test that uses returnMockCluster() (which does set Hypershift().Enabled(true)) already has the AZ-type expectation.

> I didn't introduce a shared setup helper — the file's existing convention is per-test inline mockClient.EXPECT() setup (no shared builder for the other mocks either, e.g. GetSubnetAvailabilityZone/IsLocalAvailabilityZone are duplicated the same way), so adding one mock line per test keeps it consistent rather than introducing a new abstraction for a single call.

> Full suite: go test ./pkg/machinepool/... -count=1 → ok, 259/259 specs pass.

pkg/machinepool/helper.go
markirish: Will create a follow-up ticket for this for next sprint. It seems like a true bug, but outside the scope of this PR and I don't think it can be a fast followup

markirish:
@amandahla @olucasfreitas ready for re-review, everything passing and CodeRabbit is satisfied

PR #3460, opened 2026-08-11
ROSAENG-64850 | rosa-day1-negative-f7 | fix(e2e): align registry asserts and retry AWS region flakes

amandahla:
## PR Summary

Align registry-config error assertions with lowercased CLI output (#3419), migrate OCP-76396 checks to `helper.ExpectErrorWithMessage`, and retry transient AWS regions/credentials failures in `CreateDryRun` so day1-negative specs reach their intended validation errors.

## Detailed Description of the Issue

Two E2E stability issues in `rosa-day1-negative-f7`:

1. **Registry config error casing** — PR #3419 lowercased registry-config validation messages in the CLI, but OCP-76396 / OCP-77149 assertions still expected title-case strings.
2. **AWS region/credential flakes** — `CreateDryRun` calls `GetRegionList` early; transient `Failed to retrieve AWS regions` / credential validation errors can fail negative specs before they reach the validation error under test (e.g. OCP-66761).

## Related Issues and PRs

- Jira: [ROSAENG-64850](https://issues.redhat.com/browse/ROSAENG-64850), [OCP-76396](https://issues.redhat.com/browse/OCP-76396)
- Related PR(s): #3419 (lowercased registry-config errors)

## Type of Change

- [x] fix - resolves incorrect behavior or test flakes.
- [ ] feat - adds a new user-facing capability.
- [ ] docs - updates documentation only.
- [ ] style - formatting or naming changes with no logic impact.
- [ ] refactor - code restructuring with no behavior change.
- [ ] test - adds or updates tests only.
- [ ] chore - maintenance work (tooling, housekeeping, non-product code).
- [ ] build - changes build system, packaging, or dependencies for build output.
- [ ] ci - changes CI pipelines, jobs, or automation workflows.
- [ ] perf - improves performance without changing intended behavior.

## Previous Behavior

- OCP-76396 / OCP-77149 failed on registry-config assertion casing mismatches.
- OCP-76396 used raw `ContainSubstring` on command output for error checks.
- `CreateDryRun` returned immediately on transient AWS regions/credentials failures.

## Behavior After This Change

- Registry error expectations use lowercase strings; OCP-76396 uses `helper.ExpectErrorWithMessage` (case-insensitive) for all checks in that test case.
- `CreateDryRun` retries up to 3 times (10s delay) on transient AWS regions/credentials errors via `RetryOnAWSRegionsCredentialError`.
- Unit tests cover the transient classifier and retry boundary.

### Preconditions

- AWS credentials and OCM token configured
[...]

PR #3472, opened 2026-08-17
rosa-day1-negative-f7 | fix(day1-negative): OCP-73725 OCP-73755 OCP-38857

amandahla:
## PR Summary
Align three `day1-negative` e2e specs (OCP-73725, OCP-73755, OCP-38857) with current ROSA CLI validation so periodic `rosa-day1-negative-f7` reaches the intended negative assertions.

Periodic `periodic-ci-openshift-rosa-master-e2e-rosa-day1-negative-f7` (`(day1-negative)&&!Exclude`, profile `null`) has been chronically red. Prow artifacts showed these specs failing before their target validation:

- **OCP-73725**: Generated HCP cluster names exceeded the 15-character shared-VPC limit, so the CLI failed on name length instead of network-type validation.
- **OCP-73755**: Pinning OpenShift 4.14.x with `--channel-group=stable` made OCM reject the create before the external-auth version-gate message could be asserted.
- **OCP-38857**: `rosa create cluster` without `--dry-run` did not surface flag/CIDR errors on the path the spec expects.

## Related Issues and PRs
- Jira: [ROSAENG-62387](https://issues.redhat.com/browse/ROSAENG-62387)
- Fixes: N/A
- Related PR(s): #3429 (OCP-45509), #3460 (OCP-76396 registry asserts)
- Related design/docs: N/A

## Type of Change

## Previous Behavior
- OCP-73725 failed on shared-VPC cluster name length.
- OCP-73755 failed on channel/version mismatch (`stable-4.21` vs pinned 4.14.x).
- OCP-38857 failed when `Create()` did not return errors or overlap checks hit live AWS paths.

## Behavior After This Change
- OCP-73725 uses a cluster name trimmed to ≤15 characters.
- OCP-73755 drops `--channel-group` and sets `--channel=eus-4.16` for the 4.14 negative path.
- OCP-38857 passes `--dry-run` on all `Create()` calls in the invalid-CIDR `It`.

No ROSA CLI product behavior changes.

### Preconditions
- OCM and AWS credentials for e2e (or rely on periodic `rosa-day1-negative-f7`).
- Label filter: `(day1-negative)&&!Exclude`.

### Test Steps
1. Run e2e with focus on OCP-73725, OCP-73755, and OCP-38857 (or the full day1-negative filter).
2. Confirm each `It` fails the CLI on the intended validation, not setup/channel/name-length errors.
3. Watch periodic `periodic-ci-openshift-rosa-master-e2e-rosa-day1-negative-f7` after merge.

### Expected Results
- OCP-73725 asserts network-type / `--no-cni` errors.
- OCP-73755 asserts external-auth version gate on 4.14.x.
- OCP-38857 asserts invalid/overlap CIDR messages via dry-run.

## Proof of the Fix
- Screenshots: N/A
- Videos: N/A
- Logs/CLI output: Prow job-history for `rosa-day1-negative-f7` (pre-fix failure signatures for these IDs).
- Other artifacts: N/A

## Breaking Changes
- [x] No breaking changes

N/A

- [x] `make test` passes.
- [x] `make lint` passes.
- [x] `make rosa` passes.

[ROSAENG-62387]: https://redhat.atlassian.net/browse/ROSAENG-62387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Tests**
  * Updated cluster creation validation coverage to use dry-run checks.
  * Refined HCP network validation test naming.
  * Improved external authentication validation coverage for older OpenShift versions.

### DIFF
--- a/tests/e2e/test_rosacli_cluster.go
+++ b/tests/e2e/test_rosacli_cluster.go
@@ -2124,7 +2124,7 @@ var _ = Describe("Create cluster with invalid options will",
 					"--pod-cidr":     "10111.0.0.0/16",
 				}
 				for flag, invalidValue := range illegalCIDRMap {
-					output, _, err := clusterService.Create(clusterName,
+					output, err := clusterService.CreateDryRun(clusterName,
 						flag, invalidValue,
 					)
 					Expect(err).To(HaveOccurred())
@@ -2133,7 +2133,7 @@ var _ = Describe("Create cluster with invalid options will",
 							invalidValue, flag, invalidValue))
 				}
 				By("Check the overlapped CIDR block validation")
-				output, _, err := clusterService.Create(clusterName,
+				output, err := clusterService.CreateDryRun(clusterName,
 					"--service-cidr", "1.0.0.0/16",
 					"--pod-cidr", "1.0.0.0/16",
 				)
@@ -2148,7 +2148,7 @@ var _ = Describe("Create cluster with invalid options will",
 					"--pod-cidr":     "1.0.0.0/28",
 				}
 				for flag, invalidValue := range invalidCIDRMap {
-					output, _, err := clusterService.Create(clusterName,
+					output, err := clusterService.CreateDryRun(clusterName,
 						flag, invalidValue,
 					)
 					Expect(err).To(HaveOccurred())
@@ -2167,7 +2167,7 @@ var _ = Describe("Create cluster with invalid options will",
 
 				}
 				By("Check the invalid machine CIDR for multi az")
-				output, _, err = clusterService.Create(clusterName,
+				output, err = clusterService.CreateDryRun(clusterName,
 					"--machine-cidr", "2.0.0.0/25",
 					"--multi-az",
 				)
@@ -2176,7 +2176,7 @@ var _ = Describe("Create cluster with invalid options will",
 					ContainSubstring("The allowed block size must be between a /16 netmask and /24"))
 
 				By("Check illegal host prefix")
-				output, _, err = clusterService.Create(clusterName,
+				output, err = clusterService.CreateDryRun(clusterName,
 					"--machine-cidr", "2.0.0.0/25",
 					"--host-prefix", "28",
 				)
@@ -2185,7 +2185,7 @@ var _ = Describe("Create cluster with invalid options will",
 					ContainSubstring("Subnet length should be between 23 and 26"))
 
 				By("Check invalid host prefix")
-				output, _, err = clusterService.Create(clusterName,
+				output, err = clusterService.CreateDryRun(clusterName,
 					"--machine-cidr", "2.0.0.0/25",
 					"--host-prefix", "invalid",
 				)
@@ -2706,7 +2706,7 @@ var _ = Describe("HCP cluster creation negative testing",
 		It("create HCP cluster with network type validation can work well via rosa cli - [id:73725]",
 			labels.Medium, labels.Runtime.Day1Negative,
 			func() {
-				clusterName := helper.GenerateRandomName("cluster-73725", 2)
+				clusterName := helper.GenerateRandomName("ocp-73725", 2)
 				By("Create HCP cluster with --no-cni and \"--network-type={OVNKubernetes, OpenshiftSDN}\" at the same time")
 				replacingFlags := map[string]string{
 					"-c":              clusterName,
@@ -2895,9 +2895,10 @@ var _ = Describe("HCP cluster creation negative testing",
 					"--cluster-name":  clusterName,
 					"--domain-prefix": clusterName,
 					"--version":       foundVersion,
-					"--channel-group": cg,
 				}
 				rosalCommand.ReplaceFlagValue(replacingFlags)
+				err = rosalCommand.DeleteFlag("--channel-group", true)
+				Expect(err).ToNot(HaveOccurred())
 				if !rosalCommand.CheckFlagExist("--external-auth-providers-enabled") {
 					rosalCommand.AddFlags("--external-auth-providers-enabled")
 				}


tests/e2e/test_rosacli_cluster.go
```diff
@@ -2706,7 +2712,7 @@ var _ = Describe("HCP cluster creation negative testing",
 		It("create HCP cluster with network type validation can work well via rosa cli - [id:73725]",
 			labels.Medium, labels.Runtime.Day1Negative,
 			func() {
-				clusterName := helper.GenerateRandomName("cluster-73725", 2)
+				clusterName := helper.TrimNameByLength(helper.GenerateRandomName("c73725", 2), 15)
```
jerichokeyne: I don't think `helper.TrimNameByLength` is needed here. The second argument to `helper.GenerateRandomName` specifies how many extra characters to add already, so the result should always have an exact known length already (length of first arg + 1 + second arg, since it returns something like `{prefix}-{n random characters}`)

tests/e2e/test_rosacli_cluster.go
```diff
@@ -2895,9 +2901,11 @@ var _ = Describe("HCP cluster creation negative testing",
 					"--cluster-name":  clusterName,
 					"--domain-prefix": clusterName,
 					"--version":       foundVersion,
-					"--channel-group": cg,
 				}
 				rosalCommand.ReplaceFlagValue(replacingFlags)
+				err = rosalCommand.DeleteFlag("--channel-group", true)
+				Expect(err).ToNot(HaveOccurred())
+				rosalCommand.AddFlags("--channel", "eus-4.16")
```
jerichokeyne: I really don't like the idea of hard coding a channel here. We really need to implement a helper function that will take in a version number and return the `availableChannels` info from `/api/clusters_mgmt/v1/versions/{version id}` instead

I've also noticed at least for now that just removing the `--channel` argument also works (it at least in my local testing), but I'm not sure if `--channel` is being planned to become a required argument or not (the fact that we let customers skip it in interactive mode, leads me to think that it won't be a required value for a while)

tests/e2e/test_rosacli_cluster.go
```diff
@@ -2125,6 +2125,7 @@ var _ = Describe("Create cluster with invalid options will",
 				}
 				for flag, invalidValue := range illegalCIDRMap {
 					output, _, err := clusterService.Create(clusterName,
+						"--dry-run",
```
jerichokeyne: Minor/non-blocker comment: it might make sense to use the `CreateDryRun` [function](https://github.com/openshift/rosa/blob/master/tests/utils/exec/rosacli/cluster_service.go#L31) instead. That also adds the `"--mode=auto", "--yes"` flags to the commands too, so you'd have to make sure that they work properly after the change still. Those don't usually break negative tests, but very occasionally I've seen it happen

jerichokeyne (changes requested):
Mostly just a few small changes I'd recommend, but I'd like to rethink the fix for OCP-73725. I've tried to prototype some fixes for selecting the correct channel for old OCP versions a couple times, but haven't had time to finish finding a clean solution yet, but I might be able to find those and send them your way if you want.

tests/e2e/test_rosacli_cluster.go
olucasfreitas: This is the one part of the fix that still looks brittle: we dynamically select a 4.14.x version, then hard-code `eus-4.16` just to get past earlier validation. Can we derive the channel from `availableChannels` for `foundVersion` instead so the test stays aligned with the OCM contract it is exercising?

tests/e2e/test_rosacli_cluster.go
olucasfreitas: `GenerateRandomName("c73725", 2)` is already deterministic and only 9 characters long, so the shorter prefix alone fixes the 15-character shared-VPC/domain-prefix limit. I’d treat the extra trim as optional cleanup, not a blocker.

tests/e2e/test_rosacli_cluster.go
olucasfreitas: I agree this is the part that still needs work. Even if `eus-4.16` happens to work today, it couples the test to a literal channel mapping instead of deriving whatever OCM currently advertises for the selected `foundVersion`.

tests/e2e/test_rosacli_cluster.go
olucasfreitas: I don't think this needs a code change here: `CreateDryRun()` also injects `--mode=auto --yes`, so it's not command-equivalent to adding only `--dry-run`. Keeping `Create(..., "--dry-run", ...)` preserves the current negative-path coverage while avoiding the live create path.

tests/e2e/test_rosacli_cluster.go
amandahla: What do you recommend for this test? Removing the --channel?
I think a helper to be used by another tests could be a further PR, out of scope of this one.

amandahla: I'll close this one in favor of https://github.com/openshift/rosa/pull/3472#discussion_r3798173327 since same approach is being discussed.

tests/e2e/test_rosacli_cluster.go
amandahla: Re-openinig since I didnt notice the CreateDryRun function and makes sense for consistency, good one.

tests/e2e/test_rosacli_cluster.go
amandahla: @coderabbitai create an issue for creating the helper as mentioned:
https://github.com/openshift/rosa/pull/3472#discussion_r3798173327
https://github.com/openshift/rosa/pull/3472#pullrequestreview-4953820125

for now, I removed the --channel

tests/e2e/test_rosacli_cluster.go
amandahla: CreateDryRun in place now.

tests/e2e/test_rosacli_cluster.go
amandahla: GenerateRandomName is in place now, nice.

PR #3482, opened 2026-08-21
ROSAENG-63262 | refactor: separate validation concerns and classify workflow errors for the CLI boundary

markirish:
## PR Summary
Establishes a layered validation and error-classification pattern for reusable workflows: Request.Validate() aggregates every domain violation via a flat validator list, and a shared `pkg/errors.ValidationError` type lets callers distinguish invalid input from operational failures with errors.As. Codifies error translation at the CLI boundary and demonstrates the full pattern end-to-end in the iamserviceaccount create workflow.

As part of the refactor project, the following stories need to be addressed:

- [**[ROSAENG-63262](https://redhat.atlassian.net/browse/ROSAENG-63262)**:  Separate invocation validation from domain validation](https://redhat.atlassian.net/browse/ROSAENG-63262)
- [**[ROSAENG-63263](https://redhat.atlassian.net/browse/ROSAENG-63263)**: Introduce distinguishable workflow errors](https://redhat.atlassian.net/browse/ROSAENG-63263)
- [**[ROSAENG-63264](https://redhat.atlassian.net/browse/ROSAENG-63264)**: Define error translation at the CLI boundary](https://redhat.atlassian.net/browse/ROSAENG-63264)

This PR accomplishes the following:
- **[ROSAENG-63262](https://redhat.atlassian.net/browse/ROSAENG-63262)**: Refactor C`reateIAMServiceAccountRequest.Validate()` into independently-testable validator functions joined once via errors.Join, so every violation is reported instead of only the first
- **[ROSAENG-63262](https://redhat.atlassian.net/browse/ROSAENG-63262)**: Document the CLI-layer/domain-layer validation split and the validator-list pattern in `guidelines/workflow-conventions.md`
- **[ROSAENG-63263](https://redhat.atlassian.net/browse/ROSAENG-63263)**: Add shared `pkg/errors.ValidationError{Field, Message, Err}` so callers can distinguish invalid-input failures from operational failures via `errors.As`, without parsing error text
- **[ROSAENG-63263](https://redhat.atlassian.net/browse/ROSAENG-63263)**: Document criteria for typed errors vs. sentinels, and the "Classifying validation errors" pattern, in `guidelines/workflow-conventions.md`
- **[ROSAENG-63263](https://redhat.atlassian.net/browse/ROSAENG-63263)**: Add domain and CLI tests proving validation failures are distinguishable from operational (AWS) failures via `errors.As`, using `pkg/iamserviceaccount` as the representative workflow
- **[ROSAENG-63264](https://redhat.atlassian.net/browse/ROSAENG-63264)**: Document core/CLI error-translation responsibilities (wrapping, typed errors, exit codes, remediation, diagnostics) in new `guidelines/error-conventions.md`
- **[ROSAENG-63264](https://redhat.atlassian.net/browse/ROSAENG-63264)**: Wire `cmd/create/iamserviceaccount` to call `pkg/iamserviceaccount.Service.CreateIAMServiceAccount` instead of calling `aws.Client` inline, showing command usage when the error is a validation failure

## Related Issues and PRs

- Jira: [ROSAENG-63262](https://issues.redhat.com/browse/ROSAENG-63262), [ROSAENG-63263](https://issues.redhat.com/browse/ROSAENG-63263), [ROSAENG-63264](https://issues.redhat.com/browse/ROSAENG-63264)
- Fixes: `#`
- Related PR(s):
- Related design/docs:

## Type of Change

## Previous Behavior
- rosa create iamserviceaccount validated inline in cmd.go, calling aws.Client directly: cluster/STS check, "at least one service account name," "at least one policy or inline policy," ARN format via `aws.ARNValidator` (message included the literal ARN string), and "role name required for multiple service accounts." Each check returned on the first failure.
- No validation existed for: Kubernetes service account/namespace name format (RFC 1123), inline policy JSON validity, GovCloud account ID/partition presence, or a blank (whitespace-only) explicit role name. These functions existed in `pkg/iamserviceaccount` but were never called by the CLI.
- Any error (bad input or an AWS API failure) was reported identically: printed via the reporter and the process exited 1. No usage/help text was ever shown.
- On success, "Created IAM role..." was printed immediately after role creation, before policy attachment ran.

## Behavior After This Change
- The command now delegates to pkg/iamserviceaccount.Service.CreateIAMServiceAccount, which runs the full domain Validate() in addition to the cheap CLI-side pre-checks (kept for fail-fast feedback before the OIDC/AWS lookup).
- New validations are now enforced that previously would have been silently accepted or failed later with a confusing AWS-side error: service account name/namespace format, inline policy JSON validity, GovCloud account ID/partition, and blank explicit role names.
- Domain-level violations are now aggregated and reported together in one response instead of one-at-a-time (e.g., a malformed service account name and a GovCloud misconfiguration both show up in the same error).
- The policy-ARN-format error message changed wording: it now reports the ARN's index (policy ARN at index 0 is invalid: ...) instead of embedding the literal ARN string.
- Errors are now classified: an invalid-input failure (*rosaerrors.ValidationError) additionally prints command usage/help before exiting; an operational failure (e.g., role created but policy attach failed) does not.
- The "Created IAM role..." confirmation now prints only after the entire operation succeeds, rather than before policy attachment — though aws.Client's own internal progress messages during role/policy creation are unaffected.

### Test Steps
1. <TODO:>

### Expected Results
<TODO>

## Proof of the Fix

- Screenshots:
- Videos:
- Logs/CLI output:
- Other artifacts:

## Breaking Changes
- [x] No breaking changes

Nothing breaking, but important things to consider:
1. Stricter validation could reject previously-"succeeding" calls. If anyone's automation was passing a service account name that doesn't meet Kubernetes naming rules, or a slightly-malformed inline policy JSON, IAM itself doesn't care about k8s naming conventions, so that call may have silently succeeded before (creating a role bound to a technically-invalid service account reference) and will now be rejected at the CLI.
2. The ARN-format error message text changed (index-based instead of embedding the literal ARN string). Anything scripting against that exact string breaks. Error text isn't a documented/stable interface for this CLI, but it's worth naming since it's a concrete, easy-to-hit case.

- [x] `make test` passes.
- [x] `make lint` passes.
- [x] `make rosa` passes.
- [] Any risk, limitation, or follow-up work is documented.

[ROSAENG-63262]: https://redhat.atlassian.net/browse/ROSAENG-63262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * IAM service-account creation supports inline policies provided through file references.
  * Successful IAM operations now report results more consistently.

* **Bug Fixes**
  * Added validation for malformed inputs, policy files, identifiers, role names, policy JSON, and GovCloud account details.
  * Multiple validation issues can be reported together.
  * Invalid requests fail before AWS lookups, while usage is shown for validation failures.
  * Underlying AWS and OIDC errors are preserved for clearer diagnostics.

* **Documentation**
  * Added guidance on error handling, validation, diagnostics, and exit-code conventions.

### DIFF
--- a/cmd/create/iamserviceaccount/cmd.go
+++ b/cmd/create/iamserviceaccount/cmd.go
@@ -16,14 +16,17 @@ package iamserviceaccount
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
 
+	"github.com/aws/aws-sdk-go-v2/aws/arn"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/spf13/cobra"
 
-	"github.com/openshift/rosa/pkg/aws"
+	"github.com/openshift/rosa/pkg/aws/rolebridge"
+	rosaerrors "github.com/openshift/rosa/pkg/errors"
 	"github.com/openshift/rosa/pkg/iamserviceaccount"
 	iamServiceAccountOpts "github.com/openshift/rosa/pkg/options/iamserviceaccount"
 	"github.com/openshift/rosa/pkg/rosa"
@@ -40,7 +43,7 @@ var Cmd = NewCreateIamServiceAccountCommand()
 func CreateIamServiceAccountRunner(
 	userOptions *iamServiceAccountOpts.CreateIamServiceAccountUserOptions,
 ) rosa.CommandRunner {
-	return func(ctx context.Context, r *rosa.Runtime, cmd *cobra.Command, argv []string) error {
+	return func(ctx context.Context, r *rosa.Runtime, cmd *cobra.Command, argv []string) (err error) {
 		cluster := r.FetchCluster()
 
 		// Validate cluster has STS enabled
@@ -48,40 +51,89 @@ func CreateIamServiceAccountRunner(
 			return fmt.Errorf("cluster '%s' is not an STS cluster", cluster.Name())
 		}
 
+		// Early feedback before any AWS API call (GetCreator, the OIDC
+		// provider lookup); see "Intentional duplication" in
+		// guidelines/workflow-conventions.md. CreateIAMServiceAccountRequest.
+		// Validate() checks all of this again: pkg/iamserviceaccount may
+		// eventually be called by something other than this CLI, so it must
+		// not assume these checks already ran.
+		//
+		// Usage is shown only for the checks below where a required flag was
+		// not supplied at all: that's the one case where reminding the
+		// caller of the command's flags actually helps. A value that was
+		// supplied but is malformed (an invalid ARN, an empty inline policy)
+		// is explained by its own error message; showing the full flag list
+		// alongside it reads as "your flags are wrong" when they aren't.
+		if len(userOptions.ServiceAccountNames) == 0 {
+			_ = cmd.Usage()
+			return &rosaerrors.ValidationError{
+				Field: "ServiceAccounts", Message: "at least one service account name is required",
+			}
+		}
+		if len(userOptions.PolicyArns) == 0 && userOptions.InlinePolicy == "" {
+			_ = cmd.Usage()
+			return &rosaerrors.ValidationError{Message: "at least one policy ARN or inline policy must be specified"}
+		}
+		for i, policyARN := range userOptions.PolicyArns {
+			if policyARN == "" {
+				return &rosaerrors.ValidationError{
+					Field:   "PolicyARNs",
+					Message: fmt.Sprintf("policy ARN at index %d is empty", i),
+				}
+			}
+			parsed, err := arn.Parse(policyARN)
+			if err != nil {
+				return &rosaerrors.ValidationError{
+					Field:   "PolicyARNs",
+					Message: fmt.Sprintf("policy ARN at index %d is invalid", i),
+					Err:     err,
+				}
+			}
+			if parsed.Service != "iam" || !strings.HasPrefix(parsed.Resource, "policy/") ||
+				strings.HasSuffix(parsed.Resource, "/") {
+				return &rosaerrors.ValidationError{
+					Field:   "PolicyARNs",
+					Message: fmt.Sprintf("policy ARN at index %d is not an IAM policy ARN", i),
+				}
+			}
+		}
+		if userOptions.RoleName == "" && len(userOptions.ServiceAccountNames) > 1 {
+			_ = cmd.Usage()
+			return &rosaerrors.ValidationError{
+				Field:   "RoleName",
+				Message: "role name is required when specifying multiple service accounts",
+			}
+		}
+
+		// nil means "no inline policy requested," which must be decided by
+		// whether the flag was provided, not by what it resolves to: a
+		// file:// reference that resolves to empty content is a request for
+		// an (invalid) empty inline policy, not the absence of one. Checked
+		// here, before any AWS call, as well as in domain validation; see
+		// "Intentional duplication" in guidelines/workflow-conventions.md.
+		var inlinePolicy *string
+		if userOptions.InlinePolicy != "" {
+			resolved, err := resolveInlinePolicy(userOptions.InlinePolicy)
+			if err != nil {
+				return err
+			}
+			if resolved == "" {
+				return &rosaerrors.ValidationError{
+					Field: "InlinePolicy", Message: "inline policy must not be empty when provided",
+				}
+			}
+			if !json.Valid([]byte(resolved)) {
+				return &rosaerrors.ValidationError{
+					Field: "InlinePolicy", Message: "inline policy must be valid JSON",
+				}
+			}
+			inlinePolicy = &resolved
+		}
+
 		// Get AWS creator information to determine partition for FedRAMP
 		creator, err := r.AWSClient.GetCreator()
 		if err != nil {
-			return fmt.Errorf("failed to get AWS creator information: %s", err)
-		}
-
-		// Validate service account names
-		if len(userOptions.ServiceAccountNames) == 0 {
-			return fmt.Errorf("at least one service account name is required")
-		}
-
-		// Validate that at least one policy is specified
-		if len(userOptions.PolicyArns) == 0 && userOptions.InlinePolicy == "" {
-			return fmt.Errorf("at least one policy ARN or inline policy must be specified")
-		}
-
-		// Validate policy ARNs
-		for _, arn := range userOptions.PolicyArns {
-			if err := aws.ARNValidator(arn); err != nil {
-				return fmt.Errorf("invalid policy ARN '%s': %s", arn, err)
-			}
-		}
-
-		// Generate role name if not provided
-		roleName := userOptions.RoleName
-		if roleName == "" {
-			if len(userOptions.ServiceAccountNames) > 1 {
-				return fmt.Errorf("role name is required when specifying multiple service accounts")
-			}
-			roleName = iamserviceaccount.GenerateRoleName(
-				cluster.Name(),
-				userOptions.Namespace,
-				userOptions.ServiceAccountNames[0],
-			)
+			return fmt.Errorf("failed to get AWS creator information: %w", err)
 		}
 
 		serviceAccounts := make([]iamserviceaccount.ServiceAccountIdentifier, len(userOptions.ServiceAccountNames))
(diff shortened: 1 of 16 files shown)


cmd/create/iamserviceaccount/cmd.go
```diff
@@ -40,7 +43,21 @@ var Cmd = NewCreateIamServiceAccountCommand()
...
+			// no side effects occurred, so showing usage helps the caller
+			// fix their command. An operational failure (e.g. the role was
+			// created but attaching a policy failed) is not a usage
+			// problem. This covers every return path below, not just the
+			// one from service.CreateIAMServiceAccount, so a check that
+			// fails fast before the workflow runs gets the same treatment.
+			var validationErr *rosaerrors.ValidationError
+			if errors.As(err, &validationErr) {
+				_ = cmd.Usage()
+			}
```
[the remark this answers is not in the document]
markirish: cmd.Usage() error via errors.Join. DefaultRunner only ever does Reporter.Errorf("%s", err) and exits 1 regardless of error content. Joining a near-impossible Usage() failure would just pollute the validation message shown to users, with no precedent elsewhere in the codebase for handling this error.

pkg/aws/rolebridge/rolebridge.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// AttachRolePolicy attaches a managed policy to a role.
+func (c *Client) AttachRolePolicy(_ context.Context, roleName, policyARN string) error {
+	return c.client.AttachRolePolicy(c.reporter, roleName, policyARN)
+}
+
+// PutRolePolicy creates or updates an inline policy on a role.
+func (c *Client) PutRolePolicy(_ context.Context, roleName, policyName, policy string) error {
+	return c.client.PutRolePolicy(roleName, policyName, policy)
```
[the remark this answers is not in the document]
markirish: Rejected: the underlying aws.Client interface methods (EnsureRole, AttachRolePolicy, PutRolePolicy) don't accept a context at all. The package's own doc comment already documents this as an intentional, temporary limitation to be resolved when aws.Client itself is refactored, not something to patch in this PR, and changing aws.Client's signatures is exactly the kind of broad, human-in-the-loop-required change AGENTS.md flags.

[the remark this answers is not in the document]
markirish: Rejected: pkg/iamserviceaccount/create.go (the actual caller) already wraps these with resource-specific context ("failed to attach policy '%s' to role '%s': %w"), matching the pattern in guidelines/error-conventions.md. Adding the same wrapping in rolebridge would double-wrap the error message

guidelines/error-conventions.md
```diff
@@ -0,0 +1,292 @@
...
+        return err
+    }
+    // Illustrative: operational errors could get remediation context the
+    // core layer can't add. Not yet done for this command.
+    return fmt.Errorf("%w\n\nVerify that your AWS credentials are valid and "+
+        "that your IAM user has permission to create roles", err)
+}
+```
+
+Remediation strings are CLI concerns. The core layer should never embed
```
amandahla, line 273: Just a headsup that there is a plan (a dream?) of moving messages/validations to a common library so both CLI and provider can use it
Jira ROSAENG-60583

guidelines/error-conventions.md
```diff
@@ -0,0 +1,292 @@
...
+prohibits introducing new exit codes during the migration to the target
+architecture. When that constraint is relaxed, the CLI layer would map
+error types to codes — the typed error infrastructure (`rosaerrors.ValidationError`
+and future categories) supports this, but it is not in use today.
+
+## Diagnostic Context
+
+### What to include in wrapped errors
+
+Include context that helps identify which operation failed and on what
```
amandahla, line 297: worth mention to no include sensitive information?

guidelines/error-conventions.md
```diff
@@ -0,0 +1,292 @@
...
+[ARCHITECTURE.md](ARCHITECTURE.md). For Request/Result lifecycle conventions,
+see [workflow-conventions.md](workflow-conventions.md).
+
+## Principle
+
+Core layer code (`pkg/`) returns errors. CLI layer code (`cmd/`) decides how
+to present them to the user and whether to exit the process. The core layer
+never controls the user experience of a failure.
+
+## Core Layer Error Rules
```
amandahla, line 16: minor: standard for starting with lower/upper case. In Go, error strings must start with a lowercase letter and must not end with punctuation.

guidelines/error-conventions.md
```diff
@@ -0,0 +1,292 @@
...
+}
+```
+
+Remediation strings are CLI concerns. The core layer should never embed
+guidance like "run `aws iam ...`" or "check your credentials" in its errors.
+
+### Exit codes
+
+The CLI layer decides the exit code. The core layer never returns or
+suggests an exit code. Currently, ROSA uses `0` for success and `1` for
```
amandahla, line 279: I'm considering whether this is the right place to mention preferring a 'fail-fast' approach, where errors that cause an exit code 1 are checked first. wdyt?

pkg/iamserviceaccount/create.go
```diff
@@ -50,58 +53,147 @@ type CreateIAMServiceAccountRequest struct {
...
-		if r.AccountID == "" {
-			return fmt.Errorf("account ID is required for GovCloud environments")
-		}
-		if r.Partition == "" {
-			return fmt.Errorf("partition is required for GovCloud environments")
-		}
+	return errs
+}
+
+func validateGovcloud(r *CreateIAMServiceAccountRequest) []error {
```
amandahla, line 191: Posting here since this is a good example: For checks like this one that depend on AccountID/Partition from GetCreator(), did you consider exposing them as small pure helpers (e.g. ValidateGovCloudFields(isGovcloud, accountID, partition string)) instead of only validating inside CreateIAMServiceAccountRequest.Validate()? I'm thinking ahead about that common library

guidelines/workflow-conventions.md
```diff
@@ -174,34 +176,302 @@ codebase.
 
 ### Validation
 
-The core layer validates the Request before performing any operation. Return
-a descriptive error for invalid input rather than silently correcting it or
-exiting the process.
+Validation is split between the CLI layer and the core layer. Each layer owns
+a distinct set of concerns.
+
+#### Invocation validation (CLI layer)
```
amandahla, line 182: Would it make sense to document  validation phases?  suggestion/e.g. (0) pure flag/input checks with no AWS/cluster, (1) local (file://), (2) after aws lookups (GetCreator, OIDC), (3) full Request.Validate() 
So callers know what can run fail-fast vs what needs context?

Right now it's split between CLI pre-checks (invocation) and Validate() (domain), which works for this command but might be harder to reuse consistently.

cmd/create/iamserviceaccount/cmd.go
```diff
@@ -40,48 +43,68 @@ var Cmd = NewCreateIamServiceAccountCommand()
...
+		defer func() {
+			// A validation failure means the request itself was invalid and
+			// no side effects occurred, so showing usage helps the caller
+			// fix their command. An operational failure (e.g. the role was
+			// created but attaching a policy failed) is not a usage
+			// problem. This covers every return path below, not just the
+			// one from service.CreateIAMServiceAccount, so a check that
+			// fails fast before the workflow runs gets the same treatment.
+			var validationErr *rosaerrors.ValidationError
+			if errors.As(err, &validationErr) {
```
amandahla: I'm concerned that showing the cmd usage might gave an impression to the user that something is wrong with the flag/arguments instead of something with the value. wdyt

guidelines/workflow-conventions.md
amandahla: Related: https://github.com/openshift/rosa/pull/3482#discussion_r3897539946

guidelines/error-conventions.md
amandahla: related: https://github.com/openshift/rosa/pull/3482#discussion_r3897727709

cmd/create/iamserviceaccount/cmd.go
markirish: I agree that for malformed values, printing usage will be somewhat confusing (even though the exact error should be printed after the usage print). Only in cases when values are missing should usage be printed.

pkg/iamserviceaccount/create.go
markirish: I think pulling it out now and having a ValidateXFields function makes sense, even if we currently only validate here at the moment. I pulled it out and put it in `pkg/fedramp/validate.go`

guidelines/workflow-conventions.md
markirish: Documented

cmd/create/iamserviceaccount/cmd.go
markirish: Agreed, simply posting the validation error makes more sense, except in cases when a flag or value is missing.

guidelines/error-conventions.md
markirish: Makes sense to me!

cmd/create/iamserviceaccount/cmd.go
```diff
@@ -40,48 +42,84 @@ var Cmd = NewCreateIamServiceAccountCommand()
...
+			resolved, err := resolveInlinePolicy(userOptions.InlinePolicy)
+			if err != nil {
+				return err
 			}
-			roleName = iamserviceaccount.GenerateRoleName(
-				cluster.Name(),
-				userOptions.Namespace,
-				userOptions.ServiceAccountNames[0],
-			)
+			if resolved == "" {
```
amandahla, line 120: minor: shouldnt this be an error returned by resolveInlinePolicy?

cmd/create/iamserviceaccount/cmd.go
```diff
@@ -94,71 +132,62 @@ func CreateIamServiceAccountRunner(
...
+	if err != nil {
+		return "", fmt.Errorf("failed to read policy file '%s': %w", policyPath, err)
+	}
+	return string(policyBytes), nil
+}
 
+// nilIfEmpty converts a CLI option's zero-value string into a nil pointer,
+// matching CreateIAMServiceAccountRequest's "not provided" representation
+// for optional fields.
+func nilIfEmpty(s string) *string {
```
amandahla, line 200: @coderabbitai could this be a generic helper in a common package? Does Go recent versions provide something to prevent repeating this helper here?

cmd/create/iamserviceaccount/cmd.go
```diff
@@ -40,48 +42,84 @@ var Cmd = NewCreateIamServiceAccountCommand()
...
+				return &rosaerrors.ValidationError{
+					Field: "InlinePolicy", Message: "inline policy must not be empty when provided",
+				}
+			}
+			inlinePolicy = &resolved
+		}
+
+		// Get AWS creator information to determine partition for FedRAMP
+		creator, err := r.AWSClient.GetCreator()
+		if err != nil {
```
olucasfreitas, line 135: Consider having `resolveInlinePolicy` return a `*string` (nil for empty) or an error for empty content, so file resolution and its immediate validation stay together.

cmd/create/iamserviceaccount/cmd.go
```diff
@@ -94,71 +132,62 @@ func CreateIamServiceAccountRunner(
...
+// disk when raw is a file:// reference. raw must be non-empty.
+func resolveInlinePolicy(raw string) (string, error) {
+	policyPath, ok := strings.CutPrefix(raw, "file://")
+	if !ok {
+		return raw, nil
+	}
+	policyBytes, err := os.ReadFile(policyPath)
+	if err != nil {
+		return "", fmt.Errorf("failed to read policy file '%s': %w", policyPath, err)
+	}
```
olucasfreitas, line 193: This discards the original AWS error. Consider wrapping with `%w` to preserve the cause, consistent with the `%w` convention this PR establishes.

cmd/create/iamserviceaccount/cmd.go
olucasfreitas: Agree this is worth considering. `resolveInlinePolicy` currently focuses on "read the policy content," and the empty-content check is a separate validation concern. Combining them would keep the resolution-then-validate pipeline in one place. Either way works for now — the behavior and tests are correct as-is.

cmd/create/iamserviceaccount/cmd.go
olucasfreitas: `nilIfEmpty` is 3 lines. Keeping it local avoids a new shared package for a trivial function. If more commands need the same conversion, extracting it then makes sense — but not prematurely.

guidelines/error-conventions.md
olucasfreitas: Good forward-looking note. The `ValidationError` type and validator-list pattern here are designed to be portable — no CLI dependencies, no reporter, just a struct and `errors.Join`. If ROSAENG-60583 materializes, reference this PR's pattern as the starting point.

guidelines/error-conventions.md
olucasfreitas: The "Diagnostic Context" section at lines 845-860 covers this explicitly (identifiers yes, credentials/tokens no). Worth making that cross-reference more visible from the wrapping section.

cmd/create/iamserviceaccount/cmd_test.go
```diff
@@ -217,6 +395,62 @@ var _ = Describe("Create IAM Service Account", func() {
...
+
+				providers := []aws.OidcProviderOutput{
+					{
+						Arn: "arn:aws:iam::123456789012:oidc-provider/test.example.com",
+					},
+				}
+
+				mockAWS.EXPECT().
+					GetOpenIDConnectProviderByOidcEndpointUrl("https://test.example.com").
+					Return(providers[0].Arn, nil)
```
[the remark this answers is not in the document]
olucasfreitas: Looking at the current code at HEAD (e1bdd92), `resolveInlinePolicy` is already called at lines 107-121, before `GetCreator` (line 124) and `getOIDCProviderARN` (line 137). The test correctly omits AWS mocks — the empty-file case fails before reaching them. This finding appears stale.

cmd/create/iamserviceaccount/cmd.go
amandahla: I'm thinking ahead since this might be a common thing and Go generics improved a lot specially in Go 1.27.

guidelines/workflow-conventions.md
```diff
@@ -174,34 +176,336 @@ codebase.
...
+Note that the ARN check returns `&rosaerrors.ValidationError{...}` rather
+than a plain `fmt.Errorf`, using the same type and wording as the domain
+check it duplicates (see [Classifying validation errors](#classifying-validation-errors)).
+An invocation check that fails fast still needs to be classified as an
+invalid-input error like any other, so the CLI layer can react the same way
+regardless of which layer caught it.
+
+By contrast, the STS check above and a `GetCreator` lookup failure are not
+request-invalid conditions. The request may be perfectly well-formed, and
+the environment or AWS call itself is what failed. Those stay plain wrapped
```
amandahla, line 253: This will be really helpful for tests.

pkg/iamserviceaccount/create.go
amandahla: Nice but I was thinking about helpers that expect common parameters (string/int etc) so at some point it could be easily extracted and re-used by the provider or other CLI  as well. No need to change now, just something to keep in mind in the future if you agree :)

amandahla:
/approve
There is a conflict with AGENTS.md, I can leave the lgtm once is fixed

cmd/create/iamserviceaccount/cmd.go
markirish: I think this is in the wrong spot, since the diff shows it already wraps in `%w`. But further down, around line 200, it should definitely be wrapped in `%w` to preserve context.

markirish:
@amandahla should be ready for a lgtm

cmd/create/iamserviceaccount/cmd.go
```diff
@@ -40,48 +42,84 @@ var Cmd = NewCreateIamServiceAccountCommand()
...
-				userOptions.Namespace,
-				userOptions.ServiceAccountNames[0],
-			)
+			inlinePolicy = &resolved
+		}
+
+		// Get AWS creator information to determine partition for FedRAMP
+		creator, err := r.AWSClient.GetCreator()
+		if err != nil {
+			return fmt.Errorf("failed to get AWS creator information: %w", err)
```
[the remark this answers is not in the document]
markirish: Requires a context-accepting aws.Client interface that doesn't exist, and creating an entire interface would be outside of the scope of this PR, which is focused solely on validation and error boundary.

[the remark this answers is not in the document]
markirish: Sure, create a follow up issue

guidelines/error-conventions.md
```diff
@@ -0,0 +1,336 @@
...
+        r := NewRuntime()
+        defer r.Cleanup()
+
+        if visitor != nil {
+            visitor(ctx, r, command, args)
+        }
+
+        if err := runner(ctx, r, command, args); err != nil {
+            r.Reporter.Errorf("%s", err)
+            os.Exit(1)
```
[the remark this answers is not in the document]
markirish: > error-conventions.md line 241 — cleanup skipped before os.Exit(1). I checked this is actually a real bug: defer r.Cleanup() in DefaultRunner never runs on the error path, since os.Exit() bypasses deferred calls — and worse, WithOCM()'s ocm.CreateNewClientOrExit calls os.Exit even earlier, from inside the visitor call. But: pkg/rosa/runner.go has zero diff between this branch and master, it's pre-existing code affecting every command that uses DefaultRunner, and this same doc already flags os.Exit in pkg/rosa/runtime.go as "a known legacy pattern... new code must not follow it." Fixing the actual runtime behavior is out of scope here and carries real blast radius — worth a separate ticket, not a fix folded into this PR.

@coderabbitai please create a followup issue for this note

guidelines/error-conventions.md
```diff
@@ -0,0 +1,336 @@
...
+
+`rosa.DefaultRunner` is the standard error exit point for newer commands.
+A `CommandRunner` function returns an error; `DefaultRunner` prints it via
+`reporter.Errorf` and calls `os.Exit(1)`:
+
+```go
+// pkg/rosa/runner.go
+func DefaultRunner(visitor RuntimeVisitor, runner CommandRunner) func(command *cobra.Command, args []string) {
+    return func(command *cobra.Command, args []string) {
+        ctx := context.Background()
```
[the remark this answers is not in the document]
markirish: Also out of scope, but captured in the issue you already created

markirish:
@amandahla ok, _NOW_ it should be ready for a recheck and a lgtm

PR #3547, opened 2026-09-17
ROSAENG-66596 | fix: ensure ROSA can verify and download v5 oc clients

markirish:
## PR Summary
Update the URL to download to 'latest' oc version to point to v5, which currently serves v4 but will serve v5
Ensure the regex to verify oc version will work on both v4 and v5
Add unit tests to confirm

When v5 of `oc` is released, need to ensure that `rosa download oc` and `rosa verify oc` can work with both v4 and v5 versions. There are two places in the code where v4 is hardcoded: The download URL and the regex to ensure the version number is correct. This PR fixes both, and adds a unit test to confirm both commands will work with v5.

Separately, I have smoke tested the `oc` v5 against a ROSA v4 cluster, and everything works as expected. Once v5 `oc` is available, it should be alright to point all customers to download that version, even if they are not running v5 of OpenShift

## Related Issues and PRs

- Jira: [ROSAENG-66597](https://redhat.atlassian.net/browse/ROSAENG-66597) & [ROSAENG-66598](https://redhat.atlassian.net/browse/ROSAENG-66598)
- Fixes: `#`
- Related PR(s):
- Related design/docs:

## Type of Change

## Previous Behavior

Only v4 `oc` CLI can be downloaded and verified

## Behavior After This Change

v5 `oc` can now be verified. v4 `oc` will still be downloaded from `rosa download oc`, but once v5 becomes available, it will be the default download

### Test Steps
1. run `rosa oc download` with no `oc` installed, check that the downloaded version is v4 (but URL is v5)
2. run `rosa oc verify` with a v5 version of `oc` obtained internally

### Expected Results

* v4 should be downloaded for now
* Any v5 `oc` version should be able to be verified

## Proof of the Fix

Old:
```
rosa verify oc
W: The current version (1.2.64) is not up to date with latest rosa cli released version (1.2.65).
W: It is recommended that you update to the latest version.
W: Current OpenShift Client Version: 5.0.0-rc.2
W: Your version of the OpenShift command-line tool is not supported.
Run 'rosa download oc' to download the latest version, then add it to your PATH.
```

After this PR:
```
./rosa verify oc
I: Current OpenShift Client Version: 5.0.0-rc.2

```
## Breaking Changes
- [x] No breaking changes

- [x] `make test` passes.
- [x] `make lint` passes.
- [x] `make rosa` passes.

[ROSAENG-66597]: https://redhat.atlassian.net/browse/ROSAENG-66597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated OpenShift client downloads to use the v5 client mirror.
  * OpenShift version validation now recognizes supported 4.x and 5.x versions and rejects unrelated versions.
* **Tests**
  * Added coverage for OpenShift 5.x version detection and unsupported versions.


cmd/download/oc/cmd.go
```diff
@@ -64,7 +64,7 @@ func runDownloadOC(
 	extension := extensionForGOOS(goos)
 
 	filename := fmt.Sprintf("openshift-client-%s.%s", platform, extension)
-	downloadURL := fmt.Sprintf("https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/%s", filename)
+	downloadURL := fmt.Sprintf("https://mirror.openshift.com/pub/openshift-v5/clients/ocp/latest/%s", filename)
```
olucasfreitas, line 67: This changes the default binary for every ROSA user, including existing v4 clusters, but `verify oc` cannot validate server/client compatibility. Please establish the supported v5-on-ROSA-4 contract (or retain version-aware selection) and add a test asserting the full v5 mirror URL.

cmd/verify/oc/cmd.go
```diff
@@ -68,7 +68,7 @@ func runVerifyOC(ctx context.Context, reporter rprtr.Logger, getVersion func(con
 	}
 
 	version := strings.Replace(strings.Split(string(output), "\n")[0], "\n", "", 1)
-	isCorrectVersion, err := regexp.Match(`\W4.\d*`, output)
+	isCorrectVersion, err := regexp.Match(`\W[45]\.\d*`, output)
```
olucasfreitas, line 71: This regex accepts incomplete versions such as `5.` and is not anchored to the client-version line. Please make the validation strict and add malformed/mixed-output tests so verification cannot report an invalid client as supported.

cmd/download/oc/cmd.go
olucasfreitas, line 67: Please assert the mirror path as well as the filename; otherwise this test will pass even if the v5 endpoint change regresses.

cmd/verify/oc/cmd.go
markirish: Current regex matches any string that contains:
* 1 non-character space (used to mark the beginning, but should probably use the `^` anchor after stripping whitespace)
* `4`
* 1 of _any_ character (although I think the developer intended it to be a `.` character
* 0 or more numeric characters

This change allows a match on a string that contains:
* 1 non-character space (again, should probably be `^` and string whitespace)
* `4` or `5`
* 1 `.` character
* any number of numeric characters

> This regex accepts incomplete versions such as and is not anchored to the client-version line.

I think this is by design. We really _don't_ want to get in the business of checking more than the major version for support, otherwise we are going to have a matrix of which `oc` minor and fix versions go with which `rosa` versions. At present, `oc` is hardly used by the code, so checking the major version seems like the best solution.

cmd/download/oc/cmd.go
markirish: I did smoke tests and confirmed I could do everything with a v5 rc version of `oc` on a v4 cluster that I would expect. I am still seeking official internal confirmation that `oc` can manage both v4 and v5 clusters, but my tests were fairly comprehensive for dealing with OpenShift clusters.

Currently, the v5 URL downloads the latest GA version, which is stil v4. `latest` will be updated to v5 once v5 goes GA

